PR #2, opened 2024-03-18
Update

mnonnenmacher:
Update with the latest code to prepare for moving development to the public repository.


workers/common/src/main/resources/logback.xml
```diff
@@ -26,8 +26,8 @@
     <root level="INFO">
         <appender-ref ref="STDOUT"/>
     </root>
-    <logger name="org.eclipse" level="WARN"/>
     <logger name="org.apache.http" level="WARN"/>
+    <logger name="org.eclipse.apoapsis" level="DEBUG"/>
```
kamil-bielecki-bosch, line 30: Just fo curiosity: What's reason to have DEBUG level there?

kamil-bielecki-bosch (approved):

workers/common/src/main/resources/logback.xml
mnonnenmacher: The reason is to have all logs available in the log aggregator (currently Loki), and filtering by log level can be done there. So you don't have to rerun a job to get DEBUG output.

Also see: https://github.com/eclipse-apoapsis/ort-server/commit/d56222e9813fd8b1c27f28ad80232f323b4c42ad

bs-ondem (approved):

PR #3, opened 2024-03-20
Fix failing services in Docker Compose

mmurto:
Orchestrator and workers do not spin up with the current `docker-compose.yml`. The commits fix the errors to allow for launching a development environment with compose.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -167,14 +167,16 @@ services:
       ORCHESTRATOR_SENDER_TRANSPORT_USERNAME: admin
       ORCHESTRATOR_SENDER_TRANSPORT_PASSWORD: admin
       SECRETS_PROVIDER_NAME: fileBasedSecretsProvider
-      FILE_BASED_PATH: "/etc/fileBasedSecretStorage/secrets.storage"
+      FILE_BASED_PATH: "/etc/fileBasedSecretStorage/.env.development"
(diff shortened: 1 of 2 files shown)


mnonnenmacher:
@mmurto Please check the details of the failing `eclipsefdn/eca` check, you need to create an Eclipse Foundation account and sign the ECA.

docker-compose.yml
```diff
@@ -167,14 +167,16 @@ services:
...
       GRAPHITE_PORT: 2003
       GRAPHITE_PROTOCOL: PLAINTEXT
+      CORE_SECRET_PROVIDER: secret-file
+      CORE_SECRET_FILES: /etc/fileBasedSecretStorage/.env.development
     volumes:
       - type: bind
-        source: ~/.ort/server/secrets
-        target: /etc/fileBasedSecretStorage
+        source: ./.env.development
+        target: /etc/fileBasedSecretStorage/.env.development
```
mnonnenmacher, line 179: I'm a little confused here why this is required, the database and RabbitMQ credentials should already be configured via environment variables?

mnonnenmacher: There are indeed some inconsistencies in the way secrets are handled, I will merge this as is if it fixes the issue for now and investigate separately what is the best way forward.

mnonnenmacher (approved):

PR #9, opened 2024-04-04
feat(dev): Add basic environment and requests for IDEA HTTP client

MarcelBochtler:
Add requests to read and create the basic ORT server entities with the IDEA HTTP client [1].
This setup expects a `http-client.private.env.json` file that configures the authentication.

[1]: https://www.jetbrains.com/help/idea/http-client-in-product-code-editor.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,9 @@ out/
 # Helm subcharts and install packages
 /scripts/helm/ort-server/charts/
 /scripts/helm/*.tgz
+
+# Private environment for IDEA HTTP client
+http-client.private.env.json
+
+# Private HTTP requests
+*.private.http
--- a/README.md
+++ b/README.md
@@ -96,6 +96,11 @@ without TLS.**
 | RabbitMQ       | http://localhost:15672           | admin:admin       |
 | Graphite       | http://localhost:8888            | root:root         |
 
+#### HTTP request collections
+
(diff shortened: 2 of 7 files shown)


MarcelBochtler:
Discussion:
I added only a very basic number of requests to keep it simple.
We could also import all the requests from our OpenApi spec. But I'm not sure if this would add too many and will make it more complex.

.gitignore
```diff
@@ -13,3 +13,9 @@ out/
 # Helm subcharts and install packages
 /scripts/helm/ort-server/charts/
 /scripts/helm/*.tgz
+
+# Private environment for IDEA HTTP client
+http-client.private.env.json
+
+# Private http requests
```
mnonnenmacher, line 20: Nit: HTTP (as in comment above)

mnonnenmacher (changes requested):
Could you add a README to the `requests` dir that briefly explains what these files are about and than IntelliJ Ultimate license is required to use them?

mnonnenmacher:
> Discussion: I added only a very basic number of requests to keep it simple. We could also import all the requests from our OpenApi spec. But I'm not sure if this would add too many and will make it more complex.

I think it's good to start with a small number of requests to keep it maintainable.

mnonnenmacher (approved):

PR #10, opened 2024-04-04
feat(analyzer): Enable analyzing subpaths of repositories

bs-ondem:


api/v1/mapping/src/commonMain/kotlin/Mappings.kt
```diff
@@ -258,6 +258,7 @@ fun OrtRun.mapToApi(jobs: ApiJobs) =
         index = index,
         repositoryId,
         revision,
+        path,
```
mnonnenmacher, line 261: Please add the path to the summary mapping below as well.

dao/src/main/resources/db/migration/V61__ortRunPath.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE ort_runs
+    ADD COLUMN path text NULL;
```
mnonnenmacher, line 2: Add the missing newline.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #12, opened 2024-04-08
fix(apiDocs): Fix typo in ProductsDocs

Etsija:
This PR fixes a typo in patchProductById, which leads to an erroneously named hook generated for the ORT Server UI.


Etsija:
> @Etsija, please drop the merge commit from the PR to keep a linear history (rebase instead of merge).

Yes, this was a misclick from me.

PR #18, opened 2024-04-08
ci: Fix the version of the Renovate action

mnonnenmacher:


.github/workflows/renovate.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       uses: actions/checkout@v4
 
     - name: Renovate
-      uses: renovatebot/github-action@v40
+      uses: renovatebot/github-action@v40.1.8
```
sschuberth, line 16: Could you please add a rational to the commit message? IIUC it is common practice to only pin the major version for GitHub actions.

sschuberth: The action simply lacks the required major-version only tags... see https://github.com/renovatebot/github-action/tags

oheger-bosch (approved):

PR #57, opened 2024-04-10
Initialize web UI

mmurto:
This PR adds the foundation and initial functionality of the web UI for ORT Server. The UI is a React application built with Vite.

Things still to do to get this PR ready:

- [ ] Write instructions for developing the UI in readme
- [ ] Clean up the Keycloak configuration JSON to only include additions necessary for the frontend
- [ ] Handle configuration of the Keycloak provider with environment variables, rather than hard-coding the development values


mnonnenmacher:
@mmurto Great to see you are working on the PR. Before you mark this as ready for review, is there any chance that you could split this PR into smaller chunks? I will probably have questions and comments and I'm concerned this might get hard to maintain with the current size.

mnonnenmacher:
> > @mmurto  is there any chance that you could split this PR into smaller chunks?
> 
> Sure! Based on the commits, do you have a hunch on what sort of chunkin would be the easiest for you?

Not really, maybe try to make 3 chunks of each about 10 commits, but it's hard for me to tell where to split logically so I'd leave that up to you.

Etsija:
My part of the PR (starting from [this](https://github.com/eclipse-apoapsis/ort-server/pull/57/commits/18a7379cd55801e5de8460375cfef4fc2bd43638) commit) could probably be represented with two PRs - UPDATE functionality for organizations, products and repositories, and DELETE functionality of the same items.  Or it could be even just one PR, as I don't think it's so huge, and the changes are quite isolated into a few files.

I could almost envision this PR chopped into something like
1. Initialize the UI project and set/fix the environment.
2. Setup the needed libraries and basic pages/routes structure.
3. Implement CREATE, READ of the orgs, products, repos.
4. Implement UPDATE, DELETE of the orgs, products, repos.

This is _almost_ how the PR now advances linearly - save a few commits.

mmurto:
@mnonnenmacher here are the initial commits, ready for review.

Etsija:
Two comments (very nit):
- in the README in /ui, Client details, there are no newlines between the details, while you probably intended to use them:
Capture
- also the MIT license text is missing newlines, both in source and in browser

oheger-bosch:
Hope this question is not too stupid, but would it make sense to have the UI code in a separate repository? Since a different programming language and technology is used, the setup in this repository becomes more complex. Also, there are probably different people working on the UI and on the backend.
I don't know how tight the UI is going to be coupled to the backend, but is it expected to have a separate release cycle? This would also give an indication whether to separate it or not.

sschuberth:
> Hope this question is not too stupid, but would it make sense to have the UI code in a separate repository?

The question is not stupid it all. We already discussed this in advance with @mnonnenmacher (and I was under the assumption that he shared the results of that discussion with you guys). 

The most compelling reason to keep the UI code were was that ideally we'd like to be able to auto-generate some of its code from common Kotlin code at some point, and that API refactorings could be more easily done.

> Also, there are probably different people working on the UI and on the backend.

True, but I don't regard that as a reason to split out the code. We can easily do https://github.com/eclipse-apoapsis/ort-server/issues/74 and similar things to manage.

> I don't know how tight the UI is going to be coupled to the backend, but is it expected to have a separate release cycle?

I'm not expecting this, but hard to say right now. It depends on things like how quickly code churn settles down and projects stabilize. Maybe at some point we'll have much more UI polishing that work being done on the backend, which would justify for different release cycles. But again, having the code in a single repo does not prevent that. It just could make versioning a bit trickier if SemVer is derived from Git commits.

oheger-bosch:
@sschuberth, thanks for explaining. Seems that our internal communication could be improved.

ui/.eslintrc.cjs
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
sschuberth, line 20: Commit message: Please provide *brief* rationales why these technologies were used, esp. when it comes to less obvious choices like Vite (maybe) and pnpm.

Hint: For React, we could way it's already being used at least in ORT's web-app report, and for pnpm we could say that at least it's supported by ORT's analyzer (I'd not use a technology that's not supported by ORT's analyzer).

sschuberth: How was this file created? If it was created by some React / Vite specific init command, please name that command in the commit message.

ui/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+lerna-debug.log*
```
sschuberth, line 8: Why do we need all of these if we only use pnpm?

ui/.gitignore
```diff
@@ -0,0 +1,24 @@
...
+
+# Editor directories and files
+.vscode/*
+!.vscode/extensions.json
+.idea
+.DS_Store
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
```
sschuberth, line 23: Some of these seem unnecessary, like Visual Studio solution files... or does VS Code really create these?

ui/README.md
```diff
@@ -0,0 +1,30 @@
+# React + TypeScript + Vite
+
+This template provides a minimal setup to get React working in Vite with HMR and some ESLint rules.
```
sschuberth: What is HMR? Mabe just link HMR (and also ESLint) to their homepages for completeness?

ui/.eslintrc.cjs
sschuberth: Should we talk about the directory name here, is "ui" fine with everyone? Should we maybe use "frontend" or so instead?

core/build.gradle.kts
```diff
@@ -68,6 +68,7 @@ dependencies {
     implementation(libs.ktorServerCallLogging)
     implementation(libs.ktorServerContentNegotiation)
     implementation(libs.ktorServerCore)
+    implementation(libs.ktorServerCors)
```
sschuberth, line 71: Please link to https://ktor.io/docs/server-cors.html in the commit message to explain to readers what CORS is.

core/src/main/kotlin/plugins/HTTP.kt
```diff
@@ -19,12 +19,23 @@
...
 import io.ktor.server.application.Application
 import io.ktor.server.application.install
+import io.ktor.server.plugins.cors.routing.CORS
 import io.ktor.server.plugins.defaultheaders.DefaultHeaders
 
 fun Application.configureHTTP() {
     install(DefaultHeaders) {
         header("X-Engine", "Ktor") // will send this header with each response
     }
+    install(CORS) {
```
sschuberth, line 33: We usually prefer to have an empty line before new code blocks / between previous and next code blocks.

.run/Run ORT Server.run.xml
```diff
@@ -2,9 +2,11 @@
   <configuration default="false" name="Run ORT Server" type="JetRunConfigurationType">
     <envs>
       <env name="DB_NAME" value="ort_server" />
+      <env name="DB_HOST" value="localhost" />
```
sschuberth, line 5: Please say in the commit message how you found out that "DB_SSL_MODE" is the correct name to use.

Hint: Probably say that this needs to be align with `docker-compose.yml` / `application.conf` files.

ui/package.json
```diff
@@ -10,8 +10,10 @@
     "preview": "vite preview"
   },
   "dependencies": {
+    "oidc-client-ts": "^3.0.1",
```
sschuberth, line 22: If there was a choice between different provider, please explain in the commit message why you chose this one.

ui/package.json
```diff
@@ -10,12 +10,15 @@
     "preview": "vite preview"
   },
   "dependencies": {
+    "@tanstack/react-router": "^1.22.2",
```
sschuberth, line 17: Please explain in the commit message what a "TanStack Router" is: What is it used for? What problem does it solve / feature does it implement?

ui/components.json
```diff
@@ -0,0 +1,17 @@
+{
```
sschuberth, line 1: Commit message: Please add a link to https://ui.shadcn.com/ and provide a rationale why this was chosen over other component frameworks.

sschuberth: Missing final newline in file.

ui/tailwind.config.js
```diff
@@ -0,0 +1,77 @@
+/** @type {import('tailwindcss').Config} */
```
sschuberth, line 20: Missing final newline in file.

ui/README.md
```diff
@@ -1,30 +1,23 @@
-# React + TypeScript + Vite
+# Web UI
```
sschuberth, line 1: Does it make sense to squash this change into the commit that introduced this file?

ui/.eslintrc.cjs
```diff
@@ -1,3 +1,22 @@
+/*
```
sschuberth, line 1: These should each go to the commits that introduced the respective files.

That's because we have a policy that each commit in a PR should pass all checks also on its own, not just the PR as a whole. Not having these headers would temporarily break REUSE checks.

LICENSES/MIT.txt
```diff
@@ -0,0 +1,18 @@
+MIT License
```
sschuberth, line 1: Similar here, move this to where shadcn/ui was introduced.

core/build.gradle.kts
mnonnenmacher: Could you also briefly explain why this is required, e.g. where do cross origin requests happen?

ui/README.md
mnonnenmacher: I guess this is an autogenerated README that needs to be replaced/amended later on?

.run/Run ORT Server.run.xml
mnonnenmacher: You could just say: "This is a fixup for a546302."

core/src/main/kotlin/plugins/HTTP.kt
sschuberth: ... but actually, this does not necessarily seem to be the case for the ORT server code base, also see https://github.com/eclipse-apoapsis/ort-server/pull/81/files.

ui/.eslintrc.cjs
sschuberth: > esp. when it comes to less obvious choices like Vite (maybe) and pnpm.

For Vite, you might "borrow" some ideas for a rationale from [over here](https://github.com/oss-review-toolkit/ort/pull/7952/commits/a9764049400ab26cc93160358ee200739ec98e0d).

mmurto:
Note, not all comments have been addressed yet, I'll get to them next week.

ui/package.json
mmurto: No real alternatives that I know of.

mnonnenmacher (approved):

PR #62, opened 2024-04-10
chore(scripts): Consistenly use the *.yaml file extension for Helm charts

sschuberth:


sschuberth:
@mnonnenmacher, does something need to be adjusted in your side due to this renaming?

mnonnenmacher:
> @mnonnenmacher, does something need to be adjusted in your side due to this renaming?

No, the Helm chart was only used for local testing with Minikube but has unfortunately not been maintained for a while, so I doubt that it still works correctly.

sschuberth:
> No, the Helm chart was only used for local testing with Minikube

Ok, so then the good news is that the PR could be merged without breaking anything, right? Because it will still simplify dealing with YAML file pattern as part of `.reuse/dep5`.

mnonnenmacher (approved):

Issue #74, opened 2024-04-11
Set up `.github/CODEOWNERS`


sschuberth:
We should have code ownership to auto-assign reviewers, esp. once #57 is merged and we'll have frontend / backend code.

sschuberth:
Actually, I just saw that the file is already there, but probably contains wrong groups, and does not seem to be used to auto-assign reviewers to PRs.

mnonnenmacher:
I have made an update to the CODEOWNERS, I will check if we can create visible teams for frontend or backend developers once any frontend developers have become project committers.

mnonnenmacher:
@sschuberth Regarding the auto assignment of codeowners as PR reviewers, do you suggest to enable the "Require review from Code Owners" option in the branch protection rule or how else would that work?

sschuberth:
> do you suggest to enable the "Require review from Code Owners" option

IIUC that's not necessary, a `CODEOWNERS` file should [be used also without that option](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-and-forks) as soon as it's there. At least that's how it work in ORT core, where we also do not have "Require review from Code Owners" enabled, but do get auto-assigned reviewers.

sschuberth:
Auto-assigning reviewers works now, so let's close this and maintain code ownership going forward.

PR #87, opened 2024-04-14
chore: Update CODEOWNERS

mnonnenmacher:
Set the active committers as CODEOWNERS for the whole project. Note that currently there is not team that can be used, as the automatically created team for project committers is not visible.

Relates to #74.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,4 +2,4 @@
 # last matching pattern is used, see
 # https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners#codeowners-syntax
 
-* @oss-review-toolkit/bosch
+* @bs-ondem @MarcelBochtler @mnonnenmacher @nnobelis @oheger-bosch


.github/CODEOWNERS
```diff
@@ -2,4 +2,4 @@
 # last matching pattern is used, see
 # https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners#codeowners-syntax
 
-* @oss-review-toolkit/bosch
+* @bs-ondem @MarcelBochtler @mnonnenmacher @nnobelis @oheger-bosch
```
sschuberth, line 5: Would it make to create teams already that can be referred here? Like "frontend" / "backend", or "kotlin-devs" / "typescript-devs"?

mnonnenmacher: I cannot create teams on my own, so I'd rather go with those names first until it's clarified what options we have for creating teams, e.g. if non-committers could be added to teams without write permission.

sschuberth (approved):

.github/CODEOWNERS
```diff
@@ -2,4 +2,4 @@
 # last matching pattern is used, see
 # https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners#codeowners-syntax
 
-* @oss-review-toolkit/bosch
```
oheger-bosch, line 5: Commit message nit: "...there is *no* team that can be used..."

MarcelBochtler (approved):

PR #88, opened 2024-04-15
Fix snippet choice

nnobelis:


workers/common/src/main/kotlin/common/OrtServerMappings.kt
mnonnenmacher, line 1: Commit message: Better say "Add missing snippet choice mappings" to not confuse it with actual serialization (same in next commit).
What's the reason to split this in two commits if the messages are identical?

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #89, opened 2024-04-15
ci(renovate): Configure schedule for `aws.sdk.kotlin:s3`

mnonnenmacher:
Configure Renovate to only do weekly updates [1] of `aws.sdk.kotlin:s3` as the package is updated very often.

[1]: https://docs.renovatebot.com/configuration-options/#packagerules

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -12,6 +12,10 @@
     {
       "matchManagers": ["gradle"],
       "commitMessageTopic": "{{depName}}"
+    },
+    {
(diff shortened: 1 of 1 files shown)


MarcelBochtler (approved):

sschuberth:
Just as a remark, in ORT we decided to simply [not take patch-level updates](https://github.com/oss-review-toolkit/ort/blob/c4c14f2cb29a166c6b0d5410c9cc11f6660a05d8/renovate.json#L37-L41). Maybe that would work better here as well instead of limiting by time frame?

PR #102, opened 2024-04-17
fix(reporter): Fix a version clash with YAML libraries

oheger-bosch:
The recent update of `kaml` dragged in a version of `snakeyaml` that was incompatible with jRuby and thus caused the generation of AsciiDoc reports to fail. Fix this by excluding the `kaml` library for the Reporter worker.


nnobelis (approved):

MarcelBochtler (approved):
I guess issues like this could only be avoided with full fledged end-to-end tests, right?

sschuberth:
Maybe also taking https://github.com/gradlex-org/jvm-dependency-conflict-resolution into use could have helped here...

PR #119, opened 2024-04-19
feat(ci): Upload test results as build artifacts

MarcelBochtler:
Upload the test result html to help finding issues with failing tests.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -55,6 +55,14 @@ jobs:
     - name: Run tests
       run: ./gradlew test -Dkotest.tags='!Integration'
 
+    - name: Upload test results
+      uses: actions/upload-artifact@v4
+      if: success() || failure()
+      with:
+        name: test-results
+        path: "**/build/reports/tests"
+        retention-days: 7
+
   integration-test:
     needs: build
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -55,6 +55,14 @@ jobs:
     - name: Run tests
       run: ./gradlew test -Dkotest.tags='!Integration'
 
+    - name: Upload test results
+      uses: actions/upload-artifact@v4
+      if: success() || failure()
```
sschuberth, line 60: Maybe instead use `if: always()`?

MarcelBochtler: I intentionally used `success() || failure()` because the step should not run if a user cancels the tests.

sschuberth: Interesting... I guess that's worth a code comment? Or, if you don't want to comment twice in the code, a hint in the commit message at least would be great.

mnonnenmacher (approved):

.github/workflows/build-and-test.yml
sschuberth: For reference, [these](https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions) are the different status check functions.

PR #124, opened 2024-04-22
feat(notifier): Add Jira notifier configuration

wkl3nk:
Providing a way to configure a Jira notifier to interact with a Jira server at the end of an ORT run, e.g. to create a comment to an existing Jira issue about the result of the ORT run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,13 +98,13 @@ without TLS.**
 
 ### Accessing the services
 
-| Service        | URL                              | Credentials       |
-|----------------|----------------------------------|-------------------|
-| ORT Server API | http://localhost:8080/swagger-ui |                   |
-| Keycloak       | http://localhost:8081            | admin:admin       |
-| PostgreSQL     | http://localhost:5433            | postgres:postgres |
-| RabbitMQ       | http://localhost:15672           | admin:admin       |
-| Graphite       | http://localhost:8888            | root:root         |
+| Service        | URL                              | Credentials                                             |
+|----------------|----------------------------------|---------------------------------------------------------|
+| ORT Server API | http://localhost:8080/swagger-ui |                                                         |
+| Keycloak       | http://localhost:8081            | Administrator: admin:admin<br/>User: ort-admin:password |
+| PostgreSQL     | http://localhost:5433            | postgres:postgres                                       |
+| RabbitMQ       | http://localhost:15672           | admin:admin                                             |
+| Graphite       | http://localhost:8888            | root:root                                               |
 
 #### HTTP request collections
 
--- a/workers/notifier/src/main/kotlin/notifier/NotifierRunner.kt
+++ b/workers/notifier/src/main/kotlin/notifier/NotifierRunner.kt
@@ -70,16 +70,35 @@ class NotifierRunner {
                 workerContext.configManager.getSecret(Path(it))
             }
 
-            val notifierConfiguration = if (mailServerUser != null && mailServerPassword != null) {
-                NotifierConfiguration(
-                    mail = config.mail?.mailServerConfiguration?.copy(
+            val sendMailConfiguration = if (mailServerUser != null && mailServerPassword != null) {
+                config.mail?.mailServerConfiguration?.copy(
                         username = mailServerUser,
                         password = mailServerPassword
                     )?.mapToOrt()
-                )
             } else {
-                NotifierConfiguration()
+                null
             }
+
+            val jiraRestClientUsername = config.jira?.jiraRestClientConfiguration?.username?.let {
+                workerContext.configManager.getSecret(Path(it))
+            }
+            val jiraRestClientPassword = config.jira?.jiraRestClientConfiguration?.password?.let {
+                workerContext.configManager.getSecret(Path(it))
+            }
+
+            val jiraConfiguration = if (jiraRestClientUsername != null && jiraRestClientPassword != null) {
+                config.jira?.jiraRestClientConfiguration?.copy(
+                        username = jiraRestClientUsername,
+                        password = jiraRestClientPassword
+                    )?.mapToOrt()
+            } else {
+                null
+            }
+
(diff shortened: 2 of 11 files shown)


workers/notifier/src/main/kotlin/notifier/NotifierRunner.kt
```diff
@@ -70,17 +72,36 @@ class NotifierRunner {
...
+                        username = jiraRestClientUsername,
+                        password = jiraRestClientPassword
+                    )?.mapToOrt()
+            } else {
+                JiraConfiguration()
             }
 
+            val notifierConfiguration = NotifierConfiguration(
+                mail = sendMailConfiguration,
+                jira = jiraConfiguration
```
MarcelBochtler, line 102: IMO we shouldn't use the default mail and Jira configurations, if the user didn't configure them.
I think both values should be `null` if they are not configured. This causes a different behavior in the ORT notifier, where `null` is handled separately:
https://github.com/oss-review-toolkit/ort/blob/2fcae578cfa3b5b6e451e7ab0300699c2b4d7982/notifier/src/main/kotlin/Notifier.kt#L46-L47

README.md
```diff
@@ -98,13 +98,13 @@ without TLS.**
...
-|----------------|----------------------------------|-------------------|
-| ORT Server API | http://localhost:8080/swagger-ui |                   |
-| Keycloak       | http://localhost:8081            | admin:admin       |
-| PostgreSQL     | http://localhost:5433            | postgres:postgres |
-| RabbitMQ       | http://localhost:15672           | admin:admin       |
-| Graphite       | http://localhost:8888            | root:root         |
+| Service        | URL                              | Credentials                                             |
+|----------------|----------------------------------|---------------------------------------------------------|
+| ORT Server API | http://localhost:8080/swagger-ui |                                                         |
+| Keycloak       | http://localhost:8081            | Administrator: admin:admin<br/>User: ort-admin:password |
```
MarcelBochtler, line 104: This change should be in a separate commit.

MarcelBochtler (changes requested):

workers/notifier/src/main/kotlin/notifier/NotifierRunner.kt
wkl3nk: Actually, we have no other choice than providing default Mail and Jira configurations, if the user did not explicitly configure one, because, as you already pointed out, the _Notifier_ from the oss-review-toolkit "injects" variables _mailClient_ and _jiraClient_ into the script _reporter-notification.kts_ (ort-server-config). If only **one** of these variables is not injected, then the complete kotlin script **will not compile**, and neither Mail nor Jira notifications are created at all.

_Unresolved reference: mailClient
Unresolved reference: jiraClient_

This raises the question, if there should be a refactoring in a way that there is one kotlin reporter script for Mail notifications, and another one for Jira notifications. Then, depending on which configurations are present, one or the other script could be executed or not. But IMO this would be a separate explicit change request.

sschuberth: > If only one of these variables is not injected, then the complete kotlin script will not compile

Actually, I believe these tests prove the contrary:

[Run a script without a Jira configuration](https://github.com/oss-review-toolkit/ort/blob/2fcae578cfa3b5b6e451e7ab0300699c2b4d7982/notifier/src/test/kotlin/NotifierTest.kt#L41-L49)

[Run a script without a mail configuration](https://github.com/oss-review-toolkit/ort/blob/2fcae578cfa3b5b6e451e7ab0300699c2b4d7982/notifier/src/test/kotlin/NotifierTest.kt#L61-L69)

README.md
wkl3nk: Done.

workers/notifier/src/main/kotlin/notifier/NotifierRunner.kt
wkl3nk: You convinced me. Let's stick to the contract. If a user provides a notifier script that contains notification code for both mail and jira, but forgot to take care for proper of mail and jira, then better let the whole script crash.

bs-ondem (approved):

workers/notifier/src/main/kotlin/notifier/NotifierRunner.kt
```diff
@@ -70,17 +70,36 @@ class NotifierRunner {
...
+            }
+
+            val jiraRestClientUsername = config.jira?.jiraRestClientConfiguration?.username?.let {
+                workerContext.configManager.getSecret(Path(it))
+            }
+            val jiraRestClientPassword = config.jira?.jiraRestClientConfiguration?.password?.let {
+                workerContext.configManager.getSecret(Path(it))
+            }
+
+            val jiraConfiguration = if (jiraRestClientUsername != null && jiraRestClientPassword != null) {
```
oheger-bosch, line 89: Nit: You could drop the `else { null }` part by replacing the `if` construct with `takeIf`, e.g. something like

```
val jiraConfiguration = config.jira?.jiraRestClientConfiguration?.takeIf { 
  jiraRestClientUsername != null && jiraRestClientPassword != null
}?.copy(username = jiraRestClientUsername, password = jiraRestClientPassword)?.mapToOrt()
```

workers/notifier/src/test/kotlin/NotifierRunnerTest.kt
```diff
@@ -62,6 +64,13 @@ class NotifierRunnerTest : WordSpec({
                     useSsl = false,
                     fromAddress = "no-reply@oss-review-toolkit.org"
                 )
+            ),
```
oheger-bosch, line 67: Unfortunately, the test is not really meaningful - but this is out of scope of this PR.

mnonnenmacher:
@wkl3nk No need to recreate the PR, we can just dismiss Marcel's review in such cases.

PR #134, opened 2024-04-23
ci: Add UI building to build and test workflow

### LINKED ISSUE #126: UI: Need a GitHub Action to build the UI
This issue is targeted specifically for the Double Open team: 

As an automatic means to test that Renovate Bot's dependency updates for the UI work (as in "if it builds, it works"), a GitHub Action to build the UI should be added to the workflows that are run for PRs.

Etsija:
Add UI building to the corresponding workflow, to automatically assess the buildability of the UI codebase after Renovate Bot 's suggested UI dependency updates. Use caching to reduce installation time.

This resolves #126.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -34,6 +34,38 @@ jobs:
     - name: Build all classes
       run: ./gradlew --stacktrace classes
 
+  build-ui:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v4
+
+    - name: Install Node
+      uses: actions/setup-node@v4
+      with:
+        node-version: 20
+
+    - uses: pnpm/action-setup@v3
+      name: Install pnpm
+      with:
+        version: 8
+
+    - name: Get pnpm store directory
+      shell: bash
+      run: |
+        echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
+
+    - uses: actions/cache@v4
+      name: Setup pnpm cache
+      with:
+        path: ${{ env.STORE_PATH }}
+        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+        restore-keys: |
+          ${{ runner.os }}-pnpm-store-
+
+    - name: Build UI
+      run: pnpm -C ui install build
+
   test:
     needs: build
     runs-on: ubuntu-latest


sschuberth:
What's the rationale for a whole separate workflow, instead of a separate job in the existing https://github.com/eclipse-apoapsis/ort-server/blob/main/.github/workflows/build-and-test.yml?

sschuberth:
> it ran much faster than the whole build-and-test workflow.

Note that jobs in a workflow run in parallel. So you wouldn't have to wait for the whole build-and-test workflow to complete just to see the results of your "Build UI" job.

.github/workflows/build-and-test.yml
```diff
@@ -34,6 +34,35 @@ jobs:
     - name: Build all classes
       run: ./gradlew --stacktrace classes
 
+    - name: Install Node
```
mnonnenmacher, line 37: Can you move the new steps to a separate job, e.g. `build-ui`? This provides faster feedback as the builds run in parallel and you can see immediately in a PR if the Gradle or the PNPM builds failed.

.github/workflows/build-and-test.yml
```diff
@@ -34,6 +34,35 @@ jobs:
...
+      with:
+        path: ${{ env.STORE_PATH }}
+        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+        restore-keys: |
+          ${{ runner.os }}-pnpm-store-
+
+    - name: Install UI dependencies
+      run: |
+        cd ui
+        pnpm install
```
mnonnenmacher, line 64: Maybe this should also run `pnpm build`?

mnonnenmacher (changes requested):

.github/workflows/build-and-test.yml
```diff
@@ -34,6 +34,35 @@ jobs:
...
+    - name: Install Node
+      uses: actions/setup-node@v4
+      with:
+        node-version: 20
+
+    - uses: pnpm/action-setup@v3
+      name: Install pnpm
+      with:
+        version: 8
+        run_install: false
```
sschuberth, line 46: That's the [default behavior](https://github.com/pnpm/action-setup?tab=readme-ov-file#run_install) and should be dropped.

.github/workflows/build-and-test.yml
```diff
@@ -34,6 +34,35 @@ jobs:
     - name: Build all classes
       run: ./gradlew --stacktrace classes
 
+    - name: Install Node
+      uses: actions/setup-node@v4
```
sschuberth, line 38: If Node is just installed for PNPM to run, we could consider instead enabling [standalone mode](https://github.com/pnpm/action-setup?tab=readme-ov-file#standalone) below.

.github/workflows/build-and-test.yml
sschuberth: AFAIK `pnpm` do snot have a `build` [sub-command](https://pnpm.io/cli/install).

sschuberth, line 64: Maybe making use of the [`-C` / `--dir` option](https://pnpm.io/pnpm-cli#-c-path---dir-path) would make up for a slight more elegant one-liner.

Etsija: We do have a "build" target defined for our setup.

.github/workflows/build-and-test.yml
Etsija: We have some types for Node installed, so I'm hesitant to drop it at this point.

.github/workflows/build-and-test.yml
Etsija: Done.

.github/workflows/build-and-test.yml
```diff
@@ -34,6 +34,41 @@ jobs:
...
+        path: ${{ env.STORE_PATH }}
+        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+        restore-keys: |
+          ${{ runner.os }}-pnpm-store-
+
+    - name: Install UI dependencies
+      run: |
+        cd ui
+        pnpm install
+        pnpm build
```
sschuberth, line 70: I guess this could still be combined into `pnpm -C ui install build`.

.github/workflows/build-and-test.yml
```diff
@@ -34,6 +34,41 @@ jobs:
...
+
+    - uses: actions/cache@v4
+      name: Setup pnpm cache
+      with:
+        path: ${{ env.STORE_PATH }}
+        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+        restore-keys: |
+          ${{ runner.os }}-pnpm-store-
+
+    - name: Install UI dependencies
```
sschuberth, line 66: Now that we have `build` below, this should probably just say "Build UI" (as you cannot build without installing).

.github/workflows/build-and-test.yml
Etsija: OK, I'm learning new tricks here.

sschuberth (approved):

mnonnenmacher (approved):

PR #143, opened 2024-04-24
fix(snippets): Remove the unique constraint on the snippets table

nnobelis:
In production, there are some errors when inserting snippets in the table, because the column "additional_data" is too large for the index related to the unique constraint on this table's columns. This index is also growing and is currently taking more place than the snippet data itself.
However, this column `additional_data` is also containing the FossID snippet id, which is unique for every FossID scan. Therefore, this unique constraint does not make sense and can be removed.

### DIFF
--- a/dao/src/main/kotlin/tables/SnippetsTable.kt
+++ b/dao/src/main/kotlin/tables/SnippetsTable.kt
@@ -66,8 +66,8 @@ class SnippetDao(id: EntityID<Long>) : LongEntity(id) {
                 it.mapToModel().provenance == snippet.provenance && it.additionalData?.data == snippet.additionalData
             }
 
-        fun getOrPut(snippet: Snippet): SnippetDao =
-            findBySnippet(snippet) ?: new {
+        fun put(snippet: Snippet): SnippetDao =
+            new {
                 this.purl = snippet.purl
                 this.path = snippet.location.path
                 this.startLine = snippet.location.startLine
--- a/dao/src/main/resources/db/migration/V63__removeSnippetIndex.sql
+++ b/dao/src/main/resources/db/migration/V63__removeSnippetIndex.sql
@@ -1,0 +1,2 @@
+ALTER TABLE snippets
+DROP CONSTRAINT snippets_purl_artifact_id_vcs_id_path_start_line_end_line_l_key;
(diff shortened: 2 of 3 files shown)


dao/src/main/kotlin/tables/SnippetsTable.kt
```diff
@@ -66,8 +66,8 @@ class SnippetDao(id: EntityID<Long>) : LongEntity(id) {
                 it.mapToModel().provenance == snippet.provenance && it.additionalData?.data == snippet.additionalData
```
oheger-bosch, line 66: Commit message: I did not fully understand the implications on first reading. Maybe, it could be made clearer, e.g. "Therefore, this unique constraint does not make sense, since there will never be duplicates anyway." or something like that?

nnobelis: Rephrased.

nnobelis:
@mnonnenmacher  I did a local test:

- I inserted 1000000 entries in the snippet table. I tried to replicate complex content filling up the index by having the `additional_data` column like:
```
{"data": {"entry1": "1-0e13730f-1374-437f-a1c4-3a7924908d27", "entry2": "2-a7f688db-730c-4300-80e7-ef11797962b2", "entry3": "3-ab5d41c7-3eb9-429a-90fd-be08db09db9c", "entry4": "4-8c8df2cb-2311-4c9f-9b9a-b86a87bedf89", "entry5": "5-69f34acc-5b79-4026-b5fa-64f519d22a65", "entry6": "6-ce30bccb-9871-4d81-9f1b-b99cfbac93bd", "entry7": "7-db24c477-90fe-44cf-89b0-fbd952267127", "entry8": "8-16bd1e82-33b0-41eb-9c54-57e1a6401e0c", "entry9": "9-aba13a8b-1479-483e-9b7b-7ee047204a9c", "entry10": "10-ed0bde83-3d66-41be-b8e0-1862608af6ae"}}
```
- Then I deleted the constraint with an SQL editor and it was instantaneous.

I think there is no problem. Can we merge it ?
We can alleviate the outage by doing the deployment  at the end of business hours.

nnobelis:
Test failure seems unrelated.

mnonnenmacher (approved):

PR #150, opened 2024-04-25
Avoid non-null assertions, part 1

sschuberth:

### DIFF
--- a/core/src/main/kotlin/api/RepositoriesRoute.kt
+++ b/core/src/main/kotlin/api/RepositoriesRoute.kt
@@ -36,6 +36,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.mapping.mapToApiSummary
 import org.eclipse.apoapsis.ortserver.api.v1.mapping.mapToModel
 import org.eclipse.apoapsis.ortserver.api.v1.model.CreateOrtRun
 import org.eclipse.apoapsis.ortserver.api.v1.model.CreateSecret
+import org.eclipse.apoapsis.ortserver.api.v1.model.JobSummaries
 import org.eclipse.apoapsis.ortserver.api.v1.model.Jobs
 import org.eclipse.apoapsis.ortserver.api.v1.model.PagedResponse
 import org.eclipse.apoapsis.ortserver.api.v1.model.SortDirection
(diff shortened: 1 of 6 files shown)


workers/reporter/src/main/kotlin/reporter/ReportNameMapper.kt
```diff
@@ -77,7 +77,7 @@ interface ReportNameMapper {
     }
 
     /**
-     * Return a map that assigns each element of the given collection of [reportFiles] a name. The report file is then
+     * Return a map that assigns each element of the given [reportFiles] collection by a name. The report file is then
```
mnonnenmacher, line 80: "assigns ... by a name" doesn't work, better: "...assigns a name to each element of..."

core/src/main/kotlin/api/RepositoriesRoute.kt
```diff
@@ -110,8 +110,12 @@ fun Route.repositories() = route("repositories/{repositoryId}") {
             val repositoryId = call.requireIdParameter("repositoryId")
             val pagingOptions = call.pagingOptions(SortProperty("index", SortDirection.ASCENDING))
 
-            val jobsForOrtRuns = repositoryService.getOrtRuns(repositoryId, pagingOptions.mapToModel())
-                .map { it.mapToApiSummary(repositoryService.getJobs(repositoryId, it.index)!!.mapToApiSummary()) }
+            val jobsForOrtRuns = repositoryService.getOrtRuns(repositoryId, pagingOptions.mapToModel()).mapNotNull {
+                repositoryService.getJobs(repositoryId, it.index)?.let { jobs ->
+                    it.mapToApiSummary(jobs.mapToApiSummary())
+                }
```
mnonnenmacher, line 116: Better use an empty `JobSummaries` object as fallback instead of `mapNotNull`, then at least the run is still in the list even if there is an issue with loading the jobs.

mnonnenmacher (changes requested):

workers/reporter/src/main/kotlin/reporter/ReportNameMapper.kt
sschuberth: I believe "assigns *each element* ... by a name" *does* work. I'll change it nonetheless.

core/src/main/kotlin/api/RepositoriesRoute.kt
sschuberth: IIUC `JobSummaries` does not work here, did you mean a `OrtRunSummary`? But the latter cannot be easily created empty.

sschuberth: Ok, I now got what you mean, I believe.

mnonnenmacher (approved):

PR #153, opened 2024-04-26
ui: Add Organizations

Etsija:
This PR adds CRUD (Create, Read, Update, Delete) functionality of Organizations to the ORT Server UI. Please see individual commits for details.


ui/src/components/ui/card.tsx
```diff
@@ -0,0 +1,85 @@
+/*
+ * The component is originally a part of the shadcn/ui library at https://github.com/shadcn-ui/ui.
```
sschuberth, line 2: I realize we already have similar files in `ui/src/components/ui` in `main`, but why are these *copied* from `shadcn-ui` instead of adding the latter as a dependency, @mmurto?

mmurto: That's how shadcn works - it's not installable but more of a copy-pasteable starting point, that can be customized further: https://ui.shadcn.com/docs#:~:text=Introduction-,Introduction,Use%20this%20as%20a%20reference%20to%20build%20your%20own%20component%20libraries.,-FAQ

sschuberth: Wow, I really have to get used to this, but thanks both for clarifying quickly!

mmurto: It can also be thought of we using [Radix Primitives](https://www.radix-ui.com/primitives) as our unstyled component library, and shadcn/ui is an example of how components can be styled on top of that. :)

.reuse/dep5
mnonnenmacher, line 1: Could you add a short section to `ui/README.md` with the exact command used to generate the code? This will be useful in future if someone else has to recreate it.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #158, opened 2024-04-26
Avoid non null assertions, part 2

sschuberth:

### DIFF
--- a/dao/src/main/kotlin/Exceptions.kt
+++ b/dao/src/main/kotlin/Exceptions.kt
@@ -33,7 +33,7 @@ enum class PostgresErrorCodes(val value: String) {
 
 class UniqueConstraintException(msg: String, cause: Throwable) :
     RuntimeException(msg, cause), CopyableThrowable<UniqueConstraintException> {
-    override fun createCopy() = UniqueConstraintException(message!!, cause!!)
+    override fun createCopy() = UniqueConstraintException(checkNotNull(message), cause!!)
 }
 
 /**
@@ -41,5 +41,5 @@ class UniqueConstraintException(msg: String, cause: Throwable) :
  */
 class QueryParametersException(msg: String, cause: Throwable? = null) :
     RuntimeException(msg, cause), CopyableThrowable<QueryParametersException> {
-    override fun createCopy() = QueryParametersException(message!!, cause)
+    override fun createCopy() = QueryParametersException(checkNotNull(message), cause)
 }
--- a/dao/src/main/kotlin/Exceptions.kt
+++ b/dao/src/main/kotlin/Exceptions.kt
@@ -33,7 +33,7 @@ enum class PostgresErrorCodes(val value: String) {
 
 class UniqueConstraintException(msg: String, cause: Throwable) :
     RuntimeException(msg, cause), CopyableThrowable<UniqueConstraintException> {
-    override fun createCopy() = UniqueConstraintException(checkNotNull(message), cause!!)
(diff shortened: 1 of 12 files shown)


dao/src/main/kotlin/tables/runs/repository/PackageCurationDataTable.kt
```diff
@@ -68,7 +68,7 @@ class PackageCurationDataDao(id: EntityID<Long>) : LongEntity(id) {
                 it.binaryArtifact?.mapToModel() == data.binaryArtifact &&
                         it.sourceArtifact?.mapToModel() == data.sourceArtifact &&
                         it.vcsInfoCurationData?.mapToModel() == data.vcs &&
-                        it.takeIf { it.hasAuthors }?.authors?.map { author -> author.name }?.toSet() == data.authors &&
+                        it.hasAuthors && it.authors.mapTo(mutableSetOf()) { author -> author.name } == data.authors &&
```
mnonnenmacher, line 71: This changes semantics if `it.hasAuthors == false`, previously this line could evaluate to `true` if `data.authors == null`.

dao/src/main/kotlin/Exceptions.kt
mnonnenmacher, line 1: Commit message nit: "Following" -> "Follow"

workers/common/src/main/kotlin/common/Extensions.kt
mnonnenmacher, line 1: Commit message: Why?

dao/src/main/kotlin/tables/ScanResultsTable.kt
```diff
@@ -91,20 +91,20 @@ class ScanResultDao(id: EntityID<Long>) : LongEntity(id) {
...
-                    type = RepositoryType.forName(vcsType!!),
-                    url = vcsUrl!!,
-                    revision = vcsRevision!!,
+                    type = RepositoryType.forName(vcsType.orEmpty()),
+                    url = checkNotNull(vcsUrl),
+                    revision = vcsRevision.orEmpty(),
                     path = ""
                 ),
-                resolvedRevision = vcsRevision!!
+                resolvedRevision = vcsRevision.orEmpty()
```
mnonnenmacher, line 103: Better use `checkNotNull` for all values, if one of them is null it is an indication of a bug and failing early helps with disovering it.

dao/src/main/kotlin/tables/ScanResultsTable.kt
```diff
@@ -21,8 +21,6 @@ package org.eclipse.apoapsis.ortserver.dao.tables
 
 import kotlinx.serialization.Serializable
```
mnonnenmacher, line 22: The DAO class should represent the database model completely, even if some properties are not yet used, therefore I would prefer to drop that commit.

mnonnenmacher (changes requested):

workers/common/src/main/kotlin/common/Extensions.kt
sschuberth: I've clarified in the commit message that this is necessary for *public* properties to avoid an inspection hint.

dao/src/main/kotlin/tables/runs/repository/PackageCurationDataTable.kt
sschuberth: Good catch, I missed that `data.authors` is nullable.

mnonnenmacher (approved):

PR #160, opened 2024-04-28
Add a GitHub workflow to build and publish Docker images

mnonnenmacher:

The workflow was tested in a fork: https://github.com/mnonnenmacher/ort-server
The resulting Docker images can be seen here (they are mixed with test artifacts, only those starting with "ort-server-" are relevant): https://github.com/mnonnenmacher?tab=packages

### DIFF
--- a/.github/actions/free-disk-space/action.yml
+++ b/.github/actions/free-disk-space/action.yml
@@ -1,0 +1,59 @@
+name: "Free Disk Space"
+description: "Remove unneeded preinstalled Docker images and software to free disk space."
+author: "The ORT Server Authors"
+
+runs:
+  using: "composite"
+
+  steps:
+    - name: Print Disk Space
+      shell: bash
+      run: df -h
+    - name: List Docker Images
+      if: ${{ false }} # Can be enabled if the 'Remove Unneeded Docker Images' step below needs to be updated.
+      shell: bash
+      run: docker images
+    - name: Remove Unneeded Docker Images
+      shell: bash
+      run: |
+        docker image rm \
+          node:16 \
+          node:16-alpine \
+          node:18 \
+          node:18-alpine \
+          node:20 \
+          node:20-alpine \
+          debian:10 \
+          debian:11 \
+          ubuntu:20.04 \
+          ubuntu:22.04
+    - name: Print Disk Space
+      shell: bash
+      run: df -h
+    - name: Get Size of Installed Tools
+      if: ${{ false }} # Can be enabled if the 'Remove Unneeded Tools' step below needs to be updated.
+      shell: bash
+      run: |
+        sudo du -hsc /usr/lib/*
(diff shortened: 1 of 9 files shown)


sschuberth:
FYI, I just realized your commits are not signed-off. Apparently they don't have to be... but should we still set up the DCO bot?

mnonnenmacher:
> FYI, I just realized your commits are not signed-off. Apparently they don't have to be... but should we still set up the DCO bot?

Interesting, I assume it's not required because the `eclipsefdn/eca` verifies that the contributor agreement was signed, but I have to read up on it.

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,132 @@
...
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+  build:
+    name: Build Docker Image
+    runs-on: ubuntu-22.04
+    permissions:
+      packages: write
+    strategy:
```
nnobelis, line 19: Some idea:
Use max-parallel to build some images in parallel ? 
Use fail-fast: false to build all images even if some of them fail ?

.github/actions/free-disk-space/action.yml
```diff
@@ -0,0 +1,59 @@
...
+
+runs:
+  using: "composite"
+
+  steps:
+    - name: Print Disk Space
+      shell: bash
+      run: df -h
+    - name: List Docker Images
+      if: ${{ false }} # Can be enabled if the 'Remove Unneeded Docker Images' step below needs to be updated.
```
nnobelis, line 13: What does this switch do ? Is is to disable the step ?

.github/actions/free-disk-space/action.yml
```diff
@@ -0,0 +1,59 @@
...
+      shell: bash
+      run: docker images
+    - name: Remove Unneeded Docker Images
+      shell: bash
+      run: |
```
nnobelis, line 18: Is clearing some of this an unoptimization ? If for instance ubuntu:22.04 is required for build as  base image, it now has to be downloaded everytime ?

.github/workflows/docker-build.yml
mnonnenmacher: > Use max-parallel to build some images in parallel ?

That's already the case, they are all built in parallel.

> Use fail-fast: false to build all images even if some of them fail ?

I think if one build fails there is not much use for the other images.  I would rather go in the opposite direction and prevent the publishing of all images if a single one cannot be built, but not in this PR.

.github/actions/free-disk-space/action.yml
mnonnenmacher: If we would use any of those images, yes, but I don't think we ever will.

bs-ondem (approved):

PR #162, opened 2024-04-29
Rework scheduling in Orchestrator

oheger-bosch:
This PR reworks the scheduling algorithm in Orchestrator by replacing hard-coded condition logic with a more declarative approach. This is then used to run the Reporter and Notifier workers always, even if other workers have failed before.


orchestrator/src/main/kotlin/WorkerScheduleInfo.kt
```diff
@@ -0,0 +1,222 @@
...
+ * A map allowing access to the [WorkerScheduleInfo] objects for the different worker endpoints.
+ */
+internal val scheduleInfos = mapOf(
+    AnalyzerEndpoint.configPrefix to analyzerWorkerScheduleInfo,
+    AdvisorEndpoint.configPrefix to advisorWorkerScheduleInfo,
+    ScannerEndpoint.configPrefix to scannerWorkerScheduleInfo,
+    EvaluatorEndpoint.configPrefix to evaluatorWorkerScheduleInfo,
+    ReporterEndpoint.configPrefix to reporterWorkerScheduleInfo,
+    NotifierEndpoint.configPrefix to notifierWorkerScheduleInfo
+)
```
mnonnenmacher, line 236: I find it difficult to understand that some of the functions in `WorkerScheduleInfo` depend on this top-level property which in turn holds instances of `WorkerScheduleInfo`. It seems to be a candidate for using a sealed class or an enum class, especially as the list of valid worker schedule infos is pretty stable.

For example, using an enum class would by design show that there is limited list of pre-defined schedule info, and accessing them by prefix could be done like this:
```kotlin
enum class WorkerScheduleInfo(val configPrefix: String) {
    ANALYZER("analyzer"),
    SCANNER("scanner");

    companion object  {
        private val entriesByPrefix = entries.associateBy { it.configPrefix }

        operator fun get(configPrefix: String) = entriesByPrefix[configPrefix]
    }
}

fun test() {
    WorkerScheduleInfo["analyzer"]
}
```

orchestrator/src/main/kotlin/WorkerScheduleContext.kt
```diff
@@ -53,7 +53,14 @@ internal class WorkerScheduleContext(
     /**
      * A map storing the jobs for the current run that have already been created. Keys are the endpoint names for jobs.
      */
-    val jobs: Map<String, WorkerJob>
+    val jobs: Map<String, WorkerJob>,
+
+    /**
+     * A flag indicating that the current [OrtRun] has failed. Per default, the failure state is determined by the
+     * jobs that have been run. With this flag, this mechanism can be overridden, which is necessary for workers that
+     * do not spawn jobs.
```
mnonnenmacher, line 61: I don't understand this explanation, under what circumstance would this be set to `true`?

orchestrator/src/main/kotlin/Orchestrator.kt
```diff
@@ -114,8 +112,7 @@ class Orchestrator(
         db.blockingQueryCatching(transactionIsolation = isolationLevel) {
             val ortRun = getCurrentOrtRun(configWorkerResult.ortRunId)
 
-            val context = WorkerScheduleContext(ortRun, workerRepositories, publisher, header, emptyMap())
-            context to listOf(createAnalyzerJob(ortRun).let { { scheduleAnalyzerJob(ortRun, it, header) } })
+            nextJobsToSchedule(ConfigEndpoint, ortRun.id, header, jobs = emptyMap())
```
mnonnenmacher, line 115: Not directly related to this change, but in my opinion we should also create jobs for the config worker to get rid of the requirement for the special handling. This would also allow to handle it consistently in other places, e.g. for showing that the config worker is currently running when the status of an ORT run is requested via API, or to also measure the activity of the config worker when exporting stats to Graphite.

orchestrator/src/main/kotlin/Orchestrator.kt
```diff
@@ -294,23 +291,21 @@ class Orchestrator(
     }
 
     /**
-     * Determine the next jobs that can be scheduled after the given [job][completedJob] for the given [endpoint] has
-     * completed. Use the given [header] to send messages to the worker endpoints. Return a list with the new jobs to
-     * schedule and the current [WorkerScheduleContext].
+     * Determine the next jobs that can be scheduled after a job for the given [endpoint] for the run with the given
+     * [ortRunId] has completed. Use the given [header] to send messages to the worker endpoints. Optionally,
+     * accept a map with the [jobs] that have been run Return a list with
```
mnonnenmacher, line 296: Dot missing after "run".

mnonnenmacher (changes requested):

orchestrator/src/main/kotlin/WorkerScheduleContext.kt
```diff
@@ -0,0 +1,109 @@
...
+     * Return a flag whether this [OrtRun] has failed, i.e. it has at least one job in failed state.
+     */
+    fun isFailed(): Boolean =
+        jobs.values.any { it.status == JobStatus.FAILED }
+}
+
+/**
+ * Return a flag whether this [WorkerJob] has already run.
+ */
+private fun WorkerJob.hasRun(): Boolean = status == JobStatus.FINISHED || status == JobStatus.FAILED
```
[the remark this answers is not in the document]
oheger-bosch: Renamed `hasRun()` to `isCompleted()` and dropped `isRunning()` as it can be trivially expressed in terms of `isCompleted()`.

orchestrator/src/main/kotlin/WorkerScheduleContext.kt
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Return a flag whether this [WorkerJob] is currently running.
+ */
+private fun WorkerJob.isRunning(): Boolean = status != JobStatus.FINISHED && status != JobStatus.FAILED
```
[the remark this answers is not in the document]
oheger-bosch: Dropped this function.

orchestrator/src/main/kotlin/WorkerScheduleContext.kt
```diff
@@ -0,0 +1,109 @@
...
+        val headerWithProperties = header.copy(transportProperties = ortRun.labels.selectByPrefix("transport"))
+
+        publisher.publish(to = endpoint, message = Message(headerWithProperties, payload))
+    }
+
+    /**
+     * Return a flag whether the current [OrtRun] has at least one running job.
+     */
+    fun hasRunningJobs(): Boolean =
+        jobs.values.any { it.isRunning() }
```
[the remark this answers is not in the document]
oheger-bosch: Renamed `hasJobRun()` to `isJobCompleted()`, but regarding `hasRunningJobs()` I would rather stick to this name. The meaning is here slightly different: "running" means "currently running", while "not completed" could also mean "not yet scheduled".

orchestrator/src/main/kotlin/WorkerScheduleContext.kt
oheger-bosch: The comment has been extended to mention the Config worker which is currently the only example.

orchestrator/src/main/kotlin/Orchestrator.kt
oheger-bosch: I would rather defer this for now; this can be considered together with other changes on the DB in the context of deleting ORT runs.

orchestrator/src/main/kotlin/WorkerScheduleInfo.kt
oheger-bosch: Good suggestion! I have added another commit which converts `WorkerScheduleInfo` to an enum class.

orchestrator/src/main/kotlin/Orchestrator.kt
mnonnenmacher: Yes, I didn't mean to change it now, just wanted to express the idea.

mnonnenmacher (approved):

PR #166, opened 2024-04-29
Add convenience tasks to build images

sschuberth:
I plan to improve docs further with upcoming PRs that also pass Docker build args to the Gradle tasks, but I thought this could be useful already.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,9 @@ Then build the base images for the workers which contain the external tools and 
 ```shell
 # Build all worker images at once:
 ./gradlew buildAllWorkerImages
+
+# Build all worker images at once with custom build arguments:
+./gradlew -PdockerBaseBuildArgs="TEMURIN_VERSION=11" buildAllWorkerImages
 ```
 
 *or* manually by running
--- a/README.md
+++ b/README.md
@@ -51,20 +51,6 @@ Then the Docker images can be built with [Jib](https://github.com/GoogleContaine
 ./gradlew :core:jibDockerBuild
 ```
 
-The `jibDockerBuild` task occasionally gets stuck while pulling the Eclipse Temurin base image. In this case it usually
-helps to clean the project and stop the Gradle Daemon:
-
-```shell
-./gradlew clean
-./gradlew --stop
-```
-
-When building multiple images at once it can also help to disable parallel builds:
-
-```shell
-./gradlew --no-parallel jibDockerBuild
-```
-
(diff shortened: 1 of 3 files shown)


sschuberth:
> also pass Docker build args to the Gradle tasks

Unfortunately, that's [not possible](https://docs.gradle.org/current/userguide/custom_tasks.html#limitations) purely from Kotlin DSL:

> Command line options can only be declared for custom tasks via annotation. There’s no programmatic equivalent for defining options.

So we could
- introduce a project property for Docker build args,
- live with the limitation that the convenience task cannot customize the build args.

README.md
```diff
@@ -57,20 +57,6 @@ Then the Docker images can be built with [Jib](https://github.com/GoogleContaine
...
-```shell
-./gradlew clean
-./gradlew --stop
-```
-
-When building multiple images at once it can also help to disable parallel builds:
-
-```shell
-./gradlew --no-parallel jibDockerBuild
-```
```
mnonnenmacher, line 66: Could you verify that this issue is gone on a machine were the task previously got stuck without `--no-parallel`?

mnonnenmacher:
> > also pass Docker build args to the Gradle tasks
> 
> Unfortunately, that's [not possible](https://docs.gradle.org/current/userguide/custom_tasks.html#limitations) purely from Kotlin DSL:
> 
> > Command line options can only be declared for custom tasks via annotation. There’s no programmatic equivalent for defining options.
> 
> So we could
> 
>     * introduce a project property for Docker build args,
> 
>     * live with the limitation that the convenience task cannot customize the build args.

Or implement a custom taks in `buildSrc`. But for now I think we can live with that limitation and address it once it becomes a requirement.

README.md
sschuberth: Yes, I was running into the issue before (the ORT server's `gradle.properties` has `org.gradle.parallel = true`), and could not reproduce it anymore afterwards. But of course, I could have just been lucky, and any verification is welcome.

sschuberth:
> * introduce a project property for Docker build args,

I did that now.

README.md
mnonnenmacher: I was mainly curious, because I haven't seen this mentioned in any of discussions about this issue that I found, but maybe I just missed it. I will test it on my computer at least before the merge, and if we still see this issue later we can recover those README hints from the Git history.

mnonnenmacher: At least on my machine I could not reproduce the issue with this change.

mnonnenmacher (approved):

PR #167, opened 2024-04-29
Docker chores

sschuberth:


.reuse/dep5
mnonnenmacher, line 1: Commit message: Is there any documentation why that is the case?

mnonnenmacher (changes requested):

.reuse/dep5
sschuberth: I'm not aware of any official documentation. It's just that I found it to be weird to have the `Dockerfile` as part of the context directory, as any change to the context would invalidate the Docker cache. On the other hand, if the `Dockerfile` changes, a rebuild is triggered probably anyway.

I remember having some edge-cases during development in ORT core where I did not want a `Dockerfile` change to invalidate the context, which is why we have [this](https://github.com/oss-review-toolkit/ort/blob/8f945e9f97104fffb6c0bdda9b18d352c2417907/.dockerignore#L1-L4), but I can also drop this commit if you prefer.

For the fun of it, ChatGPT says:

> Is it a problem if the Dockerfile itself is part of the Docker context?

> No, it's not a problem if the Dockerfile itself is part of the Docker context. In fact, it's quite common practice to have the Dockerfile in the same directory as the application code and include it in the Docker context. When you build a Docker image, Docker uses the context directory (along with the Dockerfile) to build the image. So having the Dockerfile in the context directory is expected and perfectly fine.

mnonnenmacher: I believe that is not required, like ChatGPT says it's common to have the Dockerfile in the context directory, and if this would break the build cache it would not have worked before this change, but I am pretty sure it does.

sschuberth: So, I guess that means to drop the commit?

sschuberth: I've dropped the commit.

mnonnenmacher: I just did a simple test locally and could verify that the cache works fine as is.

mnonnenmacher (approved):

PR #169, opened 2024-04-30
ui: Configure Server UI with environment variables

Etsija:

### DIFF
--- a/ui/src/components/providers.tsx
+++ b/ui/src/components/providers.tsx
@@ -22,11 +22,14 @@ import { QueryClientProvider } from '@tanstack/react-query';
 import { ReactNode } from 'react';
 import { AuthProvider, AuthProviderProps } from 'react-oidc-context';
 import { Toaster } from "@/components/ui/toaster";
+import { OpenAPI as OpenAPIConfig } from '@/api/requests/core/OpenAPI';
+
+OpenAPIConfig.BASE = import.meta.env.VITE_API_URL || 'http://localhost:8080';
 
 const oidcConfig = {
-  authority: 'http://localhost:8081/realms/master',
-  redirect_uri: 'http://localhost:5173/',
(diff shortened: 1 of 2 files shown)


ui/src/components/providers.tsx
```diff
@@ -22,11 +22,14 @@ import { QueryClientProvider } from '@tanstack/react-query';
...
+
+OpenAPIConfig.BASE = import.meta.env.VITE_API_URL || 'http://localhost:8080';
 
 const oidcConfig = {
-  authority: 'http://localhost:8081/realms/master',
-  redirect_uri: 'http://localhost:5173/',
-  client_id: 'react',
+  authority: import.meta.env.VITE_AUTHORITY || 'http://localhost:8081/realms/master',
+  redirect_uri: import.meta.env.VITE_WEBAPP_URL || 'http://localhost:5173/',
+  client_id: import.meta.env.VITE_CLIENT_ID || 'react',
```
sschuberth, line 32: Why are these prefixed with `VITE_`, are these really Vite-specific environment variables?

Etsija: [Env Variables and Modes](https://vitejs.dev/guide/env-and-mode)

"To prevent accidentally leaking env variables to the client, only variables prefixed with VITE_ are exposed to your Vite-processed code..."

Etsija: I also confirmed this: without the `VITE_` prefix, the environment variables are not exposed to the code.

sschuberth: Good reference, could you please add that to the commit message?

Etsija: Yes, sorry, have been busy digging out how ORT Server (back-end) is configured. Will do this shortly.

mnonnenmacher (approved):

PR #171, opened 2024-04-30
refactor(cors): Disallow anyHost from CORS

Etsija:

### DIFF
--- a/core/src/main/resources/application.conf
+++ b/core/src/main/resources/application.conf
@@ -28,6 +28,11 @@ ktor {
     watch = [ classes ]
   }
 
+  cors {
+    allowedHost = "localhost:5173"
+    allowedHost = ${?UI_HOST}
+  }
+
   application {
     modules = [ org.eclipse.apoapsis.ortserver.core.ApplicationKt.module ]
   }
--- a/core/src/test/resources/application-test-auth.conf
+++ b/core/src/test/resources/application-test-auth.conf
@@ -22,6 +22,10 @@ ktor {
     watch = [ classes ]
   }
 
+  cors {
+    allowedHost = "localhost:5173"
+  }
+  
   application {
     modules = [org.eclipse.apoapsis.ortserver.core.ApplicationTestKt.testAuthModule]
   }
--- a/core/src/test/resources/application-test.conf
+++ b/core/src/test/resources/application-test.conf
@@ -22,6 +22,10 @@ ktor {
     watch = [ classes ]
   }
 
+  cors {
+    allowedHost = "localhost:5173"
+  }
+  
   application {
     modules = [org.eclipse.apoapsis.ortserver.core.ApplicationTestKt.testModule]
   }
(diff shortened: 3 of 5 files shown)


core/src/main/resources/application.conf
```diff
@@ -28,6 +28,11 @@ ktor {
     watch = [ classes ]
   }
 
+  cors {
+    allowedHost = "localhost:5173"
+    allowedHost = ${?WEBAPP_HOST}
```
mnonnenmacher, line 33: I would prefer to call this `UI_HOST` to use the same name as in the directory structure, and to avoid any confusion with the `WebAppReporter` which is where the name "web app" is mostly used in the context of ORT.

core/src/test/resources/application-test-auth.conf
```diff
@@ -22,6 +22,11 @@ ktor {
     watch = [ classes ]
   }
 
+  cors {
+    allowedHost = "localhost:5173"
+    allowedHost = ${?WEBAPP_HOST}
```
mnonnenmacher, line 27: For the application test configs only the hard coded value is required.

core/src/test/resources/application-test.conf
```diff
@@ -22,6 +22,11 @@ ktor {
     watch = [ classes ]
   }
 
+  cors {
+    allowedHost = "localhost:5173"
+    allowedHost = ${?WEBAPP_HOST}
```
mnonnenmacher, line 27: For the application test configs only the hard coded value is required.

mnonnenmacher (changes requested):

Etsija:
Thabks @mnonnenmacher all your comments are very relevant. I'll deal with them in two days if that's ok for you - tomorrow's a Work Day holiday in Finland.

mnonnenmacher:
> Thabks @mnonnenmacher all your comments are very relevant. I'll deal with them in two days if that's ok for you - tomorrow's a Work Day holiday in Finland.

Sure, same here.

sschuberth:
> Instead of allowing any host, configure a specific UI_HOST for
> CORS [1].  If the UI_HOST parameter is not set, a sensible default
> value is used.

The commit message does not really explain *why* it makes sense to do so. maybe just end the first sentence with "... for security reasons" or so? Also a nit: Double-space after the dot of the first sentence.

Etsija:
> > Instead of allowing any host, configure a specific UI_HOST for
> > CORS [1].  If the UI_HOST parameter is not set, a sensible default
> > value is used.
> 
> The commit message does not really explain _why_ it makes sense to do so. maybe just end the first sentence with "... for security reasons" or so? Also a nit: Double-space after the dot of the first sentence.

For the nit comment: This wasn't a mistake. I am an old-school guy, and back at my time, every sentence end was followed by double-space -  that was "the standard" in Finland. Will try to remember the new rules from now on. :)

sschuberth:
> This wasn't a mistake.

Yeah, I'm aware that such a style exists, I just didn't think you'd follow it 😉 Anyway, let's keep it consistent with existing commit message and use only one space.

mnonnenmacher (approved):

PR #175, opened 2024-05-02
Add configuration options to Compose

mnonnenmacher:
Make the Docker images and the API port configurable for Docker Compose. Please see the commit messages for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,15 @@ To use a different tag, you can set the `ORT_SERVER_IMAGE_TAG` environment varia
 ORT_SERVER_IMAGE_TAG=0.1.0-SNAPSHOT-001.sha.aa4d3fa docker compose up
 ```
 
+By default, the ORT Server API is exposed on port 8080.
+If this port is already in use, it can be changed using the `ORT_SERVER_CORE_PORT` environment variable:
+
+```shell
+ORT_SERVER_CORE_PORT=8090 docker compose up
+```
+
+When using a different port, please make sure that it is not used by [another service](#accessing-the-services).
+
 #### Running with local images
 
 During development, it is useful to run the ORT Server with locally built Docker images.
--- a/README.md
+++ b/README.md
@@ -8,10 +8,36 @@ The ORT server is a standalone application to deploy the
 > Eclipse Foundation and working towards making the first release.
 > Once released, the project will use semantic versioning, until then breaking changes can occur at any time. 
 
-## Local Setup
+## Running ORT Server
 
-To start the ORT server with the required 3rd party services, you can use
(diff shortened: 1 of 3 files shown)


sschuberth (approved):

docker-compose.yml
```diff
@@ -112,7 +112,7 @@ services:
       start_period: 60s
 
   core:
-    image: ort-server-core
+    image: ${ORT_SERVER_IMAGE_PREFIX-ghcr.io/eclipse-apoapsis/}ort-server-core:${ORT_SERVER_IMAGE_TAG:-main}
```
mmurto, line 115: Didn't try to run it, but should this and others include the ":"?

```suggestion
    image: ${ORT_SERVER_IMAGE_PREFIX:-ghcr.io/eclipse-apoapsis/}ort-server-core:${ORT_SERVER_IMAGE_TAG:-main}
```

mnonnenmacher: Actually, this was on purpose: `:-` replaces the variable if it is not set or null (or empty). `-` only replaces it if it is not set. With `:-` it would not be possible to set the prefix to an empty value when using locally built images, as it would then always fall back to the default.

mnonnenmacher: I found the solution in [this Stackoverflow answer](https://unix.stackexchange.com/a/122848/89805) which has a nice table showing the different substitution options at the end. And while it might work, I have to add that I don't like that syntax, as obviously not many people are aware of it and the `-` can also appear in the default value which can be confusing.

README.md
```diff
@@ -145,9 +169,10 @@ docker compose -f docker-compose.yml -f docker-compose-maintenance.yml up flyway
 ## Publish Docker Images
 
 To publish the Docker images to a registry, first build the worker base images as described in
-[Local Setup](#local-setup). Then you can use the `jib` task to publish the images by setting the correct prefix for the
-registry. You can also configure the tag which defaults to `latest`. When publishing multiple images at once it is
-recommended to disable parallel builds.
+[Running with local images](#running-with-local-images).
+Then you can use the `jib` task to publish the images by setting the correct prefix for the registry.
+You can also configure the tag which defaults to `latest`.
+When publishing multiple images at once, it is recommended to disable parallel builds.
```
nnobelis, line 184: Maybe write a line to explain how to disable parallel builds ?

nnobelis (dismissed):

README.md
mnonnenmacher: I have removed the sentence instead, see the commit message.

nnobelis: Ah ok. Anyway I missed the lines underneath with `no-parallel`. :see_no_evil:

nnobelis (approved):

Issue #174, opened 2024-05-02
Add proper headers to endpoints with file downloads


mmurto:
Some API endpoints let users to download files, so proper headers should be set to enable download prompts in browsers. An example for how to do so can be found here: https://ktor.io/docs/server-responses.html#file

One example of such an endpoint is the report download endpoint:

https://github.com/eclipse-apoapsis/ort-server/blob/f0dada51693869a0bea6cf6fe3c3e1ffac7b55b0/core/src/main/kotlin/api/RunsRoute.kt#L71-L73

Some additional information can be found in [this Stack Overflow answer](https://stackoverflow.com/a/20509354).

sschuberth:
This was fixed by https://github.com/eclipse-apoapsis/ort-server/pull/176, but again auto-close did not work.

mnonnenmacher:
> This was fixed by #176, but again auto-close did not work.

I have created #868 to collect such cases because I still have no idea why that happens.

PR #176, opened 2024-05-02
Set Content-Disposition header when downloading reports

mnonnenmacher:
See the commit messages for details and the related issue #174.


ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -89,7 +89,15 @@ const CreateRunPage = () => {
       repositoryId: Number.parseInt(params.repoId),
       requestBody: {
         revision: values.revision,
-        jobConfigs: {},
+        jobConfigs: {
```
sschuberth, line 92: Commit message typo in "execpt".

mnonnenmacher:
@mmurto Can you please try if this fixes #174?

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
sschuberth: @Etsija, could running esp. the scanner by default be a problem for us, e.g. performance-wise?

services/report-storage/src/test/kotlin/ReportStorageServiceTest.kt
```diff
@@ -51,7 +51,11 @@ class ReportStorageServiceTest : WordSpec({
 
             val storage = mockk<Storage>()
             every { storage.containsKey(key) } returns true
-            every { storage.read(key) } returns StorageEntry.create(ByteArrayInputStream(reportData), contentType)
+            every { storage.read(key) } returns StorageEntry.create(
```
sschuberth, line 54: Commit message: Double space after "header".

storage/spi/src/main/kotlin/StorageEntry.kt
```diff
@@ -39,21 +39,24 @@ data class StorageEntry(
     val data: InputStream,
 
     /** The content type associated with this entry if any. */
-    val contentType: String?
+    val contentType: String?,
```
sschuberth, line 42: Unrelated to this commit, but now that I'm seeing this, could the meaning of a `null` `contentType` be documented here?

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
mnonnenmacher: I added that commit mainly for testing, but in the end I actually didn't find the time to test it. I could also drop it, I assume you will improve the form to trigger an ORT run anyway and make it configurable via UI which workers are executed?

storage/spi/src/main/kotlin/StorageEntry.kt
mnonnenmacher: I assume that simply means that the content type is unknown which has to be supported as the ORT reporter API does not force reporter implementations to provide the content type of the generated reports. But maybe @oheger-bosch could provide more details here.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
Etsija: Yes, I'd much rather keep the ORT run fully configurable in the UI, and am currently implementing the form to do so. 

One suggestion for improvement: I can see that currently, not running eg. scanner worker is triggered by not passing any configuration object to it in jobConfigs. This is a bit tricky in the UI sense, as I'm envisioning there would be checkboxes for all workers, and by checking a worker -> it is added to the ORT run.

My suggestion is that in jobConfig, there would be a boolean variable for each worker (default: false), which would determine whether that worker is included in the run or not.  This way, the user can in the UI still configure all workers, but just choose some of them to be run as part of ORT run. This would greatly help in debugging.

services/report-storage/src/test/kotlin/ReportStorageServiceTest.kt
Etsija: jobConfigs would then look something like:
```
{
  "analyzer": {
    "includeInRun": true,
    "allowDynamicVersions": true,
    "packageCurationProviders": [],
    "skipExcluded": true
  },
  "scanner": {
    "includeInRun": false,
    "skipExcluded": true
  },
  "reporter": {
    "includeInRun": true,
    "formats": [
      "ortresult"
    ],
    "packageConfigurationProviders": [],
    "assetFiles": [],
    "assetDirectories": []
  },
  "parameters": {}
}
```

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
Etsija: jobConfigs would then look something like:
```
{
  "analyzer": {
    "includeInRun": true,
    "allowDynamicVersions": true,
    "packageCurationProviders": [],
    "skipExcluded": true
  },
  "scanner": {
    "includeInRun": false,
    "skipExcluded": true
  },
  "reporter": {
    "includeInRun": true,
    "formats": [
      "ortresult"
    ],
    "packageConfigurationProviders": [],
    "assetFiles": [],
    "assetDirectories": []
  },
  "parameters": {}
}
```

Etsija: But all in all, I would drop this commit from the PR, as I am currently editing exactly the same file and putting in the logic to fill out those jobConfigs for different phases.

mnonnenmacher: > One suggestion for improvement: I can see that currently, not running eg. scanner worker is triggered by not passing any configuration object to it in jobConfigs. This is a bit tricky in the UI sense, as I'm envisioning there would be checkboxes for all workers, and by checking a worker -> it is added to the ORT run. With current implementation I need to ensure that when a worker is excluded from the run, the corresponding part of jobConfigs is completely deleted.

Please feel free to create an issue for that request, but I'm not sure if we want to change that. The current approach might be inconvenient in the frontend code, but it was no problem to handle that in our internal parameter validation script and it was convenient when manually creating requests. But anyway, we will make an overall review of the API also with your input, so it's helpful to create tickets for everything you would like to have changed.

> But all in all, I would drop this commit from the PR, as I am currently editing exactly the same file and putting in the logic to fill out those jobConfigs for different phases.

Sure, I will drop it, but I'll wait with editing this PR until someone can confirm that it fixes #174.

Etsija: I'm trying to verify the PR, however have some difficulties actually saving the file; back-end logs show that the request goes through OK.

mnonnenmacher: I have verified that the download at least works correctly with a REST client, I will look into the cookie topic @mmurto  mentioned below.

nnobelis (approved):

PR #179, opened 2024-05-03
feat(core): Filter organizations based on permissions

mnonnenmacher:
Change the `/organizations` endpoint to only require authentication instead of the superuser role, and to return only those organizations for which the user has `OrganizationPermission.READ`.

As the permissions are managed by Keycloak, the database does not know about them. Therefore the implementation fetches the whole list of organizations from the database and filters them afterward. For a very long list of organizations this might not perform well, so if required, a different solution has to be found later.

### DIFF
--- a/core/src/main/kotlin/api/OrganizationsRoute.kt
+++ b/core/src/main/kotlin/api/OrganizationsRoute.kt
@@ -59,8 +59,10 @@ import org.eclipse.apoapsis.ortserver.core.apiDocs.postInfrastructureServiceForO
 import org.eclipse.apoapsis.ortserver.core.apiDocs.postOrganizations
 import org.eclipse.apoapsis.ortserver.core.apiDocs.postProduct
 import org.eclipse.apoapsis.ortserver.core.apiDocs.postSecretForOrganization
+import org.eclipse.apoapsis.ortserver.core.authorization.hasPermission
 import org.eclipse.apoapsis.ortserver.core.authorization.requirePermission
 import org.eclipse.apoapsis.ortserver.core.authorization.requireSuperuser
+import org.eclipse.apoapsis.ortserver.core.utils.paginate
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/api/OrganizationsRoute.kt
```diff
@@ -78,15 +80,15 @@ fun Route.organizations() = route("organizations") {
...
         val pagingOptions = call.pagingOptions(SortProperty("name", SortDirection.ASCENDING))
 
-        val organizations = organizationService.listOrganizations(pagingOptions.mapToModel())
-        val pagedResponse = PagedResponse(
-            organizations.map { it.mapToApi() },
-            pagingOptions
-        )
+        val organizations = organizationService
+            .listOrganizations(pagingOptions.copy(limit = null, offset = null).mapToModel())
+            .filter { hasPermission(it.id, OrganizationPermission.READ) }
```
bs-ondem, line 87: Just to clarify: If a user is in the writers group, will they also have the role to read? I couldn't find that in the code quickly.

bs-ondem: Okay found it already by myself. The writers group has the write and read permission roles.

bs-ondem (approved):

core/src/main/kotlin/api/OrganizationsRoute.kt
```diff
@@ -78,15 +80,15 @@ fun Route.organizations() = route("organizations") {
...
+            .paginate(pagingOptions)
```
oheger-bosch, line 88: Will this pagination in memory work in all cases? I am not sure how the database behaves if no sort order is specified; but there is probably no guarantee that the order of results is always the same? Maybe, to be safe, always a sort order needs to be specified?

PR #180, opened 2024-05-03
feat(compose): Auto-configure Keycloak in Docker Compose for UI

mmurto:
While in longer term this should probably be done with Terraform/Pulumi, configure it with the JSON to ease development.

### DIFF
--- a/scripts/docker/keycloak/master-realm.json
+++ b/scripts/docker/keycloak/master-realm.json
@@ -84,7 +84,7 @@
       "composites" : {
         "realm" : [ "create-realm" ],
         "client" : {
-          "master-realm" : [ "impersonation", "manage-events", "manage-users", "manage-identity-providers", "view-realm", "query-users", "view-clients", "view-authorization", "manage-realm", "query-realms", "manage-clients", "view-identity-providers", "query-clients", "view-users", "create-client", "view-events", "query-groups", "manage-authorization" ]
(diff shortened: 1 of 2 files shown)


mmurto:
Only real changes are the client configuration and adding admin to the SUPERUSER group. The file was generated as instructed below, which also added some other changes.

https://github.com/eclipse-apoapsis/ort-server/blob/88204b7a1a58eb88494bbe8edaa8f219bed3a60a/scripts/docker/keycloak/init-keycloak.sh#L17-L23

scripts/docker/keycloak/master-realm.json
```diff
@@ -419,7 +420,7 @@
     "requiredActions" : [ ],
     "realmRoles" : [ "default-roles-master", "admin" ],
     "notBefore" : 0,
-    "groups" : [ ]
+    "groups" : [ "/SUPERUSERS" ]
```
mnonnenmacher, line 423: Is this required? There is already the ort-admin user which is in the superuser group and which is supposed to be used for testing. On the other hand, I guess it doesn't hurt if admin in this group as well.

mmurto: It's a little bit less friction via the UI, Keycloak and the UI share the session. If you want to change something in Keycloak and see its effect in the UI, you need to change the users in between if using the same browser.

So not required, but if there are no strong preferences from you, I'd like to keep it.

mnonnenmacher: Thanks, that makes sense.

mnonnenmacher (approved):

PR #181, opened 2024-05-03
fix(core): Fix OpenAPI representation of nullable properties

### LINKED ISSUE #165: Body parameters in OpenAPI Spec are incorrect
The generated OpenAPI spec for the endpoints contains incorrect parameter types. As an example, [`postOrtRun`](https://github.com/eclipse-apoapsis/ort-server/blob/2a9001071bfe6a0e9ad327abcffe75bf3ece03ec/core/src/main/kotlin/apiDocs/RepositoriesDocs.kt#L388-L416) should take [`CreateOrtRun`](https://github.com/eclipse-apoapsis/ort-server/blob/2a9001071bfe6a0e9ad327abcffe75bf3ece03ec/api/v1/model/src/commonMain/kotlin/OrtRun.kt#L110-L140) as JSON in the body. `CreateOrtRun` includes `AnalyzerJobConfiguration`:

https://github.com/eclipse-apoapsis/ort-server/blob/2a9001071bfe6a0e9ad327abcffe75bf3ece03ec/api/v1/model/src/commonMain/kotlin/JobConfigurations.kt#L92-L95

In the generated OpenAPI spec (JSON attached, line 2863) this is represented as object:

```json
"skipExcluded" : {
    "type" : "object"
}
```

The same seems to hold for [`repositoryConfigPath`](https://github.com/eclipse-apoapsis/ort-server/blob/2a9001071bfe6a0e9ad327abcffe75bf3ece03ec/api/v1/model/src/commonMain/kotlin/JobConfigurations.kt#L86-L90), which is a `String?`, so I wonder if the nullability makes these to be defined as an object instead of an optional value?

mnonnenmacher:
When using kotlinx.serialization with ktor-swagger-ui, the json-schema-serialization [1] library is used for generating JSON schemas [2].

For nullable values, this library generates invalid schemas which causes issues with consumer of the generated OpenAPI specification. Introduce a post-processing step to fix the schema.

Both the original json-schema-serialization library [3] and the used fork [1] are inactive, so we should not expect a fix for this issue.

A similar issue was fixed in ktor-swagger-ui [4], but the fix does not apply when using json-schema-serialization.

Fixes #165.

[1]: https://github.com/tillersystems/json-schema-serialization
[2]: https://github.com/SMILEY4/ktor-swagger-ui/wiki/Kotlinx-Multiplatform-Encoder-Example
[3]: https://github.com/Ricky12Awesome/json-schema-serialization
[4]: https://github.com/SMILEY4/ktor-swagger-ui/issues/53


mnonnenmacher:
After the change the API specs look like this:
[api.json](https://github.com/eclipse-apoapsis/ort-server/files/15202605/api.json)

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -119,3 +126,58 @@ fun Application.configureOpenApi() {
...
+ * }
+ * ```
+ */
+private fun JsonElement.replaceIfWithType(): JsonElement {
+    if (this !is JsonObject) {
+        return this
+    }
+
+    val ifElseTypes = getIfElseTypes()
+    val map = toMutableMap()
```
oheger-bosch, line 159: Nit: Maybe name this variable `properties` to make it clearer that it contains the data of the current `JsonObject`.

oheger-bosch (approved):

PR #196, opened 2024-05-07
Fix a dependency conflict

oheger-bosch:
This PR provides a work-around for the version conflict of the SnakeYaml Engine library between jRuby and Kaml (see also https://github.com/eclipse-apoapsis/ort-server/pull/129). It introduces a new utils module for YAML parsing that uses the Gradle shadow plugin to relocate the problematic packages. This is of course not a complete solution, but it fixes the problem for now; with this fix, it is again possible to analyze projects using Ruby.


sschuberth:
Just to note: As indicated [here](https://github.com/eclipse-apoapsis/ort-server/pull/129#issuecomment-2071835725) already, IMO this work-around does not scale, as we could easily run into the same problem for other dependencies as well. The "real" fix at the root would be to apply the Shadow Gradle plugin for ORT plugins to solve https://github.com/oss-review-toolkit/ort/issues/6798, not to apply the Shadow plugin here.

mnonnenmacher:
> > Just to note: As indicated [here](https://github.com/eclipse-apoapsis/ort-server/pull/129#issuecomment-2071835725) already, IMO this work-around does not scale, as we could easily run into the same problem for other dependencies as well. The "real" fix at the root would be to apply the Shadow Gradle plugin for ORT plugins to solve [oss-review-toolkit/ort#6798](https://github.com/oss-review-toolkit/ort/issues/6798), not to apply the Shadow plugin here.
> 
> Yes, I agree; I see this as a temporary work-around.
> 
> Applying the Shadow plugin to the ORT plugins would certainly be an option, but this would probably increase the size of the plugin jars and the container images significantly. It would be cool to have some classloader magic, in a similar way how application servers are able to isolate the applications deployed on them from each other.

That's very useful information in both comments, could you add that to a README.md file in the new Gradle module?

oheger-bosch:
> That's very useful information in both comments, could you add that to a README.md file in the new Gradle module?

Added a README file in a new commit that hopefully captures the essence of the discussion.

mnonnenmacher (approved):

PR #197, opened 2024-05-07
Logging improvements

oheger-bosch:
Refer to the single commits for further details.

### DIFF
--- a/transport/kubernetes/src/main/kotlin/KubernetesSenderConfig.kt
+++ b/transport/kubernetes/src/main/kotlin/KubernetesSenderConfig.kt
@@ -335,6 +335,8 @@ data class KubernetesSenderConfig(
          * annotations from them. Ignore non-existing or invalid variables, but log them.
          */
         private fun createAnnotations(variableNames: String): Map<String, String> {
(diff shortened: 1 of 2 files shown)


transport/kubernetes/src/main/kotlin/KubernetesSenderConfig.kt
```diff
@@ -335,9 +335,11 @@ data class KubernetesSenderConfig(
          * annotations from them. Ignore non-existing or invalid variables, but log them.
          */
         private fun createAnnotations(variableNames: String): Map<String, String> {
-            val (annotations, invalid) = variableNames.split(splitAnnotationVariablesRegex)
-                .map { variable -> variable to (System.getenv(variable).orEmpty()) }
-                .partition { KEY_VALUE_SEPARATOR in it.second }
+            val (annotations, invalid) = variableNames.takeUnless { it.isEmpty() }
```
sschuberth, line 338: How about instead returning early in that case, like

     variableNames.ifEmpty { return emptyMap() }

and leaving the rest of the code as-is?

mnonnenmacher (approved):

PR #200, opened 2024-05-07
feat(reporter): Add id of current ORT run as label to ORT results

wkl3nk:
Add the id of the current ORT run as label to the ORT result in order to have it available during report and notification generation.

### DIFF
--- a/workers/common/src/main/kotlin/common/OrtRunService.kt
+++ b/workers/common/src/main/kotlin/common/OrtRunService.kt
@@ -62,6 +62,7 @@ import org.eclipse.apoapsis.ortserver.model.util.asPresent
 import org.jetbrains.exposed.sql.Database
 import org.jetbrains.exposed.sql.insert
 
+import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.Repository
 import org.ossreviewtoolkit.model.ResolvedPackageCurations
 import org.ossreviewtoolkit.model.config.PackageConfiguration
@@ -88,6 +89,10 @@ class OrtRunService(
     private val scannerJobRepository: ScannerJobRepository,
     private val scannerRunRepository: ScannerRunRepository
 ) {
+    companion object {
+        private const val RUN_ID_LABEL: String = "runId"
+    }
+
     /**
      * Create an empty [ScannerRun]. This function is supposed to be called before the ORT scanner is invoked, so that
      * data can be associated to the scanner run while the ORT scanner is running.
@@ -274,6 +279,39 @@ class OrtRunService(
      */
     fun getScannerRunForOrtRun(ortRunId: Long) = db.blockingQuery {
         getScannerJobForOrtRun(ortRunId)?.let { scannerRunRepository.getByJobId(it.id) }
+    }
+
+    /**
+     * Load the results of the previous worker steps and generate an [OrtResult] from them. In addition,
+     * add a number of common labels that can be evaluated by different types of workers to obtain further information
+     * about the run.
+     *
+     * If [failIfRepoInfoMissing] is *true*, throw an [IllegalArgumentException] if the repository information
+     * is missing; otherwise, return an empty [Repository] object.
+     */
+    fun generateOrtResult(ortRun: OrtRun, failIfRepoInfoMissing: Boolean = true): OrtResult {
+        val repository = getOrtRepositoryInformation(ortRun, failIfMissing = failIfRepoInfoMissing)
+        val resolvedConfiguration = getResolvedConfiguration(ortRun)
+        val analyzerRun = getAnalyzerRunForOrtRun(ortRun.id)
+        val advisorRun = getAdvisorRunForOrtRun(ortRun.id)
+        val scannerRun = getScannerRunForOrtRun(ortRun.id)
+        val evaluatorRun = getEvaluatorRunForOrtRun(ortRun.id)
+
+        val baseResult = ortRun.mapToOrt(
+            repository = repository,
+            analyzerRun = analyzerRun?.mapToOrt(),
+            advisorRun = advisorRun?.mapToOrt(),
+            scannerRun = scannerRun?.mapToOrt(),
+            evaluatorRun = evaluatorRun?.mapToOrt(),
+            resolvedConfiguration = resolvedConfiguration.mapToOrt()
+        )
+
+        return baseResult.copy(
+            // Add common labels for all types of workers
+            labels = baseResult.labels + mapOf(
+                RUN_ID_LABEL to ortRun.id.toString()
+            )
+        )
     }
 
     /**
--- a/workers/notifier/src/main/kotlin/notifier/NotifierOrtResultGenerator.kt
+++ b/workers/notifier/src/main/kotlin/notifier/NotifierOrtResultGenerator.kt
@@ -30,7 +30,6 @@ import org.eclipse.apoapsis.ortserver.transport.EvaluatorEndpoint
 import org.eclipse.apoapsis.ortserver.transport.ReporterEndpoint
 import org.eclipse.apoapsis.ortserver.transport.ScannerEndpoint
 import org.eclipse.apoapsis.ortserver.workers.common.OrtRunService
-import org.eclipse.apoapsis.ortserver.workers.common.mapToOrt
 
 import org.ossreviewtoolkit.model.OrtResult
 
@@ -47,6 +46,7 @@ import org.ossreviewtoolkit.model.OrtResult
  * - &lt;worker&gt;JobStatus: Labels of this category allow finding out which worker jobs have been executed and their
  *   status. The values correspond to the names of the constants from the [JobStatus] enum. For instance, a label
  *   _analyzerJobStatus_ with the value _FINISHED_ means that the Analyzer job has been executed successfully.
+ * - runId: The id of the current [OrtRun].
  */
 internal class NotifierOrtResultGenerator(
     /** Reference to the service to access the current ORT run. */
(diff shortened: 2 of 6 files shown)


workers/notifier/src/main/kotlin/notifier/NotifierOrtResultGenerator.kt
```diff
@@ -64,6 +64,9 @@ internal class NotifierOrtResultGenerator(
 
         /** The separator for multiple email recipients. */
         private const val RECIPIENTS_SEPARATOR = ";"
+
+        /** The label for the runId. */
+        private const val RUN_ID_LABEL: String = "runId"
```
oheger-bosch, line 69: The class comment contains a description of the labels available to the notifier script. Please add documentation for this new label there as well.

workers/notifier/src/main/kotlin/notifier/NotifierOrtResultGenerator.kt
```diff
@@ -137,4 +143,8 @@ internal class NotifierOrtResultGenerator(
 
         return jobStatusLabels
     }
+
+    private fun getRunId(ortRun: OrtRun): Map<String, String> {
```
oheger-bosch, line 147: Better name this `getOrtRunLabels()` to align with the other functions.

workers/notifier/src/test/kotlin/notifier/NotifierOrtResultGeneratorTest.kt
```diff
@@ -67,7 +69,8 @@ class NotifierOrtResultGeneratorTest : StringSpec({
 
         val result = helper.runResultGeneratorTest()
 
-        result.labels shouldBe originalLabels
+        result.labels shouldContainAll originalLabels
+        result.labels shouldContainKey "runId"
```
oheger-bosch, line 73: The value of the label should be checked, too.

workers/reporter/src/main/kotlin/reporter/ReporterWorker.kt
```diff
@@ -38,6 +39,8 @@ import org.jetbrains.exposed.sql.Database
 
 import org.slf4j.LoggerFactory
 
+private const val RUN_ID_LABEL: String = "runId"
```
oheger-bosch, line 42: While I have no objections to add this label here for the reporter, just wondering whether there is already a concrete use case for this?
Alternatively, it would be an option to make the `NotifierOrtResultGenerator` class more generic, move it to `:workers:common` and use it for both notifier and reporter. This would reduce code duplication and provide a uniform set of labels for both workers.

workers/notifier/src/main/kotlin/notifier/NotifierOrtResultGenerator.kt
wkl3nk: Function no longer needed after latest refactoring of PR.

workers/reporter/src/main/kotlin/reporter/ReporterWorker.kt
wkl3nk: There is a feature request about having the _runId_ as label in the static web report. Therefore, it needs to be included also in the OrtResult used by the _reporter_ worker.
Refactoring: In order to get rid of the code duplication and have a place where all **common labels** can be specified, I extracted the duplicated code into a util class, which also adds the **common labels** that are used both by the _reporter_ and the _notifier_ worker.

workers/common/src/main/kotlin/common/OrtBaseResultGenerator.kt
```diff
@@ -0,0 +1,63 @@
...
+ * An internally used helper class to generate an [OrtResult] from the current [OrtRun].
+ *
+ * This class loads the results of the previous worker steps and generates an [OrtResult] from them. In addition, it
+ * adds a number of common labels that can be evaluated by different types of workers to obtain further information
+ * about the run.
+ *
+ * The following labels are added:
+ * - runId: The id of the [OrtRun].
+ */
+object OrtBaseResultGenerator {
```
oheger-bosch, line 36: I like the general direction of this refactoring. Just one thought: Would it make sense - instead of introducing a new object - to add the `generate()` function directly to `OrtRunService`? The current implementation suffers from [feature envy](https://refactoring.guru/smells/feature-envy).

workers/common/src/main/kotlin/common/OrtBaseResultGenerator.kt
```diff
@@ -0,0 +1,63 @@
...
+    private const val RUN_ID_LABEL: String = "runId"
+
+    fun generate(ortRunService: OrtRunService, ortRun: OrtRun, failIfRepoInfoMissing: Boolean = true): OrtResult {
```
oheger-bosch, line 39: The function could support a map parameter for additional labels to be added which defaults to an empty map. This would simplify its usage for clients; they would then not have to bother with copying the result and adding further labels.

oheger-bosch (commented):
As a general note: I would recommend to do the introduction of the new result generator class/function in a separate commit. Currently, the single commit does a lot of things.

workers/common/src/main/kotlin/common/OrtBaseResultGenerator.kt
wkl3nk: Yes, I had a look and it makes a lot of sense, as the OrtRunService was "injected" as a parameter and did a lot of things using the OrtRunService. So it is better to have all the code at the same place. I got rid of this OrtBaseResultGenerator and put it into OrtRunService.
Some of the test cases also had then to be moved to OrtRunServiceTest.

workers/common/src/main/kotlin/common/OrtBaseResultGenerator.kt
wkl3nk: Please let's have a separate feature request for that.

oheger-bosch: This was not meant as a new feature request, but as a means to simplify the current implementation - the result generator for the notifier would directly benefit from it. But it's fine with me, I won't insist.

workers/notifier/src/test/kotlin/notifier/NotifierOrtResultGeneratorTest.kt
```diff
@@ -202,6 +203,14 @@ private class ResultGeneratorTestHelper {
         every { getEvaluatorJobForOrtRun(ORT_RUN_ID) } returns null
         every { getScannerJobForOrtRun(ORT_RUN_ID) } returns null
         every { getReporterJobForOrtRun(ORT_RUN_ID) } returns null
+        every { generateOrtResult(ortRun, failIfRepoInfoMissing = false) } returns OrtResult.EMPTY.copy(
```
oheger-bosch, line 206: Can the other `every` declarations now be dropped? I think, on the mock only the new `generateOrtResult()` function will be called.

workers/common/src/main/kotlin/common/OrtRunService.kt
```diff
@@ -276,6 +281,39 @@ class OrtRunService(
         getScannerJobForOrtRun(ortRunId)?.let { scannerRunRepository.getByJobId(it.id) }
     }
 
+    /**
+     * Loads the results of the previous worker steps and generates an [OrtResult] from them. In addition, it
```
oheger-bosch, line 285: Nit: In contrast to Java, the convention for Kotlin comments is to use an imperative style, i.e. "Load the results...". I am personally also more used to the Java conventions, but throughout the codebase the Kotlin style is used.

workers/notifier/src/test/kotlin/notifier/NotifierOrtResultGeneratorTest.kt
wkl3nk: I am not sure if I understand correct. The `OrtRunService` is completely mocked, and it is used by test cases to check if 3 types of notifier-specific _labels_ are added. As this `generateOrtResult` function is called every time in order to add the common _labels_, I **have** to mock it, else the test does not find the function on the OrtRunService.

oheger-bosch: The `generateOrtResult()` has to be mocked for sure. I meant the other stubbing declarations, e.g. for `getXXXJobForOrtRun()`. They were necessary before, as they used to be called by the result generator. But now the result generator only calls `generateOrtResult()` and no longer the other functions.

oheger-bosch (approved):

workers/common/src/main/kotlin/common/OrtRunService.kt
```diff
@@ -88,6 +89,10 @@ class OrtRunService(
     private val scannerJobRepository: ScannerJobRepository,
     private val scannerRunRepository: ScannerRunRepository
 ) {
+    companion object {
+        private const val RUN_ID_LABEL: String = "runId"
```
sschuberth, line 93: Nit: This should drop the explicit type.

mnonnenmacher:
@oheger-bosch Please have extra attention to the ECA check before adding PRs to the merge queue, we currently cannot make it required (https://gitlab.eclipse.org/eclipsefdn/it/api/git-eca-rest-api/-/issues/161, https://github.com/eclipse-apoapsis/.eclipsefdn/pull/4).

@wkl3nk Please do not edit commits with the GitHub UI as this makes the ECA check fail: https://api.eclipse.org/git/eca/status/gh/eclipse-apoapsis/ort-server/200

oheger-bosch:
> @oheger-bosch Please pay extra attention to the ECA check before adding PRs to the merge queue, we currently cannot make it required (https://gitlab.eclipse.org/eclipsefdn/it/api/git-eca-rest-api/-/issues/161, [eclipse-apoapsis/.eclipsefdn#4](https://github.com/eclipse-apoapsis/.eclipsefdn/pull/4)).
> 
> @wkl3nk Please do not edit commits with the GitHub UI as this makes the ECA check fail: https://api.eclipse.org/git/eca/status/gh/eclipse-apoapsis/ort-server/200

What does this mean? Does the commit have to be dropped again?

mnonnenmacher:
> > @oheger-bosch Please pay extra attention to the ECA check before adding PRs to the merge queue, we currently cannot make it required (https://gitlab.eclipse.org/eclipsefdn/it/api/git-eca-rest-api/-/issues/161, [eclipse-apoapsis/.eclipsefdn#4](https://github.com/eclipse-apoapsis/.eclipsefdn/pull/4)).
> > @wkl3nk Please do not edit commits with the GitHub UI as this makes the ECA check fail: https://api.eclipse.org/git/eca/status/gh/eclipse-apoapsis/ort-server/200
> 
> What does this mean? Does the commit have to be dropped again?

No, I think as we know the commit was authered by @wkl3nk and only the wrong email was used it should be fine.
However, looking at the git log the committer was overwritten anyway by `60134199+oheger-bosch@users.noreply.github.com` as you added the commit to the merge queue.

PR #209, opened 2024-05-08
deps: Downgrade Tanstack Router to 1.31.20

Etsija:
Tanstack Router 1.31.23 has a breaking bug [1] which prevents UI to work. Downgrading to 1.31.20 fixes the issue.

[1] https://github.com/TanStack/router/issues/1570https://github.com/TanStack/router/issues/1570

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -26,7 +26,7 @@
     "@radix-ui/react-tooltip": "^1.0.7",
     "@tanstack/react-query": "^5.29.2",
     "@tanstack/react-query-devtools": "^5.32.0",
-    "@tanstack/react-router": "^1.22.2",
+    "@tanstack/react-router": "1.31.20",
     "@tanstack/router-devtools": "^1.22.2",
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.1.0",


mnonnenmacher (commented):
There's a `:` missing after `[1]` in the commit message.

mmurto (changes requested):
The lockfile should also be updated by running `pnpm install`.

sschuberth:
> The lockfile should also be updated by running `pnpm install`.

FYI, that command also has a `--lockfile-only` option, like `pnpm -C ui install --lockfile-only`.

mnonnenmacher:
> > The lockfile should also be updated by running `pnpm install`.
> 
> FYI, that command also has a `--lockfile-only` option, like `pnpm -C ui install --lockfile-only`.

While playing around with ESLint I had the impression that `pnpm install` also modified unrelated lockfile entries while `pnpm add -D lib@version` edited only the relevant parts of the lockfile. Just in case it helps, could also be that I'm wrong.

sschuberth:
> I had the impression that `pnpm install` also modified unrelated lockfile entries

I guess that's because it really (re-)installs *everything*, not just the one modified entry from `package.json`, and depending on when you ran `install` last, a lot of other dependencies might have changed as well.

Etsija:
I will fix the PR tomorrow.

mmurto:
I was wondering why the build in CI was passing even though the `package.json` and `pnpm-lock.yaml` were not in sync, [which shouldn't happen on CI](https://pnpm.io/cli/install#--frozen-lockfile), leading to #210 . Outdated lockfile should error after that is merged.

mmurto:
This was fixed in Router, so #212 should be merged instead of this.

Etsija:
Closed as another PR fixes this.

PR #210, opened 2024-05-08
Fix the build UI job

mmurto:
The current command used in the Build UI step, `pnpm -C ui install build`
doesn't `install` and `build` the UI project, but it installs a package
named `build`. Use separate steps to actually build the project.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -63,8 +63,11 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-pnpm-store-
 
+    - name: Install dependencies
+      run: pnpm -C ui install
+
     - name: Build UI
-      run: pnpm -C ui install build
+      run: pnpm -C ui build
 
   test:
     needs: build


mnonnenmacher (approved):

sschuberth:
> `pnpm -C ui install build` doesn't `install` and `build` the UI project, but it installs a package named `build`.

Ouch, that was based on [my proposal](https://github.com/eclipse-apoapsis/ort-server/pull/134#discussion_r1576224455), thanks for fixing it @mmurto. And of course there exists a package just called "build"...

mmurto:
> Ouch, that was based on [my proposal](https://github.com/eclipse-apoapsis/ort-server/pull/134#discussion_r1576224455), thanks for fixing it @mmurto. And of course there exists a package just called "build"...

Based on the fact that the latest version of [`build` was published 12 years ago and it has 45 000 weekly downloads](https://www.npmjs.com/package/build?activeTab=readme), I'd guess that's a decently common one. :)

PR #223, opened 2024-05-12
Prettier codegen

sschuberth:

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -9,7 +9,7 @@
     "build": "tsc && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "preview": "vite preview",
-    "codegen": "openapi-rq -i http://localhost:8080/swagger-ui/api.json -o src/api"
+    "codegen": "openapi-rq -i http://localhost:${ORT_SERVER_CORE_PORT:-8080}/swagger-ui/api.json -o src/api"
   },
   "dependencies": {
     "@hookform/resolvers": "^3.3.4",
--- a/ui/package.json
+++ b/ui/package.json
@@ -9,7 +9,7 @@
     "build": "tsc && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "preview": "vite preview",
-    "codegen": "openapi-rq -i http://localhost:${ORT_SERVER_CORE_PORT:-8080}/swagger-ui/api.json -o src/api"
+    "codegen": "openapi-rq -i http://localhost:${ORT_SERVER_CORE_PORT:-8080}/swagger-ui/api.json -o src/api --format prettier"
   },
   "dependencies": {
     "@hookform/resolvers": "^3.3.4",
--- a/ui/package.json
+++ b/ui/package.json
@@ -9,7 +9,7 @@
     "build": "tsc && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "preview": "vite preview",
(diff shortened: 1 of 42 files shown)


ui/package.json
```diff
@@ -9,7 +9,7 @@
     "build": "tsc && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "preview": "vite preview",
-    "codegen": "openapi-rq -i http://localhost:8080/swagger-ui/api.json -o src/api"
+    "codegen": "openapi-rq -i http://localhost:${ORT_SERVER_CORE_PORT:-8080}/swagger-ui/api.json -o src/api"
```
sschuberth: FYI, there are sill the following occurrences of `localhost:8080` in the code base, but I'm unsure which of these should probably also get addressed as part of this commit:

```shell
$ git grep localhost:8080
README.md:| ORT Server API | http://localhost:8080/swagger-ui |                                                         |
config/github/src/test/kotlin/GitHubConfigFileProviderFactoryTest.kt:            GitHubConfigFileProvider.GITHUB_API_URL to "http://localhost:8080",
docker-compose.yml:      test: [ "CMD-SHELL", "wget --no-verbose --tries=1 http://localhost:8080/api/v1/liveness" ]
scripts/docker/keycloak/HealthCheck.java:        String url = "http://localhost:8080/health/live";
scripts/docker/keycloak/master-realm.json:    "redirectUris" : [ "http://localhost:8081/*", "http://localhost:8080/*" ],
scripts/requests/env/http-client.env.json:    "host": "http://localhost:8080/api/v1",
transport/kubernetes/src/test/resources/kubeconfig:    server: localhost:8080
ui/src/api/requests/core/OpenAPI.ts:    BASE: "http://localhost:8080",
ui/src/components/providers.tsx:OpenAPIConfig.BASE = import.meta.env.VITE_API_URL || 'http://localhost:8080';
workers/reporter/src/main/resources/application.conf:  reportDownloadLinkPrefix = "http://localhost:8080"
```

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -79,7 +79,8 @@ fun Application.configureOpenApi() {
         }
 
         server {
-            url = "http://localhost:8080"
+            val port = System.getenv("ORT_SERVER_CORE_PORT") ?: "8080"
+            url = "http://localhost:$port"
```
mnonnenmacher, line 83: If you change the port here you also have to adapt the port mapping in docker-compose.yml.

Also, I think the actual issue is that the port that the generated code uses cannot be set at runtime, if this cannot be fixed, this variable will never work without rebuilding the docker image for the UI, once we have it.

Having to set the port (and URL) at compile time makes it impossible to publish reusable binaries.

mnonnenmacher: This could help: https://github.com/7nohe/openapi-react-query-codegen/issues/94#issuecomment-2079344928

sschuberth: Let me just drop this hunk instead. My main point anyway was to make `pnpm codegen` use the correct port.

ui/package.json
mmurto: While there's probably no harm in this, IMO we should (and will need for deployment) set up an environment variable for the API URL, and use it as described [here](https://github.com/7nohe/openapi-react-query-codegen/issues/94#issuecomment-2079344928). Then the effect of this can be achieved by setting the `ORT_SERVER_CORE_URL` etc. with `http://localhost:8080` or `http://localhost:8081.`

MarcelBochtler (approved):

PR #224, opened 2024-05-13
fix(ui): Adapt the UI to the latest API changes

Etsija:
The PRs #157, #181 and #223 introduced some major changes to the generated UI API which all need to be adapted in the UI codebase.

To assist with further similar PRs and clearly separate manual changes from auto-generated ones, we do not use the "every commit must pass all checks" convention in this PR.


sschuberth:
Note that IMO https://github.com/eclipse-apoapsis/ort-server/pull/223 should be merged first, and then the API code in here be regenerated afterwards, to reflect the new `prettier` formatting.

Etsija:
The Build and test check failing is because of a bug in `openapi-react-query-codegen`, [which I reported](https://github.com/7nohe/openapi-react-query-codegen/issues/112), so let's hope it gets fixed soon.  Meanwhile, I've tried a temporary fix by putting 
```
/* tslint:disable:no-unused-variable */
```
to the lines preceding the error-causing unused variable declarations.  I couldn't get this fix to work, and am really reluctant to do it, anyway, as changing manually a part of software which is auto-generated isn't really wise.

The failing test doesn't appear to affect the query client's functionality in any way.

mmurto:
> The "Build and test" check failing is because of a bug in `openapi-react-query-codegen`, [which I reported](https://github.com/7nohe/openapi-react-query-codegen/issues/112), so let's hope it gets fixed soon.

A few options:

1. add the `@ts-nocheck` manually, document it in the readme and wait for/fix the issue upstream,
2. disable the rule in tsconfig, which also disables it for our code, 
3. package the API client in a separate package to not trigger the rule.

I'd say 1 is probably the best in short-term. If the manual addition of the headers is cumbersome, we could write a small bash script that is executed with the generator script.

Etsija:
> > The "Build and test" check failing is because of a bug in `openapi-react-query-codegen`, [which I reported](https://github.com/7nohe/openapi-react-query-codegen/issues/112), so let's hope it gets fixed soon.
> 
> A few options:
> 
>     1. add the `@ts-nocheck` manually, document it in the readme and wait for/fix the issue upstream,
> 
>     2. disable the rule in tsconfig, which also disables it for our code,
> 
>     3. package the API client in a separate package to not trigger the rule.
> 
> 
> I'd say 1 is probably the best in short-term. If the manual addition of the headers is cumbersome, we could write a small bash script that is executed with the generator script.

Well it depends on how often the OpenAPI spec will change (I would guess it is still volatile), and how long does the bug fixing in upstream take. I can surely add those lines manually for some time.

sschuberth:
I know we had talked about this, but now that I see the diff of this PR I yet wonder whether it would be feasible to separate purely auto-generated code from the manual changes required. 

While that would break with the every-commit-must-pass-all-checks-convention, it would more clearly show what manual changes were needed, which helps if anything similar needs to be done in the future again.

So I'd expected three commits currently:
1. Auto-generated code
2. Manual changes needed to `ui/src/routes`
3. Manual changes for `_queryKey`

MarcelBochtler (approved):

PR #227, opened 2024-05-13
fix(workers): Fix NuGet config file generation path

MarcelBochtler:
Change path of NuGet config generation to one that complies to the NuGet specification [1].

[1]: https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file

Authored-by: Kamil Bielecki <kamil.bielecki@pl.bosch.com>

### DIFF
--- a/workers/analyzer/src/test/kotlin/AnalyzerEndpointTest.kt
+++ b/workers/analyzer/src/test/kotlin/AnalyzerEndpointTest.kt
@@ -170,7 +170,7 @@ class AnalyzerEndpointTest : KoinTest, StringSpec() {
 
         "The build environment should contain a NuGet.Config file" {
             runEnvironmentTest { homeFolder ->
-                val nuGetFile = homeFolder.resolve("NuGet.Config")
+                val nuGetFile = homeFolder.resolve(".nuget/NuGet/NuGet.Config")
                 val content = nuGetFile.readText()
 
                 content shouldContain "packageSources"
--- a/workers/common/src/main/kotlin/common/env/NuGetGenerator.kt
+++ b/workers/common/src/main/kotlin/common/env/NuGetGenerator.kt
@@ -30,7 +30,7 @@ import org.eclipse.apoapsis.ortserver.workers.common.env.definition.NuGetDefinit
 class NuGetGenerator : EnvironmentConfigGenerator<NuGetDefinition> {
     companion object {
         /** The name of the configuration file created by this generator. */
-        private const val TARGET = "NuGet.Config"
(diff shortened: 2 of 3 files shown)


nnobelis (commented):
In the linked documentation I don't see the path to `.nuget/Nuget/`.

Quite the opposite:
```
Add a nuget.config file in the root of your project repository. 
```

Can someone explain ?

kamil-bielecki-bosch:
> In the linked documentation I don't see the path `.nuget/Nuget/`.
> 
> Quite the opposite:
> 
> ```
> Add a nuget.config file in the root of your project repository. 
> ```
> 
> Can someone explain ?

There is another NuGet documentation page:
https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior
In section "On Mac/Linux, the user-level config file location varies by tooling" there is another path advised.

workers/common/src/main/kotlin/common/env/NuGetGenerator.kt
```diff
@@ -30,7 +30,7 @@ import org.eclipse.apoapsis.ortserver.workers.common.env.definition.NuGetDefinit
 class NuGetGenerator : EnvironmentConfigGenerator<NuGetDefinition> {
     companion object {
         /** The name of the configuration file created by this generator. */
-        private const val TARGET = "NuGet.Config"
+        private const val TARGET = ".nuget/NuGet/NuGet.Config"
```
sschuberth, line 33: Should this be made `internal` so it could be used in the tests instead of hard-coding the same string again?

workers/analyzer/src/test/kotlin/AnalyzerEndpointTest.kt
```diff
@@ -287,6 +287,7 @@ class AnalyzerEndpointTest : KoinTest, StringSpec() {
      * should create the configuration files declared for the environment. The specified [test][block] can then
      * check whether these files have been created correctly.
      */
+    @Suppress("LongMethod")
```
sschuberth, line 290: Why is this necessary now if the method was not modified?

workers/common/src/main/kotlin/common/env/NuGetGenerator.kt
kamil-bielecki-bosch: Proper file placement is checked in two kind of tests (unit and integration ones), so that's why this string is used twice.
Also - in case of sharing this value, in fact there should be separate test to check if string content is valid.

oheger-bosch (approved):

PR #230, opened 2024-05-14
feat(backend): Add URL validation for `Repository`s 

bs-ondem:

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/Repository.kt
+++ b/api/v1/model/src/commonMain/kotlin/Repository.kt
@@ -19,7 +19,13 @@
 
 package org.eclipse.apoapsis.ortserver.api.v1.model
 
+import io.konform.validation.Validation
+
+import java.net.URI
+
 import kotlinx.serialization.Serializable
+
+import org.eclipse.apoapsis.ortserver.api.v1.model.validation.ValidatorFunc
 
 /**
  * Response object for the repository endpoint.
@@ -33,7 +39,16 @@ data class Repository(
 
     /** The url to the repository. */
     val url: String
-)
+) {
+    companion object {
+        const val INVALID_URL_MESSAGE = "The repository URL is malformed."
+        const val USER_INFO_MESSAGE = "The repository URL must not contain userinfo."
+
+        fun isValidUrl(url: String): Boolean = runCatching { URI.create(url) }.isSuccess
+
+        fun hasUserInfo(url: String): Boolean = runCatching { URI.create(url) }.getOrNull()?.userInfo != null
+    }
+}
 
 /**
  * Request object for the create repository endpoint.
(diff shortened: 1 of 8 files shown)


api/v1/model/src/commonMain/kotlin/Repository.kt
```diff
@@ -33,7 +39,16 @@ data class Repository(
 
     /** The url to the repository. */
     val url: String
-)
+) {
+    companion object {
+        const val INVALID_URL_MESSAGE = "The entity url is malformed."
+        const val USER_INFO_MESSAGE = "The entity url must not contain userinfo."
+
+        fun isValidUrl(url: String): Boolean = runCatching { URL(url) }.isSuccess
```
sschuberth, line 47: Note that constructing a `URL` only accepts a limited set of protocols and might actually make network requests. If this is just about a syntax check, use `URI.create()` instead.

bs-ondem: Thanks for the hint, I didn't know the limitations of `URL`. I replaced it with `URI.create()`.

api/v1/model/src/commonMain/kotlin/Repository.kt
```diff
@@ -33,7 +39,16 @@ data class Repository(
 
     /** The url to the repository. */
     val url: String
-)
+) {
+    companion object {
+        const val INVALID_URL_MESSAGE = "The entity url is malformed."
```
oheger-bosch, line 44: Not sure whether these constants and functions should be public; they then become part of the official API. Therefore, it is probably better to have them as private members of the file.

api/v1/model/src/commonMain/kotlin/Repository.kt
```diff
@@ -33,7 +39,16 @@ data class Repository(
...
+        const val USER_INFO_MESSAGE = "The entity url must not contain userinfo."
```
oheger-bosch, line 45: Since the messages are specific to repositories, they could run "The repository url..." instead of "entity url". And should URL be written in uppercase?

api/v1/model/src/commonMain/kotlin/Repository.kt
bs-ondem: To maintain consistency with the other data classes, I made these variables and functions public. Do you think we should change their visibility to private as well?

dao/src/main/resources/db/migration/V65__removeUserInfoFromRepositoryUrl.sql
```diff
@@ -0,0 +1,31 @@
...
+                '\1'
+        ) AS clean_url,
+        ROW_NUMBER() OVER (
+            PARTITION BY
+                regexp_replace(url, '^(https?://)([^/@:]+(:[^/@]*)?@)?', '\1'),
+                product_id
+            ORDER BY id
+        ) AS rn
+    FROM repositories)
+DELETE
```
oheger-bosch, line 21: Will this DELETE work if the repository is referenced from ORT runs? This would violate a foreign key relation, wouldn't it?

api/v1/model/src/commonMain/kotlin/Repository.kt
oheger-bosch: Good point. No strong feelings about it, but I would tend to say yes. The public API should be as small as possible.

dao/src/main/resources/db/migration/V65__removeUserInfoFromRepositoryUrl.sql
bs-ondem: Good point. As discussed, I will change my SQL query so that no repositories will be deleted. Instead, it will replace the user information with something like 'xxxx' (from https://username@github.com/org/repo.git to https://xxxx@github.com/org/repo.git).

api/v1/model/src/commonMain/kotlin/Repository.kt
oheger-bosch: Ok, probably not our most pressing issue currently.

oheger-bosch:
The database migration script exceeds my SQL fu. Has it been tested on a real database containing real ORT run data?

oheger-bosch (approved):

PR #233, opened 2024-05-16
Format and Lint UI

mmurto:
See individual commit messages for details.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -83,6 +83,41 @@ jobs:
     - name: Run ESLint
       run: pnpm -C ui lint 
 
+  prettier:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v4
+
+    - name: Install Node
+      uses: actions/setup-node@v4
+      with:
+        node-version: 20
+
+    - uses: pnpm/action-setup@v4
+      name: Install pnpm
+      with:
+        version: 9
+
+    - name: Get pnpm store directory
+      shell: bash
+      run: |
+        echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
+
+    - uses: actions/cache@v4
+      name: Setup pnpm cache
+      with:
+        path: ${{ env.STORE_PATH }}
+        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+        restore-keys: |
+          ${{ runner.os }}-pnpm-store-
+
+    - name: Install dependencies
+      run: pnpm -C ui install --dev
+
+    - name: Run Prettier
+      run: pnpm -C ui format:check 
+
   renovate-validation:
     runs-on: ubuntu-22.04
     steps:
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -48,6 +48,41 @@ jobs:
     - name: Check for Detekt Issues with Type Resolution
       run: ./gradlew --stacktrace detektMain detektTest
 
+  eslint:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v4
+
+    - name: Install Node
+      uses: actions/setup-node@v4
+      with:
+        node-version: 20
+
+    - uses: pnpm/action-setup@v4
+      name: Install pnpm
+      with:
+        version: 9
+
+    - name: Get pnpm store directory
+      shell: bash
+      run: |
+        echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
+
+    - uses: actions/cache@v4
+      name: Setup pnpm cache
+      with:
+        path: ${{ env.STORE_PATH }}
+        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+        restore-keys: |
+          ${{ runner.os }}-pnpm-store-
+
+    - name: Install dependencies
+      run: pnpm -C ui install --dev
+
+    - name: Run ESLint
+      run: pnpm -C ui lint 
+
   renovate-validation:
     runs-on: ubuntu-22.04
     steps:
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -96,7 +96,7 @@ Files: */resources/META-INF/services/*
 Copyright: 2022 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
 License: Apache-2.0
 
-Files: ui/.prettierrc
+Files: ui/.prettier*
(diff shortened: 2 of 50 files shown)


ui/package.json
```diff
@@ -50,7 +50,7 @@
     "@typescript-eslint/parser": "^7.1.1",
     "@vitejs/plugin-react": "^4.2.1",
     "autoprefixer": "^10.4.19",
-    "eslint": "^9.0.0",
+    "eslint": "^8",
```
sschuberth, line 55: Just a style question: Should this be `"^8.0.0"`, similar to like in line 38? Or does that even have a different meaning?

ui/src/components/toast-error.tsx
```diff
@@ -17,14 +17,25 @@
  * License-Filename: LICENSE
  */
 
+import { ApiError } from '@/api/requests';
```
sschuberth, line 20: Please add a bit more background information to the commit message, maybe also referring to 0fd185826671f3c6c4bda47bcefb00d44c6936a6 which introduced those `any` casts just recently.

ui/.prettierignore
```diff
@@ -0,0 +1 @@
+pnpm-lock.yaml
```
sschuberth, line 1: I'm surprised that this needs to be ignored explicitly... doesn't `prettier` only look at JS / TS code files? If not (by default), would it instead make sense to configure it to only look at JS / TS code files?

sschuberth: I guess https://prettier.io/docs/en/options.html#parser is the proper reference.

ui/src/components/ui/alert-dialog.tsx
```diff
@@ -7,32 +7,32 @@
  * SPDX-License-Identifier: MIT
  */
 
-import * as React from "react"
-import * as AlertDialogPrimitive from "@radix-ui/react-alert-dialog"
+import * as React from 'react';
```
sschuberth, line 10: Just an observation: Interesting that apparently more files than just those in `api/queries` were not formatted correctly by the `codegen`'s `--format` option.

.github/workflows/static-analysis.yml
```diff
@@ -48,6 +48,41 @@ jobs:
...
+        path: ${{ env.STORE_PATH }}
+        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+        restore-keys: |
+          ${{ runner.os }}-pnpm-store-
+
+    - name: Install dependencies
+      run: pnpm -C ui install
+
+    - name: Run ESLint
+      run: cd ui && pnpm lint
```
sschuberth, line 84: Nit: This could / should use the `-C ui` from line 81.

.github/workflows/static-analysis.yml
```diff
@@ -83,6 +83,41 @@ jobs:
...
+        path: ${{ env.STORE_PATH }}
+        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+        restore-keys: |
+          ${{ runner.os }}-pnpm-store-
+
+    - name: Install dependencies
+      run: pnpm -C ui install
+
+    - name: Run Prettier
+      run: cd ui && pnpm format:check
```
sschuberth, line 119: Again, consider using `-C ui`.

sschuberth (changes requested):
I believe this PR should now also remove `--format` from the `codegen` script as it's not working correctly anyway. Instead, `codegen` should probably call `format` automatically.

ui/package.json
mmurto: `"^8"` and `"^8.0.0"` have identical meaning AFAIK - reference [here](https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept). But now that I think it through again, specifying minimum of the currently latest version `8.x.x` is probably more accurate, as it hasn't been tested with older versions.

ui/.prettierignore
mmurto: [Prettier supports other languages as well](https://prettier.io/docs/en/), and for us it IMO makes sens to also format at least the JSON files in the UI directory to, as there's currently no formatter for JSON. If the project as a whole (on root level) starts using some other formatted for other files than TS/JS, it would probably then make sense to use prettier only for TS/JS, or what do you think?

ui/src/components/ui/alert-dialog.tsx
mmurto: My guess is that `codegen`'s `--format` only applied to the generated files (though it missed some), and not other files in the project.

.github/workflows/static-analysis.yml
mmurto: Ah, tried with `-c` and wondered why it didn't work, let's go with `-C`.

ui/.prettierignore
sschuberth: Yeah, I was concerned that running prettier on the whole directory recursively would be overly broad. But it really seems like we'd want *all* supported files in the directory to get formatted, so let' keep it as-is.

.github/workflows/static-analysis.yml
```diff
@@ -48,6 +48,41 @@ jobs:
...
+    - uses: actions/cache@v4
+      name: Setup pnpm cache
+      with:
+        path: ${{ env.STORE_PATH }}
+        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+        restore-keys: |
+          ${{ runner.os }}-pnpm-store-
+
+    - name: Install dependencies
+      run: pnpm -C ui install
```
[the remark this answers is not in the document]
mmurto: Yep, should work with it. This shares the cache with the Build UI job, so it's only symlinking for the dependencies, but it's a little bit more correct I'd say. 🤓

ui/package.json
```diff
@@ -8,8 +8,9 @@
     "dev": "vite",
     "build": "tsc && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "format": "prettier --write .",
```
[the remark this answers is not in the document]
sschuberth: Hmm, sorry for the nudging, but also this commit does now more than "Add a script for formatting with Prettier":

1. It does also call `prettier` explicitly instead of using `--format`,
2. and reformats the code in `api/queries`.

IMO 1. should be done is a separate commit and 2. should be squashed to the next commit.

.reuse/dep5
```diff
@@ -96,6 +96,10 @@ Files: */resources/META-INF/services/*
 Copyright: 2022 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
 License: Apache-2.0
 
+Files: ui/.prettierignore
```
sschuberth, line 99: How about instead generalizing the below entry to `ui/.prettier*`?

sschuberth (changes requested):

sschuberth (approved):
LGTM, mind approving @MarcelBochtler or @oheger-bosch?

mnonnenmacher (approved):

mmurto:
Rebased to solve conflict in lockfile.

mnonnenmacher (approved):

PR #234, opened 2024-05-16
feat(reporter): Only include specified categories in the sources bundle

MarcelBochtler:
Similar to the ORT CLI [1], the SourceCode Bundle should only include packages from specified license categories.

[1]: https://github.com/oss-review-toolkit/ort/blob/8181666dc5d8bfd739b1260c171c4e707c0c7568/model/src/main/resources/reference.yml#L163-L165

### DIFF
--- a/workers/reporter/src/main/kotlin/reporter/SourceCodeBundleReporter.kt
+++ b/workers/reporter/src/main/kotlin/reporter/SourceCodeBundleReporter.kt
@@ -34,11 +34,18 @@ import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageType
 import org.ossreviewtoolkit.model.config.DownloaderConfiguration
 import org.ossreviewtoolkit.model.config.PluginConfiguration
+import org.ossreviewtoolkit.model.licenses.LicenseCategorization
+import org.ossreviewtoolkit.model.licenses.LicenseClassifications
+import org.ossreviewtoolkit.model.licenses.LicenseInfoResolver
+import org.ossreviewtoolkit.model.licenses.LicenseView
+import org.ossreviewtoolkit.model.licenses.ResolvedLicenseInfo
+import org.ossreviewtoolkit.model.utils.createLicenseInfoResolver
 import org.ossreviewtoolkit.reporter.Reporter
 import org.ossreviewtoolkit.reporter.ReporterInput
 import org.ossreviewtoolkit.utils.common.encodeOrUnknown
 import org.ossreviewtoolkit.utils.common.packZip
 import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
+import org.ossreviewtoolkit.utils.spdx.SpdxLicenseChoice
 
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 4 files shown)


workers/reporter/src/main/kotlin/reporter/SourceCodeBundleReporter.kt
```diff
@@ -162,4 +192,25 @@ class SourceCodeBundleReporter(
...
+        val resolvedLicenseInfo = licenseInfoResolver.resolveLicenseInfo(id)
+        val effectiveLicenses = resolvedLicenseInfo.effectiveLicense(
+            LicenseView.CONCLUDED_OR_DECLARED_AND_DETECTED,
+            licenseChoices
+        )?.decompose().orEmpty()
+
+        return licenseCategorizations
+            .filter { it.id in effectiveLicenses }
+            .flatMap { it.categories }
+            .toSet()
```
sschuberth, line 214: See https://github.com/oss-review-toolkit/ort/pull/8654 😉

workers/reporter/src/main/kotlin/reporter/SourceCodeBundleReporter.kt
```diff
@@ -92,6 +112,16 @@ class SourceCodeBundleReporter(
                 log.info("Found ${packages.size} packages(s) in the ORT result.")
                 addAll(packages)
             }
+        }.filter { pkg ->
+            if (includedLicenseCategories.isEmpty()) { return@filter true }
```
sschuberth, line 116: Consider only filtering at all if `includedLicenseCategories` is non-empty, like

        val filteredPackages = packages.takeIf { includedLicenseCategories.isEmpty() } ?: packages.filter { pkg ->
            val pkgLicenseCategories = pkg.getLicenseCategories(
                licenseClassifications.categorizations,
                ortResult.createLicenseInfoResolver(),
                ortResult.getPackageLicenseChoices(pkg.id)
            )

            pkgLicenseCategories.any { it in includedLicenseCategories }
        }

and then using `filteredPackages` below.

MarcelBochtler: While it is obviously better to not applying the filter if there is nothing to filter, I don't like that your approach would introduce a new variable where the old one (`packages`), should not be used in the code below.

workers/reporter/src/main/kotlin/reporter/SourceCodeBundleReporter.kt
```diff
@@ -94,13 +114,23 @@ class SourceCodeBundleReporter(
             }
         }
 
+        val filteredPackages = allPackages.takeIf { includedLicenseCategories.isEmpty() } ?: allPackages.filter { pkg ->
+            val pkgLicenseCategories = pkg.getLicenseCategories(
+                licenseClassifications.categorizations,
+                ortResult.createLicenseInfoResolver(),
```
oheger-bosch, line 120: Not sure what exactly `ortResult.createLicenseInfoResolver()` does, but I assume such a resolver should be created only once and not for each element to be filtered.

oheger-bosch (approved):

PR #238, opened 2024-05-17
Add support for generating a .git-credentials file

oheger-bosch:
This PR makes the handling of the credentials for infrastructure services more generic by introducing a new `credentialsTypes` property. The property controls, in which (dynamically generated) configuration files the credentials of services should be listed. While in the past only a `.netrc` file was supported, it is now also possible to have credentials added to a `.git-credentials` file.

This extension of the concept became necessary because we encountered some Git repositories that could not be cloned by the Git command line, even if correct credentials were present in the `.netrc` file. Only after the credentials were made available in a `.git-credentials` file, cloning was possible. This happened with repositories hosted on a Microsoft Teams Foundation Server.

Note that this PR changes the data model for infrastructure services and will therefore affect the UI.

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/CredentialsType.kt
+++ b/api/v1/model/src/commonMain/kotlin/CredentialsType.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.api.v1.model
+
+/**
+ * An enumeration class defining the supported credential types that are available for infrastructure services.
+ * Depending on the type(s) set, the credentials of the service are added to different configuration files. Note that
+ * infrastructure services can be assigned multiple credential types. It is also possible that they have no type;
+ * then they are ignored when generating configuration files for credentials.
+ */
+enum class CredentialsType {
+    /**
+     * Credentials type indicating that the credentials of the service should be added to the _.netrc_ file. This is
+     * in most cases the desired option, since it allows access to the service from most external tools.
+     */
+    NETRC_FILE,
+
+    /**
+     * Credentials type indicating that the credentials of the service should be added to the _.git-credentials_ file.
+     * For normal cases, it should not be necessary to use this type, since Git should be able to obtain the
+     * credentials from the _.netrc_ file. There are, however, rare cases when Git is not able to authenticate against
+     * a repository based on the information in the _.netrc_ file. This was observed for instance with repositories
+     * hosted on Microsoft Azure DevOps servers.
+     */
+    GIT_CREDENTIALS_FILE
+}
--- a/workers/common/src/main/kotlin/common/env/GitCredentialsGenerator.kt
+++ b/workers/common/src/main/kotlin/common/env/GitCredentialsGenerator.kt
@@ -1,0 +1,103 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.workers.common.env
+
+import java.net.URI
+
+import org.eclipse.apoapsis.ortserver.model.CredentialsType
+import org.eclipse.apoapsis.ortserver.model.InfrastructureService
+import org.eclipse.apoapsis.ortserver.workers.common.env.definition.EnvironmentServiceDefinition
+
+import org.slf4j.LoggerFactory
+
+/**
+ * A specialized generator class to generate files with special credentials for Git.
+ *
+ * This generator is required only for corner cases. Normally, the Git CLI should be able to obtain credentials from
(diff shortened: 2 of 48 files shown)


api/v1/model/src/commonMain/kotlin/CredentialsType.kt
```diff
@@ -0,0 +1,43 @@
...
+    NETRC,
+
+    /**
+     * Credentials type indicating that the credentials of the service should be added to the _.git-credentials_ file.
+     * For normal cases, it should not be necessary to use this type, since Git should be able to obtain the
+     * credentials from the _.netrc_ file. There are, however, rare cases when Git is not able to authenticate against
+     * a repository based on the information in the _.netrc_ file. This was observed for instance with repositories
+     * hosted on Microsoft Azure DevOps servers.
+     */
+    GIT
```
sschuberth, line 42: How about a slight more specific name like `GIT_CREDENTIALS`, or maybe even `GIT_CREDENTIALS_FILE` in preparation for non-file types? Thinking about it, then the above should probably also be named `NETRC_FILE`.

workers/common/src/main/kotlin/common/env/GitCredentialsGenerator.kt
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * A specialized generator class to generate files with special credentials for Git.
+ *
+ * This generator is required only for corner cases. Normally, the Git CLI should be able to obtain credentials from
+ * the _.netrc_ file. However, in some special cases, this mechanism is not working, and authentication against the
+ * repository is only possible if the credentials are placed in a _.git-credentials_ file. This generator is able to
+ * produce such a file, together with a Git configuration file that references it. It only processes infrastructure
+ * services whose credentials types contain [CredentialsType.GIT].
+ */
+class GitCredentialsGenerator : EnvironmentConfigGenerator<EnvironmentServiceDefinition> {
```
sschuberth, line 39: Maybe "GitCredentials*File*Generator"?

workers/common/src/main/kotlin/common/env/GitCredentialsGenerator.kt
```diff
@@ -0,0 +1,101 @@
...
+    companion object {
+        /** The name of the file storing the actual credentials. */
+        private const val GIT_CREDENTIALS_FILE = ".git-credentials"
```
sschuberth, line 42: Maybe "GIT_CREDENTIALS_FILE*NAME*"? Similar below.

workers/common/src/main/kotlin/common/env/GitCredentialsGenerator.kt
```diff
@@ -0,0 +1,101 @@
...
+
+        private val logger = LoggerFactory.getLogger(GitCredentialsGenerator::class.java)
+
+        /**
+         * Return a string with the URL of this [InfrastructureService] with the credentials embedded as needed within
+         * the _.git-credentials_ file. Use [builder] to obtain secret references. Return *null* if the URL is invalid.
+         */
+        private fun InfrastructureService.urlWithCredentials(builder: ConfigFileBuilder): String? = runCatching {
+            buildString {
+                val serviceUrl = URI(url).toURL()
```
sschuberth, line 55: IIRC, in some recent Java version the `URI()` constructor got deprecated, and we should use `URI.create()` instead.

sschuberth: Also see https://github.com/eclipse-apoapsis/ort-server/pull/239.

workers/common/src/main/kotlin/common/env/GitCredentialsGenerator.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Generate the (static) content of the _.gitconfig_ file using the given [builder]. In this file, only the
+     * credential store is enabled.
+     */
+    private suspend fun generateGitConfig(builder: ConfigFileBuilder) {
+        builder.buildInUserHome(GIT_CONFIG_FILE) {
+            println("[credential]")
+            println("\thelper = store")
```
nnobelis, line 98: When I look in my `.gitconfig` file there is:
`helper = store --file ~/.git-credentials`

Should we put this `--file` ?

oheger-bosch: `~/.git-credentials` is the standard location. I would therefore only add the parameter if we - for whatever reasons - deviate from this standard.

workers/common/src/main/kotlin/common/env/GitCredentialsGenerator.kt
oheger-bosch: The name is in-line with the other generator classes, for instance `MavenSettingsGenerator`, `NetRcGenerator`, `NpmRcGenerator`.

api/v1/model/src/commonMain/kotlin/CredentialsType.kt
```diff
@@ -0,0 +1,43 @@
...
+     * in most cases the desired option, since it allows access to the service from most external tools.
+     */
+    NETRC_FILE,
+
+    /**
+     * Credentials type indicating that the credentials of the service should be added to the _.git-credentials_ file.
+     * For normal cases, it should not be necessary to use this type, since Git should be able to obtain the
+     * credentials from the _.netrc_ file. There are, however, rare cases when Git is not able to authenticate against
+     * a repository based on the information in the _.netrc_ file. This was observed for instance with repositories
+     * hosted on Microsoft Azure DevOps servers.
```
MarcelBochtler, line 40: > This was observed for instance with repositories hosted on Microsoft Azure DevOps servers.

Wasn't this just observed on Microsoft TFS? AFAIK Azure DevOps works fine with the .netrc file.

workers/common/src/main/kotlin/common/env/definition/EnvironmentServiceDefinition.kt
```diff
@@ -40,15 +41,15 @@ open class EnvironmentServiceDefinition(
     val service: InfrastructureService,
 
     /**
-     * A flag to indicate whether the associated [InfrastructureService] should be excluded from the _.netrc_ file.
-     * If this flag is defined, it overrides the corresponding flag from the service.
+     * A set determining the locations where the credentials of the associated [InfrastructureService] should be
+     * used. If it is defined, it overrides the corresponding property from the service.
      */
-    val excludeServiceFromNetrc: Boolean? = null
+    val credentialsTypes: Set<CredentialsType>? = null
```
MarcelBochtler, line 47: What would it mean if this value is `null`, and what would it mean if it is an empty Set?
Does it mean the same? Is this handled differently?

Could the value be non-nullable and use an emptySet as default?

api/v1/model/src/commonMain/kotlin/CredentialsType.kt
oheger-bosch: I am not sure which server the affected customer actually has. I only read that TFS is now named Azure DevOps Server and took this name in the comment. I could change this to "Microsoft TFS servers" or maybe just "Microsoft servers" ?

workers/common/src/main/kotlin/common/env/definition/EnvironmentServiceDefinition.kt
oheger-bosch: The service definition allows overriding the corresponding property from the infrastructure service if `credentialsType` is not *null*. So, there is a difference between a *null* value and an empty set: *null* means that the value from the infrastructure service is used; otherwise, the value from the definition is used. The comment says "If it is defined, it overrides the corresponding property from the service." Do you think, this needs further clarification?

api/v1/model/src/commonMain/kotlin/CredentialsType.kt
MarcelBochtler: I thought TFS is the on premises counterpart to the SaaS Azure DevOps.
But I don't think that this is relevant.

MarcelBochtler (approved):

PR #265, opened 2024-05-21
Use Kotlin's Base64 coding

sschuberth:

### DIFF
--- a/config/github/src/test/kotlin/GitHubConfigFileProviderTest.kt
+++ b/config/github/src/test/kotlin/GitHubConfigFileProviderTest.kt
@@ -47,7 +47,8 @@ import io.ktor.http.HttpStatusCode
 import io.mockk.every
 import io.mockk.mockk
 
-import java.util.Base64
+import kotlin.io.encoding.Base64
+import kotlin.io.encoding.ExperimentalEncodingApi
 
 import org.eclipse.apoapsis.ortserver.config.ConfigException
 import org.eclipse.apoapsis.ortserver.config.ConfigManager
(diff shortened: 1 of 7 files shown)


config/github/src/test/kotlin/GitHubConfigFileProviderTest.kt
```diff
@@ -47,7 +47,8 @@ import io.ktor.http.HttpStatusCode
 import io.mockk.every
 import io.mockk.mockk
 
-import java.util.Base64
+import kotlin.io.encoding.Base64
```
MarcelBochtler, line 50: What is the advantage from using Kotlin's Base64 encoder in code that not required to be multiplatform compatible?
Isn't it worse to use it while it is considered an `ExperimentalEncodingApi`?

sschuberth: The above aside, the code is even a bit simpler now.

sschuberth: @MarcelBochtler, can we move this forward, please?

MarcelBochtler: I'm not blocking this PR, but your arguments didn't convince me.
I would like to hear other opinions, e.g. from @mnonnenmacher or @oheger-bosch.

I will be fine with any decision.

mnonnenmacher: To me the risk is minimal, the worst that can happen is that we have to do some minor adaptations if the API changes with a Kotlin upgrade. And I would agree that it is in general a good practice to prefer Kotlin APIs over Java APIs when possible. Based on that I'll approve the change and merge later this afternoon if there are no objections.

mnonnenmacher (approved):

config/github/src/test/kotlin/GitHubConfigFileProviderTest.kt
MarcelBochtler: No objections from my side.

PR #271, opened 2024-05-21
refactor: Simplify the secrets API via an enum for the id's entity

sschuberth:


dao/src/main/kotlin/repositories/DaoSecretRepository.kt
```diff
@@ -43,23 +44,17 @@ import org.slf4j.LoggerFactory
...
-    }
+    override fun create(path: String, name: String, description: String?, entity: Entity, id: Long) =
+        db.blockingQuery {
+            SecretDao.new {
+                this.path = path
+                this.name = name
+                this.description = description
+                this.organization = id.takeIf { entity == Entity.ORGANIZATION }?.let { OrganizationDao[it] }
+                this.product = id.takeIf { entity == Entity.PRODUCT }?.let { ProductDao[it] }
+                this.repository = id.takeIf { entity == Entity.REPOSITORY }?.let { RepositoryDao[it] }
```
mnonnenmacher, line 55: `if` would be shorter and to me also more readable:

```kotlin
this.organization = if (entity == Entity.ORGANIZATION) OrganizationDao[it] else null
```

But it should also work with `when`:
```kotlin
when (entity) {
  Entity.ORGANIZATION -> this.organization = OrganizationDao[id]
  ...
}
```

secrets/spi/src/main/kotlin/SecretsProvider.kt
```diff
@@ -50,8 +52,8 @@ interface SecretsProvider {
     fun removeSecret(path: Path)
 
     /**
-     * Generate a [Path] for the secret belonging to the given [organizationId], [productId] and [repositoryId], as well
-     * as the [secretName].
+     * Generate a [Path] for the secret belonging to the given [entity] with [id], as well as the [secretName].
```
mnonnenmacher, line 55: Please also describe the new default implementation.

core/src/main/kotlin/api/OrganizationsRoute.kt
```diff
@@ -238,9 +239,8 @@ fun Route.organizations() = route("organizations") {
                         createSecret.name,
                         createSecret.value,
                         createSecret.description,
-                        organizationId,
-                        null,
-                        null
+                        Entity.ORGANIZATION,
+                        organizationId
```
mnonnenmacher, line 245: For most callers it would be nicer to have convenience functions like `createRepositorySecret` or `createProductSecret` (probably with default implementations in the interface as implementations would be identical).

core/src/main/kotlin/api/OrganizationsRoute.kt
```diff
@@ -221,7 +223,7 @@ fun Route.organizations() = route("organizations") {
                     val organizationId = call.requireIdParameter("organizationId")
                     val secretName = call.requireParameter("secretName")
 
-                    secretService.deleteSecretByOrganizationAndName(organizationId, secretName)
+                    secretService.deleteSecret(Entity.ORGANIZATION, organizationId, secretName)
```
mnonnenmacher, line 226: Similar to the comment on the first commit, it would be nice to have convenience functions like `getRepositorySecret`, `deleteRepositorySecret`, or `listRepositorySecrets`.

mnonnenmacher (changes requested):

dao/src/main/kotlin/repositories/DaoSecretRepository.kt
sschuberth: What I usually like about the `takeIf` approach is that the thing to assign from is listed first, instead of last / in the middle with the `if` approach.

Anyway, if `when` works here as entities are `null` by default, that's yet nicer. I'll use that.

secrets/spi/src/main/kotlin/SecretsProvider.kt
sschuberth: Done. For the record, the default implementation is not "new" in the sense that it would be doing something different than before. It's just a simpler, centralized implementation of the original logic.

I decided to move this refactoring to another preceding commit to make that more clear.

core/src/main/kotlin/api/OrganizationsRoute.kt
sschuberth: Where's the convenient in using these functions that basically just encode the entity parameter as part of the function name? They are hardly shorter, and create a maintenance burden if more entities (like business units, groups of organizations etc.) would be added in the future.

dao/src/main/kotlin/repositories/DaoSecretRepository.kt
mnonnenmacher: > What I usually like about the `takeIf` approach is that the thing to assign from is listed first, instead of last / in the middle with the `if` approach.

It's a matter of taste and being familiar with the pattern, but to me it's actually the opposite, close to an anti-pattern, especially if the condition does not refer to the thing `takeIf` is called on, like in this case. The if condition I can read from left to right like human language and understand on the first read what happens: "If the condition is fulfilled do this, otherwise that."
With `takeIf` in combination with `let` I always need additional thinking steps, I have to jump back to the beginning after understanding the condition and then see how the thing `takeIf` was called on is used in the `let` block, and remember that if the condition is false the whole expression evaluates to null.

> Anyway, if `when` works here as entities are `null` by default, that's yet nicer. I'll use that.

I only hope it works but I'm not sure. I guess you'll find out.

sschuberth: > I guess you'll find out.

I hope the exitsing test coverage does for me 😉

core/src/main/kotlin/api/OrganizationsRoute.kt
mnonnenmacher: I guess it's a matter of taste as well, personally I'd rather call `getOrganizationSecrets(orgId)` than `getSecrets(Entity.ORGANIZATION, orgId)` because it has less parameters.

sschuberth: So, not a *must* to address? Because I'd prefer not to.

dao/src/main/kotlin/repositories/DaoSecretRepository.kt
mmurto: Just a point, for someone not too familiar with Kotlin, of the options above, the one below is by far the simplest to quickly comprehend for me. 

> ```kotlin
> this.organization = if (entity == Entity.ORGANIZATION) OrganizationDao[it] else null
> ```

core/src/main/kotlin/api/OrganizationsRoute.kt
mnonnenmacher: No, not a must, but as this is still a draft maybe there is some time to get an opinion from another contributor (@oheger-bosch @MarcelBochtler @bs-ondem)?

oheger-bosch:
Not that convinced from the current proposal. If we change something in this API, we should go a step further and make it type-safe. This could be achieved by introducing value classes for the IDs of organizations, products, and repositories.

I am not sure, does Kotlin support the inheritance of value classes from a sealed interface? Then we could have something like
```
sealed interface HierarchyId
```
The repository then only needs one `createSecret(id: HierarchyId)` function and can decided based on the ID type on which level to add the new secret.

sschuberth:
> does Kotlin support the inheritance of value classes from a sealed interface?

It does, and I like that idea. I'll probably follow-up on that later.

sschuberth:
> > does Kotlin support the inheritance of value classes from a sealed interface?
> 
> It does, and I like that idea. I'll probably follow-up on that later.

I've finally implemented your idea @oheger-bosch, for now limit to secrets provider related code.

core/src/main/kotlin/api/OrganizationsRoute.kt
sschuberth: Outdated, due to another approach chosen now.

core/src/main/kotlin/api/OrganizationsRoute.kt
sschuberth: Outdated, due to another approach chosen now.

oheger-bosch:
Looks good to me.

Just one question regarding the commit messages: Since we are still at version 0.x, do we use the "!" to mark breaking changes? This could screw up the automated calculation of the next version number.

sschuberth:
> Since we are still at version 0.x, do we use the "!" to mark breaking changes? This could screw up the automated calculation of the next version number.

AFAIK, the plugin that we use to calculate versions automatically takes care of this, and does not bump the major version for breaking changes if the version is below 1.0.0.

I've verified this by running `./gradlew printVersion` on my branch which shows `0.9.0-RC.007.sha.6e55e37`.

sschuberth:
Also, we had breaking changes before, like efadae752d97db1b23256bb6ca52e5b293b87154 or 23514f71b414749c391832fdfbfef9630e95cb1a.

oheger-bosch:
> AFAIK, the plugin that we use to calculate versions automatically takes care of this, and does not bump the major version for breaking changes if the version is below 1.0.0.
> 
> I've verified this by running `./gradlew printVersion` on my branch which shows `0.9.0-RC.007.sha.6e55e37`.

Thanks for checking this. I had something in mind, but are not up-to-date here.

oheger-bosch (approved):

PR #275, opened 2024-05-22
build(ui): Remove generated routes from Git to generate them on the spot

mmurto:
TanStack Router auto-generates `routeTree.gen.ts` to provide type-safe
routing. Storing this in Git may require manual generation when Renovate
tries to update Router. This commit removes the generated file from Git and
generates it with build script.


ui/package.json
```diff
@@ -6,11 +6,12 @@
   "license": "Apache-2.0",
   "scripts": {
     "dev": "vite",
-    "build": "tsc && vite build",
+    "build": "pnpm generate:routes && tsc && vite build",
```
mnonnenmacher, line 9: Could we do the same change for `codegen`?

sschuberth:
I'll work on resolving the conflicts as I need this commit on another branch anyway.

ui/package.json
mmurto: We could, but we need the Core running for that to get the OpenAPI JSON. If it runs and returns the spec without any of the services (db, keycloak), then it could be relatively simple.

mnonnenmacher (approved):

PR #277, opened 2024-05-22
Generate UI client on the fly

mnonnenmacher:
Add a Gradle task to generate the openapi.json and use it to generate the UI client on the fly. See the commit messages for details.


core/src/test/kotlin/utils/GenerateOpenApiSpec.kt
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * This fake test starts the [ortServerTestApplication] to get the generated OpenAPI specification and writes it to
+ * `build/openapi/openapi.json`. This is used by the `generateOpenApiSpec` Gradle task to generate the OpenAPI
+ * specification in the file system to make it accessible for generating the client for the ui module.
+ *
+ * This test is not executed during normal test runs and only enabled if the `generateOpenApiSpec` system property is
+ * set to `true`.
+ */
+class GenerateOpenApiSpec : StringSpec({
+    "generate the OpenAPI specification".config(enabled = System.getProperty("generateOpenApiSpec") == "true") {
```
sschuberth, line 46: Nit: Maybe use `.toBoolean()` instead of `== "true"`.

core/src/test/kotlin/utils/GenerateOpenApiSpec.kt
```diff
@@ -0,0 +1,66 @@
...
+
+            response shouldHaveStatus 200
+
+            val outputDir = File("build/openapi")
+            outputDir.mkdirs()
+
+            val outputFile = outputDir.resolve("openapi.json")
+            outputFile.writeText(response.bodyAsText())
+
+            outputFile should exist()
```
sschuberth, line 63: Prefer `shouldBe aFile()`.

core/src/test/kotlin/utils/GenerateOpenApiSpec.kt
```diff
@@ -0,0 +1,66 @@
...
+            additionalConfigs = mapOf("jwt.issuer" to "https://example.org")
+        ) {
+            application { configureOpenApi() }
+
+            val response = createJsonClient().get("/swagger-ui/api.json")
+
+            response shouldHaveStatus 200
+
+            val outputDir = File("build/openapi")
+            outputDir.mkdirs()
```
sschuberth, line 58: Nit: Maybe use `val outputDir = File("build/openapi").apply { mkdirs() }`. Similar below.

mnonnenmacher:
@mmurto @Etsija Please test if this works for you.

.github/workflows/build-and-test.yml
```diff
@@ -40,6 +40,20 @@ jobs:
     - name: Checkout Repository
       uses: actions/checkout@v4
 
+    - name: Setup Java
```
sschuberth, line 43: Commit message:
- "it now" -> "it is now" or "it now is"
- maybe change "ui" to "UI"

ui/package.json
```diff
@@ -6,7 +6,7 @@
   "license": "Apache-2.0",
   "scripts": {
     "dev": "vite",
-    "build": "pnpm generate:routes && tsc && vite build",
+    "build": "pnpm generate:api && pnpm generate:routes && tsc && vite build",
```
sschuberth, line 9: Does pnpm accept multiple scripts to run in one invocation, like `pnpm generate:api generate:routes`?

mmurto: [It does](https://pnpm.io/cli/run#running-multiple-scripts), but I'd say the current is clearer than regex, so I'd go with it.

sschuberth: > clearer than regex

Huh? What regex?

sschuberth: Ah, I though you were referring to my example, which does not use a regex, but lists both scripts explicitly. If that's not supported, but only via regex, it could indeed be clearer to keep the current code.

mmurto: Yeah, I don't think just lining them up works, as you can give further arguments to scripts. Though I haven't tried out, but it's not something I've bumped into in the wild.

sschuberth: Although `pnpm run "/^generate:.*/"` does have some charm, as it nicely runs any current and future generate scripts.

BTW, why do the docs say "pnpm *run*", but we don't in line 9?

mmurto: https://github.com/pnpm/pnpm/issues/2169 but apparently not in the web docs.

core/src/test/kotlin/utils/GenerateOpenApiSpec.kt
mnonnenmacher: I think I like the current way better for simplicity.

Etsija:
I tested your PR locally. Here are my findings:
1. Ran up the Docker Compose environment for UI testing: `docker compose pull && docker compose up`
2. From main dir: `./gradlew :core:generateOpenApiSpec`
- this produced `core/build/openapi/openapi.json` as expected
- as a side effect, it also produced `.kotlin/sessions/kotlin-compiler-[number].salive`, which isn't in `.gitignore` so needed to be deleted manually. What is the file, and should the whole `.kotlin/`directory be in `.gitignore` as part of this PR?
3. From ui/ dir: `pnpm build`
- this produced the UI query client in `ui/src/api/` as expected
- for some reason, all files in `ui/src/api/queries/` and (at least from what I can see) the file `ui/src/api/requests/services.gen.ts` are unformatted, as if Prettier hasn't run at all for them
- compared to the main branch, where the whole `ui/src/api/` folder is Prettier'd correctly, this feels like a regression. Even if the whole `ui/src/api/` folder is ignored by Git now, we'd still like the query client to be properly formatted for clarity, because we need to revisit the query client quite often in UI work
4. From ui/ dir: `pnpm dev`
- checked the UI client: seems to be working as expected. All CRUD operations working, new ORT runs are created and run as expected, API error responses are shown as expected

sschuberth:
> as a side effect, it also produced `.kotlin/sessions/kotlin-compiler-[number].salive`

This is a side effect of the K2 migration, I'll make a PR.

Etsija:
Addendum to point 3. in my findings: found out that Prettier actually ignores all files in .gitignore, so that's why the regenerated query client is not Prettier'd.  I've not yet found a neat solution for this.

sschuberth:
@mmurto pointed us at https://github.com/prettier/prettier/issues/15438, which could help.

Etsija:
Yes, specifically [this](https://github.com/prettier/prettier/issues/15438#issuecomment-1731842806). The comment also has a valid point about two different tools having a conflict of interests.

mnonnenmacher:
@Etsija Thanks for checking, I have applied the suggested fix for prettier.

MarcelBochtler (approved):

PR #278, opened 2024-05-22
refactor(config): Migrate from `constructorArgs` to `providedProperties`

sschuberth:
The `constructorArgs` are deprecated to pass values to scripts. Use the generic `providedProperties` mechanism instead.


oheger-bosch:
Looks good. Just to make sure: This does not require any changes on the scripts that are executed, right? (Probably not, since the tests are all successful.)

sschuberth:
> This does not require any changes on the scripts that are executed, right?

Correct.

oheger-bosch:
> > This does not require any changes on the scripts that are executed, right?
> 
> Correct.

Thanks for confirming.

oheger-bosch (approved):

PR #280, opened 2024-05-23
style(ui): Tune prettier to align with the rest of the codebase

Etsija:
Please have a look at the individual commits for details.


ui/.prettierrc
```diff
@@ -3,5 +3,8 @@
   "jsxSingleQuote": true,
   "trailingComma": "es5",
   "useTabs": false,
-  "plugins": ["@ianvs/prettier-plugin-sort-imports"]
+  "plugins": [
```
sschuberth, line 6: Commit message nit: We do not use a double-space to separate sentences in a paragraph.

Etsija: Sry, it still comes from flesh memory from time to time...fixed now.

mmurto:
It could make sense to [configure separation between import groups ](https://github.com/trivago/prettier-plugin-sort-imports?tab=readme-ov-file#importorderseparation)to align with Kotlin code.

Etsija:
> It could make sense to [configure separation between import groups ](https://github.com/trivago/prettier-plugin-sort-imports?tab=readme-ov-file#importorderseparation)to align with Kotlin code.

I don't think the plugin I'm using supports that.

mmurto:
> > It could make sense to [configure separation between import groups ](https://github.com/trivago/prettier-plugin-sort-imports?tab=readme-ov-file#importorderseparation)to align with Kotlin code.
> 
> I don't think the plugin I'm using supports that.

Oh, didn't notice it was different, trivago's seems to be more popular. [But your choice supports is as well](https://github.com/IanVS/prettier-plugin-sort-imports?tab=readme-ov-file#3-add-spaces-between-import-groups).

Etsija:
> > > It could make sense to [configure separation between import groups ](https://github.com/trivago/prettier-plugin-sort-imports?tab=readme-ov-file#importorderseparation)to align with Kotlin code.
> > 
> > 
> > I don't think the plugin I'm using supports that.
> 
> Oh, didn't notice it was different, trivago's seems to be more popular. [But your choice supports is as well](https://github.com/IanVS/prettier-plugin-sort-imports?tab=readme-ov-file#3-add-spaces-between-import-groups).

Yes, but then we need to invent an import order instead of using the default one, which seems to work well. I'm open to suggestions as to the definition of the import order.

mmurto:
> Yes, but then we need to invent an import order instead of using the default one, which seems to work well. I'm open to suggestions as to the definition of the import order.

A good start would be the default but with lines between. :)

ui/src/components/providers.tsx
```diff
@@ -17,12 +17,12 @@
...
  */
 
+import { OpenAPI as OpenAPIConfig } from '@/api/requests/core/OpenAPI';
+import { Toaster } from '@/components/ui/toaster';
 import { queryClient } from '@/lib/query-client';
 import { QueryClientProvider } from '@tanstack/react-query';
 import { ReactNode } from 'react';
 import { AuthProvider, AuthProviderProps } from 'react-oidc-context';
-import { Toaster } from '@/components/ui/toaster';
-import { OpenAPI as OpenAPIConfig } from '@/api/requests/core/OpenAPI';
```
mmurto, line 25: Just an example but this happens across the files. The sorting does not work, as it considers the imports from local code with the alias `@/` to be third-party code. Not sure what the best way to fix it is, but the differentiator is that third party code from namespaces doesn't have `/` after `@`.

mmurto (changes requested):

mmurto (approved):

ui/.prettierrc
```diff
@@ -2,5 +2,14 @@
   "singleQuote": true,
   "jsxSingleQuote": true,
   "trailingComma": "es5",
-  "useTabs": false
```
MarcelBochtler, line 4: Commit message:
```
To align the better with the rest
```

This should probably be:
```
To align better with the rest
```

PR #288, opened 2024-05-23
refactor(secrets): Make common `createPath()` code the default

sschuberth:


secrets/spi/src/main/kotlin/SecretsProvider.kt
```diff
@@ -55,14 +55,14 @@ interface SecretsProvider {
      */
     fun createPath(organizationId: Long?, productId: Long?, repositoryId: Long?, secretName: String): Path {
         val secretType = when {
-            organizationId != null -> "organization"
-            productId != null -> "product"
-            repositoryId != null -> "repository"
+            organizationId != null -> "organization_$organizationId"
```
oheger-bosch, line 58: What I dislike on the new approach is that the separator ('_') between level and ID is now repeated multiple times, making it harder to change the way paths are generated.

An alternative could be having a nested function like
```
fun levelId(level: String, id: Long?): String? =
  id?.let { "${level}_$id" }
```
and then do something like

```
return listOfNotNull(
    levelId("organization", organizationId),
    levelId("product", productId),
    levelId("repository", repositoryId)
)..singleOrNull() ?: throw IllegalArgumentException(...)
```

sschuberth: > the separator ('_') between level and ID is now repeated multiple times, making it harder to change the way paths are generated.

Actually, I think that could be even an advantage, as no one says that the separator between path components and the final name should be the same. While implementing a new secrets provider, I was actually struggling with the fact that from a given `Path`, you cannot easily tell anymore which was the "name" part only (as the name could also contain an underscore itself).

I've found a good way to solve that now, but I'm still not sure whether we should *enforce* the separator to be the same between all path components.

> An alternative could be having a nested function like

My goal was to make the code simpler, not more complex / longer 😉

sschuberth: Anyway, I expect the code to change again with a new iteration of https://github.com/eclipse-apoapsis/ort-server/pull/271.

oheger-bosch: > Anyway, I expect the code to change again with a new iteration of #271.

So, I see no point in changing this now. Could this commit be dropped?

sschuberth: > So, I see no point in changing this now. Could this commit be dropped?

I see no point in spending work to drop it as it's already there. Can it stay?

oheger-bosch (approved):

PR #306, opened 2024-05-27
Opt-in to `ExperimentalSerializationApi` automatically

sschuberth:

### DIFF
--- a/api/v1/mapping/build.gradle.kts
+++ b/api/v1/mapping/build.gradle.kts
@@ -21,7 +21,6 @@ import io.gitlab.arturbosch.detekt.Detekt
 
 plugins {
     alias(libs.plugins.kotlinMultiplatform)
-    alias(libs.plugins.kotlinxSerialization)
 }
 
 group = "org.eclipse.apoapsis.ortserver.api.v1"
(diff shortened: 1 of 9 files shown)


api/v1/mapping/build.gradle.kts
mnonnenmacher, line 1: Just to be sure, the compiler plugin is only required for code generation in modules that use the `@Serializable` annotation, but not in modules that just consume serializable classes, right?

build.gradle.kts
```diff
@@ -164,9 +164,12 @@ subprojects {
     }
 
     tasks.withType<KotlinCompile>().configureEach {
+        val hasSerializationPlugin = plugins.hasPlugin(libs.plugins.kotlinxSerialization.get().pluginId)
+
         compilerOptions {
             allWarningsAsErrors = true
             jvmTarget = maxKotlinJvmTarget
+            if (hasSerializationPlugin) optIn = listOf("kotlinx.serialization.ExperimentalSerializationApi")
```
mnonnenmacher, line 172: Could we just globally define this opt in, or would it fail in projects not depending on kotlinx.serialization?

api/v1/mapping/build.gradle.kts
sschuberth: Correct.

build.gradle.kts
sschuberth: It would fail.

mnonnenmacher (approved):

PR #307, opened 2024-05-28
feat(reporter): Add support for how-to-fix texts

bs-ondem:

### DIFF
--- a/workers/reporter/src/main/kotlin/reporter/ReporterRunner.kt
+++ b/workers/reporter/src/main/kotlin/reporter/ReporterRunner.kt
@@ -63,11 +63,13 @@ import org.ossreviewtoolkit.model.utils.setResolutions
 import org.ossreviewtoolkit.plugins.packageconfigurationproviders.api.PackageConfigurationProviderFactory
 import org.ossreviewtoolkit.plugins.packageconfigurationproviders.api.SimplePackageConfigurationProvider
 import org.ossreviewtoolkit.reporter.DefaultLicenseTextProvider
+import org.ossreviewtoolkit.reporter.HowToFixTextProvider
 import org.ossreviewtoolkit.reporter.LicenseTextProvider
 import org.ossreviewtoolkit.reporter.Reporter
 import org.ossreviewtoolkit.reporter.ReporterInput
 import org.ossreviewtoolkit.utils.common.safeMkdirs
 import org.ossreviewtoolkit.utils.ort.ORT_COPYRIGHT_GARBAGE_FILENAME
+import org.ossreviewtoolkit.utils.ort.ORT_HOW_TO_FIX_TEXT_PROVIDER_FILENAME
 import org.ossreviewtoolkit.utils.ort.ORT_LICENSE_CLASSIFICATIONS_FILENAME
 import org.ossreviewtoolkit.utils.ort.ORT_RESOLUTIONS_FILENAME
 import org.ossreviewtoolkit.utils.ort.showStackTrace
@@ -152,6 +154,19 @@ class ReporterRunner(
                 resolvedOrtResult = resolvedOrtResult.setResolutions(resolutionProvider)
             }
 
+            val howToFixTextProviderScript = config.howToFixTextProviderFile
+                ?.let { configManager.getFileAsString(context.resolvedConfigurationContext, Path(it)) }
+                ?: configManager.getFileAsString(
+                    context.resolvedConfigurationContext,
+                    Path(ORT_HOW_TO_FIX_TEXT_PROVIDER_FILENAME)
+                )
+
+            val howToFixTextProvider = if (howToFixTextProviderScript.isNotEmpty()) {
+                HowToFixTextProvider.fromKotlinScript(howToFixTextProviderScript, resolvedOrtResult)
+            } else {
+                HowToFixTextProvider.NONE
+            }
+
             val outputDir = context.createTempDir()
 
             val (successes, failures) = runBlocking(Dispatchers.IO) {
@@ -172,6 +187,7 @@ class ReporterRunner(
                         copyrightGarbage = copyrightGarbage,
                         licenseClassifications = licenseClassifications,
                         licenseTextProvider = createLicenseTextProvider(context, config),
+                        howToFixTextProvider = howToFixTextProvider
                     )
                 }
 
--- a/workers/reporter/src/test/kotlin/reporter/ReporterRunnerTest.kt
+++ b/workers/reporter/src/test/kotlin/reporter/ReporterRunnerTest.kt
@@ -40,7 +40,7 @@ import io.mockk.mockk
 import io.mockk.mockkObject
 import io.mockk.runs
 import io.mockk.slot
-import io.mockk.unmockkObject
+import io.mockk.unmockkAll
 import io.mockk.verify
 
 import java.io.File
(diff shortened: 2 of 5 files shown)


workers/reporter/src/main/kotlin/reporter/ReporterRunner.kt
```diff
@@ -152,6 +154,23 @@ class ReporterRunner(
...
+                    context.resolvedConfigurationContext,
+                    Path(howToFixTextProviderFile)
+                )
+
+                if (script.isNotEmpty()) {
+                    HowToFixTextProvider.fromKotlinScript(script, resolvedOrtResult)
+                } else {
+                    HowToFixTextProvider.NONE
+                }
+            }.getOrDefault(HowToFixTextProvider.NONE)
```
oheger-bosch, line 172: If an exception occurs in this block, it should at least be logged.

workers/reporter/src/main/kotlin/reporter/ReporterRunner.kt
```diff
@@ -152,6 +154,23 @@ class ReporterRunner(
                 resolvedOrtResult = resolvedOrtResult.setResolutions(resolutionProvider)
             }
 
+            val howToFixTextProviderFile =
+                config.howToFixTextProviderFile ?: ORT_HOW_TO_FIX_TEXT_PROVIDER_FILENAME.also {
+                    logger.info("No how-to-fix Kotlin script path provided, using default path '$it'.")
+                }
+            val howToFixTextProvider = runCatching {
```
oheger-bosch, line 161: Tests are missing for this new functionality.

oheger-bosch (changes requested):

workers/reporter/src/test/kotlin/reporter/ReporterRunnerTest.kt
```diff
@@ -696,6 +698,52 @@ class ReporterRunnerTest : WordSpec({
...
+
+            val reporter = reporterMock(format)
+            every { reporter.generateReport(any(), any(), any()) } returns listOf(tempfile())
+
+            mockReportersAll(format to reporter)
+
+            // Mock the HowToFixTextProvider.fromKotlinScript function to return the mocked object.
+            val mockHowToFixTextProvider = HowToFixTextProvider { "A test How-To-Fix text." }
+            mockkObject(HowToFixTextProvider)
+            every { HowToFixTextProvider.fromKotlinScript(any(), any()) } returns mockHowToFixTextProvider
```
oheger-bosch, line 718: Can the parameters be more specific? It is not very meaningful to have `any()` everywhere.

workers/reporter/src/test/kotlin/reporter/ReporterRunnerTest.kt
```diff
@@ -696,6 +698,52 @@ class ReporterRunnerTest : WordSpec({
...
+            )
+
+            val reporter = reporterMock(format)
+            every { reporter.generateReport(any(), any(), any()) } returns listOf(tempfile())
+
+            mockReportersAll(format to reporter)
+
+            // Mock the HowToFixTextProvider.fromKotlinScript function to return the mocked object.
+            val mockHowToFixTextProvider = HowToFixTextProvider { "A test How-To-Fix text." }
+            mockkObject(HowToFixTextProvider)
```
oheger-bosch, line 717: The `afterEach` block needs to be extended to unmock this object again to avoid interference with other tests. Using `unmockAll` in `afterEach` would also be an option.

oheger-bosch (changes requested):
Some minor issues in the test.

One question regarding error handling: Now the exception is logged when the provider cannot be created, which is fine. But I wonder whether this is sufficient. If there is a problem with the configuration or anything else, we will probably not notice, and then the users will get no how-to-fix hints. So, should we rather fail in this case?

bs-ondem:
> So, should we rather fail in this case?

So, you would rather remove the `runCatching` block and let `getFileAsString` throw any occurring exception? I also considered this approach but wasn't entirely sure. However, you're right. If there's a compilation issue while creating the `HowToFixTextProvider`, that exception would also be swallowed.

oheger-bosch:
> So, you would rather remove the `runCatching` block and let `getFileAsString` throw any occurring exception? I also considered this approach but wasn't entirely sure. However, you're right. If there's a compilation issue while creating the `HowToFixTextProvider`, that exception would also be swallowed.

I am also not sure, but have a tendency to let it fail. If the script is broken, this is a severe bug on our side and should be noticed and fixed fast. The downside is that then no reports can be created.

bs-ondem:
What about following the same approach like for the other configuration files? If there is an issue while reading (`ConfigException) the file make a log output, and if there is an compilation error while creating the `HowToFixTextProvider` throw the exception?

oheger-bosch:
> > What about following the same approach like for the other configuration files? If there is an issue while reading (`ConfigException`) the file make a log output, and if there is an compilation error while creating the `HowToFixTextProvider` throw the exception?
> 
> I would agree that the behavior should be the same as for the other configuration files. The logic used there is explained in the documentation of the `ConfigManager.readConfigFileWithDefault()` extension function:
...

The approach taken here should be consistent with the handling of other files.

That said, I have to admit that I find the logic applied for other configurations not really intuitive. What is the advantage of falling back to default names for configuration files and then having additional logic to handle the case that they are missing? It would be more straight-forward to use the following approach:
* If the name of a configuration file is provided, read it and throw an exception if this fails.
* If no name is provided, use the default value for the configuration.

mnonnenmacher:
> That said, I have to admit that I find the logic applied for other configurations not really intuitive. What is the advantage of falling back to default names for configuration files and then having additional logic to handle the case that they are missing? It would be more straight-forward to use the following approach:
> 
>     * If the name of a configuration file is provided, read it and throw an exception if this fails.
>     * If no name is provided, use the default value for the configuration.

The default values are used in case there is no configuration file at the default location. For example, if no `copyright-garbage.yml` exists, the fallback value is an empty `CopyrightGarbage` object because `ReporterInput.copyrightGarbage` is not nullable. For the how to fix text provider the fallback should be `HowToFixTextProvider.NONE`.

oheger-bosch:
> The default values are used in case there is no configuration file at the default location. For example, if no `copyright-garbage.yml` exists, the fallback value is an empty `CopyrightGarbage` object because `ReporterInput.copyrightGarbage` is not nullable. For the how to fix text provider the fallback should be `HowToFixTextProvider.NONE`.

Sure. What I question is whether there needs to be a default location for optional configuration files. Read the file from the provided location or use the default value if no location is defined.

mnonnenmacher:
> > The default values are used in case there is no configuration file at the default location. For example, if no `copyright-garbage.yml` exists, the fallback value is an empty `CopyrightGarbage` object because `ReporterInput.copyrightGarbage` is not nullable. For the how to fix text provider the fallback should be `HowToFixTextProvider.NONE`.
> 
> Sure. What I question is whether there needs to be a default location for optional configuration files. Read the file from the provided location or use the default value if no location is defined.

Default locations are a convenience feature for me, without them the user would have to supply all config file locations for every single request.

oheger-bosch:
> Default locations are a convenience feature for me, without them the user would have to supply all config file locations for every single request.

The current API is not really convenient; I guess, there will be some refactoring done. The drawback of the default locations is that a caller then cannot turn off configuration if the files are present.

oheger-bosch (approved):

PR #310, opened 2024-05-28
Add a Scaleway secrets provider

sschuberth:


sschuberth:
Note that we need this urgently, and thus no sophisticated tests are added that do not work without a real Scaleway account. But I've run the tests with my credentials and they pass. 

There probably also is not much point in capturing responses via WireMock of so, as the Scaleway API is still in beta, and things might change. So we'll best notice this in production 😬 

Anyway, this implementation does not affect anyone who's not actively using the Scaleway secrets manager. So the risk to merge should be very low.

secrets/scaleway/src/main/kotlin/ScalewayConfiguration.kt
```diff
@@ -34,6 +39,31 @@ data class ScalewayConfiguration(
 
         val DEFAULT_API_VERSION = ScalewayApiVersion.V1_BETA1
         val DEFAULT_REGION = ScalewayRegion.FRANCE_PARIS
+
+        fun create(configManager: ConfigManager): ScalewayConfiguration =
+            with(serializer().descriptor) {
+                val apiVersionName = getElementName(1)
```
mnonnenmacher, line 45: What's the benefit of referencing the properties by their index instead of name? It looks like you are just replacing hard-coded names with magic numbers.

sschuberth: In a way you're right, but at least this way, if a property name changes, the hard-coded string does not need to be renamed as well. I saw all the repetition via constants in the `VaultConfiguration` and didn't like it, so I was looking for a different approach.

mnonnenmacher: But the code looks more cryptic and if the property order changes you would have the same issue which might be even harder to find. In both cases there should be tests that prevent errors when properties change.

> I saw all the repetition via constants in the VaultConfiguration and didn't like it,

I don't like that either, maybe we could think about a generic approach that uses reflection and does not depend on kotlinx.serialization in future.

Anyway, as this code is isolated and it's urgent for you I don't want to block this here.

sschuberth: > I don't like that either, maybe we could think about a generic approach that uses reflection and does not depend on kotlinx.serialization in future.

Avoid reflection (which would not work multi-platform) by leveraging kotlinx.serialization was exactly the charming aspect of it. But I agree that a generic solution built into ConfigManager or so would be nicer.

mnonnenmacher:
You have to add the config properties to the `secretsProvider` sections in all `application.conf` files that have this section, otherwise there is no way to configure the new provider. If you are touching this anyway, maybe take the chance and add some comments for better overview, like this:
```
secretsProvider {
  name = "vault"
  name = ${?SECRETS_PROVIDER_NAME}

  # Properties for the vault provider.
  vaultUri = ${?VAULT_URI}
  vaultRoleId = ${?VAULT_ROLE_ID}
  vaultSecretId = ${?VAULT_SECRET_ID}
  vaultRootPath = ${?VAULT_ROOT_PATH}
  vaultPrefix = ${?VAULT_PREFIX}
  vaultNamespace = ${?VAULT_NAMESPACE}

  # Properties for the file based provider.
  fileBasedPath = ${?FILE_BASED_PATH}

  # Properties for the scaleway provider.
  ...
}
```

secrets/scaleway/src/main/kotlin/ScalewayConfiguration.kt
sschuberth: > I you prefer hard-coding the property names, I can do that quickly.

Done.

sschuberth:
> You have to add the config properties to the `secretsProvider` sections in all `application.conf` files that have this section

Ok, this means I should use a prefix to group / disambiguate property names in the config, which in turn means config properties are named differently than class properties, and thus my kontlix.serialization approach would not have worked anyway.

mnonnenmacher:
> > You have to add the config properties to the `secretsProvider` sections in all `application.conf` files that have this section
> 
> Ok, this means I should use a prefix to group / disambiguate property names in the config, which in turn means config properties are named differently than class properties, and thus my kontlix.serialization approach would not have worked anyway.

Right, that might well be the case. I'm only just getting a deeper look into the configuration code because of the Slack discussion we had today.

oheger-bosch:
Could you then add a disclaimer to the class that it accesses a preview API and use on your own risk or something like that. Also the test should maybe have some TODO comments that it should be reworked later. Would it make sense to add the code to an `experimental` package?

sschuberth:
> Could you then add a disclaimer to the class that it accesses a preview API and use on your own risk or something like that. Also the test should maybe have some TODO comments that it should be reworked later.

Done.

> Would it make sense to add the code to an `experimental` package?

I'd prefer not to, to keep the refactoring effort a bit lower once the API becomes stable. Also, I'd not consider the client implementation per se to be experimental, but the server-side API it talks to.

oheger-bosch:
> Done.
> 
> > Would it make sense to add the code to an `experimental` package?
> 
> I'd prefer not to, to keep the refactoring effort a bit lower once the API becomes stable. Also, I'd not consider the client implementation per se to be experimental, but the server-side API it talks to.

Ok, thanks. For me it's fine now. Leaving the approval to @mnonnenmacher .

mnonnenmacher (approved):

PR #314, opened 2024-05-28
Publish UI Docker image

mnonnenmacher:
Publish UI Docker image and add it to Docker Compose.

The image build was tested in a fork, the resulting image can be found here:
https://github.com/mnonnenmacher/ort-server/pkgs/container/ort-server-ui

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -18,50 +18,66 @@ jobs:
       packages: write
     strategy:
       matrix:
+        # Define the Docker images to build. The job first builds an image using Docker which is used for the ui and for
+        # worker base images. Then the job builds an image using Jib. Both steps are optional, e.g., not all Jib builds
+        # need a base image, and the ui build does not need a Jib build.
+        #
+        # If the image is built with Docker:
+        # preparationTask (optional): Gradle task to run before building the image.
+        # image: Name of the image, used for the Docker image name.
+        # context: Path to the Docker context directory.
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-build.yml
```diff
@@ -134,4 +156,4 @@ jobs:
           -Djib.console=plain \
           -Djib.container.labels="$(echo "${{ steps.meta.outputs.labels }}" | tr '\n' ',' | sed 's/,$//')" \
           -Djib.to.tags="$(echo "${{ steps.meta.outputs.tags }}" | tr '\n' ',' | sed 's/,$//')"
-        docker push ${{ env.REGISTRY }}/${{ github.repository_owner }}/ort-server-${{ matrix.docker.image }} --all-tags
+        docker push ${{ env.REGISTRY }}/${{ github.repository_owner }}/ort-server-${{ matrix.docker.jibImage }} --all-tags
```
bs-ondem, line 159: Just curious: Why are you pushing the images with `docker push` instead of using `Jib`'s push functionality?

mnonnenmacher: I had some issues with using the `jib` task for pushing, but I don't remember the details. I think the problem was that when pushing multiple tags with `jib` they appeared as separate images in the GitHub packages UI. When pushing multiple tags with docker only one image with multiple tags is shown.

bs-ondem (approved):

PR #315, opened 2024-05-28
fix(ui): Identify ORT runs by their index instead of id

mnonnenmacher:
ORT runs have a global id and a repository specific index. Use the index instead of the id when navigating to a run, as the index is used by the API endpoint to get ORT run details. For report downloads still the global id has to be used.

This fixes the issue that the links to ORT runs were broken if runs for multiple repositories exist.

### DIFF
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -69,7 +69,7 @@ export const Header = () => {
   const runMatch = matches.find(
     (match) =>
       match.routeId ===
-      '/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runId'
+      '/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex'
   );
 
   return (
--- a/ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.index.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.index.tsx
@@ -222,13 +222,13 @@ const RepoComponent = () => {
                       <div>
                         <Link
                           to={
-                            '/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runId'
+                            '/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex'
                           }
                           params={{
                             orgId: params.orgId,
                             productId: params.productId,
                             repoId: repo.id.toString(),
-                            runId: run.id.toString(),
+                            runIndex: run.index.toString(),
                           }}
                           className='font-semibold text-blue-400 hover:underline'
                         >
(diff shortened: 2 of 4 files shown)


Etsija:
I went through the code and seems OK. I would like to test this out with a small enough GitHub repository that has an `.ort.yml` in it, as I've only created ORT runs for DOS repository so far, so need to create another product and do an ORT run for it. Any suggestions? An ORT run for ORT itself or ORT Server might take hours to complete. jshttp/mime-types would be an excellent repo to use, but is there a way to exclude devDependencies from the run without the `.ort.yml` present in that repository?

mnonnenmacher:
> I went through the code and seems OK. I would like to test this out with a small enough GitHub repository that has an `.ort.yml` in it, as I've only created ORT runs for DOS repository so far, so need to create another product and do an ORT run for it. Any suggestions? An ORT run for ORT itself or ORT Server might take hours to complete. jshttp/mime-types would be an excellent repo to use, but is there a way to exclude devDependencies from the run without the `.ort.yml` present in that repository?

I'm using this repository for testing a lot, it does not have an .ort.yml but only a few dependencies so the first scan will take a while but not too long: https://github.com/oss-review-toolkit/ort-test-data-npm

Etsija (approved):

Etsija:
@mnonnenmacher 
- created a new org + product
- created a new repository for that product (used your suggestion)
- created an ORT run, which was successful, and the run index is 1 as expected
- was able to download run results

All in all, this PR is OK. Left my approval.

mnonnenmacher:
> I don't understand this commit.
> 
> How can a link that uses a ORT run global ID be broken if there is one unique ID for one ORT run ?

That's the crucial sentence from the commit message: "Use the index instead of the id when navigating to a run, as the index is used by the API endpoint to get ORT run details." So the bug was that the id was passed to that endpoint instead of the index. This worked by accident in case only a single repository was used, because then the id and index are identical. As soon as there ORT runs for more than one repository the links were broken.

nnobelis:
I understand.

Maybe add to the commit message this part "This worked by accident in case only a single repository was used, because then the id and index are identical. As soon as there ORT runs for more than one repository the links were broken."

I think it helps the understanding.

Etsija:
Also I clearly see a need to start adding some e2e UI tests to our project, as these kind of CRUD errors would've been caught - but so far I've only done the testing manually, with a single repository.

Anyway, I'd still like to postpone this kind of tests until also ORT runs can be deleted, as without it, the CRUD tests wouldn't be complete, and when running the tests locally I would end up with an increasing number of runs which I cannot delete.

(As a side note, I would rank the ORT run deletion feature quite high on my list.)

bs-ondem (approved):
Since @Etsija tested the functionality and your commit message has been improved, I will approve your PR.

PR #318, opened 2024-05-29
feat(dev) : Add some example HTTP requests

nnobelis:


sschuberth:
Commit message: The space before the colon needs to go away. And maybe use "scripts" instead of "dev" as the scope?

mnonnenmacher:
> And maybe use "scripts" instead of "dev" as the scope?

@MarcelBochtler used `feat(requests)` for the last change to this file.

nnobelis:
@mnonnenmacher no he used `feat(dev)`: https://github.com/eclipse-apoapsis/ort-server/commit/1f4d5a5f5698235a2d84e1c9199559020b0c8a6d

mnonnenmacher:
> @mnonnenmacher no he used `feat(dev)`: [1f4d5a5](https://github.com/eclipse-apoapsis/ort-server/commit/1f4d5a5f5698235a2d84e1c9199559020b0c8a6d)

Only for the first commit, later he used `feat(requests)`: 4bf76e3301377c8e3be540ee19178f32af7646ab

scripts/requests/organizations.http
```diff
@@ -15,3 +15,42 @@ Content-Type: application/json
   "name": "Example Organization",
   "description": "This is an example organization"
 }
+
+### Create an Organization secret
+POST {{host}}/{{organizationId}}/secrets
+
```
mnonnenmacher, line 21: Nit: For consistency remove the empty lines above `Authorization`.

scripts/requests/repositories.http
```diff
@@ -32,13 +32,35 @@ Content-Type: application/json
...
+      "useClearlyDefined": "true",
+      "recipientAddresses": "recipient@example.com",
+      "fossIdServer": "https://fossid.example.com/instance/",
+      "fossIdProjectName": "FossID-Projects",
+      "fossIdWaitForResult": "true",
+      "fossIdKeepFailedScans": "false",
+      "fossIdUseDeltaScans":  "true",
+      "fossIdDeltaScanLimit": "5",
+      "snippetReport": "true",
+      "fossIdFetchSnippetMatchedLines": "true"
```
mnonnenmacher, line 60: These are all Bosch internal parameters that require a special `.params.kts` script. If we want to have request templates for internal testing we should create them in an internal repository.

mmurto:
> > @mnonnenmacher no he used `feat(dev)`: [1f4d5a5](https://github.com/eclipse-apoapsis/ort-server/commit/1f4d5a5f5698235a2d84e1c9199559020b0c8a6d)
> 
> Only for the first commit, later he used `feat(requests)`: [4bf76e3](https://github.com/eclipse-apoapsis/ort-server/commit/4bf76e3301377c8e3be540ee19178f32af7646ab)
> 
> In general the scopes should refer to modules in the project, from this perspective `scripts` or `requests` make more sense than `dev`. But that's really just a nit, we haven't even documented yet what scopes should be used.

Semi-related, and it hasn't been documented for the project yet either, but for [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), which seem to be used for the project, a `feat` usually meanse introduction of a new feature to the codebase, correlating with `MINOR` in Semantic Versioning. Determining version numbers and writing changelogs is IMO the primary point of Conventional Commits.

If used like that, it makes no sense to mark this kind of commits as `feat`, as this commit should not really increment the version number of ORT Server, if it is published with a version number. This kind of commits should IMO be either `chore` or `docs`.

scripts/requests/repositories.http
```diff
@@ -32,13 +32,35 @@ Content-Type: application/json
...
+      "skipExcluded": true
+    },
+    "advisor": {
+    },
+    "scanner": {
+    },
+    "evaluator": {
+    },
+    "reporter": {
+      "skipExcluded": true
```
mnonnenmacher, line 47: The reporter does not have a `skipExcluded` option, only the analyzer, advisor and scanner have.

mnonnenmacher (changes requested):

scripts/requests/repositories.http
nnobelis: Can we put the options from ORT `reference.yml` instead ?

mnonnenmacher: You could do that, but then that request would not help you for internal testing.

nnobelis:
Since the examples requests contain some Bosch specifics, they will be added to an internal project instead.

PR #322, opened 2024-05-30
ui: Include advisor in an ORT Run

Etsija:
Add Advisor job to the ORT runs (enabled by default, default advisors OSV and VulnerableCode).

It has been verified with [this repository](https://github.com/oss-review-toolkit/ort-test-data-npm.git) that the Advisor is started, it runs parallel to the Scanner, and the job running parallel is indicated correctly in the UI by "pulsing traffic lights" both for the Advisor and Scanner jobs. The Advisor produces Vulnerabilities into WebApp as indicated in this screenshot:

Capture

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ First, ensure to have [Docker BuildKit](https://docs.docker.com/build/buildkit/)
 To build the Docker image for the UI, the OpenAPI specification must be generated first:
 
 ```shell
-./gradlew :ui:generateOpenApiSpec
+./gradlew :core:generateOpenApiSpec
 ```
 
 Then build the base images for the workers which contain the external tools and required configuration either via the Gradle task
--- a/ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
@@ -64,6 +64,11 @@ const formSchema = z.object({
       skipExcluded: z.boolean(),
       enabledPackageManagers: z.array(z.string()),
     }),
+    advisor: z.object({
+      enabled: z.boolean(),
+      skipExcluded: z.boolean(),
+      advisors: z.array(z.string()),
+    }),
     scanner: z.object({
       enabled: z.boolean(),
       skipConcluded: z.boolean(),
@@ -78,6 +83,21 @@ const formSchema = z.object({
     }),
   }),
 });
+
+const advisors = [
+  {
+    id: 'OssIndex',
+    label: 'OSS Index',
+  },
+  {
+    id: 'OSV',
+    label: 'OSV',
+  },
+  {
+    id: 'VulnerableCode',
+    label: 'VulnerableCode',
+  },
+] as const;
 
 const packageManagers = [
   {
@@ -193,6 +213,11 @@ const CreateRunPage = () => {
           skipExcluded: true,
           enabledPackageManagers: packageManagers.map((pm) => pm.id),
         },
+        advisor: {
+          enabled: true,
+          skipExcluded: true,
+          advisors: ['OSV', 'VulnerableCode'],
+        },
         scanner: {
           enabled: true,
           skipConcluded: true,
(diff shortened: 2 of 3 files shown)


ui/.gitignore
```diff
@@ -8,3 +8,4 @@ dist
 
 src/api/
 src/routeTree.gen.ts
+build/
```
sschuberth, line 11: Commit message nits:
- "openApi" -> "OpenAPI"
- "when running theformatting checks" -> "when running the `format:check` script"
- Put `.gitignore` into backticks to improve formatting.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -64,6 +64,11 @@ const formSchema = z.object({
       skipExcluded: z.boolean(),
       enabledPackageManagers: z.array(z.string()),
     }),
+    advisor: z.object({
```
sschuberth, line 67: Commit message:

> The default advisors are OSV and VulnerableCode, 

That's not strictly correct; ORT core itself has no default advisors. It's just the Jenkins example pipeline that configures these by default. And this makes sense, but we should change the word to:

"Enable the advisor job with OSV and VulnerableCode advisors by default to get some meaningful vulnerability results."

Or so.

sschuberth (changes requested):

ui/.gitignore
mnonnenmacher: You could just add the file to the `.prettierignore` instead to avoid the "duplication" in the `.gitignore`.

README.md
mnonnenmacher, line 1: Commit message: Please use `docs(README): ` instead, `fix` should only be used for code fixes. We still need to document that.

mnonnenmacher (changes requested):

ui/.gitignore
mmurto: IMO the logic is more sound if `.gitignore` includes all gitignored files and Prettier ignores files that are gitignored based on that. For files that are tracked in Git but for some reason should still not be formatted, like lockfiles, `.prettierignore` makes sense.

mnonnenmacher: Does adding it to `.gitignore` actually work? Because in 1320b42a954b59c0958559bbafe776c70745248a I have added the `--ignore-file .prettierignore` option so that the generated API code is still formatted even though it is in `.gitignore`.
And I tend to agree with the creator of https://github.com/prettier/prettier/issues/15438 that using `.gitignore` for Prettier is problematic, because it mixes the configuration of different tools for different purposes, potentially leading to conflicts.

Etsija: I believe you're right, Martin, we did discuss this earlier, and I actually would prefer your proposal over the `.gitignore` one.

mmurto: > Does adding it to `.gitignore` actually work? Because in [1320b42](https://github.com/eclipse-apoapsis/ort-server/commit/1320b42a954b59c0958559bbafe776c70745248a) I have added the `--ignore-file .prettierignore` option so that the generated API code is still formatted even though it is in `.gitignore`.

It would work - by setting the `--ignore-file` the default ignore files (`.gitignore,.prettierignore`) are not used, so `.gitignore` is not used for that command at all.

> And I tend to agree with the creator of [prettier/prettier#15438](https://github.com/prettier/prettier/issues/15438) that using `.gitignore` for Prettier is problematic, because it mixes the configuration of different tools for different purposes, potentially leading to conflicts.

Yep, looks problematic, though it works as it works. We could surely use `--ignore-file .prettierignore` with all formatting calls and if we want to save computer cycles, copy `node_modules` etc. to `.prettierignore`.

> Mikko, do you succumb?

Sure, no strong opinions, just feels more logical to me to use `.prettierignore` for non-gitignored files, but totally fine by me either way. :)

Etsija: I decided to redo this the way Martin suggested; now done.

mnonnenmacher (approved):
@Etsija Please give a thumbs up to review comments you have addressed and use the "Re-request review" button next to the reviewer names on the top right. This makes it easier for the reviewer to see if you are done with addressing comments or still working on it.

Etsija:
> @Etsija Please give a thumbs up to review comments you have addressed and use the "Re-request review" button next to the reviewer names on the top right. This makes it easier for the reviewer to see if you are done with addressing comments or still working on it.

Thanks for the advice - been reluctant to ping busy colleagues.

PR #323, opened 2024-05-30
chore(gitignore): Ignore some directories at the top-level only

sschuberth:
Reduce potential side-effects due to matching in sub-directories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
-.gradle/
-.idea/
-.kotlin/
+/.gradle/
+/.idea/
+/.kotlin/
 build/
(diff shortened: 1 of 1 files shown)


sschuberth:
@mmurto @Etsija should `.vscode/` also only be ignored at the root level? Or could it be created i the `ui` sub-directory?

mmurto:
> @mmurto @Etsija should `.vscode/` also only be ignored at the root level? Or could it be created i the `ui` sub-directory?

Root only should be enough.

mnonnenmacher (approved):

Issue #256, opened 2024-05-20
Align on either Markdown or Asciidoc documentation


sschuberth:
The project currently makes use of both Markdown (4 files) and Asciidoc (22 files) for documentation. For consistency, documentation should be aligned on either format.

mnonnenmacher:
I would prefer Markdown over Asciidoc for the following reasons:
* Markdown is far more popular for documentation on GitHub, so more developers are familiar with the syntax.
* The two tools I would prefer to render the documentation to a website both use markdown (Docusaurus, because it uses React like our frontend, and because it's used for ORT; mkdocs-material, because it seems more streamlined than Docusaurus).
* GitHub supports rendering Mermaid diagrams in Markdown, but not PUML diagrams in Asciidoc.

One thing where Asciidoc is clearly better than Markdown is the table syntax, that is terrible in Markdown.

sschuberth:
Another plus-side of using Asciidoc is that it's also what [ORT's template reporter](https://github.com/oss-review-toolkit/ort/blob/main/plugins/reporters/asciidoc/src/main/kotlin/AsciiDocTemplateReporter.kt) uses, so getting familiar with the syntax could be beneficial.

oheger-bosch:
My preference is Asciidoc, because in my opinion it offers a great balance between complexity and feature richness - meaning that for typical use cases it is easy to use (and learn) while offering really nice functionality. Its syntax is clear and stringent; some quirks of the Markdown syntax have already been mentioned.

I am not familiar with Mermaid, but from looking at the website, I am not sure whether it supports all the diagram types we use in the documentation. For instance, PlantUML supports complex deployment diagrams with default elements for nodes, databases, message queues, etc.

Another benefit I see is the support in tools like [docToolchain](https://doctoolchain.org/docToolchain/v2.0.x/). Here multiple documents can be combined to generate documentation in different formats, including websites. We use this internally to deploy our arc42 architecture documentation to GitHub pages. Such an approach could be used to solve the problem that PUML is not rendered natively by GitHub.

mnonnenmacher:
> My preference is Asciidoc, because in my opinion it offers a great balance between complexity and feature richness - meaning that for typical use cases it is easy to use (and learn) while offering really nice functionality. Its syntax is clear and stringent; some quirks of the Markdown syntax have already been mentioned.

Are there any specific issues with the Markdown syntax except for tables?

[...]

PR #328, opened 2024-05-30
docs: Convert README files from AsciiDoc to Markdown

mnonnenmacher:
Convert all readme files in the repository from AsciiDoc to Markdown to use the same lanugage for all of them. Markdown was chosen over AsciiDoc because it is the more commonly used format in open source projects and most developers are familiar with it.

While at it, format all of those files to use the one sentence per line approach to minimize the diffs when the files are edited.

The conversion was done by converting the files using asciidoctor and pandoc:

  asciidoctor -b docbook5 README.adoc
  pandoc -f docbook -t gfm README.xml -o README.md \
    --base-header-level=2 --wrap=preserve

Afterwards any problems were corrected manually.

Relates to #256.


sschuberth:
> Relates to #256.

Why only "Relates" and not "Resolves"?

mnonnenmacher:
> > Relates to #256.
> 
> Why only "Relates" and not "Resolves"?

The `docs` folder still contains adoc files. Those files I do not want to just convert but largely rewrite as they are pretty outdated.

bs-ondem (approved):

mmurto:
> While at it, format all of those files to use the one sentence per line approach to minimize the diffs when the files are edited.

One option (that I personally prefer, but opinions of course vary :)) would be to use Prettier for markdown files with a set line width like 80 or 100. That would make all Markdown files consistent, althogh the diffs may not be that good, even though IMO readable enough. It could also be nice to enable [Markdownlint](https://github.com/DavidAnson/markdownlint).

oheger-bosch:
Since I have contributed a significant part of the documentation, I would have been nice if I had been asked for my opinion as well.

mnonnenmacher:
> Since I have contributed a significant part of the documentation, it would have been nice if I had been asked for my opinion as well.

@oheger-bosch Sorry if that was against your preference, please add your perspective to #256, I have just added a comment explaining why I prefer Markdown.

PR #337, opened 2024-05-31
feat(transport): Add AWS SQS as a transport

sschuberth:
This is specifically for use with the Scaleway implementation of SQS, see [1].

[1]: https://www.scaleway.com/en/developers/api/messaging-and-queuing/sqs-api/


transport/sqs/src/main/kotlin/SqsConfig.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.eclipse.apoapsis.ortserver.config.Path
+
+internal const val SERVER_URI_PROP_NAME = "serverUri"
+internal const val ACCESS_KEY_ID_PROP_NAME = "accessKeyId"
+internal const val SECRET_ACCESS_KEY_PROP_NAME = "secretAccessKey"
+internal const val QUEUE_NAME_PROP_NAME = "queueName"
+
+/**
+ * This class defines the configuration properties for an SQS transport. Specifically, this targets the implementation
+ * at Scaleway, see https://www.scaleway.com/en/developers/api/messaging-and-queuing/sqs-api/.
```
mnonnenmacher, line 36: The message of the last sentence is a bit unclear to me. Does it mean that for AWS a different configuration will be required? Or was it just not tested with AWS?

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
```diff
@@ -0,0 +1,125 @@
...
+
+        val receiveRequest = ReceiveMessageRequest {
+            this.queueUrl = queueUrl
+
+            // Opt-in to the message attributes to receive.
+            messageAttributeNames = org.eclipse.apoapsis.ortserver.transport.sqs.messageAttributeNames
+
+            // Enable "long polling" to eliminate empty responses, see
+            // https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/working-with-messages.html#setting-up-long-polling
+            waitTimeSeconds = 1
```
mnonnenmacher, line 79: The linked docs say that in most cases this can be set to 20 seconds, why did you chose 1? IIUC, if a message arrives within the 20 seconds the function would still return the message immediately. Or would it always wait the full time before returning anything?

Also, I was thinking if we should explicitly set [`maxNumberOfMessages`](https://sdk.amazonaws.com/kotlin/api/latest/sqs/aws.sdk.kotlin.services.sqs.model/-receive-message-request/index.html) to 1, even if it is the default, to explicitly state it here.

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
```diff
@@ -0,0 +1,125 @@
...
+
+                runCatching {
+                    handler(Message(attrs.toMessageHeader(), serializer.fromJson(body)))
+
+                    val deleteRequest = DeleteMessageRequest {
+                        this.queueUrl = queueUrl
+                        receiptHandle = message.receiptHandle
+                    }
+
+                    client.deleteMessage(deleteRequest)
```
mnonnenmacher, line 102: I saw that there is a [`visibilityTimeout`](https://sdk.amazonaws.com/kotlin/api/latest/sqs/aws.sdk.kotlin.services.sqs.model/-receive-message-request/index.html) option to configure how long a message is invisible to subsequent retrieve requests, but at least the docs for the Kotlin SDK do not explain the behavior when this is set to `null` and what is the default. Could it happen that the message is received by another receiver if the `handler` call takes very long?

transport/sqs/src/test/kotlin/SqsMessageSenderFactoryTest.kt
```diff
@@ -0,0 +1,136 @@
...
+
+        client.createQueue(CreateQueueRequest { queueName = TEST_QUEUE_NAME })
+
+        shouldThrow<IOException> {
+            MessageSenderFactory.createSender(OrchestratorEndpoint, configManager)
+        }
+
+        verify { mockClient.close() }
+
+        unmockkStatic(::createSqsClient)
```
mnonnenmacher, line 134: This is not called if the test fails before which could potentially affect other tests. We usually call `unmockkAll()` in `afterTest` or `afterSpec` to prevent that. `try/finally` is also an option, but we only use it in helper functions.

transport/sqs/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+```
+endpoint {
+  sender/receiver: {
+    type = "SQS"
+    serverUri = "https://sqs.hostname.com:32803/"
+    queueName = "my_message_queue"
+  }
+}
+```
```
oheger-bosch, line 32: The documentation should probably also mention the secrets that are evaluated by the `SqsConfig` class.

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
oheger-bosch: I have the same question. Since message processing can take really long if a worker gets executed, can it have a negative impact if the message is deleted only afterward? Should deletion be done before calling the handler and how does this play together with error handling and message re-delivery?

transport/sqs/src/main/kotlin/SqsConfig.kt
sschuberth: Is was just not tested with AWS directly. I've clarified the sentence.

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
sschuberth: > why did you chose 1?

Simply because I was unsure how much delay would be acceptable for ORT server.

> Or would it always wait the full time before returning anything?

My understanding so far was that if the poll interval is set to 20s, that would mean what messages might not be received for 20s. But reading thought [this](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-short-and-long-polling.html) again it says

> Long polling - [...] sending a response once at least one message is available

So it seems that using the recommended 20s should be ok, I'll make that change.

> Also, I was thinking if we should explicitly set [maxNumberOfMessages](https://sdk.amazonaws.com/kotlin/api/latest/sqs/aws.sdk.kotlin.services.sqs.model/-receive-message-request/index.html) to 1, even if it is the default, to explicitly state it here.

Should we rather do the opposite and increase this, to allow for more message to be delivered at once, possibly increasing performance?

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
sschuberth: Counter-question: What is ORT server's expectation here? Should every message be only retceived once, by exactly one receiver?

sschuberth: Here's more documentation about the setting: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html

transport/sqs/src/test/kotlin/SqsMessageSenderFactoryTest.kt
sschuberth: I went for the `afterEach` approach.

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
mnonnenmacher: > Should we rather do the opposite and increase this, to allow for more message to be delivered at once, possibly increasing performance?

I think it's better to process messages immediately. For example, if a worker has finished, the orchestrator should start he next one immediately and not wait for other messages. And workers should not handle more than one message at a time to not limit parallelism (even though in Kubernetes the Kubernetes transport would typcially be used for "sending messages" from the orchestrator to the workers).

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
mnonnenmacher: > Counter-question: What is ORT server's expectation here? Should every message be only retceived once, by exactly one receiver?

In general, yes. If a receiver fails while processing a message it might make sense that another receiver can pick it up later on, but this whole scenario is a bit unclear to me, and without looking at the code I don't know if other transport implementations acknowledge messages immediately or only after handling them.

mnonnenmacher: The RabbitMQ docs linked above state that a message should be acknowledged after processing it:

> Consumer acknowledgements are not aware of publishers: their goal is to confirm to a RabbitMQ node that a given delivery was successfully received and processed successfully, so the delivered message can be marked for future deletion.

However, our RabbitMQ implementation uses autoAck which acknowledges the message as soon as it was received, so it would be consistent to do the same in other transports.

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
sschuberth: > I think it's better to process messages immediately.

I agree, but I'm not reading the docs for [maxNumberOfMessages](https://sdk.amazonaws.com/kotlin/api/latest/sqs/aws.sdk.kotlin.services.sqs.model/-receive-message-request/index.html) so that it would always wait for that number of messages:

> "Amazon SQS never returns more messages than this value (however, fewer messages might be returned)."

> And workers should not handle more than one message at a time

IMO they can't away, as the signature of the `handler` function only allows one message to be passed at a time. So the question rather is: How quickly is it allowed to call the `handler` function in succession? In case of multiple received messages, I'm just looping over them, and calling the handler for each separately.

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
sschuberth: > So I'll add code to delete the message after retrieval.

I totally forgot that this [is already in place](https://github.com/eclipse-apoapsis/ort-server/pull/337/files#diff-b5c5be3aad7e14978470edd9afcb0c5d7b9063767e7820cdf2b35e4d656001ebR100-R105) 😸

oheger-bosch: But then we are back to my question: Should the message be deleted before or after calling the handler? As it is currently, the message is not deleted if the handler crashes. Will it then be picked up again, and - if the problem persists - can an endless loop be entered?

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
oheger-bosch: This is transparent for the transport implementation. The handler function is blocking, so this means that it can handle the next message when it returns. For transport implementations based on message queues, this probably means that in order to process messages in parallel, a corresponding number of receivers will have to be deployed.

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
sschuberth: Quoting @mnonnenmacher:

> If a receiver fails while processing a message it might make sense that another receiver can pick it up later on

So I assumed the current behavior to delete the message only if the handler does not crash is the desired one.

> Will it then be picked up again, and - if the problem persists - can an endless loop be entered?

I believe we'd eventually have some repeated handling, but not real endless loop, as at some point the aforementioned [visibility timeout](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html) will make the message disappear. We could even [lower the visibility timeout](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html) after a message has been received but before it gets handled / deleted to reduce the number of potential repeated handlings if the handler crashes.

Just tell me how it should be implemented, and I'll do so.

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
sschuberth: > This is transparent for the transport implementation.

Exactly. The (my) question is not about processing messages in parallel or not, but about whether there's a difference (for the handler) to be called 10 times for 10 individually received messages, or 10 times for 10 messages received at once.

For the handler, this should not make a difference, as stated. But again my question remains, could it still be more efficient fro the AWS SDK perspective to allow receiving messages in bulk, even if they are passed one-by-one to the (blocking) handler?

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
sschuberth: > at some point the aforementioned [visibility timeout](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html) will make the message disappear.

Scratch that, I was misreading how it works: The message will just not get delivered during that time, but will never be automatically deleted from the queue.

sschuberth: Reading [more docs](https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/SQS/MSQSDeleteMessageDeleteMessageRequest.html): 

> Amazon SQS automatically deletes messages left in a queue longer than the retention period configured for the queue.

So should we simply rely on the queue's retention period here, or should we always delete the message before it's being handled?

sschuberth: I'm deleting the message now before it's being passed to the handler.

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
oheger-bosch: When the handler calls a worker which can take minutes for its execution, then I doubt that it has any measurable effect if the fetching of messages is done slightly more or less efficiently. It may make a difference if message processing was really fast. At least that would be my understanding.

oheger-bosch (dismissed):
Looks good to me in its current state.

sschuberth:
> Looks good to me in its current state.

I had to do another minor update to the exception handling.

sschuberth:
> Execution failed for task ':workers:notifier:test'.

This test failure seems to be unrelated. Interestingly, I cannot find any actually failing test in the log output.

oheger-bosch (approved):

PR #340, opened 2024-05-31
Transport test refactoring

sschuberth:

### DIFF
--- a/transport/activemqartemis/src/test/kotlin/ArtemisTestContainer.kt
+++ b/transport/activemqartemis/src/test/kotlin/ArtemisTestContainer.kt
@@ -36,10 +36,10 @@ private const val ARTEMIS_CONTAINER = "quay.io/artemiscloud/activemq-artemis-bro
 private const val ARTEMIS_PORT = 61616
 
 /**
- * Extension function to start an Artemis broker in a test container for testing a factory of the given
- * [transportType]. The resulting [ConfigManager] can be used to connect to the broker in the container.
+ * Extension function to start an Artemis broker in a test container for testing a factory of the given [consumerName]
+ * and [transportType]. The resulting [ConfigManager] can be used to connect to the broker in the container.
  */
-fun Spec.startArtemisContainer(transportType: String): ConfigManager {
+fun Spec.startArtemisContainer(consumerName: String, transportType: String): ConfigManager {
     val containerEnv = mapOf("AMQ_USER" to "admin", "AMQ_PASSWORD" to "admin")
     val artemisContainer = install(
         ContainerExtension(
@@ -52,7 +52,7 @@ fun Spec.startArtemisContainer(transportType: String): ConfigManager {
         )
     )
 
-    val keyPrefix = "orchestrator.$transportType"
+    val keyPrefix = "$consumerName.$transportType"
     val configMap = mapOf(
         "$keyPrefix.serverUri" to "amqp://${artemisContainer.host}:${artemisContainer.firstMappedPort}",
         "$keyPrefix.queueName" to "testQueue",
--- a/transport/activemqartemis/src/test/kotlin/ArtemisTestContainer.kt
+++ b/transport/activemqartemis/src/test/kotlin/ArtemisTestContainer.kt
@@ -19,14 +19,12 @@
 
 package org.eclipse.apoapsis.ortserver.transport.artemis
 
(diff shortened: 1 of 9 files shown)


transport/rabbitmq/src/test/kotlin/RabbitMqMessageReceiverFactoryTest.kt
```diff
@@ -156,17 +155,18 @@ class RabbitMqMessageReceiverFactoryTest : StringSpec() {
...
-        )
 
-        return ConfigManager.create(ConfigFactory.parseMap(configMap))
+        return createConfigManager(
+            "orchestrator",
+            "receiver",
+            "rabbitMQ",
+            "amqp://${rabbitMq.host}:${rabbitMq.firstMappedPort}",
+            "orchestrator.receiver.queueName" to queueName,
+        )
```
mnonnenmacher: Please use named arguments for readability.

transport/rabbitmq/src/test/kotlin/RabbitMqMessageSenderFactoryTest.kt
```diff
@@ -167,17 +166,18 @@ class RabbitMqMessageSenderFactoryTest : StringSpec() {
...
-        )
 
-        return ConfigManager.create(ConfigFactory.parseMap(configMap))
+        return createConfigManager(
+            "orchestrator",
+            "sender",
+            "rabbitMQ",
+            "amqp://${rabbitMq.host}:${rabbitMq.firstMappedPort}",
+            configProvidersMap
+        )
```
mnonnenmacher: Please use named arguments for readability.

transport/activemqartemis/src/test/kotlin/ArtemisTestContainer.kt
```diff
@@ -53,11 +51,10 @@ fun Spec.startArtemisContainer(consumerName: String, transportType: String): Con
...
-    val keyPrefix = "$consumerName.$transportType"
-    val configMap = mapOf(
-        "$keyPrefix.serverUri" to "amqp://${artemisContainer.host}:${artemisContainer.firstMappedPort}",
-        "$keyPrefix.queueName" to TEST_QUEUE_NAME,
-        "$keyPrefix.type" to "activeMQ"
+    return createConfigManager(
+        consumerName,
+        transportType,
+        "activeMQ",
+        "amqp://${artemisContainer.host}:${artemisContainer.firstMappedPort}"
```
mnonnenmacher, line 58: Please use named arguments for readability.

transport/rabbitmq/src/test/kotlin/RabbitMqTestContainer.kt
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Extension function to start an RabbitMq broker in a test container for testing a factory of the given [consumerName]
+ * and [transportType]. The resulting [ConfigManager] can be used to connect to the broker in the container.
+ */
+fun Spec.startRabbitMqContainer(consumerName: String, transportType: String): ConfigManager {
+    val rabbitMq = install(
+        ContainerExtension(
+            RabbitMQContainer("rabbitmq")
+        )
+    )
```
mnonnenmacher, line 41: How does it work, does Kotest automatically stop the container after the test if it is installed inside a test?

mnonnenmacher (changes requested):

mnonnenmacher (approved):

transport/rabbitmq/src/test/kotlin/RabbitMqTestContainer.kt
Kantis: Here's the source for [ContainerExtension](https://github.com/kotest/kotest-extensions-testcontainers/blob/master/src/main/kotlin/io/kotest/extensions/testcontainers/ContainerExtension.kt).

Looks like the default lifecycle is that the installed extension is active for the rest of the project. You might want to set `mode = ContainerLifecycleMode.Spec`

sschuberth: Thanks @Kantis for chiming in! However, I believe the default `Project` mode is the right one to use right now because

> This mode can be used with multiple test containers by using separate instances of the container extensions.

and we *are* currently using multiple test container instances in a single spec.

PR #358, opened 2024-06-04
ui: Add initial admin dashboard

lamppu:
This PR adds an initial admin dashboard for superusers in the UI. Please see the individual commits for details.


scripts/docker/keycloak/master-realm.json
```diff
@@ -963,6 +963,7 @@
       "consentRequired" : false,
       "config" : {
         "user.attribute" : "foo",
+        "id.token.claim": "true",
```
mnonnenmacher, line 966: Users can have a LOT of roles, we deliberately do not add the roles to the tokens, otherwise the header size limit could quickly be reached (and requests would become slow).
If this admin UI should only be accessible to users with the admin role, we could think about different solutions, like an endpoint to check if the users has the admin role, or adding just a specific role to the token (if that is possible).

lamppu: Yes, I did notice that, however this would only be needed for the id token, and not the access token that is used in the header of the requests.

lamppu: Of course other solutions can be thought of too, however I am wondering if there will be a need to hide/show some buttons etc. based on user roles for other users too later on, so it might not just be about who's an admin and who's not.

mnonnenmacher: > Yes, I did notice that, however this would only be needed for the id token, and not the access token that is used in the header of the requests.

I missed that detail, for the id token it should indeed be fine to add the roles.

But I also noticed that the roles are still added to the access token in the compose setup, it seems we only changed that in our internal config. That will at some point become an issue when more repositories are added.

mnonnenmacher (commented):
Code looks good to me. @Etsija @mmurto Could one of you please review as well?

@lamppu Could you share what your vision for the admin UI is? Meaning, what kind of features should it offer?

My vision for the UI structure is strongly influenced by GitHub, because I use it so much. Based on that my expectation was that in addition to the product list there would be a settings tab in the normal organization view where organization settings can be changed (only available with the required permissions of course). Similar for the product and repository pages.

So what I would expect in the admin UI would be global configuration options, e.g. to configure available advisors (later on), but not a list of organizations.

sschuberth:
> My vision for the UI structure is strongly influenced by GitHub

@mmurto will be pleased to hear that, as he's also repeatedly referred to GitHub's UI as a reference 😸 And for good reason!

mnonnenmacher:
> @mnonnenmacher I think our vision is quite similar, that the admin dashboard would allow you to configure some more global things, and which tools are available for which organizations and such. The reason for the organization list being there was to later move the creating and deleting organizations exclusively there, however there was some debate about this, so it's definitely up for discussion (as is everything else for that matter). And yes, I would expect that the organization/product/repository specific settings would be somehow configurable in their respective views.

To me it would be cleaner to keep the options to create and delete organizations where they currently are but hide them for non-admin users (which should be possible now with the changes from this PR). Maintaining two organization lists sounds like unnecessary effort. So how about not adding it, even if that means that the admin UI only has the user management tab for now?

lamppu:
> > @mnonnenmacher I think our vision is quite similar, that the admin dashboard would allow you to configure some more global things, and which tools are available for which organizations and such. The reason for the organization list being there was to later move the creating and deleting organizations exclusively there, however there was some debate about this, so it's definitely up for discussion (as is everything else for that matter). And yes, I would expect that the organization/product/repository specific settings would be somehow configurable in their respective views.
> 
> To me it would be cleaner to keep the options to create and delete organizations where they currently are but hide them for non-admin users (which should be possible now with the changes from this PR). Maintaining two organization lists sounds like unnecessary effort. So how about not adding it, even if that means that the admin UI only has the user management tab for now?

Makes sense to me, but double checking with @mmurto and @sschuberth if they have any thoughts on this

sschuberth:
> To me it would be cleaner to keep the options to create and delete organizations where they currently are but hide them for non-admin users

Could you explain a bit *why* this is cleaner for you? Isn't it cleaner if all admin settings are gathered in one place rather than spread across the whole UI, which then also changes on a per-role basis (which could be confusing to users with multiple roles)?

mnonnenmacher:
> > To me it would be cleaner to keep the options to create and delete organizations where they currently are but hide them for non-admin users
> 
> Could you explain a bit _why_ this is cleaner for you? Isn't it cleaner if all admin settings are gathered in one place rather than spread across the whole UI, which then also changes on a per-role basis (which could be confusing to users with multiple roles)?

For consistency, the buttons to create products or repositories are located at the product and repository lists, so it's more intuitive for the user if the button to create organizations is at a similar location. I would find it confusing to have two organization lists with different feature sets. For other admin options that's different, but currently I cannot think of any. The user management links make sense, but all other admin options I can think of still need to be implemented.

mmurto:
A quick comment on the organization list:

We discussed briefly whether it makes sense to have the organization list at `/` as prominently as it currently exists, or if some organization switcher in the navbar or user dropdown would make more sense. This is because, and now I'm guessing, the big majority of users will belong to only one organization. If that's the case, it probably doesn't make sense to reserve the home page for a list of one organization where the user always just clicks the org to see its products.

lamppu:
> > So shall i go ahead and merge it?
> 
> In any case the conflict in the lockfile needs to be resolved first.

Yeah I'll rebase now

mnonnenmacher (approved):

PR #369, opened 2024-06-06
feat(core): Set Swagger UI server URL from config

haikoschol:
This change introduces three new configuration variables that are used for setting the "server" value in the Swagger UI configuration. The new variables have default values that keep the existing behaviour.

This enables using Swagger UI in environments with some kind of proxy or load balancer in front of the ORT Server API.

### DIFF
--- a/core/src/main/kotlin/plugins/OpenApi.kt
+++ b/core/src/main/kotlin/plugins/OpenApi.kt
@@ -79,7 +79,10 @@ fun Application.configureOpenApi() {
         }
 
         server {
-            url = "http://localhost:8080"
+            val scheme = config.property("ktor.deployment.publicScheme").getString()
+            val fqdn = config.property("ktor.deployment.publicFqdn").getString()
+            val port = config.property("ktor.deployment.publicPort").getString()
+            url = "$scheme://$fqdn:$port"
             description = "Local ORT server"
         }
 
--- a/core/src/main/resources/application.conf
+++ b/core/src/main/resources/application.conf
@@ -23,6 +23,12 @@ configManager {
 
 ktor {
   deployment {
+    publicScheme = "http"
+    publicScheme = ${?PUBLIC_SCHEME}
+    publicFqdn = "localhost"
+    publicFqdn = ${?PUBLIC_FQDN}
+    publicPort = 8080
+    publicPort = ${?PUBLIC_PORT}
     port = 8080
     port = ${?PORT}
     watch = [ classes ]
--- a/core/src/test/resources/application-test-auth.conf
+++ b/core/src/test/resources/application-test-auth.conf
@@ -17,6 +17,9 @@
 
 ktor {
   deployment {
+    publicScheme = "http"
+    publicFqdn = "localhost"
+    publicPort = 8080
     port = 8080
     port = ${?PORT}
     watch = [ classes ]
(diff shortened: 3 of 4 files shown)


haikoschol:
I'm not a fan of the `public` prefix. Naming-wise I'd prefer changing `port` to `listenPort` instead, but that would be a breaking change.

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -79,7 +79,10 @@ fun Application.configureOpenApi() {
         }
 
         server {
-            url = "http://localhost:8080"
+            val scheme = config.property("ktor.deployment.publicScheme").getString()
```
sschuberth, line 82: Should we maybe instead use `tryGetString()` and apply the default / fallback here? Like

    val scheme = config.tryGetString("ktor.deployment.publicScheme") ?: "http"

haikoschol: Why handle this one different from all other config vars?

sschuberth: I didn't really have a look at the other config vars, but I agree we should do it consistently.

haikoschol: The default values and environment variable names are [defined in the application.conf](https://github.com/haikoschol/ort-server/blob/77387624a9b3e3f0cfeba3fc20d4d634143f716d/core/src/main/resources/application.conf#L26) file for `core`.

core/src/main/resources/application.conf
```diff
@@ -23,6 +23,12 @@ configManager {
 
 ktor {
   deployment {
+    publicScheme = "http"
+    publicScheme = ${?SCHEME}
+    publicFqdn = "localhost"
+    publicFqdn = ${?FQDN}
```
mnonnenmacher, line 29: Calling the variables `PUBLIC_SCHEME` and `PUBLIC_FQDN` would make it more obvious that they belong together with `PUBLIC_PORT`.

core/src/main/resources/application.conf
```diff
@@ -23,6 +23,12 @@ configManager {
 
 ktor {
   deployment {
+    publicScheme = "http"
+    publicScheme = ${?PUBLIC_SCHEME}
+    publicFqdn = "localhost"
+    publicFqdn = ${?PUBLIC_FQDN}
+    publicPort = 8080
+    publicPort = ${?PUBLIC_PORT}
```
mnonnenmacher, line 31: You need to add those new properties to the config files used for testing at `core/src/test/resources` as well (without the environment variable lines as they are not used anyway).

mnonnenmacher (changes requested):

core/src/main/resources/application.conf
haikoschol: Thanks! I hadn't gotten around to look why the tests failed but now that I did I also wouldn't have found the issue in the output. :)

mnonnenmacher (approved):

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -79,7 +79,10 @@ fun Application.configureOpenApi() {
...
+            val fqdn = config.property("ktor.deployment.publicFqdn").getString()
+            val port = config.property("ktor.deployment.publicPort").getString()
+            url = "${scheme}://${fqdn}:${port}"
```
mnonnenmacher, line 85: Detekt complains about the redundant curly braces here.

PR #408, opened 2024-06-12
chore: Remove unused property from MessageHeader

haikoschol:
<strike>The header of messages exchanged between the various endpoints contains a field called "token". This change sets that field with the JWT from the "Authorization" HTTP header.</strike>

This change removes the "token" property from the header of messages exchanged between the various ORT Server endpoints. It is currently left empty and there are no changes planned that would make use of it.

### DIFF
--- a/core/src/main/kotlin/services/OrchestratorService.kt
+++ b/core/src/main/kotlin/services/OrchestratorService.kt
@@ -59,11 +59,9 @@ class OrchestratorService(
             ortRunRepository.create(repositoryId, revision, path, jobConfig, jobConfigContext, labels.orEmpty())
         }
 
-        // TODO: Set the correct token.
         orchestratorSender.send(
             Message(
                 header = MessageHeader(
-                    token = "",
                     traceId = UUID.randomUUID().toString(),
                     ortRunId = ortRun.id
                 ),
--- a/orchestrator/src/test/kotlin/OrchestratorEndpointTest.kt
+++ b/orchestrator/src/test/kotlin/OrchestratorEndpointTest.kt
@@ -62,7 +62,6 @@ import org.koin.test.mock.declareMock
 
 class OrchestratorEndpointTest : KoinTest, StringSpec() {
     private val msgHeader = MessageHeader(
-        token = "token",
         traceId = "traceId",
         ortRunId = 17
     )
--- a/orchestrator/src/test/kotlin/OrchestratorTest.kt
+++ b/orchestrator/src/test/kotlin/OrchestratorTest.kt
@@ -97,7 +97,6 @@ import org.eclipse.apoapsis.ortserver.transport.ReporterEndpoint
 @Suppress("LargeClass")
 class OrchestratorTest : WordSpec() {
     private val msgHeader = MessageHeader(
-        token = "token",
         traceId = "traceId",
         ortRunId = RUN_ID,
     )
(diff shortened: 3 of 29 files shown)


haikoschol:
Turns out the SQS transport does not tolerate leaving this field empty. This was not caught by the integration tests because they [set the token to a non-empty string](https://github.com/eclipse-apoapsis/ort-server/blob/bb48b3cc32821dce11e139d3b7c12e8dcac6c559/transport/sqs/src/test/kotlin/SqsMessageSenderFactoryTest.kt#L70).

core/src/main/kotlin/services/OrchestratorService.kt
```diff
@@ -53,17 +53,17 @@ class OrchestratorService(
...
         val ortRun = db.dbQuery {
             ortRunRepository.create(repositoryId, revision, path, jobConfig, jobConfigContext, labels.orEmpty())
         }
 
-        // TODO: Set the correct token.
         orchestratorSender.send(
             Message(
                 header = MessageHeader(
-                    token = "",
+                    token = jwt.orEmpty(),
```
sschuberth, line 66: Hmm, if empty tokens are not supported, we probably should not fall back to one here, or?

mnonnenmacher, line 66: @oheger-bosch This change seems to be correct according to the documentation of the property, but since we never made use of it I wonder if we should remove it instead?

sschuberth:
> Turns out the SQS transport does not tolerate leaving this field empty.

Should I try to modify the SQS implementation to tolerate empty tokens? Does ORT server require that to be supported, @oheger-bosch?

core/src/main/kotlin/services/OrchestratorService.kt
haikoschol: I expect that sooner or later it might be necessary to check permissions in components behind `core`. But I can't come up with an example on the spot and it can probably just as easily be added back in then.

haikoschol: They are supported in all transports except SQS.

sschuberth: > according to the documentation of the property

To me, it's not clearly documented whether the `token` property may be empty. Still awaiting and answer to what so I can make changes accordingly.

oheger-bosch: The original idea behind this token was to support a transport implementation based on direct HTTP calls that use a token for authentication. As @mnonnenmacher said, we never implemented this and had never a use case for the token. Nothing within ORT Server makes use of the token, and I am pretty sure that there are places in the code were the token is not correctly propagated when another endpoint is invoked.

So, I would also conclude that the cleanest solution is to drop this property. If an actual use case comes up later, then it could be implemented in a consistent way.

haikoschol: > So, I would also conclude that the cleanest solution is to drop this property. If an actual use case comes up later, then it could be implemented in a consistent way.

I'll change my PR to remove the token then (tomorrow).

sschuberth:
> Without having tried it, you could probably leave `TOKEN_PROPERTY` out of the map if `token` is empty [here](https://github.com/eclipse-apoapsis/ort-server/blob/e2366c3b044f27e0ae3c8338ffff3db34be42da2/transport/sqs/src/main/kotlin/SqsMessageSender.kt#L61).

Yes, that's exactly how I was going to implement it, *if* somebody confirms to me that empty tokens should be supported.

oheger-bosch:
> > Without having tried it, you could probably leave `TOKEN_PROPERTY` out of the map if `token` is empty [here](https://github.com/eclipse-apoapsis/ort-server/blob/e2366c3b044f27e0ae3c8338ffff3db34be42da2/transport/sqs/src/main/kotlin/SqsMessageSender.kt#L61).
> 
> Yes, that's exactly how I was going to implement it, _if_ somebody confirms to me that empty tokens should be supported.

I think, the current agreement is to drop the token property altogether.

mnonnenmacher (approved):

PR #409, opened 2024-06-12
Yarn proxy support

oheger-bosch:
Refer to the individual commits for further information.

### DIFF
--- a/workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
+++ b/workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
@@ -58,6 +58,26 @@ class ConfigFileBuilder(val context: WorkerContext) {
         }
 
         /**
+         * Execute the given [block] to generate a proxy configuration section if necessary. This function checks
+         * whether a proxy is (at least partially) defined based on the typical environment variables. If this is the
+         * case, the given [block] is invoked with a corresponding [ProxyConfig] and can write the specific proxy
+         * configuration.
+         */
+        fun PrintWriter.printProxySettings(block: PrintWriter.(ProxyConfig) -> Unit) {
+            val proxySettings = ProxyVariables.entries.associateWith { it.getValue() }.filterValues { it != null }
+
+            if (proxySettings.isNotEmpty()) {
+                val proxyConfig = ProxyConfig(
+                    httpProxy = proxySettings[ProxyVariables.HTTP_PROXY],
+                    httpsProxy = proxySettings[ProxyVariables.HTTPS_PROXY],
+                    noProxy = proxySettings[ProxyVariables.NO_PROXY]
+                )
+
+                this.block(proxyConfig)
+            }
+        }
+
+        /**
          * Generate a unique name for a secret reference. The name is based on a random number. Therefore, it should
          * not appear in other parts of the generated file, and no escaping needs to be implemented.
          */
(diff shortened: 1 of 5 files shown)


workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
```diff
@@ -47,6 +47,9 @@ class ConfigFileBuilder(val context: WorkerContext) {
     companion object {
         private val logger = LoggerFactory.getLogger(ConfigFileBuilder::class.java)
 
+        /** The names of environment variables containing the proxy configuration. */
+        private val PROXY_VARIABLES = listOf("HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY")
```
nnobelis, line 51: IMO a solution based on enum would be better: replace this with an enum,
add a `getEnv` function to the enum, build a EnumMap of enum to their env value and replace array accesses underneath with map lookup.
The problem is, if somebody change this array without adapting `printProxySettings`, it will bug.
Also, using enum is better for finding usages in the IDE.
What do you think ?

workers/common/src/main/kotlin/common/env/NpmRcGenerator.kt
```diff
@@ -51,21 +56,11 @@ class NpmRcGenerator : EnvironmentConfigGenerator<NpmDefinition> {
...
-        /**
-         * Generate proxy settings to be included in the generated configuration file based on existing environment
-         * variables.
+         * Generate a part of the proxy configuration based on the given [key] and [value]. Do not output anything if
+         * the value is undefined.
          */
-        private fun generateProxySettings(): List<String> = proxySettings.mapNotNull { (key, variable) ->
-            settingFromEnv(key, variable)
+        private fun PrintWriter.printProxySetting(key: String, value: String?) {
+            value?.let { println("$key=$it") }
```
nnobelis, line 63: What is the added value of this intermediate function ? Can't it be inlined ?

workers/common/src/main/kotlin/common/env/YarnRcGenerator.kt
```diff
@@ -31,12 +34,25 @@ class YarnRcGenerator : EnvironmentConfigGenerator<YarnDefinition> {
     companion object {
         /** The name of the configuration file created by this generator. */
         private const val TARGET = ".yarnrc.yml"
+
+        /**
+         * Print a line for the given [key] and [value] if the value is not null.
+         */
+        private fun PrintWriter.printOptionalSetting(key: String, value: String?) {
```
nnobelis, line 41: Same comment regarding inlining.

workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
oheger-bosch: Good idea. I have changed the implementation accordingly.

workers/common/src/main/kotlin/common/env/NpmRcGenerator.kt
oheger-bosch: It reduces some redundancy; without the function, the *null* check would have to be done for each property of the proxy configuration.

workers/common/src/main/kotlin/common/env/YarnRcGenerator.kt
oheger-bosch: See above.

nnobelis (approved):

PR #410, opened 2024-06-12
fix(docker): Use sudo when changing permissions

MarcelBochtler:


nnobelis (approved):

oheger-bosch:
The same construct is used in multiple container images; so these should be changed, too. Also, could you give a reasoning why this change is necessary or which issue is fixed by this?

MarcelBochtler:
> The same construct is used in multiple container images; so these should be changed, too. Also, could you give a reasoning why this change is necessary or which issue is fixed by this?

Without this change I wasn't able to build the Analyzer base image on my machine.
It failed with a permissions error when trying to change the directory permissions.

PR #421, opened 2024-06-14
reporter: Do not fail if no how to fix text provider script exists

mnonnenmacher:

### DIFF
--- a/workers/common/src/main/kotlin/common/Extensions.kt
+++ b/workers/common/src/main/kotlin/common/Extensions.kt
@@ -35,36 +35,37 @@ import org.slf4j.LoggerFactory
 val logger: Logger = LoggerFactory.getLogger(ConfigManager::class.java)
 
 /**
- * If [path] is not `null`, read the file using the provided [context]. If a [ConfigException] occurs while reading the
- * file it is rethrown. If [path] is `null`, the file at [defaultPath] is read instead. If the file cannot be read, the
- * [fallbackValue] is returned.
+ * If [path] is not `null`, read the YAML configuration file using the provided [context] and deserialize its value. If
+ * a [ConfigException] occurs while reading the file it is rethrown. If [path] is `null`, the file at [defaultPath] is
+ * read instead. If the file cannot be read, the [fallbackValue] is returned.
+ *
  * This function realizes the contract that if a specific config file is requested, not being able to read it leads to
  * an exception, but the default file is allowed to not exist. Not being able to deserialize the file to the return type
  * [T] always leads to an exception.
  */
-inline fun <reified T> ConfigManager.readConfigFileWithDefault(
+inline fun <reified T> ConfigManager.readConfigFileValueWithDefault(
     path: String?,
     defaultPath: String,
     fallbackValue: T,
     context: Context?
 ): T = if (path != null) {
-    readConfigFile(path, context) {
+    readConfigFileValue(path, context) {
         logger.error("Could not read config file from path '$path'.")
         throw it
     }
 } else {
-    readConfigFile(defaultPath, context) {
+    readConfigFileValue(defaultPath, context) {
         logger.warn("Could not read config file from default path '$defaultPath'.")
         fallbackValue
     }
 }
 
 /**
- * Read the YAML configuration file at [path] using the provided [context]. If a [ConfigException] occurs while reading
- * the file, the [exceptionHandler] is invoked which rethrows the exception by default. If another exception occurs
- * while reading the file it is rethrown.
+ * Read the YAML configuration file at [path] using the provided [context] and deserialize its content. If a
+ * [ConfigException] occurs while reading the file, the [exceptionHandler] is invoked which rethrows the exception by
+ * default. If another exception occurs while reading the file, it is rethrown.
  */
-inline fun <reified T> ConfigManager.readConfigFile(
+inline fun <reified T> ConfigManager.readConfigFileValue(
     path: String,
     context: Context?,
     exceptionHandler: (ConfigException) -> T = { throw it }
--- a/workers/common/src/main/kotlin/common/Extensions.kt
+++ b/workers/common/src/main/kotlin/common/Extensions.kt
@@ -21,6 +21,8 @@ package org.eclipse.apoapsis.ortserver.workers.common
 
 import com.fasterxml.jackson.module.kotlin.readValue
 
+import java.io.InputStream
+
 import org.eclipse.apoapsis.ortserver.config.ConfigException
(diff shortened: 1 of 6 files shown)


sschuberth:
There's a [test failure](https://github.com/eclipse-apoapsis/ort-server/actions/runs/9514361889/job/26226456864?pr=421#step:5:4994).

mnonnenmacher:
> There's a [test failure](https://github.com/eclipse-apoapsis/ort-server/actions/runs/9514361889/job/26226456864?pr=421#step:5:4994).

I know, I finally found the time to fix it.

workers/common/src/main/kotlin/common/Extensions.kt
```diff
@@ -42,18 +42,18 @@ val logger: Logger = LoggerFactory.getLogger(ConfigManager::class.java)
  * an exception, but the default file is allowed to not exist. Not being able to deserialize the file to the return type
  * [T] always leads to an exception.
  */
-inline fun <reified T> ConfigManager.readConfigFileWithDefault(
+inline fun <reified T> ConfigManager.deserializeConfigFileWithDefault(
```
sschuberth, line 48: Commit message: "that *they* not only read"

sschuberth: IMO "deserialize" is an awfully technical term for such a function name... even KxS itself uses "decode" instead. Better naming pairs could be e.g. "decode" / "read", "parse" / "load" etc.

workers/common/src/main/kotlin/common/Extensions.kt
```diff
@@ -38,6 +40,7 @@ val logger: Logger = LoggerFactory.getLogger(ConfigManager::class.java)
  * If [path] is not `null`, read the file using the provided [context]. If a [ConfigException] occurs while reading the
  * file it is rethrown. If [path] is `null`, the file at [defaultPath] is read instead. If the file cannot be read, the
  * [fallbackValue] is returned.
+ *
```
sschuberth, line 43: Unrelated change.

workers/common/src/main/kotlin/common/Extensions.kt
```diff
@@ -47,31 +50,83 @@ inline fun <reified T> ConfigManager.deserializeConfigFileWithDefault(
...
     }
 }
 
 /**
- * Read the YAML configuration file at [path] using the provided [context]. If a [ConfigException] occurs while reading
- * the file, the [exceptionHandler] is invoked which rethrows the exception by default. If another exception occurs
- * while reading the file it is rethrown.
+ * Deserialize the YAML configuration file at [path] using the provided [context]. If a [ConfigException] occurs while
+ * reading the file, the [exceptionHandler] is invoked which rethrows the exception by default. If another exception
+ * occurs while reading the file, it is rethrown.
```
sschuberth, line 99: This should be part of the renaming in the previous commit.

workers/reporter/src/main/kotlin/reporter/ReporterRunner.kt
```diff
@@ -43,6 +43,7 @@ import org.eclipse.apoapsis.ortserver.workers.common.context.WorkerContext
 import org.eclipse.apoapsis.ortserver.workers.common.context.WorkerContextFactory
 import org.eclipse.apoapsis.ortserver.workers.common.deserializeConfigFileWithDefault
 import org.eclipse.apoapsis.ortserver.workers.common.mapToOrt
+import org.eclipse.apoapsis.ortserver.workers.common.readConfigFileWithDefault
```
sschuberth, line 46: Commit message nit: Maybe "how-to-fix text".

workers/common/src/main/kotlin/common/Extensions.kt
```diff
@@ -47,31 +50,83 @@ inline fun <reified T> ConfigManager.deserializeConfigFileWithDefault(
...
     fallbackValue: T,
     context: Context?
+): T = getConfigFileWithDefault<T>(path, defaultPath, fallbackValue, context, ::deserializeConfigFile)
+
+/**
+ * If [path] is not `null`, read the file using the provided [context]. If a [ConfigException] occurs while reading the
+ * file it is rethrown. If [path] is `null`, the file at [defaultPath] is read instead. If the file cannot be read, the
+ * [fallbackValue] is returned.
+ *
+ * This function realizes the contract that if a specific config file is requested, not being able to read it leads to
```
oheger-bosch, line 60: As I have stated elsewhere, IMHO we are trying to be too clever here. There are too many options to specify or not specify configuration paths making the whole mechanism hard to understand.

workers/common/src/main/kotlin/common/Extensions.kt
mnonnenmacher: How about "readConfigFileValueWithDefault"?

workers/common/src/main/kotlin/common/Extensions.kt
mnonnenmacher: For the user there are just two options:
* Specify a path -> Use the file or fail if it does not exist.
* Do not speficy a path -> Use the default file if it exists.

workers/common/src/main/kotlin/common/Extensions.kt
sschuberth: Fine with me.

workers/common/src/main/kotlin/common/Extensions.kt
```diff
@@ -48,14 +50,45 @@ inline fun <reified T> ConfigManager.readConfigFileValueWithDefault(
...
+ * an exception, but the default file is allowed to not exist.
+ */
+fun ConfigManager.readConfigFileWithDefault(
+    path: String?,
+    defaultPath: String,
+    fallbackValue: String,
+    context: Context?
+): String = getConfigFileWithDefault(path, defaultPath, fallbackValue, context, ::readConfigFile)
```
sschuberth, line 68: Can't we just get rid of this function by making `::readConfigFile` the default argument for `getConfigFileWithDefault`? Similar for `readConfigFile` below.

When doing do, I'd then rename `getConfigFileWithDefault` to `readConfigFileWithDefault` and `getConfigFile` to `readConfigFile` in order to not also have read vs. get variants of a function, but only "File" vs "FileValue" and optional "WithDefault" variants of the functions.

mnonnenmacher: I think the reason I kept the functions separate was that the `get` function returns an input stream like the `getConfigFile` function of the `ConfigManager`. And I wanted to deserialize the input stream directly without converting it to a string first. If you want to optimize something, can you make a follow-up PR?

sschuberth (approved):

Issue #406, opened 2024-06-11
Consolidate API endpoints for ORT runs


mnonnenmacher:
The endpoints to get information about ORT runs are currently at two different locations:

* `/repositories/{id}/runs/{index}`
* `/runs/{id}/reporter/{filename}`
* `/runs/{id}/logs`

Move the first endpoint to `/runs/{id}` to make the API more consistent.

wkl3nk:
I agree, ```/runs/{id}```sounds reasonable and consistent.

oheger-bosch:
Yes, it makes sense to have the run-related information under the path `/runs`.

Just wondering how to treat the `/repositories/{id}/runs` endpoint that returns the runs of a specific repository. Should it stay as is or be moved, too, maybe to something like `/runs/repository/{id}` ?

mnonnenmacher:
> Just wondering how to treat the `/repositories/{id}/runs` endpoint that returns the runs of a specific repository. Should it stay as is or be moved, too, maybe to something like `/runs/repository/{id}` ?

I would leave that where it is, it's consistent with other endpoints that return associated objects like `organizations/{id}/products` or `repositories/{id}/secrets`.

mmurto:
> > Just wondering how to treat the `/repositories/{id}/runs` endpoint that returns the runs of a specific repository. Should it stay as is or be moved, too, maybe to something like `/runs/repository/{id}` ?
> 
> I would leave that where it is, it's consistent with other endpoints that return associated objects like `organizations/{id}/products` or `repositories/{id}/secrets`.

Another option would be to expose them from `/runs` by taking in a filter parameter (query/body, not path) for `repositoryId`.

mnonnenmacher:
> Another option would be to expose them from /runs by taking in a filter parameter (query/body, not path) for repositoryId.

That's an option, but it's harder to implement because the permission model is based on organizations, products, and repositories, so filtering the full list of runs by permission requires a lot of permission checks and in my opinion it's not worth it if we don't have a real use case for it. Filtering is straight forward if the repository is known in advance.

After looking a bit deeper into the topic I decided that moving the endpoint altogether makes things unnecessarily complicated for the frontend where the repository id and ORT run index are known from path parameters, but not the run id. Instead, I have changed #425 to only add the new endpoint at `/rund/{id}` as this is useful for other scenarios, but not remove the old endpoint at `/repositories/{id}/runs/{index}`.

mmurto:
[...]

PR #425, opened 2024-06-14
Add `/runs/{id}` endpoint

mnonnenmacher:
Add an endpoint to get the details of an ORT run at `/runs/{id}`.

For context see #406.

### DIFF
--- a/core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
+++ b/core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
@@ -53,15 +53,20 @@ import kotlin.time.Duration.Companion.minutes
 import kotlinx.datetime.Clock
 import kotlinx.datetime.Instant
 
+import org.eclipse.apoapsis.ortserver.api.v1.mapping.mapToApi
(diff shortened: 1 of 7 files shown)


core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
```diff
@@ -198,6 +208,52 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
...
+                    null,
+                    JobConfigurations(),
+                    "testContext",
+                    labelsMap
+                )
+
+                val response = superuserClient.get("/api/v1/runs/${run.id}")
+
+                response shouldHaveStatus HttpStatusCode.OK
+                response shouldHaveBody run.mapToApi(Jobs())
```
oheger-bosch, line 244: This is basically the same test as the previous one. Where are job details tested?

mnonnenmacher: I just copied the existing test cases and didn't notice those tests were identical. I have added commits to address that.

oheger-bosch (approved):

PR #426, opened 2024-06-14
feat(workers): Create <proxies> section in settings.xml

wkl3nk:
Create _proxies_ section in Maven configuration file _settings.xml_ as one additional method (next to environment variables like _HTTP_PROXY_ and system properties like _http.proxyHost_) for Java tools like Eclipse Tycho to find out which proxy to use for outgoing connections.

Such a <proxies> section typically looks as follows:

```
<settings ... >
...
    <proxies>
        <proxy>
            <id>http-proxy</id>
            <active>true</active>
            <protocol>http</protocol>
            <host>proxy.example.com</host>
            <port>8080</port>
            <nonProxyHosts>*.example.com|*.svc.cluster.local</nonProxyHosts>
        </proxy>
        <proxy>
            <id>https-proxy</id>
            <active>true</active>
            <protocol>https</protocol>
            <host>proxy.example.com</host>
            <port>8080</port>
            <nonProxyHosts>*.example.com|*.svc.cluster.local</nonProxyHosts>
        </proxy>
    </proxies>
</settings>
```

### DIFF
--- a/workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
+++ b/workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
@@ -78,6 +78,39 @@ class ConfigFileBuilder(val context: WorkerContext) {
         }
 
         /**
+         * Execute the given [block] to generate a proxy configuration if necessary.
+         * This variant is based on system properties that define https(s) proxy settings rather than
+         * environment variables. It is typically used by tools that have a strong connection to the Java ecosystem.
+         * If one of httpProxyHost or httpsProxyHost is defined, the [block] is invoked with a corresponding
+         * [ProxyConfigFromSystemProperties].
+         */
+        fun PrintWriter.printProxySettingsFromSystemProperties(
+            block: PrintWriter.(ProxyConfigFromSystemProperties) -> Unit
+        ) {
+            val httpProxy = ProxySystemProperties.HTTP_PROXY_HOST.getValue()?.let { host ->
+                ProxySystemProperties.HTTP_PROXY_PORT.getValue()?.let { port ->
+                    ProxyFromSystemProperties(host, port)
+                }
+            }
+
+            val httpsProxy = ProxySystemProperties.HTTPS_PROXY_HOST.getValue()?.let { host ->
+                ProxySystemProperties.HTTPS_PROXY_PORT.getValue()?.let { port ->
+                    ProxyFromSystemProperties(host, port)
+                }
+            }
+
+            if (httpProxy != null || httpsProxy != null) {
+                block(
+                    ProxyConfigFromSystemProperties(
+                        httpProxy = httpProxy,
+                        httpsProxy = httpsProxy,
+                        nonProxyHosts = ProxySystemProperties.NON_PROXY_HOSTS.getValue()
+                    )
+                )
+            }
+        }
+
+        /**
          * Generate a unique name for a secret reference. The name is based on a random number. Therefore, it should
          * not appear in other parts of the generated file, and no escaping needs to be implemented.
          */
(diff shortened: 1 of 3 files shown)


workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
```diff
@@ -77,6 +82,42 @@ class ConfigFileBuilder(val context: WorkerContext) {
...
         }
 
+        /**
+         * Execute the given [block] to generate a proxy configuration if necessary.
+         * This variant is based on system properties that define https(s) proxy settings rather than
+         * environment variables. It is typically used by tools that have a strong connection to the Java ecosystem.
+         * If one of httpProxyHost or httpsProxyHost is defined, the [block] is invoked with a corresponding
+         * [ProxiesFromSystemProperties].
+         *
+         * @param block A lambda function that is an extension function on PrintWriter and takes a
```
oheger-bosch, line 92: Nit: In the codebase, typically no `@param` tags are used, but the Kotlin convention to mention parameter names in the text of the comment.

workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
```diff
@@ -77,6 +82,42 @@ class ConfigFileBuilder(val context: WorkerContext) {
...
+         * [ProxiesFromSystemProperties] object as a parameter.
+         */
+        fun PrintWriter.printProxySettingsFromSystemProperties(
+            block: PrintWriter.(ProxiesFromSystemProperties) -> Unit
+        ) {
+            val httpProxy = HTTP_PROXY_HOST.getValue()?.let { host ->
```
oheger-bosch, line 98: Since there are now multiple enumerations involved, I would find it more readable if the constants were qualified, i.e. `ProxySystemProperties.HTTP_PROXY_HOST` etc.

workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
```diff
@@ -162,3 +216,14 @@ data class ProxyConfig(
     val httpsProxy: String?,
     val noProxy: String?
 )
+
+data class ProxiesFromSystemProperties(
```
oheger-bosch, line 220: Public classes should have a comment. Here, it would be especially useful to describe the differences to `ProxyConfig`, since semantically both classes are equivalent. I think, ideally, only a single class would be used to hold the proxy configuration; but I understand that different representations of the data are needed.

BTW, should the class be named `ProxyConfigFromSystemProperties` ?

workers/common/src/test/kotlin/common/env/MavenSettingsGeneratorTest.kt
```diff
@@ -110,6 +120,136 @@ class MavenSettingsGeneratorTest : WordSpec({
             lines[lines.size - 2] shouldBe "    </servers>"
             lines[lines.size - 1] shouldBe "</settings>"
         }
+
+        "generates no proxy section if system properties are not set" {
+            mockkObject(HTTP_PROXY_HOST)
```
oheger-bosch, line 125: KoTest has a [helper function](https://kotest.io/docs/extensions/system_extensions.html#system-property-extension) to temporarily override system properties. Using this is probably simpler than mocking the proxy constants. If you use mocking, please place the `unmockAll()` call in an `afterTest` function to make sure that it is always invoked, even in case of a test failure.

workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
wkl3nk: Fixed.

workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
wkl3nk: Yes, added ProxyStemProperties as proposed.

workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
wkl3nk: Added comments and renamed the class.

workers/common/src/test/kotlin/common/env/MavenSettingsGeneratorTest.kt
wkl3nk: Thanks for the hint. That makes the test look a lot of nicer.

workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
```diff
@@ -162,3 +208,25 @@ data class ProxyConfig(
     val httpsProxy: String?,
     val noProxy: String?
 )
+
+/**
+ * A data class holding proxy information taken vom system properties.
+ * In contract to [ProxyConfig], which takes this information from environment variables, this class uses the
```
oheger-bosch, line 214: Nit: Typo "In contrast ..."

oheger-bosch (commented):
One typo, otherwise looks good.

workers/common/src/main/kotlin/common/env/ConfigFileBuilder.kt
wkl3nk: Typo fixed.

oheger-bosch (approved):

PR #431, opened 2024-06-17
feat(workers): Log the ORT Server version when a worker starts

nnobelis:
image

The Ascii art uses a library from a small project "Kotlin Ink":
https://github.com/Derrick-Mwendwa/kotlin-ink

Is it ok to use it in ORT Server ? I reviewed the source code and it seems clean.

I also opened an issue for better integration:
https://github.com/Derrick-Mwendwa/kotlin-ink/issues/3

### DIFF
--- a/transport/spi/src/main/kotlin/EndpointComponent.kt
+++ b/transport/spi/src/main/kotlin/EndpointComponent.kt
@@ -30,6 +30,9 @@ import org.koin.core.module.Module
 import org.koin.core.module.dsl.singleOf
 import org.koin.dsl.module
 
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
 /**
  * An abstract base class providing functionality useful for components implementing endpoints in the ORT server.
  *
@@ -50,6 +53,8 @@ abstract class EndpointComponent<T : Any>(
     val configManager: ConfigManager = ConfigManager.create(ConfigFactory.load())
 ) : KoinComponent {
     abstract val endpointHandler: EndpointHandler<T>
+
+    protected val logger: Logger = LoggerFactory.getLogger(this::class.java)
 
     /**
      * Start this endpoint and perform necessary initialization, so that incoming messages can be received and
--- a/transport/spi/src/main/kotlin/EndpointComponent.kt
+++ b/transport/spi/src/main/kotlin/EndpointComponent.kt
@@ -66,6 +66,7 @@ abstract class EndpointComponent<T : Any>(
             modules(customModules())
         }
 
+        logger.logVersion()
         MessageReceiverFactory.createReceiver(endpoint, configManager, endpointHandler)
     }
 
--- a/transport/spi/src/main/kotlin/Utils.kt
+++ b/transport/spi/src/main/kotlin/Utils.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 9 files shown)


workers/advisor/src/main/kotlin/advisor/AdvisorWorker.kt
```diff
@@ -44,6 +45,8 @@ internal class AdvisorWorker(
     private val contextFactory: WorkerContextFactory
 ) {
     fun run(jobId: Long, traceId: String): RunResult = runCatching {
+        logVersion()
```
oheger-bosch, line 48: Instead of adding the `logVersion()` call to all workers, I think it would be possible to add it to the `EndpointComponent` class. Then the information would also be printed for core and orchestrator.

workers/common/src/main/kotlin/common/Utils.kt
```diff
@@ -19,11 +19,39 @@
...
+    val oldStream = System.out
+    val contentStream = ByteArrayOutputStream()
+
+    PrintStream(contentStream).apply {
+        System.setOut(this)
+
+        KInk.say("ORT-Server")
+        println("ORT Server version: $ORT_SERVER_VERSION")
+
+        System.setOut(oldStream)
```
oheger-bosch, line 51: Resetting the old stream should better be done in a `try`-`finally` block.

oheger-bosch (commented):
Regarding using of a library, I am not sure whether this is actually needed. The Ascii art is (currently ?) not generated dynamically. So it would be an option to have a pre-generated text file that is just dumped by `logVersion()`.

Actually, it would be nice to have different texts for the different components, for instance something like "ORT Server - Analyzer", "ORT Server - Orchestrator", etc. This could be achieved by loading a text file with a well-known name from the resources.

workers/advisor/src/main/kotlin/advisor/AdvisorWorker.kt
nnobelis: To which function of `EndpointComponenent` should I add it then? It feels `start()` is only called once at initialization.

Also if the logging is done these, will it be correctly filtered by Grafana's workers filter ?

nnobelis:
> Regarding using of a library, I am not sure whether this is actually needed. The Ascii art is (currently ?) not generated dynamically. So it would be an option to have a pre-generated text file that is just dumped by `logVersion()`.
> 
> Actually, it would be nice to have different texts for the different components, for instance something like "ORT Server - Analyzer", "ORT Server - Orchestrator", etc. This could be achieved by loading a text file with a well-known name from the resources.

I agree we could have a static text instead.
However `ORT Server - Analyzer` in Ascii is IMO too much output.

How about:

```
<ORT-Server in static Ascii art>
ORT Server Analyzer version: 1.0.0-SNAPSHOT+2378.sha.3bad08f 
```

What do you think ?

workers/advisor/src/main/kotlin/advisor/AdvisorWorker.kt
nnobelis: But in this case, you would have to go maybe (way) back in the logs to find this information. Isn't it more practical to see it in every worker run ?

oheger-bosch:
> I agree we could have a static text instead. However `ORT Server - Analyzer` in Ascii is IMO too much output.
> 
> How about:
> 
> ```
> <ORT-Server in static Ascii art>
> ORT Server Analyzer version: 1.0.0-SNAPSHOT+2378.sha.3bad08f 
> ```
> 
> What do you think ?

Alternatively, the variable part could be in Ascii art, e.g. only "Analyzer", but I am not the visual guy and do not care that much.

workers/advisor/src/main/kotlin/advisor/AdvisorWorker.kt
oheger-bosch: Being a bit egoistic, for us as users of the Kubernetes transport, this does not make a difference. Here the message is printed for each worker run.

nnobelis: Ok. I hope I got it correctly. Please review the changes.

oheger-bosch:
There is a detekt issue; otherwise looks good to me. IIUC, the message no longer contains the name of the component, but this should be displayed via the logger, right?

nnobelis:
I fixed the detekt issue.

> IIUC, the message no longer contains the name of the component, but this should be displayed via the logger, right?
 
Exactly.

oheger-bosch (approved):

PR #443, opened 2024-06-18
fix(ui): Redirect back to same path after signing in

### LINKED ISSUE #435: Links to routes do not work
When opening a URL like `/organizations/1` or `/organizations/1/products/1/repositories/1/runs/1` the ui loads the organizations list instead of the requested page. This was tested in the Docker Compose setup. The issue only occurs if the ui was not previously loaded in the same browser tab.

mmurto:
To sing the users in, the UI briefly redirects the browser to Keycloak when
loading the UI in browser. Keycloak redirects users back to the UI, so use
the current URL as the redirect URI to get the user to the same page they
tried to access before signing in.

Fixes #435.

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -71,7 +71,9 @@ export const App = () => {
       !auth.isLoading &&
       !hasTriedSignin
     ) {
-      auth.signinRedirect();
+      auth.signinRedirect({
+        redirect_uri: window.location.href,
+      });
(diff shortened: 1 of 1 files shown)


sschuberth (approved):

mmurto:
> Found a typo in the commit message, I'll fix.

Done, no talk about singing anymore. 🎵

mnonnenmacher (approved):

PR #447, opened 2024-06-19
Job monitor delete server jobs only

oheger-bosch:
Refer to the single commits for further information.

### DIFF
--- a/transport/kubernetes-jobmonitor/src/main/kotlin/JobHandler.kt
+++ b/transport/kubernetes-jobmonitor/src/main/kotlin/JobHandler.kt
@@ -80,6 +80,13 @@ internal class JobHandler(
         private const val TRACE_LABEL_PREFIX = "trace-id-"
 
         /**
+         * A label selector to find only jobs for ORT Server components. Only those are handled when looking for
+         * completed or failed jobs.
+         */
+        private val workerJobsLabelSelector = "ort-worker in " +
+                Endpoint.entries().joinToString(",", prefix = "(", postfix = ")") { it.configPrefix }
+
+        /**
          * Return a flag whether this job has failed. For jobs that are still running the result is *false*.
          */
         fun V1Job.isFailed(): Boolean = status?.conditions.orEmpty().any { it.type == FAILED_CONDITION }
@@ -135,7 +142,7 @@ internal class JobHandler(
      * Return a list with all currently existing jobs that have been completed before the given [time].
      */
     fun findJobsCompletedBefore(time: OffsetDateTime): List<V1Job> =
-        listJobs().filter { it.completedBefore(time) }
+        listJobs(workerJobsLabelSelector).filter { it.completedBefore(time) }
 
     /**
      * Return a list with all currently active jobs for the worker defined by the given [endpoint].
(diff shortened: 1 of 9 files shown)


transport/spi/src/main/kotlin/Endpoint.kt
```diff
@@ -51,21 +51,34 @@ sealed class Endpoint<T : Any>(
     val configPrefix: String
 ) {
     companion object {
+        /**
+         * A map holding the existing [Endpoint] instances using their configuration prefixes as keys.
+         */
+        private val endpointsByConfigPrefix by lazy {
```
sschuberth, line 57: Is there really anything expensive in creating this list, so that it's worth to create it only lazily?

transport/kubernetes-jobmonitor/src/main/kotlin/JobHandler.kt
```diff
@@ -79,6 +79,13 @@ internal class JobHandler(
         /** A prefix for the name of a label storing a part of the trace ID. */
         private const val TRACE_LABEL_PREFIX = "trace-id-"
 
+        /**
+         * A label selector to find only jobs for ORT Server components. Only those are handled when looking for
+         * completed or failed jobs.
+         */
+        private val workerJobsLabelSelector = "ort-worker in " +
```
sschuberth, line 86: Would it make sense to introduce a constant for "ort-worker in ", similar to the above constants?

transport/kubernetes-jobmonitor/src/test/kotlin/JobHandlerTest.kt
```diff
@@ -374,6 +376,63 @@ class JobHandlerTest : WordSpec({
...
+                    capture(slotLabel),
+                    null,
+                    null,
+                    null,
+                    null,
+                    false
+                )
+            }
+
+            val labelSelectorRegex = Regex("""ort-worker in \((.+)\)""")
```
sschuberth, line 421: If introducing a constant, it should be used here as well.

transport/spi/src/main/kotlin/Endpoint.kt
oheger-bosch: Without the `lazy`, an exception during class initialization is thrown. This is because the objects to be added to the map have not yet been fully initialized.

transport/kubernetes-jobmonitor/src/main/kotlin/JobHandler.kt
oheger-bosch: This property is already (semantically) a constant that contains the label selector. I don't think it is an improvement to extract the prefix to another constant.

transport/kubernetes-jobmonitor/src/test/kotlin/JobHandlerTest.kt
oheger-bosch: In the test, I want to explicitly state the expectations. Therefore, I would not reuse a constant from the class under test.

oheger-bosch:
Added one more commit to handle notifier jobs in the lost job detection. Those were missing.

mnonnenmacher (approved):

PR #450, opened 2024-06-19
test: Disable scanning for Kotest project config in third-party JARs

sschuberth:
This should reduce test startup times yet more, see [1].

[1]: https://kantis.github.io/posts/Faster-Kotest-startup/


sschuberth:
This is the same change as done for ORT core [here](https://github.com/oss-review-toolkit/ort/pull/8774).

oheger-bosch (approved):

PR #454, opened 2024-06-19
Improve configuration of UI Docker image

mnonnenmacher:

### DIFF
--- a/ui/docker/entrypoint.sh
+++ b/ui/docker/entrypoint.sh
@@ -24,6 +24,7 @@
 : "${UI_CLIENT_ID:=ort-server-ui}"
 
 # Replace placeholders with actual environment variables in JavaScript files.
+find /usr/share/nginx/html/assets -name '*.js' -exec sed -i "s#VITE_CLIENT_ID||\"ort-server-ui-dev\"#VITE_CLIENT_ID||\"$UI_CLIENT_ID\"#g" {} +
 find /usr/share/nginx/html/assets -name '*.js' -exec sed -i "s#VITE_UI_URL||\"http://localhost:5173/\"#VITE_UI_URL||\"$UI_URL\"#g" {} +
 
 if [ -n "$UI_API_URL" ]; then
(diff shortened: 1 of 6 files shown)


ui/docker/entrypoint.sh
```diff
@@ -34,9 +34,7 @@ if [ -n "$VITE_AUTHORITY" ]; then
   find /usr/share/nginx/html/assets -name '*.js' -exec sed -i "s#VITE_AUTHORITY||\"http://localhost:8081/realms/master\"#VITE_AUTHORITY||\"$VITE_AUTHORITY\"#g" {} +
 fi
 
-if [ -n "$VITE_CLIENT_ID" ]; then
-  find /usr/share/nginx/html/assets -name '*.js' -exec sed -i "s#VITE_CLIENT_ID||\"ort-server-ui-dev\"#VITE_CLIENT_ID||\"$VITE_CLIENT_ID\"#g" {} +
-fi
+find /usr/share/nginx/html/assets -name '*.js' -exec sed -i "s#VITE_CLIENT_ID||\"ort-server-ui-dev\"#VITE_CLIENT_ID||\"$VITE_CLIENT_ID\"#g" {} +
```
sschuberth: Commit message nits:
- Maybe put "if" in backticks in the title, or say "if-condition" to avoid "if" being misread as a word of the sentence.
- "varibale" -> "variable"
- "was redundant" -> "is redundant"

ui/docker/entrypoint.sh
```diff
@@ -24,6 +24,7 @@
 : "${UI_CLIENT_ID:=ort-server-ui}"
 
 # Replace placeholders with actual environment variables in JavaScript files.
+find /usr/share/nginx/html/assets -name '*.js' -exec sed -i "s#VITE_CLIENT_ID||\"ort-server-ui-dev\"#VITE_CLIENT_ID||\"$UI_CLIENT_ID\"#g" {} +
```
sschuberth, line 31: Commit message nit: Maybe "*a* better overview"

sschuberth:
@mmurto any comments here?

mnonnenmacher:
> @mmurto any comments here?

If you are not deploying the UI to a subpath (like we do) the only thing that you have to update should be the port mapping for the change from port 80 to 8080.

ui/src/app.tsx
```diff
@@ -39,6 +39,7 @@ export interface RouterContext {
 // Create a new router instance
 const router = createRouter({
   routeTree,
+  basepath: import.meta.env.VITE_BASEPATH || '/',
```
bs-ondem, line 42: I must admit that my expertise in TypeScript/Frontend development is still developing, so I might be missing some context or best practices here. But do we need this fallback here since we are setting `UI_BASEPATH` always with a default value?

mnonnenmacher: That's only the case in Docker, when running the UI differently the `VITE_BASEPATH` variable might not be set.

bs-ondem (approved):

Issue #453, opened 2024-06-19
Add pagination to lists


mnonnenmacher:
The lists of organizations, products, repositories, and runs use the default limit of 20 items. If there are more entries, they are currently not visible.

Add support for pagination to all lists, so that all entries can be accessed.

Note that currently the API responses do not contain the number of pages or the total number of entries, so it would not be possible to show the total number of pages in the UI. Please also document which properties would be required to properly implement pagination so that the API can be improved later on.

sschuberth:
> Please also document which properties would be required to properly implement pagination so that the API can be improved later on.

As @Etsija is currently on vacation, maybe @mmurto can provide some input here.

mmurto:
For pagination adding the total number of rows should be enough - offset seems to be already implemented.

mnonnenmacher:
> For pagination adding the total number of rows should be enough - offset seems to be already implemented.

I saw two different kinds of pagination data in APIs. One used page number, page size and total number of pages, the other offset, limit and total count. The latter would course be easier to implement for us as it requires less changes, but if the first option would be significantly better for the UI code I would be open to change it.

mmurto:
> > For pagination adding the total number of rows should be enough - offset seems to be already implemented.
> 
> I saw two different kinds of pagination data in APIs. One used page number, page size and total number of pages, the other offset, limit and total count. The latter would course be easier to implement for us as it requires less changes, but if the first option would be significantly better for the UI code I would be open to change it.

I don't think exposing page number from the API is required, as page number is pretty much just `offset / limit`. Sure the calculation could be done in the back end, but I don't know if there is value in it.

Another option would be [cursor pagination](https://slack.engineering/evolving-api-pagination-at-slack/), but I think offset should work at least for now.

Etsija:
So we need total number of entries for all GET requests that return lists to paginate, of which I created an issue to back-end.

mnonnenmacher:
@Etsija With #633 all list endpoints should now provide the total entity count.

Etsija:
@mnonnenmacher @sschuberth I'm scanning the status of the older UI tickets, and I think this can be closed now, as all existing list endpoints provide the total entity count, and we have implemented pagination for all pages that use list endpoints. There are some list endpoints not yet implemented, like `.../{runId}/issues` but I believe you will provide the total entity count for those in the future?

PR #455, opened 2024-06-19
feat(ui): Increase the limits in API requests

mnonnenmacher:
Increase the limits when requesting list of organizations, products, repositories, and runs from the default 20 to 1000.

This mitigates the issue that not all entities can be accessed from the UI until #453 is implemented.

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId.index.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId.index.tsx
@@ -86,6 +86,7 @@ const OrganizationComponent = () => {
         queryFn: async () =>
           await ProductsService.getOrganizationProducts({
             organizationId: Number.parseInt(params.orgId),
+            limit: 1000,
           }),
       },
     ],
@@ -252,6 +253,7 @@ export const Route = createFileRoute('/_layout/organizations/$orgId/')({
         queryFn: () =>
           ProductsService.getOrganizationProducts({
             organizationId: Number.parseInt(params.orgId),
+            limit: 1000,
           }),
       }),
     ]);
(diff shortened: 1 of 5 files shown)


mmurto:
Code is good, but there's a commit-lint error that I can't spot by eye.

mnonnenmacher:
> Code is good, but there's a commit-lint error that I can't spot by eye.

The problem was that commitlint treats every line that contains an issue reference as a footer. Not sure if that can be changed.

ui/src/routes/_layout/organizations/$orgId.index.tsx
```diff
@@ -86,6 +86,7 @@ const OrganizationComponent = () => {
         queryFn: async () =>
           await ProductsService.getOrganizationProducts({
             organizationId: Number.parseInt(params.orgId),
+            limit: 1000,
```
mmurto, line 89: Actually this should be adjusted, or more specifically the limit should be added [here](https://github.com/eclipse-apoapsis/ort-server/blob/0158f02c36ff15b66865393418d843b5ce1b10e4/ui/src/routes/_layout/organizations/%24orgId.index.tsx#L251-L258) as well, and the same is true for other pages. This way [loader](https://tanstack.com/router/v1/docs/framework/react/guide/external-data-loading#using-loaders-to-ensure-data-is-loaded) uses the same query as the query inside the component.

Not related to this PR, but this should be refactored to use `queryOptions` as shown [here](https://tanstack.com/router/v1/docs/framework/react/guide/external-data-loading#a-more-realistic-example-using-tanstack-query) to not require repetition of the query options.

mmurto (dismissed):

wkl3nk (approved):

sschuberth (approved):

PR #507, opened 2024-06-25
Caching for GitHubConfigFileProvider

oheger-bosch:
`GitHubConfigFileProvider` downloads multiple configuration files from a configurable GitHub repository during an ORT run. When there are many ORT runs, this can be an issue, since the GitHub REST API has some rate limiting applied (5000 requests per hour).

This PR therefore adds support for caching to this provider. This is rather effective for this use case, as files under a specific revision are immutable.

For now, configuration files are cached in a local directory, which can be a network share. This is going to be extended by also caching the content of folders and also by an automatic clean-up of old entries in the cache.

### DIFF
--- a/config/github/src/main/kotlin/GitHubConfigCache.kt
+++ b/config/github/src/main/kotlin/GitHubConfigCache.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.config.github
+
+import io.ktor.utils.io.ByteReadChannel
+
+import java.io.InputStream
+
+/**
+ * Definition of an interface for a cache to store GitHub configuration data.
+ *
+ * During an ORT run, [GitHubConfigFileProvider] typically loads many configuration files from the configured GitHub
+ * repository. This can be problematic for multiple reasons:
+ * - The GitHub API has a rate limit of 5000 requests per hour, which can be exceeded quickly.
+ * - There is a strong dependency to the availability of the GitHub API.
+ * - The same data is requested multiple times during an ORT run, which is inefficient.
+ *
+ * To address these issues, this caching interface is introduced. Currently, it is not intended to support custom
+ * implementations, but there will be a default implementation using a file-based cache and a no-op implementation for
+ * turning off caching.
+ */
+interface GitHubConfigCache {
+    /**
+     * Return an [InputStream] to access the file at the given [path] from the given [revision]. Use the given [load]
+     * function to obtain the file content if necessary. The [load] function returns a channel with the content of the
+     * file to be cached; this is typically obtained via an HTTP request to the GitHub API. A concrete cache
+     * implementation persists this data and returns an [InputStream] to access it (which is also the return type
+     * for files requested from a config file provider).
+     */
+    suspend fun getOrPutFile(revision: String, path: String, load: suspend () -> ByteReadChannel): InputStream
+}
--- a/config/github/src/main/kotlin/GitHubConfigFileCache.kt
+++ b/config/github/src/main/kotlin/GitHubConfigFileCache.kt
@@ -1,0 +1,168 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.config.github
+
+import io.ktor.utils.io.ByteReadChannel
+import io.ktor.utils.io.core.isEmpty
+import io.ktor.utils.io.core.readByteBuffer
+
+import java.io.File
+import java.io.FileInputStream
+import java.io.InputStream
+import java.io.RandomAccessFile
+import java.nio.channels.FileLock
+
+import kotlin.time.Duration
+
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.withContext
+
+import org.slf4j.LoggerFactory
+
+private val logger = LoggerFactory.getLogger(GitHubConfigFileCache::class.java)
+
+/**
+ * The path under which files downloaded from GitHub are stored in the cache. This is a subdirectory of the folder for
+ * the specific revision.
+ */
+private const val FILES_PATH = "tree"
+
+/**
+ * A class implementing a file cache for data fetched from the GitHub configuration repository.
+ *
+ * Caching of content downloaded from GitHub should be rather effective, as the revision can be used as a criterion to
+ * construct cache keys. It is guaranteed that the data of a specific revision does not change. By applying caching,
+ * the number of requests sent to the GitHub REST API can be reduced significantly. This is especially important since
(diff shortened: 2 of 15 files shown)


config/github/src/main/kotlin/GitHubConfigFileProvider.kt
```diff
@@ -224,13 +252,33 @@ class GitHubConfigFileProvider(
...
-        httpClient.get(requestUrl) {
+        val response = httpClient.get(requestUrl) {
             header("Accept", contentType)
         }
+
+        val rateLimitReset = checkForRateLimitRetry(response)
+        return if (rateLimitReset != null) {
+            val resetAt = Instant.fromEpochSeconds(rateLimitReset)
+            val delay = resetAt - Clock.System.now()
+            logger.warn("Rate limit exceeded. Retrying in {} seconds.", delay)
```
nnobelis, line 292: Shouldn't a limit be put in place to avoid waiting for too long ?

config/github/src/main/kotlin/GitHubConfigCache.kt
```diff
@@ -0,0 +1,45 @@
...
+ * implementations, but there will be a default implementation using a file-based cache and a no-op implementation for
+ * turning off caching.
+ */
+interface GitHubConfigCache {
+    /**
+     * Return an [InputStream] to access the file at the given [path] from the given [revision]. Use the given [load]
+     * function to obtain the file content if necessary.
+     */
+    suspend fun getOrPutFile(revision: String, path: String, load: suspend () -> ByteReadChannel): InputStream
+}
```
nnobelis, line 45: I don't understand the difference between the returned `InputStream` and the parameter of the lambda. Aren't they both the file content?
EDIT: looking at the next commit, `load` is the function that loads the content that will then be returned as an `InputStream`.
IMO the wording of the documentation should be improved.

config/github/src/main/kotlin/GitHubConfigDisabledCache.kt
```diff
@@ -0,0 +1,49 @@
...
+
+import java.io.ByteArrayInputStream
+import java.io.ByteArrayOutputStream
+import java.io.InputStream
+
+/**
+ * An implementation of [GitHubConfigCache] that does not cache anything, but always delegates to the function that
+ * fetches the requested data.
+ */
+class GitHubConfigDisabledCache : GitHubConfigCache {
```
nnobelis, line 33: Suggestion: `GitHubConfigNoCache`

config/github/src/main/kotlin/GitHubConfigFileCache.kt
```diff
@@ -0,0 +1,168 @@
...
+                lock = runCatching {
+                    this@waitForReadLock.channel.tryLock(0, Long.MAX_VALUE, true)
+                }.getOrNull()
+
+                if (lock == null) {
+                    delay(lockCheckInterval)
+                }
+            } while (lock == null)
+
+            lock.close()
```
nnobelis, line 165: As `Filelock` implements `AutoCloseable`, can you use `.use{}` here ?

config/github/src/main/kotlin/GitHubConfigFileCache.kt
```diff
@@ -0,0 +1,168 @@
...
+ * The implementation has to deal with some race conditions. Since the cache is used by multiple JVMs, it has to be
+ * ensured that concurrent access does not cause any problems. For instance, if a file which is not yet present in the
+ * cache is requested by two JVM instances at the same time, it should be requested from GitHub only once, and it can
+ * be read only after it has been fully downloaded. The implementation achieves this by using exclusive and shared
+ * locks on the accessed files.
+ *
+ * Note: This implementation focuses on the specific requirements of [GitHubConfigFileProvider]. It may be possible to
+ * extract common functionality of a file-based cache for more generic use cases later.
+ */
+class GitHubConfigFileCache(
```
nnobelis, line 65: Another point I don't see mentioned: this cache is only growing and entries never seem to be removed.
Shouldn't it be mentioned in the documentation ?
Do you plan on using a "real" caching library with some eviction mechanism such as, from the top of my head, EhCache ?

config/github/src/main/kotlin/GitHubConfigFileProvider.kt
```diff
@@ -154,6 +180,17 @@ class GitHubConfigFileProvider(
...
         }
 
+        /**
+         * Create the cache for configuration data based on the given [config].
+         */
+        private fun createCache(config: Config): GitHubConfigCache =
+            config.getStringOrNull(CACHE_DIRECTORY)?.let { cacheDir ->
+                logger.debug("Using file-based cache in directory '{}'.", cacheDir)
+
+                val lockCheckInterval = config.getInt(LOCK_CHECK_INTERVAL)
```
nnobelis, line 190: Can you have a suggested default value for this property ?

config/github/src/main/kotlin/GitHubConfigFileProvider.kt
```diff
@@ -93,6 +101,19 @@ class GitHubConfigFileProvider(
...
+         * uses a [GitHubConfigFileCache] object to cache the fetched configuration files. Otherwise, caching is
+         * disabled.
+         */
+        const val CACHE_DIRECTORY = "gitHubCacheDirectory"
+
+        /**
+         * Configuration property that defines the interval for obtaining read locks (in seconds) when using the
+         * file-based configuration cache. This is only evaluated if a cache directory is defined.
+         */
+        const val LOCK_CHECK_INTERVAL = "gitHubCacheLockCheckInterval"
```
nnobelis, line 115: In would be nice to have the unit in the name of the property.

config/github/src/main/kotlin/GitHubConfigFileProvider.kt
oheger-bosch: A limit would not really help, the next request would then fail again. Note that the restrictions for this mechanism are documented in the commit message.

config/github/src/main/kotlin/GitHubConfigCache.kt
oheger-bosch: Extended the documentation.

config/github/src/main/kotlin/GitHubConfigDisabledCache.kt
oheger-bosch: Renamed the class.

config/github/src/main/kotlin/GitHubConfigFileCache.kt
oheger-bosch: This does not play well with the loop.

config/github/src/main/kotlin/GitHubConfigFileCache.kt
oheger-bosch: I am going to add clean-up logic in a follow-up PR. I don't plan to use a caching library here. I think the use case is rather special, since a fix revision guarantees that files will not become invalid in the cache. Therefore, the standard eviction strategies are probably not suitable.

config/github/src/main/kotlin/GitHubConfigFileProvider.kt
oheger-bosch: Added the suffix "Sec" to the property name.

config/github/src/main/kotlin/GitHubConfigFileProvider.kt
oheger-bosch: A default value is set in the `application.conf` files.

config/github/src/main/kotlin/GitHubConfigFileCache.kt
nnobelis: I was thinking more like a time-based eviction policy, e.g. "if this file was not accessed since x months, every it from the cache".

config/github/src/main/kotlin/GitHubConfigFileProvider.kt
nnobelis: Could you add it to the Kotlin property name too ?

oheger-bosch: Sure, but is this really needed? The unit is mentioned in the comment and is also obvious when the constant is used.

config/github/src/main/kotlin/GitHubConfigFileProvider.kt
nnobelis: I saw but I don't feel comfortable having the default value only in a `.conf` file. If anyone instantiate your class without it, it will crash.
Just a gut feeling, I won't block the PR over it.

nnobelis: Another question out of curiosity: `GitHubConfigFileCache` expects the `lockCheckInterval` to be a `duration`.
But you read it with `config.getInt`. How does it work ?

oheger-bosch: A conversion to a duration is done via the `lockCheckInterval.seconds` extension property.

config/github/src/main/kotlin/GitHubConfigFileProvider.kt
nnobelis: I read somewhere is best practice.
But you can leave it if you want.

config/github/src/main/kotlin/GitHubConfigFileProvider.kt
oheger-bosch: > I saw but I don't feel comfortable having the default value only in a `.conf` file. If anyone instantiate your class without it, it will crash. Just a gut feeling, I won't block the PR over it.

I made the cache implementations *internal*; then they cannot be instantiated by clients.

config/github/src/main/kotlin/GitHubConfigFileProvider.kt
oheger-bosch: I changed it - if it is best practice.

config/github/src/main/kotlin/GitHubConfigFileCache.kt
oheger-bosch: Eviction will apply to whole revisions. When the `main` branch moves forward, older revisions are typically no longer used.

nnobelis (approved):
Thanks for your patience!

PR #510, opened 2024-06-25
feat(api): Add an endpoint to return the ORT-Server version

### LINKED ISSUE #95: Add an API endpoint to get the ORT Server version / Git revision
E.g. to easily verify that one is running the latest / expected ORT Server version, it would be great if there was an API endpoint to get the version / Git revision (for display in the frontend).

mnonnenmacher:
Add an endpoint `/versions` to return the ORT-Server version. The endpoint returns a map as later on more tool versions will be returned, for example, the version of ORT that is in use.

The endpoint requires authentication to not expose the version to arbitrary clients for security reasons.

Fixes #95.

### DIFF
--- a/core/src/main/kotlin/api/VersionsRoute.kt
+++ b/core/src/main/kotlin/api/VersionsRoute.kt
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.core.api
+
+import io.github.smiley4.ktorswaggerui.dsl.get
+
+import io.ktor.server.application.call
+import io.ktor.server.response.respond
+import io.ktor.server.routing.Route
+
+import org.eclipse.apoapsis.ortserver.core.apiDocs.getVersions
+import org.eclipse.apoapsis.ortserver.model.ORT_SERVER_VERSION
+
+fun Route.versions() = get("versions", getVersions) {
+    call.respond(
+        mapOf(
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/api/VersionsRoute.kt
```diff
@@ -0,0 +1,37 @@
...
+import io.ktor.server.response.respond
+import io.ktor.server.routing.Route
+
+import org.eclipse.apoapsis.ortserver.core.apiDocs.getVersions
+import org.eclipse.apoapsis.ortserver.model.ORT_SERVER_VERSION
+
+fun Route.versions() = get("versions", getVersions) {
+    call.respond(
+        mapOf(
+            "ORT-Server" to ORT_SERVER_VERSION
```
sschuberth, line 34: I'm not sure about any conventions, but other code that uses `ORT_SERVER_VERSION` seems to refer to it as the version for "ORT Server" (without the dash), so maybe omit it here, too?

mmurto: I would go with camelCase, as the object will be transmitted as JSON to clients, where it will be deserialized. This will likely require the least boilerplate of the parsing in clients.

core/src/main/kotlin/api/VersionsRoute.kt
```diff
@@ -0,0 +1,37 @@
...
+import io.ktor.server.response.respond
+import io.ktor.server.routing.Route
+
+import org.eclipse.apoapsis.ortserver.core.apiDocs.getVersions
+import org.eclipse.apoapsis.ortserver.model.ORT_SERVER_VERSION
+
+fun Route.versions() = get("versions", getVersions) {
+    call.respond(
+        mapOf(
+            "ORT Server" to ORT_SERVER_VERSION
```
mmurto, line 34: I would use camelCase (so either `ORTServer` or `ortServer`) as discussed [here](https://github.com/eclipse-apoapsis/ort-server/pull/510#discussion_r1653005515).

mmurto (changes requested):

core/src/main/kotlin/api/VersionsRoute.kt
sschuberth: Hmm, but this is a free-from key name in a map with arbitrary entries, i.e. it's not a property name that's part of a schema, but more something that we'd display as-is in the UI, or?

mmurto: I guess that makes sense.

mmurto (approved):

core/src/main/kotlin/api/VersionsRoute.kt
sschuberth: So I guess that means you're with going with the more readable "Ort Server" than "ortServer".

sschuberth (approved):

core/src/main/kotlin/api/VersionsRoute.kt
mmurto: Yep!

mnonnenmacher: It's easy to change if we have a need for another format later on, for now I would go with the native formatting of the tool name.

PR #516, opened 2024-06-26
ui: Add parameters, labels and job config context to ORT Run

### LINKED ISSUE #312: Add parameters and labels to ORT Run creation form
Key-value (strings) pairs should be added.

Etsija:
With this PR, parameters and labels can be added to an ORT run. If Rerun functionality for an old ORT run is used, the parameters and labels used for the old run are also pre-populated in the run creation form.

In anticipation of future implementation, job configuration context is also added to the run creation form.

Capture1

All added properties of the ORT run are shown in the run details page.

Capture2

This resolves #312.

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
@@ -19,7 +19,8 @@
 
 import { zodResolver } from '@hookform/resolvers/zod';
 import { createFileRoute, useNavigate } from '@tanstack/react-router';
-import { useForm } from 'react-hook-form';
+import { PlusIcon, TrashIcon } from 'lucide-react';
+import { useFieldArray, useForm } from 'react-hook-form';
 import { z } from 'zod';
 
 import { useRepositoriesServicePostOrtRun } from '@/api/queries';
@@ -55,6 +56,11 @@ import { Separator } from '@/components/ui/separator';
 import { Switch } from '@/components/ui/switch';
 import { useToast } from '@/components/ui/use-toast';
 
+const keyValueSchema = z.object({
+  key: z.string().min(1),
+  value: z.string(), // Allow empty values for now
+});
+
 const formSchema = z.object({
   revision: z.string(),
   path: z.string(),
@@ -82,7 +88,9 @@ const formSchema = z.object({
       enabled: z.boolean(),
       formats: z.array(z.string()),
     }),
+    parameters: z.array(keyValueSchema).optional(),
   }),
+  labels: z.array(keyValueSchema).optional(),
   jobConfigContext: z.string().optional(),
 });
 
@@ -293,7 +301,22 @@ const CreateRunPage = () => {
               ortRun.jobConfigs.reporter?.formats ||
               baseDefaults.jobConfigs.reporter.formats,
           },
+          // Convert the parameters object map coming from the back-end to an array of key-value pairs.
+          // This needs to be done because the useFieldArray hook requires an array of objects.
+          parameters: ortRun.jobConfigs.parameters
+            ? Object.entries(ortRun.jobConfigs.parameters).map(([k, v]) => ({
+                key: k,
+                value: v,
+              }))
+            : [],
         },
+        // Convert the labels object map coming from the back-end to an array of key-value pairs.
+        labels: ortRun.labels
+          ? Object.entries(ortRun.labels).map(([k, v]) => ({
+              key: k,
+              value: v,
+            }))
+          : [],
         jobConfigContext:
           ortRun.jobConfigContext || baseDefaults.jobConfigContext,
       }
@@ -302,6 +325,24 @@ const CreateRunPage = () => {
   const form = useForm<z.infer<typeof formSchema>>({
     resolver: zodResolver(formSchema),
     defaultValues: defaultValues,
+  });
+
+  const {
+    fields: parametersFields,
+    append: parametersAppend,
+    remove: parametersRemove,
+  } = useFieldArray({
+    name: 'jobConfigs.parameters',
+    control: form.control,
+  });
+
+  const {
+    fields: labelsFields,
+    append: labelsAppend,
+    remove: labelsRemove,
+  } = useFieldArray({
+    name: 'labels',
+    control: form.control,
   });
 
   async function onSubmit(values: z.infer<typeof formSchema>) {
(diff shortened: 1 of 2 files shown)


ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -369,7 +374,30 @@ const CreateRunPage = () => {
...
+              name='jobConfigContext'
+              render={({ field }) => (
+                <FormItem className='pt-4'>
+                  <FormLabel>Job configuration context</FormLabel>
+                  <FormControl>
+                    <Input {...field} />
+                  </FormControl>
+                  <FormDescription>
+                    An optional context to be used when obtaining configuration
+                    for this ORT run. This context is passed to the
```
sschuberth, line 460: I believe we should not make too much use of "ORT" in descriptions like this. To the user, it's an implementation detail that ORT is used underneath. So maybe just replace "ORT run" with "job".

sschuberth: The next two sentences I'd probably drop, and instead of giving concrete examples, simply state generally like over [here](https://github.com/eclipse-apoapsis/ort-server/blob/fe810f368363e9a8ec4a4ab2604cb938a5bcf90e/config/spi/src/main/kotlin/Context.kt#L26) that "the meaning of the context is up for interpretation by the implementation of the configuration provider". (Or so.)

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -19,7 +19,8 @@
 
 import { zodResolver } from '@hookform/resolvers/zod';
 import { createFileRoute, useNavigate } from '@tanstack/react-router';
-import { useForm } from 'react-hook-form';
+import { PlusIcon, TrashIcon } from 'lucide-react';
```
sschuberth, line 22: Commit message nits:
- If referring to ORT, do so as "ORT", not "ort". However, we could also just say e.g. "run creation form", without "ORT".
- Avoid work duplication in "Add the ability of adding", maybe just "Allow to specify parameters..."

sschuberth: Just curious about the validation: As I guess every input can be accepted as a string, what kind of input would fail? The only case I can think of right now where it should fail is an empty key. I guess we should allow empty values, though, as for some use-cases just the presence of a key is enough.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -55,6 +56,11 @@ import { Separator } from '@/components/ui/separator';
 import { Switch } from '@/components/ui/switch';
 import { useToast } from '@/components/ui/use-toast';
 
+const keyValueSchema = z.object({
+  key: z.string().min(1),
+  value: z.string().min(1),
```
sschuberth, line 61: So given my above comment, this line should probably drop `.min(1)`. But I'm curious about @mnonnenmacher's opinion here: Do you have a use-case for empty label values?

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -402,12 +467,11 @@ const CreateRunPage = () => {
                 </FormItem>
               )}
             />
-
             <FormField
               control={form.control}
               name='path'
               render={({ field }) => (
-                <FormItem>
+                <FormItem className='pt-4'>
```
sschuberth, line 474: Unrelated change? Some goes for the line removal above.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.runs.$runIndex.index.tsx
```diff
@@ -115,14 +115,13 @@ const RunComponent = () => {
...
-              <TableHead>Run parameter</TableHead>
-              <TableHead>Value</TableHead>
+              <TableHead>Run details</TableHead>
               <TableHead className='sr-only'>Actions</TableHead>
             </TableRow>
           </TableHeader>
           <TableBody>
             <TableRow>
-              <TableCell>Created At</TableCell>
+              <TableCell>Created at</TableCell>
```
sschuberth, line 124: Strictly speaking these (good) spelling changes here and below should go to a separate commit, as they are unrelated to the main topic of this commit.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -421,6 +485,135 @@ const CreateRunPage = () => {
               )}
             />
 
+            <h3 className='mt-4'>Parameters</h3>
+            <div className='text-sm text-gray-500'>
+              A map with custom parameters for the whole ORT run. The parameters
```
sschuberth, line 490: IMO this needs some rewording, similar as commented on a previous description (i.e. drop "ORT" and concrete examples). But I'd like @mnonnenmacher to comment on this.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -421,6 +485,135 @@ const CreateRunPage = () => {
...
+              can be evaluated by the validation script executed by the config
+              worker. The script can convert these parameters to specific job
+              configurations understood by ORT Server. The first column is the
+              key and the second is the value.
```
sschuberth, line 494: Regarding this last sentence, can we instead have "Key" and "Value" headers for the columns?

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -421,6 +485,135 @@ const CreateRunPage = () => {
...
+                pAppend({ key: '', value: '' });
+              }}
+            >
+              Add parameter
+              <PlusIcon className='ml-1 h-4 w-4' />
+            </Button>
+
+            <h3 className='mt-4'>Labels</h3>
+            <div className='text-sm text-gray-500'>
+              The labels of this run. The first column is the key and the second
```
sschuberth, line 556: Labels are strictly speaking not attached to runs, but to results (ORT result files). Again, I'd prefer @mnonnenmacher to share his view for a good wording. Also here, instead of the last sentence, I'd prefer column headers.

mnonnenmacher: In the server context it's fine to say labels are associated with a run, the result files are merely a report format here.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
mnonnenmacher: @sschuberth I will try to go through the comments later today, but I would also be fine if you merge this without my review and we improve the wording in a separate PR.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -304,6 +327,24 @@ const CreateRunPage = () => {
...
+    remove: pRemove,
+  } = useFieldArray({
+    name: 'jobConfigs.parameters',
+    control: form.control,
+  });
+
+  const {
+    fields: lFields,
+    append: lAppend,
+    remove: lRemove,
```
mmurto, line 342: I'd use `parametersFields` and `labelsFields` for these so it's easier to comprehend what they are.

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
Etsija: Not unrelated, padding was needed due to moving the Path element up in the UI, to preserve clear spacing of the elements.

mmurto (approved):

PR #519, opened 2024-06-27
Fix creating ORT runs concurrently

mnonnenmacher:

### DIFF
--- a/core/src/main/kotlin/services/OrchestratorService.kt
+++ b/core/src/main/kotlin/services/OrchestratorService.kt
@@ -57,6 +57,7 @@ class OrchestratorService(
         labels: Map<String, String>?
     ): OrtRun {
         val ortRun = db.dbQuery(transactionIsolation = Connection.TRANSACTION_SERIALIZABLE) {
+            maxAttempts = 20
             ortRunRepository.create(repositoryId, revision, path, jobConfig, jobConfigContext, labels.orEmpty())
         }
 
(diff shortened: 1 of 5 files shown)


nnobelis (commented):
LGTM but please wait for other opinions.

mnonnenmacher:
> LGTM but please wait for other opinions.

I need to fix the failing tests anyway.

core/src/main/kotlin/services/OrchestratorService.kt
```diff
@@ -57,6 +57,7 @@ class OrchestratorService(
         labels: Map<String, String>?
     ): OrtRun {
         val ortRun = db.dbQuery(transactionIsolation = Connection.TRANSACTION_SERIALIZABLE) {
+            maxAttempts = 20
```
oheger-bosch, line 60: Just curious, how did you come to this value of 20? Did you do some experiments? I would be interested in the effect it has to increase this value. For instance, if going from 3 repetitions to 5 or maybe 10, how many ORT runs could then be started in parallel?

On a side note, if somebody wants to start 50 parallel runs on a single repository, there should probably some rate limiting be applied.

mnonnenmacher: It's a bit random, I chose 50 for the test case to use a value so high that we will probably never see it in production. With 10 max attempts that still failed for a few of them, so I went with 20 to be on the safe side.

oheger-bosch: Thanks for explaining.

oheger-bosch (dismissed):

PR #521, opened 2024-06-27
ci(github): Publish test results as job summaries

sschuberth:


.github/workflows/build-and-test.yml
```diff
@@ -104,6 +104,12 @@ jobs:
     - name: Run tests
       run: ./gradlew test -Dkotest.tags='!Integration'
 
+    - name: Create Test Summary
+      uses: test-summary/action@v2
+      with:
+        paths: "**/test-results/**/TEST-*.xml"
+      if: success() || failure()
```
nnobelis, line 111: Out of curiosity: why this condition is present ?

sschuberth: Because we'd like to publish the summary even if the previous step failed due to failing tests. Also see [this discussion](https://github.com/eclipse-apoapsis/ort-server/pull/119#discussion_r1572104378) about the difference with using `always()`.

mnonnenmacher (approved):

PR #525, opened 2024-06-27
chore(workers): Remove an unused receiver from an extension function

sschuberth:


workers/common/src/main/kotlin/common/Extensions.kt
```diff
@@ -75,7 +75,7 @@ fun ConfigManager.readConfigFileWithDefault(
  * This function realizes the contract that if a specific config file is requested, not being able to read it leads to
```
nnobelis, line 75: Commit message typo: ..._unsed_...

nnobelis (approved):

workers/common/src/main/kotlin/common/Extensions.kt
```diff
@@ -75,7 +75,7 @@ fun ConfigManager.readConfigFileWithDefault(
...
  * an exception, but the default file is allowed to not exist.
  */
-inline fun <reified T> ConfigManager.getConfigFileWithDefault(
+inline fun <reified T> getConfigFileWithDefault(
```
mnonnenmacher, line 78: My intention was that this function doesn't leak to the top-level but has to be called on the `ConfigManager` like the other similar functions, because it cannot be made private.

sschuberth: If you prefer, I could also revert this, but then suppress the inspection hint about the receiver being unused.

sschuberth: I've found [a better solution](https://github.com/eclipse-apoapsis/ort-server/pull/531).

PR #528, opened 2024-06-28
fix(compose): Use `ORT_SERVER_CORE_PORT` also for the UI

mnonnenmacher:
Use the `ORT_SERVER_CORE_PORT` variable also for configuring the UI service to make sure it expects the API at the correct port.


sschuberth:
> @sschuberth Can you verify that this works for you?

It does, thanks a lot!

sschuberth (approved):

Issue #532, opened 2024-06-28
No logs about authentication errors from GitHub config file provider


Etsija:
Currently, when setting up the GitHub configuration file provider with wrong bearer token, there are no ERROR messages about the failure to authenticate in the server log. Instead, there are ConfigExceptions coming from either [this](https://github.com/eclipse-apoapsis/ort-server/blob/2946f0c099edd65ea201ca7949f818a26c75d557/config/github/src/main/kotlin/GitHubConfigFileProvider.kt#L212) or [this](https://github.com/eclipse-apoapsis/ort-server/blob/2946f0c099edd65ea201ca7949f818a26c75d557/config/github/src/main/kotlin/GitHubConfigFileProvider.kt#L218) line, which is confusing, as the exceptions don't immediately tell the user that authorization with the GitHub REST API has failed.

I think `resolveContext()` of `GitHubConfigFileProvider` could be improved to include some helpful logs in case of auth failures. @oheger-bosch do you see any possibilites for such improvements? In my case, I just spent several hours of debugging why the branch or SHA-1 aren't found from our config repository, only to find out that I had copied wrong bearer token to the secrets, so the server couldn't access the repository at all.

oheger-bosch:
I have added checks for the response status of GitHub requests and better log and exception messages.

PR #534, opened 2024-06-28
GitHub caching

oheger-bosch:
This PR further improves the support for caching for `GitHubConfigFileProvider`. The content of folders is now cached as well, and the cache is periodically cleaned.


oheger-bosch:
> While at it, would it be possible to also add a commit to resolve #532, @oheger-bosch?

Added a commit which produces better logs and exception messages in case of errors from the GitHub API.

PR #536, opened 2024-06-28
fix(api): Remove CVS as a repository type

sschuberth:
CVS support was removed from ORT core in [1].

[1]: https://github.com/oss-review-toolkit/ort/pull/6932

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/Repository.kt
+++ b/api/v1/model/src/commonMain/kotlin/Repository.kt
@@ -110,6 +110,5 @@ enum class RepositoryType {
     GIT,
     GIT_REPO,
     MERCURIAL,
-    SUBVERSION,
-    CVS
+    SUBVERSION
 }


sschuberth:
As a related question, why does the API currently require the type to be explicitly configured at all, as ORT's downloader can auto-detect it from the URL?

api/v1/model/src/commonMain/kotlin/Repository.kt
```diff
@@ -110,6 +110,5 @@ enum class RepositoryType {
     GIT,
     GIT_REPO,
     MERCURIAL,
-    SUBVERSION,
-    CVS
+    SUBVERSION
```
mnonnenmacher, line 113: In theory this would also require a database migration to make sure that any existing entries are removed to avoid runtime errors. But I believe in this case the probability that they exist in any production databases is zero because support was removed over a year ago.

mnonnenmacher:
> As a related question, why does the API currently require the type to be explicitly configured at all, as ORT's downloader can auto-detect it from the URL?

The auto-detection is not 100% reliable, or is it?

sschuberth:
> The auto-detection is not 100% reliable, or is it?

I believe it is reliable as the URL is probed, for all implementations except GitRepo.

api/v1/model/src/commonMain/kotlin/Repository.kt
sschuberth: So, good to merge?

mnonnenmacher:
> > The auto-detection is not 100% reliable, or is it?
> 
> I believe it is reliable as the URL is probed, for all implementations except GitRepo.

Then we could consider to make this an optional property.

mnonnenmacher (approved):

PR #538, opened 2024-06-28
Fix config worker without validation script

mnonnenmacher:

### DIFF
--- a/workers/config/src/main/kotlin/ConfigComponent.kt
+++ b/workers/config/src/main/kotlin/ConfigComponent.kt
@@ -61,7 +61,7 @@ class ConfigComponent : EndpointComponent<ConfigRequest>(ConfigEndpoint) {
             }
 
             else -> {
-                logger.error("Unexpected result of Config worker: $result")
+                logger.error("Unexpected result of Config worker for run '$runId': $result")
                 ConfigWorkerError(runId)
             }
         }
--- a/workers/config/src/main/kotlin/ConfigWorker.kt
+++ b/workers/config/src/main/kotlin/ConfigWorker.kt
@@ -111,7 +111,7 @@ class ConfigWorker(
         } else {
             logger.info("Skipping validation as no script exists.")
 
-            RunResult.Ignored to Triple(
+            RunResult.Success to Triple(
                 context.ortRun.jobConfigs.asPresent(),
                 OptionalValue.Absent,
                 OptionalValue.Absent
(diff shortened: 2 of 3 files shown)


workers/config/src/main/kotlin/ConfigWorker.kt
```diff
@@ -105,6 +107,8 @@ class ConfigWorker(
                 )
             }
         } else {
+            logger.info("Skipping validation script as it does not exist.")
```
sschuberth, line 110: Nit: Maybe "Skipping validation as no script exists."

workers/config/src/main/kotlin/ConfigWorker.kt
```diff
@@ -83,6 +83,8 @@ class ConfigWorker(
         // TODO: Currently the path to the validation script is hard-coded. It may make sense to have it configurable.
         val validationScriptExists = configManager.containsFile(resolvedJobConfigContext, VALIDATION_SCRIPT_PATH)
         val (result, validationResult) = if (validationScriptExists) {
+            logger.info("Running validation script.")
```
sschuberth, line 86: Is there some else of useful information that we could log here, like the `ortRunId` or the `resolvedJobConfigContext`?

workers/config/src/main/kotlin/ConfigComponent.kt
```diff
@@ -50,7 +50,7 @@ class ConfigComponent : EndpointComponent<ConfigRequest>(ConfigEndpoint) {
 
         val runId = message.payload.ortRunId
         val responsePayload = when (val result = runBlocking { configWorker.run(runId) }) {
-            is RunResult.Success -> {
+            is RunResult.Success, RunResult.Ignored -> {
```
oheger-bosch, line 53: How about just returning `RunResult.Success` also in case the config script is missing? This should be a normal case, so does not require any special return value.

mnonnenmacher: Yes, that was the other option I considered, I wasn't sure if there was a reason to us `Ingored` that I didn't see.

oheger-bosch: Not sure why `Ignored` was used initially, but I think `Success` would be the correct result. Other workers use `Ignored` when they cannot find the current ORT run. This is not the case here.

workers/config/src/main/kotlin/ConfigWorker.kt
mnonnenmacher: I have added another log statement at the top, before the run id was only logged at the end.

oheger-bosch (approved):

Etsija (approved):

oheger-bosch (approved):

Issue #316, opened 2024-05-28
Secret properties in application.conf cannot be set via environment variables


mnonnenmacher:

mnonnenmacher:
The current implementation has the benefit that the RabbitMQ secrets have to be configured only once if the `secret-file` provider is used, as they are always requested from the same path, no matter which property in `application.conf` they refer to.

One option to fix this issue while keeping this behavior would be, that if a secret provider is configured, the values of the properties reference the name of the secret instead of the value.

For example, if no provider is configured, `analyzer.receiver.password` would contain the RabbitMQ password. If a provider is configured, the property would contain the path of the secret instead, for example `"rabbitMqPassword"`. The provider would then use this value to look up the actual secret value.

oheger-bosch:
That the configuration still contains properties for passwords is wrong IMHO; this is a remainder from an early stage before the introduction of secret providers. The mode that allows reading secrets from the configuration was just a quick work-around, but it was never fully functional due to the mismatches between secret names and configuration property names.

Wouldn't be the correct way to solve this issue to implement a secret provider that reads secrets from environment variables? (And if it was actually needed to have multiple sources for secrets, there could be a composite secret provider that queries a configurable set of child providers.)

mmurto:
I've understood and imagined the secret management to be like this:

- secrets required for deploying ORT Server are given to the containers via environment variables,
- secrets required for ORT steps to run within ORT Server, like VCS credentials for the analyzed projects are managed with the secret providers, so stored in Vault for example.

Is this incorrect?

mnonnenmacher:
@mmurto 

>     * secrets required for deploying ORT Server are given to the containers via environment variables,

Incorret, based on @oheger-bosch's response. As described in this issue it "accidentally" works for some secrets where the secret paths match those of the properties in the config files, but in general they should be loaded from the config secret provider.

>     * secrets required for ORT steps to run within ORT Server, like VCS credentials for the analyzed projects are managed with the secret providers, so stored in Vault for example.

Correct.


@oheger-bosch 

> Wouldn't be the correct way to solve this issue to implement a secret provider that reads secrets from environment variables? 

That would allow to use environment variables, but there would still be an inconsistency between config properties and secrets:
[...]

PR #540, opened 2024-06-28
feat(ui): Add some configuration options to the evaluator job

Etsija:
Add the possibility to specify evaluator rules, license classifications, copyright garbage, and resolutions files to the evaluator job.  The user can also specify a path within the repository where the configuration file is located.

Capture

Using `GitHubConfigFileProvider`, it was verified that all files (with their specific paths inside the repository) are properly passed to the Evaluator and Reporter.

By providing fallback files (copied from [oss-review-toolkit/ort-config](https://github.com/oss-review-toolkit/ort-config)) at the root directory of the GitHub config file provider, it was also verified that in case of empty fields, the files from the default directory were taken into the jobs correctly.

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
@@ -83,6 +83,10 @@ const formSchema = z.object({
     }),
     evaluator: z.object({
       enabled: z.boolean(),
+      ruleSet: z.string().optional(),
+      licenseClassificationsFile: z.string().optional(),
+      copyrightGarbageFile: z.string().optional(),
+      resolutionsFile: z.string().optional(),
     }),
     reporter: z.object({
       enabled: z.boolean(),
@@ -292,6 +296,14 @@ const CreateRunPage = () => {
             enabled:
               ortRun.jobConfigs.evaluator !== undefined &&
               ortRun.jobConfigs.evaluator !== null,
+            ruleSet: ortRun.jobConfigs.evaluator?.ruleSet || undefined,
+            licenseClassificationsFile:
+              ortRun.jobConfigs.evaluator?.licenseClassificationsFile ||
+              undefined,
+            copyrightGarbageFile:
+              ortRun.jobConfigs.evaluator?.copyrightGarbageFile || undefined,
+            resolutionsFile:
+              ortRun.jobConfigs.evaluator?.resolutionsFile || undefined,
           },
           reporter: {
             enabled:
@@ -373,7 +385,18 @@ const CreateRunPage = () => {
         }
       : undefined;
     const evaluatorConfig = values.jobConfigs.evaluator.enabled
-      ? {}
+      ? {
+          // Only include the config parameter structures if the corresponding form fields are not empty.
+          // In case they are empty, the default path from the config file provider will be used to
+          // resolve the corresponding files.
+          ruleSet: values.jobConfigs.evaluator.ruleSet || undefined,
+          licenseClassificationsFile:
+            values.jobConfigs.evaluator.licenseClassificationsFile || undefined,
+          copyrightGarbageFile:
+            values.jobConfigs.evaluator.copyrightGarbageFile || undefined,
+          resolutionsFile:
+            values.jobConfigs.evaluator.resolutionsFile || undefined,
+        }
       : undefined;
     const reporterConfig = values.jobConfigs.reporter.enabled
       ? {
@@ -987,7 +1010,82 @@ const CreateRunPage = () => {
                 <AccordionItem value='evaluator' className='flex-1'>
                   <AccordionTrigger>Evaluator</AccordionTrigger>
                   <AccordionContent>
-                    No job configurations yet implemented for Evaluator.
+                    <div className='text-sm text-gray-500'>
+                      In case any input field is left empty, the default path
+                      from the config file provider will be used for the
+                      corresponding file.
+                    </div>
+                    <FormField
+                      control={form.control}
+                      name='jobConfigs.evaluator.ruleSet'
+                      render={({ field }) => (
+                        <FormItem className='pt-4'>
+                          <FormLabel>Evaluator rules</FormLabel>
+                          <FormControl>
+                            <Input {...field} />
+                          </FormControl>
+                          <FormDescription>
+                            The path to the rules file to get from the
+                            configuration provider. Note: In case this field is
+                            left empty, the default path of the config file
+                            provider has to include a rules file, otherwise the
+                            Evaluator job will fail.
+                          </FormDescription>
+                          <FormMessage />
+                        </FormItem>
+                      )}
+                    />
+                    <FormField
+                      control={form.control}
+                      name='jobConfigs.evaluator.licenseClassificationsFile'
+                      render={({ field }) => (
+                        <FormItem className='pt-4'>
+                          <FormLabel>License classifications</FormLabel>
+                          <FormControl>
+                            <Input {...field} />
+                          </FormControl>
+                          <FormDescription>
+                            The path to the license classifications file to get
+                            from the configuration provider.
+                          </FormDescription>
+                          <FormMessage />
+                        </FormItem>
+                      )}
+                    />
+                    <FormField
+                      control={form.control}
+                      name='jobConfigs.evaluator.copyrightGarbageFile'
+                      render={({ field }) => (
+                        <FormItem className='pt-4'>
+                          <FormLabel>Copyright garbage</FormLabel>
+                          <FormControl>
+                            <Input {...field} />
+                          </FormControl>
+                          <FormDescription>
+                            The path to the copyright garbage file to get from
+                            the configuration provider.
+                          </FormDescription>
+                          <FormMessage />
+                        </FormItem>
+                      )}
+                    />
+                    <FormField
+                      control={form.control}
+                      name='jobConfigs.evaluator.resolutionsFile'
+                      render={({ field }) => (
+                        <FormItem className='pt-4'>
+                          <FormLabel>Resolutions</FormLabel>
+                          <FormControl>
+                            <Input {...field} />
+                          </FormControl>
+                          <FormDescription>
+                            The path to the resolutions file to get from the
+                            configuration provider.
+                          </FormDescription>
+                          <FormMessage />
+                        </FormItem>
+                      )}
+                    />
                   </AccordionContent>
                 </AccordionItem>
               </div>


sschuberth:
Thinking a bit about the order, I believe we should sort from most important / often used to least important / seldom used options. So that would mean (also note the slightly different titles I propose to use):

1. "Evaluator rules"
2. "License classifications"
3. "Copyright garbage"

mmurto:
Regarding the commit message

> If a suitable configuration file provider is used (for example, `GitHubConfigFileProvider`), the user can also specify a path (relative to the root directory of the config repository) where the configuration file is located.

does this only apply to come config file providers, or does it work for all? If it works for all, I think the message could be changed a little.

Etsija:
> Regarding the commit message
> 
> > If a suitable configuration file provider is used (for example, `GitHubConfigFileProvider`), the user can also specify a path (relative to the root directory of the config repository) where the configuration file is located.
> 
> does this only apply to come config file providers, or does it work for all? If it works for all, I think the message could be changed a little.

Well to be honest, I don't know. IIUC there is some legacy behaviour from ORT core that all configuration files needed to be hosted in the same directory, in form of true files or symlinks, so I don't actually know if ORT Server's local configuration file provider has such a requirement. Maybe @mnonnenmacher could shed some light to the issue. But not knowing if all config file providers has the freedom explained in the commit message, I wanted to be careful of giving such a statement.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -236,6 +239,9 @@ const CreateRunPage = () => {
       },
       evaluator: {
         enabled: true,
+        copyrightGarbageFile: '',
+        licenseClassificationsFile: '',
+        ruleSet: '',
```
sschuberth, line 244: Let's align with the other names and call this `rulesFile`. Also because the term "rule*Set*" reminds me too much of `evaluator/src/main/kotlin/RuleSet.kt` which is a class that *can* be used as part of rules, but does not *have* to be.

In general, I think it's a good idea to look at the respective CLI tool's options and their help when choosing names and descriptions. So in this case, I'd look at [these](https://github.com/oss-review-toolkit/ort/blob/146ab3bc6e5f790d304e1d6f4187f072961d8541/plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt#L110-L139).

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -987,7 +1008,63 @@ const CreateRunPage = () => {
...
+                      control={form.control}
+                      name='jobConfigs.evaluator.ruleSet'
+                      render={({ field }) => (
+                        <FormItem>
+                          <FormLabel>Evaluator rules</FormLabel>
+                          <FormControl>
+                            <Input {...field} />
+                          </FormControl>
+                          <FormDescription>
+                            The path to the ruleset file to use for the
```
sschuberth, line 1021: Also here: "rules file".

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -987,7 +1008,63 @@ const CreateRunPage = () => {
...
+                            evaluation, which is resolved from the configured
```
sschuberth, line 1022: You don't mention the "for the evaluation" part in the other descriptions, so I'd drop it here, too.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
mmurto: `rulesFile` might be a good variable name, but I think it's better natural language if it's called a rule (singular) file here.

sschuberth: Doesn't that somewhat imply that only a single rule can be implemented per file? I'd like to avoid that impression.

mmurto: I don't think that's the implication. There are "rule books" for sports that include a lot of rules, and even closer "rule set" that can include many rules.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
Etsija: `ruleSet` comes from back-end API and cannot be changed in front-end.

sschuberth: @mnonnenmacher do you think it makes sense to rename that backend property?

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -987,7 +1008,62 @@ const CreateRunPage = () => {
...
+                      render={({ field }) => (
+                        <FormItem>
+                          <FormLabel>Evaluator rules</FormLabel>
+                          <FormControl>
+                            <Input {...field} />
+                          </FormControl>
+                          <FormDescription>
+                            The path to the rules file which is resolved from
+                            the configured configuration source. If this is left
+                            empty, the default path from ORT will be used.
```
sschuberth, line 1023: BTW, what are the convention to wrap (or not) such texts? Should we maybe start to follow the one-sentence-per-line rule that we aim to follow for Markdown?

mmurto: Prettier automatically formats the files and wraps the text based on something, so I'd just let it do it.

mmurto: This part of ["Why Prettier?"](https://prettier.io/docs/en/why-prettier#building-and-enforcing-a-style-guide) may be relevant. ;)

> By far the biggest reason for adopting Prettier is to stop all the on-going debates over styles. [It is generally accepted that having a common style guide is valuable for a project and team](https://www.smashingmagazine.com/2012/10/why-coding-style-matters/) but getting there is a very painful and unrewarding process. People get very emotional around particular ways of writing code and nobody likes spending time writing and receiving nits.

Etsija: I never do line breaking manually in the code files, as Prettier does far better job for me!

sschuberth: > nobody likes spending time writing and receiving nits.

That's absolutely not true 🤣

Etsija:
Can I kindly ask for a decision from the reviewers, as I believe I have made the requested changes where possible, and given answers otherwise.

sschuberth:
> Can I kindly ask for a decision from the reviewers, as I believe I have made the requested changes where possible, and given answers otherwise.

I'm still [awaiting feedback](https://github.com/eclipse-apoapsis/ort-server/pull/540#discussion_r1660751278) from @mnonnenmacher, but we could also merge this now and improve later.

mmurto:
> > Can I kindly ask for a decision from the reviewers, as I believe I have made the requested changes where possible, and given answers otherwise.
> 
> I'm still [awaiting feedback](https://github.com/eclipse-apoapsis/ort-server/pull/540#discussion_r1660751278) from @mnonnenmacher, but we could also merge this now and improve later.

I wonder if the [commit message](https://github.com/eclipse-apoapsis/ort-server/pull/540#issuecomment-2198345713) can be a little misleading, if in fact all providers support the functionality, and it can't be changed afterwards. Maybe it would be clear engouh (and not misleading) if the whole "If a suitable configuration provider" sentence was dropped?

Etsija:
> > > Can I kindly ask for a decision from the reviewers, as I believe I have made the requested changes where possible, and given answers otherwise.
> > 
> > 
> > I'm still [awaiting feedback](https://github.com/eclipse-apoapsis/ort-server/pull/540#discussion_r1660751278) from @mnonnenmacher, but we could also merge this now and improve later.
> 
> I wonder if the [commit message](https://github.com/eclipse-apoapsis/ort-server/pull/540#issuecomment-2198345713) can be a little misleading, if in fact all providers support the functionality, and it can't be changed afterwards. Maybe it would be clear engouh (and not misleading) if the whole "If a suitable configuration provider" sentence was dropped?

From the code, it's not at all clear that a `LocalConfigFileProvider` and `ConfigSecretFileProvider` would support paths.

mmurto:
> > > > Can I kindly ask for a decision from the reviewers, as I believe I have made the requested changes where possible, and given answers otherwise.
> > > 
> > > 
> > > I'm still [awaiting feedback](https://github.com/eclipse-apoapsis/ort-server/pull/540#discussion_r1660751278) from @mnonnenmacher, but we could also merge this now and improve later.
> > 
> > 
> > I wonder if the [commit message](https://github.com/eclipse-apoapsis/ort-server/pull/540#issuecomment-2198345713) can be a little misleading, if in fact all providers support the functionality, and it can't be changed afterwards. Maybe it would be clear engouh (and not misleading) if the whole "If a suitable configuration provider" sentence was dropped?
> 
> From the code, it's not at all clear to me that a `LocalConfigFileProvider` and `ConfigSecretFileProvider` would support paths, and I am reluctant to state something which isn't confirmed in the commit message.

`ConfigSecretFileProvider` implements the `ConfigSecretProvider`, not `ConfigFileProvider`. There was some discussion of the distinction in [this comment](https://github.com/eclipse-apoapsis/ort-server/issues/316#issuecomment-2154506925) and in the thread more widely.

The Evaluator Runner [seems to call all of the paths ](https://github.com/eclipse-apoapsis/ort-server/blob/4da8897e37b8452d58e70fc3db54b2a110860641/workers/evaluator/src/main/kotlin/evaluator/EvaluatorRunner.kt#L76-L93)regardless of the `ConfigFileProvider`.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
mnonnenmacher: We had this discussion internally before but could not agree on another name. `ruleSetFile` and `rulesFile` would both be fine with me, but mainly I would like if the suffix was aligned with the other properties.

mnonnenmacher:
> > I wonder if the [commit message](https://github.com/eclipse-apoapsis/ort-server/pull/540#issuecomment-2198345713) can be a little misleading, if in fact all providers support the functionality, and it can't be changed afterwards. Maybe it would be clear engouh (and not misleading) if the whole "If a suitable configuration provider" sentence was dropped?
> 
> From the code, it's not at all clear to me that a `LocalConfigFileProvider` and `ConfigSecretFileProvider` would support paths, and I am reluctant to state something which isn't confirmed in the commit message.

Even if it's not mentioned explicitly in the `ConfigFileProvider` interface the intention is that implementations have to support subdirectories. That's why it uses `path` instead of `name` for the file parameters and has a `listFiles` function that takes a `path`.

mnonnenmacher (dismissed):
I wonder why you did not add support for the remaining `resolutionsFile` option as well, but feel free to merge anyway, that could also be added later.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
sschuberth: > I would like if the suffix was aligned with the other properties.

Exactly my thought as well.

Etsija:
Commit message edited, ready now for merge?

Etsija:
> Did you test without passing any values? There's some error happening for me when these are empty strings

OK I have to do some more to this. Returning to Draft then. Didn't think a feature this small would be so...resistant, but oh well. Might as well add the resolutions file support to it.

mmurto:
> > Did you test without passing any values? There's some error happening for me when these are empty strings
> 
> OK I have to do some more to this. Returning to Draft then. Didn't think a feature this small would be so...resistant, but oh well. Might as well add the resolutions file support to it.

My quick guess is that the Evaluator uses defaults if no paths are passed, but currently empty strings are passed as empty strings, and Evaluator can't find the files from `./`. As I believe empty string is never a valid option, a solution could be to handle it in `onSubmit` to pass `undefined` instead of `""` if the string is empty (`""`).

Etsija:
> > > Did you test without passing any values? There's some error happening for me when these are empty strings
> > 
> > 
> > OK I have to do some more to this. Returning to Draft then. Didn't think a feature this small would be so...resistant, but oh well. Might as well add the resolutions file support to it.
> 
> My quick guess is that the Evaluator uses defaults if no paths are passed, but currently empty strings are passed as empty strings, and Evaluator can't find the files from `./`. As I believe empty string is never a valid option, a solution could be to handle it in `onSubmit` to pass `undefined` instead of `""` if the string is empty (`""`).
> 
> To emphasize, this is pure guessing without trying or looking at the code. :)

Sadly, that wasn't the case, as I actually switched to use `undefined` instead of ''.  Seems as if `GitHubConfigFileProvider` won't tolerate empty values for the evaluator files we specify in the UI - as if it doesn't have anything to fall back into, although the logs hint it does:

```
ort-server-evaluator-worker-1  | 2024-07-01 15:33:07.748 [pool-1-thread-9] level=INFO  o.e.a.o.w.evaluator.EvaluatorRunner - No rule set path provided, using default path 'evaluator.rules.kts'.
ort-server-evaluator-worker-1  | 2024-07-01 15:33:08.119 [pool-1-thread-9] level=ERROR o.e.a.o.c.g.GitHubConfigFileProvider - Error response from GitHub API request: 404 .
ort-server-evaluator-worker-1  | 2024-07-01 15:33:08.120 [pool-1-thread-9] level=INFO  o.e.a.o.c.g.GitHubConfigFileProvider - Response body: {"message":"Not Found","documentation_url":"https://docs.github.com/rest/repos/contents#get-repository-content","status":"404"}
ort-server-evaluator-worker-1  | 2024-07-01 15:33:08.120 [pool-1-thread-9] level=ERROR o.e.a.o.w.evaluator.EvaluatorWorker - Error while processing message with traceId '8193e024-1c7e-40c4-9fd9-64b83132b13f': Error response from GitHub API request: 404 .
ort-server-evaluator-worker-1  | 2024-07-01 15:33:08.120 [pool-1-thread-9] level=ERROR o.e.a.o.w.e.EvaluatorComponent - Evaluator job '14' failed.
```

Etsija:
That was my hunch. So, it would be tolerable then to note that in the explanation of those parameters that in case of the said provider, the files _have_ to exist in the provider, so no empty values?

It would of course be best if this is validated in the form, but I don't think the UI even knows which config file provider is in use.

mmurto:
> That was my hunch. So, it would be tolerable then to note that in the explanation of those parameters that in case of the said provider, the files _have_ to exist in the provider, so no empty values?

I think empty values are a valid input, in which case it should fall back to the default path.

> It would of course be best if this is validated in the form, but I don't think the UI even knows which config file provider is in use.

At some point we could return the valid paths in the chosen provider and use them to validate it.

Etsija:
> > That was my hunch. So, it would be tolerable then to note that in the explanation of those parameters that in case of the said provider, the files _have_ to exist in the provider, so no empty values?
> 
> I think empty values are a valid input, in which case it should fall back to the default path.
> 
> > It would of course be best if this is validated in the form, but I don't think the UI even knows which config file provider is in use.
> 
> At some point we could return the valid paths in the chosen provider and use them to validate it.

_What_ default path?

mmurto:
> > > That was my hunch. So, it would be tolerable then to note that in the explanation of those parameters that in case of the said provider, the files _have_ to exist in the provider, so no empty values?
> > 
> > 
> > I think empty values are a valid input, in which case it should fall back to the default path.
> > > It would of course be best if this is validated in the form, but I don't think the UI even knows which config file provider is in use.
> > 
> > 
> > At some point we could return the valid paths in the chosen provider and use them to validate it.
> 
> _What_ default path?

https://github.com/eclipse-apoapsis/ort-server/blob/4da8897e37b8452d58e70fc3db54b2a110860641/workers/evaluator/src/main/kotlin/evaluator/EvaluatorRunner.kt#L76-L93

Etsija:
I have now fixed the empty fields problem (see the PR description for details), and also added support for the resolutions file.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -373,7 +385,25 @@ const CreateRunPage = () => {
...
+          ...(values.jobConfigs.evaluator.licenseClassificationsFile && {
+            licenseClassificationsFile:
+              values.jobConfigs.evaluator.licenseClassificationsFile,
+          }),
+          ...(values.jobConfigs.evaluator.resolutionsFile && {
+            resolutionsFile: values.jobConfigs.evaluator.resolutionsFile,
+          }),
+          ...(values.jobConfigs.evaluator.ruleSet && {
+            ruleSet: values.jobConfigs.evaluator.ruleSet,
+          }),
```
mmurto, line 405: I don't think there's need to spread here, I beliebe the key can exist if `undefined` is given. This could be a little simplere like

```ts
copyrightGarbageFile: values.jobConfigs.evaluator.copyrightGarbageFile || undefined
```

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -83,6 +83,10 @@ const formSchema = z.object({
     }),
     evaluator: z.object({
       enabled: z.boolean(),
+      copyrightGarbageFile: z.string().optional(),
```
sschuberth, line 86: Nit: For consistency, I'd prefer if the order of the form fields was also reflected in this hunk, and in the two hunks below.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -373,7 +385,25 @@ const CreateRunPage = () => {
         }
       : undefined;
     const evaluatorConfig = values.jobConfigs.evaluator.enabled
-      ? {}
+      ? {
+          // Only include the config param structures if the corresponding form fields are not empty.
```
sschuberth, line 389: Nit: Spell out "param" as "parameter".

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -987,7 +1017,83 @@ const CreateRunPage = () => {
...
+                      name='jobConfigs.evaluator.ruleSet'
+                      render={({ field }) => (
+                        <FormItem className='pt-4'>
+                          <FormLabel>Evaluator rules</FormLabel>
+                          <FormControl>
+                            <Input {...field} />
+                          </FormControl>
+                          <FormDescription>
+                            The path to the rules file which is resolved from
+                            the configured configuration source. Note: in case
```
sschuberth, line 1036: Start capitalized after ":", so "Note: In ...".

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
```diff
@@ -987,7 +1017,83 @@ const CreateRunPage = () => {
...
+                      name='jobConfigs.evaluator.resolutionsFile'
+                      render={({ field }) => (
+                        <FormItem className='pt-4'>
+                          <FormLabel>Resolutions</FormLabel>
+                          <FormControl>
+                            <Input {...field} />
+                          </FormControl>
+                          <FormDescription>
+                            The path to the resolutions file which is resolved
+                            from the configured configuration source.
```
sschuberth, line 1091: Esp. in this description there's a few word duplications: "resolutions" / "resolved", "configured" / "configuration". I propose to just say: "The path to the resolutions file to get from the configuration provider."

Similar above.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
Etsija: Already tried, didn't work.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
mmurto: Are we Americans now?! :D

Etsija: My thoughts exactly.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
mmurto: Alright, probably a bug in the API then.

Etsija:
What is this renovate validation failure that all of a sudden arose!?

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
sschuberth: We use American, not British, English in all text. So yes 😉

sschuberth:
> What is this renovate validation failure that all of a sudden arose!?

Probably some network flakyness. I've retriggered the job, and the failure is gone now.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
mmurto: I tried this with `copyrightGarbageFile: values.jobConfigs.evaluator.copyrightGarbageFile || undefined` and it did work for me. How did you get it to not work?

Etsija: Try evaluation rules file.

sschuberth (approved):

sschuberth:
@mmurto please also re-review.

ui/src/routes/_layout/organizations/$orgId.products.$productId.repositories.$repoId.create-run.tsx
mmurto: That was just for example, the code looks like:

```ts
    const evaluatorConfig = values.jobConfigs.evaluator.enabled
      ? {
          // Only include the config param structures if the corresponding form fields are not empty.
          // In case they are empty, the default path from the config file provider will be used to
          // resolve the corresponding files.
            copyrightGarbageFile:
              values.jobConfigs.evaluator.copyrightGarbageFile || undefined,
            licenseClassificationsFile:
              values.jobConfigs.evaluator.licenseClassificationsFile || undefined,
            resolutionsFile: values.jobConfigs.evaluator.resolutionsFile || undefined,
            ruleSet: values.jobConfigs.evaluator.ruleSet || undefined,
        }
      : undefined;
```

Etsija: OK I must've then been wrong in my tests, I'll make the changes and retest.

Etsija: Done and working with your suggested changes, thanks for verifying this!

mmurto (approved):

PR #553, opened 2024-07-01
ci(docker): Enable `latest` tag for published images

mmurto:
The `latest` tag is widely used and expected in the Docker ecosystem.
Tag the latest builds from the main branch with the tag.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -121,6 +121,7 @@ jobs:
           type=raw,value=${{ env.ORT_SERVER_VERSION }}
           type=ref,event=branch
           type=sha
+          type=raw,value=latest,enable={{is_default_branch}}
 
     - name: Build ${{ matrix.docker.image }} Image
       if: ${{ matrix.docker.dockerfile != '' }}
@@ -143,6 +144,7 @@ jobs:
           type=raw,value=${{ env.ORT_SERVER_VERSION }}
           type=ref,event=branch
           type=sha
+          type=raw,value=latest,enable={{is_default_branch}}
 
     - name: Build ${{ matrix.docker.jibImage }} Image
       if: ${{ matrix.docker.task != '' }}


sschuberth:
Cross-posting from an internal discussion:

> I believe the idea was to use Docker the main tag for builds from the main Git branch, and later a Docker latest tag for the image corresponding to the latest "real" release.

So I'm not sure this is how @mnonnenmacher intends it to work.

mmurto:
> Cross-posting from an internal discussion:
> 
> > I believe the idea was to use Docker the main tag for builds from the main Git branch, and later a Docker latest tag for the image corresponding to the latest "real" release.
> 
> So I'm not sure this is how @mnonnenmacher intends it to work.

If that's the intention, it makes sense to use latest for the releases when releases exist, but IMO makes sense to use latest for main before they exist.

mmurto:
> > I believe the idea was to use Docker the main tag for builds from the main Git branch, and later a Docker latest tag for the image corresponding to the latest "real" release.
> 
> That is correct, while there is no official definition of what the "latest" tag means, the common usage is that it points to the latest release. I did not add the latest tag to not create the false impression that there are releases.

I don't imagine someone interprets the latest tag existing to mean that real releases exist, or does it? I do think it's best used as the latest stable version, but as there are no stable version, having it for the most recent version would make sense. As it currently doesn't exist, for example running `docker run ghcr.io/eclipse-apoapsis/ort-server-core` doesn't work, as `docker run` [defaults to latest](https://docs.docker.com/engine/reference/run/#image-references) when the tag is omitted. The place where I ran into it was when setting up Renovate to create PRs to update our images. If you specify the Docker image by digest and want to have Renovate update it, it checks for updated versions [under the latest tag](https://docs.renovatebot.com/modules/manager/kustomize/#:~:text=without%20a%20version%2C%20digests%20are%20tracked%20as%20%3Alatest). Sure it is workable around via `newTag: main@digest`, but this is an example that many tools expect the tag to exist.

If you disagree and feel latest before proper releases do more harm than good, feel free to close this PR.

mnonnenmacher:
I the latest tag helps with getting Renovate to update our images that's reason enough for me.

mnonnenmacher (approved):

PR #562, opened 2024-07-02
refactor(vault): Remove an unneeded `runBlocking` call

oheger-bosch:
Rework the token handling to not fetch a token initially, but deal with a missing token during request execution. While this slightly makes the logic more complex, it has the benefits that during creation of a `VaultSecretsProvider` instance no request is sent, and the `runBlocking` call to fetch the initial token can be dropped.

### DIFF
--- a/secrets/vault/src/main/kotlin/VaultSecretsProvider.kt
+++ b/secrets/vault/src/main/kotlin/VaultSecretsProvider.kt
@@ -36,10 +36,9 @@ import io.ktor.client.request.setBody
 import io.ktor.http.ContentType
 import io.ktor.http.HttpStatusCode
 import io.ktor.http.contentType
+import io.ktor.http.encodedPath
 import io.ktor.http.isSuccess
 import io.ktor.serialization.kotlinx.json.json
-
-import java.util.concurrent.atomic.AtomicReference
 
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.sync.Mutex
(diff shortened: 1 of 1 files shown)


sschuberth:
Commit message nits:
- "While this slightly makes the logic more complex" -> "While this makes the logic slightly more complex"
- Probably singular "benefit"

secrets/vault/src/main/kotlin/VaultSecretsProvider.kt
```diff
@@ -153,27 +155,36 @@ class VaultSecretsProvider(
...
-                    }
-                }
-
-                request.headers[TOKEN_HEADER] = token.get()
+            if (request.url.encodedPath == LOGIN_PATH) {
+                // This is the login request; execute it directly.
                 execute(request)
             } else {
-                call
+                val currentToken = mutex.withLock { token }
```
mnonnenmacher, line 166: IIUC the reason to use the lock here is to wait if there is a running token request. Could you add a code comment to explain that?

mnonnenmacher (changes requested):

secrets/vault/src/main/kotlin/VaultSecretsProvider.kt
oheger-bosch: Well, basic protection against concurrent access to shared data. I have added a comment to the `token` variable.

mnonnenmacher (approved):

PR #574, opened 2024-07-02
chore(renovate): Update Node once a week

mmurto:
Node Docker image specified in UI's Dockerfile updates very often. It's
only used to build UI before serving it from NGINX, so update it only once
a week to reduce spam from Renovate.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -27,6 +27,10 @@
     {
       "matchPackageNames": ["eslint"],
       "rangeStrategy": "in-range-only"
+    },
+    {
+      "matchPackageNames": ["node"],
+      "schedule": ["after 9pm on sunday"]
     }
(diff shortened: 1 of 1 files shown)


mmurto:
May be something out of the ordinary as the rate hasn't been that fast before, but there were three updates today: https://github.com/eclipse-apoapsis/ort-server/pulls?q=is:pr+update+node.js+is:closed

mnonnenmacher:
> May be something out of the ordinary as the rate hasn't been that fast before, but there were three updates today: https://github.com/eclipse-apoapsis/ort-server/pulls?q=is:pr+update+node.js+is:closed

IIRC we had it a least once before that there were multiple updates on a single day.

mnonnenmacher (approved):

PR #584, opened 2024-07-03
fix(job-monitor): Include trace ID in lost job msg

haikoschol:
Previous to this change, using Kubernetes Job Monitor together with SQS transport failed when sending a lost job notification. This happened because trace ID in the message header was left empty. This change adds a dummy trace ID to the header.

### DIFF
--- a/transport/kubernetes-jobmonitor/src/main/kotlin/FailedJobNotifier.kt
+++ b/transport/kubernetes-jobmonitor/src/main/kotlin/FailedJobNotifier.kt
@@ -63,7 +63,7 @@ internal class FailedJobNotifier(
      * Orchestrator about jobs that disappeared in Kubernetes.
      */
     fun sendLostJobNotification(ortRunId: Long, endpoint: Endpoint<*>) {
-        val header = MessageHeader(traceId = "", ortRunId)
+        val header = MessageHeader(traceId = "<lost-job>", ortRunId)
         val message = Message(header, WorkerError(endpoint.configPrefix))
 
         sendToOrchestrator(message)


mnonnenmacher:
> Same issue as in #408. First I tried [not adding the trace ID if it is empty](https://github.com/eclipse-apoapsis/ort-server/pull/408#issuecomment-2163270448), but this caused an exception on the receiving side when `SqsMessageReceiverFactory` calls `checkNotNull` on the attributes. At this point there is no way to know whether a trace ID should be present or not.

So is the root issue that the SQS transport cannot handle messages without a trace id then? Because this issue does not occur with the RabbitMQ transport.

sschuberth:
> > Same issue as in #408. First I tried [not adding the trace ID if it is empty](https://github.com/eclipse-apoapsis/ort-server/pull/408#issuecomment-2163270448), but this caused an exception on the receiving side when `SqsMessageReceiverFactory` calls `checkNotNull` on the attributes. At this point there is no way to know whether a trace ID should be present or not.
> 
> So is the root issue that the SQS transport cannot handle messages without a trace id then? Because this issue does not occur with the RabbitMQ transport.

I basically have the same question as I had [here](https://github.com/eclipse-apoapsis/ort-server/pull/408#issuecomment-2163270448) for token also for the trace ID then: Should it be allowed to be empty? Given that it works so with RabbitMQ, I assume yes, and will instead make the SQS implementation also tolerate an empty trace ID.

sschuberth:
> and will instead make the SQS implementation also tolerate an empty trace ID.

I just checked, and both the RabbitMQ and SQS should already be able to handle empty (but not non-existing) trace IDs.

haikoschol:
> > and will instead make the SQS implementation also tolerate an empty trace ID.
> 
> I just checked, and both the RabbitMQ and SQS should already be able to handle empty (but not non-existing) trace IDs. So is there anything to do?

How did you check that?

If there is a lost job (row in the db with a non-terminal status, but no matching `Job` resource in the cluster) and SQS transport is used for communicating back to orchestrator, Kubernetes Job Monitor fails with the following log output:

```
2024-07-04 03:31:59.160 [DefaultDispatcher-worker-2] level=WARN  o.e.a.o.t.k.j.LostJobsFinder - Found 1 lost jobs for analyzer.
2024-07-04 03:31:59.160 [DefaultDispatcher-worker-2] level=DEBUG o.e.a.o.t.k.j.LostJobsFinder - Lost jobs: [AnalyzerJob(id=42, ortRunId=50, createdAt=2024-07-04T03:30:22Z, startedAt=null, finishedAt=null, configuration=AnalyzerJobConfiguration(allowDynamicVersions=true, disabledPackageManagers=null, enabledPackageManagers=[Bundler, Cargo, GoMod, GradleInspector, Maven, NPM, PIP, Pipenv, PNPM, Poetry, Yarn, Yarn2], environmentConfig=null, packageCurationProviders=[], packageManagerOptions=null, repositoryConfigPath=null, skipExcluded=true), status=SCHEDULED)]
2024-07-04 03:31:59.227 [DefaultDispatcher-worker-2] level=INFO  o.e.a.o.t.k.j.FailedJobNotifier - Sending 'WorkerError' message to 'OrchestratorEndpoint'. TraceID: ''.
2024-07-04 03:31:59.520 [DefaultDispatcher-worker-2] level=ERROR o.e.a.o.t.k.jobmonitor.Scheduler - Exception while executing scheduled action.
aws.sdk.kotlin.services.sqs.model.SqsException: The request parameter "MessageAttribute / traceId" must be set to a non-empty value.
	at aws.sdk.kotlin.services.sqs.serde.SendMessageOperationDeserializerKt.throwSendMessageError(SendMessageOperationDeserializer.kt:86)
	at aws.sdk.kotlin.services.sqs.serde.SendMessageOperationDeserializerKt.access$throwSendMessageError(SendMessageOperationDeserializer.kt:1)
	at aws.sdk.kotlin.services.sqs.serde.SendMessageOperationDeserializer.deserialize(SendMessageOperationDeserializer.kt:37)
	at aws.sdk.kotlin.services.sqs.serde.SendMessageOperationDeserializer.deserialize(SendMessageOperationDeserializer.kt:32)
	at aws.smithy.kotlin.runtime.http.operation.DeserializeHandler.call(SdkOperationExecution.kt:341)
	at aws.smithy.kotlin.runtime.http.operation.DeserializeHandler$call$1.invokeSuspend(SdkOperationExecution.kt)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:277)
	at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:95)
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:69)
	at kotlinx.coroutines.BuildersKt.runBlocking(Unknown Source)
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(Builders.kt:48)
	at kotlinx.coroutines.BuildersKt.runBlocking$default(Unknown Source)
	at org.eclipse.apoapsis.ortserver.transport.sqs.SqsMessageSender.send(SqsMessageSender.kt:54)
	at org.eclipse.apoapsis.ortserver.transport.kubernetes.jobmonitor.FailedJobNotifier.sendToOrchestrator(FailedJobNotifier.kt:82)
	at org.eclipse.apoapsis.ortserver.transport.kubernetes.jobmonitor.FailedJobNotifier.sendLostJobNotification(FailedJobNotifier.kt:69)
	at org.eclipse.apoapsis.ortserver.transport.kubernetes.jobmonitor.LostJobsFinder.checkForLostWorkerJobs(LostJobsFinder.kt:144)
	at org.eclipse.apoapsis.ortserver.transport.kubernetes.jobmonitor.LostJobsFinder.checkForLostJobs(LostJobsFinder.kt:119)
	at org.eclipse.apoapsis.ortserver.transport.kubernetes.jobmonitor.LostJobsFinder.run_HG0u8IE$checkForLostJobs(LostJobsFinder.kt:109)
	at org.eclipse.apoapsis.ortserver.transport.kubernetes.jobmonitor.LostJobsFinder.access$run_HG0u8IE$checkForLostJobs(LostJobsFinder.kt:57)
	at org.eclipse.apoapsis.ortserver.transport.kubernetes.jobmonitor.LostJobsFinder$run$1.invoke(LostJobsFinder.kt:109)
	at org.eclipse.apoapsis.ortserver.transport.kubernetes.jobmonitor.LostJobsFinder$run$1.invoke(LostJobsFinder.kt:109)
	at org.eclipse.apoapsis.ortserver.transport.kubernetes.jobmonitor.Scheduler$schedule$1$1.emit(Scheduler.kt:60)
	at org.eclipse.apoapsis.ortserver.transport.kubernetes.jobmonitor.Scheduler$schedule$1$1.emit(Scheduler.kt:58)
	at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke(SafeCollector.kt:11)
	at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke(SafeCollector.kt:11)
	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:113)
	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:82)
	at org.eclipse.apoapsis.ortserver.transport.kubernetes.jobmonitor.SchedulerKt$createTickerFlow$1.invokeSuspend(Scheduler.kt:83)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
	at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:111)
	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:99)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:584)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:811)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:715)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:702)
```

sschuberth:
> How did you check that?

By looking at the code 😸 But...

> Kubernetes Job Monitor fails with the following log output:

...ah, the exception actually comes from AWS SQS itself, not our SQS implementation. So apparently the former does not allow *any* empty message attributes. Let's discuss potential work-arounds today.

haikoschol:
> So is the root issue that the SQS transport cannot handle messages without a trace id then? Because this issue does not occur with the RabbitMQ transport.

The AWS SQS library does not accept empty key-value pairs in messages.

sschuberth:
> The AWS SQS library does not accept empty key-value pairs in messages.

Although [this](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html) says: "Amazon SQS doesn't interpret, validate, or use the appended data."

sschuberth:
And [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_MessageAttributeValue.html) it says "`Name`, `type`, `value` and the message body must not be empty or null.".

So the question probably boils down to: Should we by convention forbid empty trace IDs in ORT server, or should we en-/decode an actually empty value as something special for SQS?

oheger-bosch:
In this specific situation that a Kubernetes job got lost, the job monitor simply does not have a valid trace ID it could add to the message. Since the trace ID is not an essential piece of information, I would not want to enforce its presence. So I would opt for supporting this corner case in the SQS implementation.

sschuberth:
> So I would opt for supporting this corner case in the SQS implementation.

I'll prepare a PR.

haikoschol:
Somehow I get the impression that you're not very fond of my proposed solution. 😂

sschuberth:
> Somehow I get the impression that you're not very fond of my proposed solution. 😂

It would have been fine with me, as it's a corner case anyway 🤷🏻 

> I'll prepare a PR.

See https://github.com/eclipse-apoapsis/ort-server/pull/600.

PR #589, opened 2024-07-03
ci: Make the Docker build a callable workflow

nnobelis:
Hence, users can reuse this workflow in their (private) repositories. Workflow inputs will be added in a future commit.


nnobelis:
Another strategy with be used for our builds.

PR #590, opened 2024-07-03
fix(ui): Hide Router devtools in production

### LINKED ISSUE #484: Hide TanStack Router debug view in production builds
The TanStack Router debug view on the bottom left should not appear in production builds created with `pnpm build`.

mmurto:
Fixes #484.

### DIFF
--- a/ui/src/routes/__root.tsx
+++ b/ui/src/routes/__root.tsx
@@ -18,15 +18,26 @@
  */
 
 import { createRootRouteWithContext, Outlet } from '@tanstack/react-router';
-import { TanStackRouterDevtools } from '@tanstack/router-devtools';
+import React, { Suspense } from 'react';
 
 import { RouterContext } from '@/app';
+
+// Don't use Router devtools in production.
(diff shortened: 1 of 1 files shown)


ui/src/routes/__root.tsx
```diff
@@ -18,15 +18,26 @@
...
 import { RouterContext } from '@/app';
 
+// Don't use Router devtools in production.
+const TanStackRouterDevtools = import.meta.env.PROD
+  ? () => null
+  : React.lazy(() =>
+      import('@tanstack/router-devtools').then((res) => ({
+        default: res.TanStackRouterDevtools,
+      }))
+    );
```
sschuberth, line 32: Crazy how complicated this is 🤯

sschuberth (approved):

ui/src/routes/__root.tsx
mmurto: It could done simpler by just not rendering the devtools if the environment is production, but that way the component would still get bundled into the production bundle. This guard saves probably a few kB on each load. :D

PR #593, opened 2024-07-03
Improvements to user hooks and admin guarding

mmurto:
See commits.

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -19,10 +19,11 @@
 
 import { createRouter, RouterProvider } from '@tanstack/react-router';
 import { useEffect, useState } from 'react';
-import { hasAuthParams, useAuth } from 'react-oidc-context';
+import { hasAuthParams } from 'react-oidc-context';
 
 import { LoadingIndicator } from '@/components/loading-indicator';
 import { OpenAPI } from './api/requests/index.ts';
+import { useUser } from './hooks/useUser.ts';
 import { queryClient } from './lib/query-client.ts';
 import { routeTree } from './routeTree.gen';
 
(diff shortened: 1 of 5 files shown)


ui/src/app.tsx
```diff
@@ -19,10 +19,11 @@
 
 import { createRouter, RouterProvider } from '@tanstack/react-router';
 import { useEffect, useState } from 'react';
-import { hasAuthParams, useAuth } from 'react-oidc-context';
+import { hasAuthParams } from 'react-oidc-context';
 
 import { LoadingIndicator } from '@/components/loading-indicator';
 import { OpenAPI } from './api/requests/index.ts';
+import { useUser } from './hooks/useUser.ts';
```
Etsija, line 26: I guess it would be better code practise to use module path aliases `@/` for all imports inside the UI project, to avoid using relative paths; at least that's what we do in the Clearance UI code.

mmurto: That could be a style choice we can make, there seems to be an Eslint plugin for it as well. Currently relative imports are used in many places, even just above the line you commented.

Etsija: I'd personally prefer module path aliases in all imports. Could be that VS Code's auto import uses relative paths by default. The way we've done with Clearance UI has been to fix these as they come along in files.

mmurto: I think it makes sense to allow relative imports to at least subpahts of the current directory. If you're working on a file and want to split something from it that only that file uses, it makes sense to co-locate it next to it (or a subdir) and then referencing it with for example `./helper` instead of `@/lib/auth/permissions/helper` is cleaner and works after copy pasting the whole dir somewhere else. I agree that for parents `../some/helper` it probably almost always makes sense to use the alias import.

Etsija (approved):

PR #597, opened 2024-07-03
Rename `OrtIssue` to `Issue`

mnonnenmacher:
Align with the renaming in ORT, see the commit messages for details.


oheger-bosch (approved):
Note: This change breaks our config validation script. Also, the update on the production system should probably be done during a maintenance window since the renaming of database structures is likely to break ORT runs in progress.

PR #603, opened 2024-07-04
refactor(ui): Move utility types in create run page to separate file

mmurto:
This helps declutter the page component. The `-` prefix in routes makes
files not create routes even though they're in the routes directory,
allowing to colocate helpers and components not used anywhere else.

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/create-run.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/create-run.tsx
@@ -55,6 +55,7 @@ import { Label } from '@/components/ui/label';
 import { Separator } from '@/components/ui/separator';
 import { Switch } from '@/components/ui/switch';
 import { useToast } from '@/components/ui/use-toast';
+import { advisors, packageManagers, reportFormats } from './-types';
 
 const keyValueSchema = z.object({
   key: z.string().min(1),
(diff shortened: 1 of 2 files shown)


ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/create-run.tsx
```diff
@@ -55,6 +55,7 @@ import { Label } from '@/components/ui/label';
 import { Separator } from '@/components/ui/separator';
 import { Switch } from '@/components/ui/switch';
 import { useToast } from '@/components/ui/use-toast';
+import { advisors, packageManagers, reportFormats } from './-types';
```
mmurto, line 58: @Etsija I think this for example is a good use for relative imports.

Etsija: This was a really elegant way for decluttering. Related to this PR, I have some thoughts about more refactoring: as the create-run page reuses similar form components, they should probably be extracted into their own respective components. I could think of at least `components/form/(input-field.tsx, toggle-field.tsx, checkbox-field.tsx)`.

Etsija (approved):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/create-run.tsx
mmurto: Sounds good! It would be great to have a unified style for forms accross the UI, and shared components would very much help in that.

PR #604, opened 2024-07-04
feat(Docker): Enable Bazel in the Analyzer Docker build

nnobelis:


workers/analyzer/docker/Analyzer.Dockerfile
```diff
@@ -22,6 +22,7 @@
 # License-Filename: LICENSE
 
 ARG ANDROID_CMD_VERSION=11076708
+ARG BAZEL_VERSION=7.0.1
```
sschuberth, line 25: Why not the latest 7.2.1? If there's a reason, please provide it in the commit message.

workers/analyzer/docker/Analyzer.Dockerfile
```diff
@@ -408,6 +409,25 @@ RUN mkdir -p $DOTNET_HOME/bin \
...
+
+ENV BAZEL_HOME=/opt/bazel
+
+RUN mkdir -p $BAZEL_HOME/bin \
+    && if [ "$(arch)" = "aarch64" ]; then \
+    curl -L https://github.com/bazelbuild/bazel/releases/download/$BAZEL_VERSION/bazel-$BAZEL_VERSION-linux-arm64 -o $BAZEL_HOME/bin/bazel; \
+    else \
+    curl -L https://github.com/bazelbuild/bazel/releases/download/$BAZEL_VERSION/bazel-$BAZEL_VERSION-linux-x86_64 -o $BAZEL_HOME/bin/bazel; \
+    fi \
+    && chmod a+x $BAZEL_HOME/bin/bazel
```
sschuberth, line 426: @haikoschol is this still the "right" way to install Bezel, or should maybe Bazelisk be used instead by now?

workers/analyzer/docker/Analyzer.Dockerfile
nnobelis: This is just aligned with ORT :shrug: 

What do you think we should do?

sschuberth: Hmm, I just checked, and running ORT core tests with Bazel 7.2.1 requires changes to test assets, so I'd propose to stick both to Bazel 7.0.1 for now, but mention the alignment in the commit message.

workers/analyzer/docker/Analyzer.Dockerfile
sschuberth: Let's go ahead with this and eventually improve later, when doing the Bazel 7.2.1 (or later) upgrade.

sschuberth (approved):

workers/analyzer/docker/Analyzer.Dockerfile
haikoschol: This is [what Bazelisk does](https://github.com/bazelbuild/bazelisk?tab=readme-ov-file#where-does-bazelisk-get-bazel-from) essentially.

I don't think Bazelisk supports installing a specific version. The point of using it is to allow [specifying the Bazel version to run in a repository/workspace](https://github.com/bazelbuild/bazelisk?tab=readme-ov-file#how-does-bazelisk-know-which-bazel-version-to-run).

PR #606, opened 2024-07-04
ci: Fix Docker images tags

nnobelis:

image


mnonnenmacher:
> Please have a look at the individual commit messages for the details.
> 
> image

Looking at that image I some another problem: The `main` and `sha-xxx` tags are now assigned only to the last pushed image which is `jdk11`. Instead, the `main` tag should be assigned to the `jdk17` build as before. I think the cleanest solution would to add the `-jdk11` and `-jdk21` suffixes also to the `main` and `sha-xxx` tags.

nnobelis:
image

@mnonnenmacher

mnonnenmacher (approved):

PR #607, opened 2024-07-04
feat(api): Also expose the core ORT version as part of the endpoint

sschuberth:


core/src/main/kotlin/api/VersionsRoute.kt
```diff
@@ -28,10 +28,13 @@ import io.ktor.server.routing.Route
...
 import org.eclipse.apoapsis.ortserver.model.ORT_SERVER_VERSION
 
+import org.ossreviewtoolkit.utils.ort.Environment
+
 fun Route.versions() = get("versions", getVersions) {
     call.respond(
         mapOf(
-            "ORT Server" to ORT_SERVER_VERSION
+            "ORT Server" to ORT_SERVER_VERSION,
+            "ORT Core" to Environment.ORT_VERSION
```
mnonnenmacher, line 37: Should this be called "ORT Core"? I think we use "Core" nowhere else in the context of ORT. How about just "ORT" or spell it out to "OSS Review Toolkit"?

mnonnenmacher (changes requested):

core/src/main/kotlin/api/VersionsRoute.kt
sschuberth: > I think we use "Core" nowhere else in the context of ORT.

Not until now at least, but the need arose with the advent of ORT Server. I somehow wanted to mention "ORT", but at the same time make clear that this is not ORT, the CLI, that people are used to refer to, but the sum of the ORT libraries.

So, I think it's justified to "invent" the new term "ORT Core" here. @mmurto any opinion?

sschuberth: Among these options, I'd prefer to stick with "ORT Core", and "ORT" goes second.

mnonnenmacher (approved):

core/src/main/kotlin/api/VersionsRoute.kt
mmurto: Just popped in to mind that "ORT Core" may be a little confusing sometimes as the API component of ORT Server is also called "Core".

PR #616, opened 2024-07-05
feat(ui): Add entity names to some toast messages

### LINKED ISSUE #552: Show entity name in delete confirmation "toast"
After deleting e.g. an organization, a "toast" scrolls in at the bottom right that confirms the deletion. However, it only contains a generic message. We should name the entity that was deleted as part of the confirmation.

Etsija:
Please see the commit message for details.

### DIFF
--- a/ui/src/routes/_layout/create-organization.tsx
+++ b/ui/src/routes/_layout/create-organization.tsx
@@ -55,10 +55,10 @@ const CreateOrganizationPage = () => {
   const { toast } = useToast();
 
   const { mutateAsync, isPending } = useOrganizationsServicePostOrganizations({
-    onSuccess() {
+    onSuccess(data) {
       toast({
         title: 'Create Organization',
-        description: 'New organization created successfully.',
+        description: `New organization "${data.name}" created successfully.`,
       });
       navigate({
         to: '/',
--- a/ui/src/routes/_layout/organizations/$orgId/create-product.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/create-product.tsx
@@ -57,10 +57,10 @@ const CreateProductPage = () => {
   const { toast } = useToast();
 
   const { mutateAsync, isPending } = useProductsServicePostProduct({
-    onSuccess() {
+    onSuccess(data) {
       toast({
         title: 'Create Product',
-        description: 'New product created successfully.',
+        description: `New product "${data.name}" created successfully.`,
       });
       navigate({
         to: '/organizations/$orgId',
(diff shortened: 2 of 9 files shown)


ui/src/routes/_layout/create-organization.tsx
```diff
@@ -58,7 +58,10 @@ const CreateOrganizationPage = () => {
     onSuccess() {
       toast({
         title: 'Create Organization',
-        description: 'New organization created successfully.',
+        description:
+          'New organization "' +
+          form.getValues().name +
+          '" created successfully.',
```
sschuberth, line 64: Just wondering, can / should [template literals](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals) be used here to improve readability via string interpolation? Like

    `New organization "${form.getValues().name}" created successfully.`

Etsija: I've never been a fan of those, as I'm not seeing why they would be more readable.

sschuberth: We should have a clear line here, and do it (or not) consistently, if at all. I'm curious about @mmurto's advice about best practices here.

Etsija: As I'm sure he's pro-literals, I've already made the changes.

sschuberth (dismissed):

sschuberth:
> Resolves #552.

Actually this bit should go to the commit message, not in the PR discussion.

ui/src/routes/_layout/organizations/$orgId/create-product.tsx
```diff
@@ -60,7 +60,7 @@ const CreateProductPage = () => {
     onSuccess() {
       toast({
         title: 'Create Product',
-        description: 'New product created successfully.',
+        description: `New product "${form.getValues().name}" created successfully.`,
```
mmurto, line 63: I would change all of these to use the return value of the mutations to get the names/etc. from there. I believe the toast will show a wrong name with the current implementation if someone clicks submit and writes something to the input while the submit is processing. The return value should be available to the onSuccess callback.

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId/create-product.tsx
sschuberth: No, I've removed this PR from the merge queue, so you can still fix it.

mmurto (approved):

PR #619, opened 2024-07-05
Add parent ids to API entities

mnonnenmacher:
Add all parent ids to products, repositories, and ORT runs. See the commit messages for details.


Etsija:
LGTM, but I'd like @sschuberth to have a look at it, too.

sschuberth (approved):

PR #631, opened 2024-07-08
Add shortcut to ORT run details page

### LINKED ISSUE #433: Add route to ORT run details page by ORT run id
Add a route at `/runs/{id}` that forwards to the run details page at `organizations/{id}/products/{id}/repositories/{id}/runs/{index}`. For this the route has to load the ORT run details via the `/runs/{id}` endpoint (#425) to get the required information.

This route would be useful for short links to an ORT run and for situations where only the ORT run id is known. In addition it would be nice to have a text field to enter the ORT run id in the UI that forwards to the new route, for example, in a prominent place, like the title bar.

Etsija:

Capture1

Capture2

Capture3

This PR resolves #433.

### DIFF
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -17,9 +17,13 @@
  * License-Filename: LICENSE
  */
 
-import { Link, useRouterState } from '@tanstack/react-router';
+import { zodResolver } from '@hookform/resolvers/zod';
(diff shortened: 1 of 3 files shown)


mmurto (approved):

sschuberth:
A proposal for future improvement: How about displaying an "error toast" for non-existing runs instead of navigating away from the current page to a page that just shows the error, and requires another click to get back?

PR #633, opened 2024-07-08
fix(analyzer): Resolve secrets of package curation providers

mnonnenmacher:
Resolve the secrets in the configurations of the package curation providers from the configured secret storage, instead of using the secret values directly.

As the `AnalyzerRunner` does not have access to the worker context when running in a fork, the secrets need to be resolved before creating the fork.

### DIFF
--- a/workers/analyzer/src/test/kotlin/AnalyzerRunnerTest.kt
+++ b/workers/analyzer/src/test/kotlin/AnalyzerRunnerTest.kt
@@ -36,6 +36,7 @@ import io.kotest.matchers.shouldNotBe
 import io.kotest.matchers.string.shouldContain
 
 import io.mockk.coEvery
+import io.mockk.coVerify
 import io.mockk.every
 import io.mockk.just
 import io.mockk.mockk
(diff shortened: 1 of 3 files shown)


sschuberth (dismissed):

workers/analyzer/src/test/kotlin/AnalyzerRunnerTest.kt
```diff
@@ -422,6 +431,130 @@ class AnalyzerRunnerTest : WordSpec({
...
+                every { command() } returns listOf("some", "command")
+            }
+
+            val config = AnalyzerJobConfiguration(packageCurationProviders = packageCurationProviderConfigs)
+
+            val runner = spyk(AnalyzerRunner(ConfigFactory.empty()))
+
+            coEvery { runner.createProcessBuilder(any(), any(), any(), any()) } returns processBuilder
+
+            runCatching {
```
oheger-bosch, line 536: The `runCatching` should be removed. Otherwise, exceptions and test failures are silently swallowed.

oheger-bosch (changes requested):

workers/analyzer/src/test/kotlin/AnalyzerRunnerTest.kt
mnonnenmacher: The important part of the test is the `coVerify` block below which verifies that the `runForked` function is called with resolved secrets. The `runBlocking` is currently required because `run` throws an exception, but I did not want to add more setup code which is required for the test.

oheger-bosch: This is not really obvious. Could you put the `run` call in a `shouldThrow` block instead and add a corresponding comment? Then there is at least some control about what is happening.

sschuberth (approved):

PR #636, opened 2024-07-08
Add total counts to organization endpoints that return lists

mnonnenmacher:
Add the total entity counts to organization endpoints to enable proper pagination in the UI. Please see the commit messages for details.

This is a first step to implement the requirements for #453, the remaining endpoints will be aligned in separate PRs.

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/PagedResponse.kt
+++ b/api/v1/model/src/commonMain/kotlin/PagedResponse.kt
@@ -33,6 +33,18 @@ data class PagedResponse<E>(
     val options: PagingOptions
 )
 
+/**
+ * A response object for returning paged lists of entities.
+ */
+@Serializable
+data class PagedResponse2<E>(
+    /** The list of returned entities. */
+    val data: List<E>,
+
+    /** The pagination information for the result set. */
+    val pagination: PagingData
+)
+
 @Serializable
 data class PagingOptions(
     /** An optional limit for the number of items to return. */
--- a/api/v1/model/src/commonMain/kotlin/PagedResponse.kt
+++ b/api/v1/model/src/commonMain/kotlin/PagedResponse.kt
@@ -60,7 +60,19 @@ data class PagingOptions(
      * An optional list of properties by which the result set should be sorted.
      */
     val sortProperties: List<SortProperty>? = null
-)
+) {
+    /**
+     * Convert this object to [PagingData] with the provided [totalCount]. This requires that all properties of this
+     * object are not `null`, an exception is thrown otherwise.
+     */
+    fun toPagingData(totalCount: Long): PagingData {
+        checkNotNull(limit)
+        checkNotNull(offset)
+        checkNotNull(sortProperties)
+
+        return PagingData(limit, offset, totalCount, sortProperties)
+    }
+}
 
 @Serializable
 data class PagingData(
--- a/api/v1/model/src/commonMain/kotlin/PagedResponse.kt
+++ b/api/v1/model/src/commonMain/kotlin/PagedResponse.kt
@@ -51,6 +51,21 @@ data class PagingOptions(
 )
 
 @Serializable
+data class PagingData(
+    /** The limit for the number of items to return. */
+    val limit: Int,
+
+    /** The offset for the items to return. */
+    val offset: Long,
+
+    /** The total count of items. */
+    val totalCount: Long,
+
+    /** The properties by which the result set was sorted. */
+    val sortProperties: List<SortProperty>
+)
+
+@Serializable
 /**
  * A data class defining a property by which a query result should be sorted.
  */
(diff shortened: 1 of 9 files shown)


api/v1/model/src/commonMain/kotlin/PagedResponse.kt
```diff
@@ -50,6 +50,21 @@ data class PagingOptions(
     val sortProperties: List<SortProperty>? = null
 )
 
+@Serializable
+data class PagingData(
+    /** The limit for the number of items to return. */
```
sschuberth, line 79: Has 0 a special meaning, like "no limit"?

api/v1/model/src/commonMain/kotlin/PagedResponse.kt
```diff
@@ -50,6 +50,21 @@ data class PagingOptions(
...
+    val limit: Int,
```
sschuberth, line 80: Why is this just an `Int` if the below two are `Long`s?

api/v1/model/src/commonMain/kotlin/PagedResponse.kt
```diff
@@ -50,6 +50,21 @@ data class PagingOptions(
...
+
+    /** The offset for the items to return. */
+    val offset: Long,
+
+    /** The total count of items. */
+    val totalCount: Long,
+
+    /** The properties by which the result set was sorted. */
+    val sortProperties: List<SortProperty>,
```
sschuberth, line 89: Is the trailing comma intended?

api/v1/model/src/commonMain/kotlin/PagedResponse.kt
```diff
@@ -33,6 +33,18 @@ data class PagedResponse<E>(
     val options: PagingOptions
 )
 
+/**
```
sschuberth, line 36: Commit message:
- I guess "pages" -> "paged"
- "having to *apply* the changes"

api/v1/model/src/commonMain/kotlin/PagedResponse.kt
mnonnenmacher: No.

api/v1/model/src/commonMain/kotlin/PagedResponse.kt
mnonnenmacher: Because requesting more than max int entities would not be feasible, but the database could contain more entries.

api/v1/model/src/commonMain/kotlin/PagedResponse.kt
mnonnenmacher: No.

api/v1/model/src/commonMain/kotlin/PagedResponse.kt
sschuberth: Should we consider doing so then? Are there use-cases where you definitely want to get all entries at once?

sschuberth: But doing so would require to make `limit` a `Long` as well, which you [below](https://github.com/eclipse-apoapsis/ort-server/pull/636#discussion_r1668899703) said to be infeasible (as it would allow requesting more than max int entities). So I guess we should keep it as-is.

mnonnenmacher: I don't see a need for that currently, and it could be done by providing a very large limit. The API currently enforces a default limit of 20 if the user does not provide a limit, but we probably should also consider setting an upper bound to prevent that a request can use up all memory (the lists are not streamed).

core/src/main/kotlin/api/OrganizationsRoute.kt
```diff
@@ -82,13 +84,17 @@ fun Route.organizations() = route("organizations") {
...
             .filter { hasPermission(it.id, OrganizationPermission.READ) }
-            .paginate(pagingOptions)
+
+        val pagedOrganizations = filteredOrganizations.paginate(pagingOptions)
             .map { it.mapToApi() }
 
-        val pagedResponse = PagedResponse(organizations, pagingOptions)
+        val pagedResponse = PagedResponse2(
+            pagedOrganizations,
+            pagingOptions.toPagingData(filteredOrganizations.size.toLong())
```
sschuberth, line 96: This converts an `Int`-typed `size` to `Long`. So is there even a point in making the `totalCount` a `Long`?

oheger-bosch: I wonder the same. Especially, as before the whole list of entities is loaded into memory to apply the permissions filter.

mnonnenmacher: That's a special case for this endpoint because permission based filtering cannot be done by the database query. All other endpoints (at least currently) will take the total count from the database where it can be a `Long`.

sschuberth (approved):

core/src/main/kotlin/api/OrganizationsRoute.kt
mmurto: > That's a special case for this endpoint because permission based filtering cannot be done by the database query.

Not super deep in how the server does auth checks, but I assume it's possible to get a list of organization ids that the user has the correct roles for in Keycloak? The database query could then be filtered based on this, couldn't it?

PR #639, opened 2024-07-08
Minor about menu improvements

sschuberth:

### DIFF
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -37,6 +37,7 @@ import {
   DropdownMenu,
   DropdownMenuContent,
   DropdownMenuItem,
+  DropdownMenuSeparator,
   DropdownMenuTrigger,
 } from './ui/dropdown-menu';
 import { Input } from './ui/input';
@@ -218,6 +219,7 @@ export const Header = () => {
                 <DropdownMenuItem>Admin</DropdownMenuItem>
               </Link>
             )}
+            <DropdownMenuSeparator />
             <DropdownMenuItem
               onClick={async () => {
                 await user.signoutRedirect();
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -210,6 +210,9 @@ export const Header = () => {
             </Button>
           </DropdownMenuTrigger>
           <DropdownMenuContent align='end'>
+            <Link to='/about'>
+              <DropdownMenuItem>About</DropdownMenuItem>
+            </Link>
             {user.hasRole('superuser') && (
               <Link to='/admin'>
                 <DropdownMenuItem>Admin</DropdownMenuItem>


ui/src/components/header.tsx
```diff
@@ -210,11 +210,15 @@ export const Header = () => {
...
           <DropdownMenuContent align='end'>
+            <Link to='/about'>
+              <DropdownMenuItem>About</DropdownMenuItem>
+            </Link>
             {user.hasRole('superuser') && (
               <Link to='/admin'>
                 <DropdownMenuItem>Admin</DropdownMenuItem>
               </Link>
             )}
+            <hr />
```
mmurto, line 221: Replace the generic `hr` with [`DropdownMenuSeparator`](https://github.com/eclipse-apoapsis/ort-server/blob/2c90f97e9543e7cba6d37ffd98e481ad4a78cf18/ui/src/components/ui/dropdown-menu.tsx#L171-L181).

mmurto (changes requested):

ui/src/components/header.tsx
sschuberth: Mind elaborating why / what the effect is? Isn't generic better?

mmurto: Sure! Generic in this UI component context means _base unstyled HTML component_, so no CSS or JS. By using the UI components instead of these we gain styles (the linked separator component has Tailwind classes `-mx-1 my-1 h-px bg-muted`, designed to fit in with other dropdown elements) and everything that the underlying [Radix UI component](https://www.radix-ui.com/primitives/docs/components/dropdown-menu) does. This is probably not that much in the case of a separator, but what it often does is provide keyboard navigation, make things more accessible and handle different edge-cases.

Does that make sense?

sschuberth: Makes sense. Though why not https://ui.shadcn.com/docs/components/separator?

mmurto: [If we look at it](https://github.com/eclipse-apoapsis/ort-server/blob/2c90f97e9543e7cba6d37ffd98e481ad4a78cf18/ui/src/components/ui/separator.tsx#L28-L29) compared to the dropdown separator, we see that it is different color (`bg-muted` vs. `bg-border`) and it has no margins. That means that someone has designed the dropdown separator to fit in with other dropdown elements so it's easier to get a polished look.

There may be some differences with the underlying Radix components as well, but probably not huge in the case of a separator.

mmurto: Can you spot any differences? 🤓 

`DropdownMenuSeparator`: 
image

`Separator`:
image

`hr`:
image

sschuberth: I can, though I probably wouldn't say to like `` DropdownMenuSeparator`   best 😉

mmurto: Using component libraries is IMO not really about what each individual would like the best, but consistency. But it's [completely editable](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/components/ui/dropdown-menu.tsx), feel free to make a PR. 🤓

Issue #618, opened 2024-07-05
Support the notifier job configuration


mnonnenmacher:
Add support for the notifier job configuration to the form to create a new ORT run. Also show the notifier job configuration in the run details.

Etsija:
Can I ask you about the `username` and `password` fields of Notifier configuration: are they really not secrets?
```
    "notifier": {
      "notifierRules": "reporter.notifications.kts",
      "resolutionsFile": "resolutions.yml",
      "mail": {
        "recipientAddresses": [
          "mail@example.com",
          "info@example.com"
        ],
        "mailServerConfiguration": {
          "hostName": "smtp.example.com",
          "port": 587,
          "username": "user",
          "password": "password",
          "useSsl": true,
          "fromAddress": "no-reply@example.com"
        }
      },
      "jira": {
        "jiraRestClientConfiguration": {
          "serverUrl": "https://jira.example.com",
          "username": "user",
          "password": "password"
        }
      }
    },
```

Etsija:
ORT run creation form is already over 1000 lines long, so we're probably going to do some refactoring to it before adding new configurations.

mnonnenmacher:
> Can I ask you about the `username` and `password` fields of Notifier configuration: are they really not secrets; so they are exposed in the UI?

Good point, they are treated as secrets, so the values are paths to secrets in the secret storage.

Etsija:
> > Can I ask you about the `username` and `password` fields of Notifier configuration: are they really not secrets; so they are exposed in the UI?
> 
> Good point, they are treated as secrets, so the values are paths to secrets in the secret storage.

Thanks, so it's safe for us to expose them in the form as ordinary text inputs.

PR #648, opened 2024-07-09
ui: Refactor ORT Run creation form

Etsija:
ORT run creation form is rapidly becoming difficult to understand and maintain, so this PR does some crude code refactoring work, mainly to separate job configurations into their own respective components. This will make it easier to add new jobs to the form, for example the Notifier, as requested in #618.


mmurto:
@Etsija if you're writing with the `any`, this example could maybe be helpful: https://codesandbox.io/s/kind-wildflower-1kzege?file=/src/App.tsx.

ui/src/components/form/checkbox-field.tsx
```diff
@@ -0,0 +1,121 @@
...
+  TName extends FieldPathByValue<TFieldValues, Array<string>>,
+> = {
+  form: UseFormReturn<TFieldValues, TName>;
+  name: TName;
+  label?: string;
+  description?: React.ReactNode;
+  items: readonly { id: string; label: string }[];
+};
+
+export const CheckboxField = <
```
mmurto, line 51: This should have a more descriptive name, as `CheckboxField` doesn't convey it being a multiselect.

ui/src/components/form/checkbox-field.tsx
```diff
@@ -0,0 +1,121 @@
...
+                    ? 'indeterminate'
+                    : false
+              }
+              onCheckedChange={(checked) => {
+                const enabledItems = checked
+                  ? items.map((item) => item.id)
+                  : [];
+                form.setValue(
+                  name,
+                  enabledItems as FieldPathValue<TFieldValues, TName>
```
mmurto, line 85: This should have a code comment on why `as` can be used here.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/-components/advisor-fields.tsx
```diff
@@ -0,0 +1,98 @@
...
+              </FormItem>
+            )}
+          />
+          <CheckboxField
+            form={form}
+            name='jobConfigs.advisor.advisors'
+            label='Enabled advisors'
+            description={
+              <>
+                Select the <strong>advisors</strong> enabled for this ORT Run.
```
mmurto, line 89: Is this supposed to be `strong`?

ui/src/components/form/checkbox-field.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+type CheckboxFieldProps<
+  TFieldValues extends FieldValues,
+  TName extends FieldPathByValue<TFieldValues, Array<string>>,
+> = {
+  form: UseFormReturn<TFieldValues, TName>;
+  name: TName;
+  label?: string;
+  description?: React.ReactNode;
+  items: readonly { id: string; label: string }[];
```
mmurto, line 48: `items` doesn't really tell what it is, maybe `options` would be better?

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/create-run.tsx
```diff
@@ -99,6 +95,8 @@ const formSchema = z.object({
   jobConfigContext: z.string().optional(),
 });
 
+export type FormValues = ReturnType<typeof formSchema.parse>;
```
mmurto, line 98: As this is exported, it should have a descriptive name like `CreateRunFormValues`.

mmurto (changes requested):

ui/src/components/form/multi-select-field.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    : false
+              }
+              onCheckedChange={(checked) => {
+                const enabledItems = checked
+                  ? options.map((option) => option.id)
+                  : [];
+                form.setValue(
+                  name,
+                  // Here, we use the `FieldPathValue` type to ensure that the value is an array of strings.
+                  enabledItems as FieldPathValue<TFieldValues, TName>
```
mmurto, line 90: Maybe something like:

TypeScript doesn't get this, but `TName extends FieldPathByValue<TFieldValues, Array<string>>`, so the field behind TName is always an `Array<string>` and `options.map((option) => option.id)` is also `Array<string>`, so this type cast is safe.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/-components/advisor-fields.tsx
```diff
@@ -0,0 +1,94 @@
...
+                  />
+                </FormControl>
+              </FormItem>
+            )}
+          />
+          <MultiSelectField
+            form={form}
+            name='jobConfigs.advisor.advisors'
+            label='Enabled advisors'
+            description={<>Select the advisors enabled for this ORT Run.</>}
```
mmurto, line 87: Just an info-nit, string can be directly passed as `ReactNode`, so no need to wrap them in Fragment.

mmurto (changes requested):

ui/src/components/form/multi-select-field.tsx
Etsija: BTW is there any benefit for using backticks in code comments, or should they just be plain text?

mmurto: To continue, `as` doesn't really ensure anything, but it rather tells TS compiler that whatever it thinks is wrong, here's the correct type. It doesn't do anything to the value, so if you do `const niceNumber = "I'm a number" as number`, you'll get autocompletions and type checking as if it really was a number, but at runtime it's still the same string and can cause bugs.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/-components/advisor-fields.tsx
Etsija: Yeah I know, but I would like to keep it as it is, to hint the user of the component that this is a ReactNode, so descriptions can include JSX.

ui/src/components/form/multi-select-field.tsx
mmurto: > BTW is there any benefit for using backticks in code comments, or should they just be plain text?

Some editors may display them differently, if you put them in docstrings like:

```typescript
/**
 * This is code: `code`
 */
function foo()
```

For regular inline comments probably no difference, just stylistic choice and communicates that it is code.

mmurto (approved):

PR #652, opened 2024-07-09
Gradle build improvements

sschuberth:

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -57,12 +57,6 @@ if (version == Project.DEFAULT_VERSION) {
 logger.lifecycle("Building ORT Server version $version.")
 
 allprojects {
-    buildscript {
-        repositories {
-            mavenCentral()
-        }
-    }
-
     apply(plugin = rootProject.libs.plugins.mavenPublish.get().pluginId)
 
     repositories {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -66,8 +66,8 @@ allprojects {
             forRepository {
                 maven("https://jitpack.io")
             }
+
             filter {
-                includeGroup("com.github.oss-review-toolkit.ort")
                 includeModule("com.github.Ricky12Awesome", "json-schema-serialization")
             }
         }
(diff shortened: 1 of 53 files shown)


build.gradle.kts
```diff
@@ -44,7 +44,15 @@ semver {
...
-version = semver.semVersion
+// Only override a default version (which usually is "unspecified"), but not a custom version.
+if (version == Project.DEFAULT_VERSION) {
+    version = semver.semVersion.takeIf { it.isPreRelease }
+        // To get rid of a build part's "+" prefix because Docker tags do not support it, use only the original "build"
+        // part as the "pre-release" part.
+        ?.toString()?.replace("${semver.defaultPreRelease}+", "")
+            // Fall back to a plain version without pre-release or build parts.
+        ?: semver.version
+}
```
mnonnenmacher, line 55: I find it difficult to understand the impact of this change. Would it affect the Docker tags or the versions of the Maven packages? What exactly changes compared to before?

mnonnenmacher (changes requested):

build.gradle.kts
mnonnenmacher: Thanks, but it's still not totally clear to me what the difference between before and after is for a pre release version. I think what would help me would be an example.
Actually, the reason I did not copy this code from ORT was that I was not sure if doing the fix here for Docker could negatively impact the versions of any other release artifacts.

If this is changed, probably this line from the Docker build workflow needs to be adapted:
https://github.com/eclipse-apoapsis/ort-server/blob/a72980ee6d6294a27edc9345991a216e68f9ce78/.github/workflows/docker-build.yml#L128

Commit message: Typo in "chnage"

mnonnenmacher (changes requested):

build.gradle.kts
sschuberth: > If this is changed, probably this line from the Docker build workflow needs to be adapted:

I believe it does not *need* to be adjusted, but the final ` | sed 's/+/-/` could be dropped. Then the code would be equals to ORT's.

> Commit message: Typo in "chnage"

Fixed.

mnonnenmacher (approved):

build.gradle.kts
sschuberth: > I think what would help me would be an example.

The code changes a pre-release version that would usually read like `26.0.0-SNAPSHOT+010.sha.72d9291` to read `26.0.0-010.sha.72d9291` instead.

PR #661, opened 2024-07-11
Add Vitest as a test framework for UI code

sschuberth:


ui/vite.config.ts
```diff
@@ -31,4 +31,7 @@ export default defineConfig({
       '@': path.resolve(__dirname, './src'),
     },
   },
+  test: {
+    includeSource: ['src/**/*.{js,ts}'],
```
sschuberth, line 35: Should I drop "js" from here? Should "tsx" be added instead?

mmurto: Can be dropped, there probably should not be js files with tests. IMO tsx files should not include tests, so add it later if actually needed.

mmurto (approved):

PR #662, opened 2024-07-11
ui: Add support for the Notifier

Etsija:
This PR adds support for the Notifier job, by adding an own component for it to the ORT run creation form. The job is disabled by default.

Information about the Notifier job should also be seen in the run results, similar to the other jobs included in the run.  

Note:
- job information is so far only a JSON structure. Results showing will be fine-tuned in a later PR
- as I've never configured or used Notifier, the PR is not tested with an ORT run
- to me it is unclear what Notifier configuration parameters are needed, so I have made all configuration parameters mandatory. This means that if the user enables Notifier, then all configuration parameters need to have sensible values

For the above reasons, I'm expecting some conversation and guidance to this PR from the interested persons.

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/create-run.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/create-run.tsx
@@ -49,6 +49,7 @@ import { useToast } from '@/components/ui/use-toast';
 import { AdvisorFields } from './-components/advisor-fields';
 import { AnalyzerFields } from './-components/analyzer-fields';
 import { EvaluatorFields } from './-components/evaluator-fields';
+import { NotifierFields } from './-components/notifier-fields';
 import { ReporterFields } from './-components/reporter-fields';
 import { ScannerFields } from './-components/scanner-fields';
 import { packageManagers } from './-types';
@@ -88,6 +89,29 @@ const formSchema = z.object({
     reporter: z.object({
       enabled: z.boolean(),
       formats: z.array(z.string()),
+    }),
+    notifier: z.object({
+      enabled: z.boolean(),
+      notifierRules: z.string().optional(),
+      resolutionsFile: z.string().optional(),
+      mail: z.object({
+        recipientAddresses: z.array(z.object({ email: z.string() })),
+        mailServerConfiguration: z.object({
+          hostName: z.string(),
+          port: z.coerce.number().int(),
+          username: z.string(),
+          password: z.string(),
+          useSsl: z.boolean(),
+          fromAddress: z.string(),
+        }),
+      }),
+      jira: z.object({
+        jiraRestClientConfiguration: z.object({
+          serverUrl: z.string(),
+          username: z.string(),
+          password: z.string(),
+        }),
+      }),
     }),
     parameters: z.array(keyValueSchema).optional(),
   }),
@@ -154,6 +178,27 @@ const CreateRunPage = () => {
       reporter: {
         enabled: true,
         formats: ['ortresult', 'WebApp'],
+      },
+      notifier: {
+        enabled: false,
+        mail: {
+          recipientAddresses: [{ email: '' }],
+          mailServerConfiguration: {
+            hostName: 'localhost',
+            port: 587,
+            username: '',
+            password: '',
+            useSsl: true,
+            fromAddress: '',
+          },
+        },
+        jira: {
+          jiraRestClientConfiguration: {
+            serverUrl: '',
+            username: '',
+            password: '',
+          },
+        },
       },
     },
     jobConfigContext: 'main',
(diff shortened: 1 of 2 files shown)


ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/create-run.tsx
```diff
@@ -223,6 +271,72 @@ const CreateRunPage = () => {
...
+              ortRun.jobConfigs.notifier !== undefined &&
+              ortRun.jobConfigs.notifier !== null,
+            notifierRules:
+              ortRun.jobConfigs.notifier?.notifierRules || undefined,
+            resolutionsFile:
+              ortRun.jobConfigs.notifier?.resolutionsFile || undefined,
+            mail: {
+              //recipientAddresses:
+              //  ortRun.jobConfigs.notifier?.mail?.recipientAddresses ||
+              //  baseDefaults.jobConfigs.notifier.mail.recipientAddresses,
```
mmurto, line 285: This is probably leftover from something?

mmurto (changes requested):

mmurto:
@mnonnenmacher I guess there currently is no way to configure Notifier's mail server and jira somewhere else, but the details need to be sent here? For mail server it probably makes sense to configure one per ORT Server instance, and for Jira they should probably be configurable at organization level, maybe overrideable for products and repos. What do you think?

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/create-run.tsx
```diff
@@ -89,6 +90,32 @@ const formSchema = z.object({
...
+      notifierRules: z.string().optional(),
+      resolutionsFile: z.string().optional(),
+      mail: z.object({
+        recipientAddresses: z.array(z.object({ email: z.string() })),
+        mailServerConfiguration: z.object({
+          hostName: z.string(),
+          // This primitive type integer) should be coerced to avoid the error
+          // "Expected number but received string" when the form is submitted.
+          // See: https://github.com/react-hook-form/resolvers/issues/73#issuecomment-1647321322
+          port: z.coerce.number().int(),
```
mmurto, line 104: `coerce` is quite basic Zod functionality, so this code comment could maybe be removed.

mmurto (changes requested):

mnonnenmacher:
> @mnonnenmacher I guess there currently is no way to configure Notifier's mail server and jira somewhere else, but the details need to be sent here? 

Correct, but you can also configure it in the validation script, then you don't have to provide the values for each run.

> For mail server it probably makes sense to configure one per ORT Server instance, and for Jira they should probably be configurable at organization level, maybe overrideable for products and repos. What do you think?

We integrated the ORT notifier because we needed a quick solution to keep our notifications running. But on the long term I would prefer to make a new event based notification system for the server and I don't think that the Kotlin script based approach from the ORT CLI works well here.

mnonnenmacher:
>     * to me it is unclear what Notifier configuration parameters are needed, so I have made all configuration parameters mandatory. This means that if the user enables Notifier, then all configuration parameters need to have sensible values

To match the ORT data model, `mail` and `jira` should be optional so that you could configure only one of them.

Etsija:
> > ```
> > * to me it is unclear what Notifier configuration parameters are needed, so I have made all configuration parameters mandatory. This means that if the user enables Notifier, then all configuration parameters need to have sensible values
> > ```
> 
> To match the ORT data model, `mail` and `jira` should be optional so that you could configure only one of them. However, in our case I believe we always set both options anyway, so the PR should also work for us as is.

I guess it could be merged, and in case any troubles arise, I can do a speedy follow-up/fix PR. I'm saying this just for the fact that I could not verify the functionality myself, as I always do for my PRs.  But I did review the `jobConfig.notifier` control structure, and it seems to be a match with the specs (back-end code + Swagger).

mnonnenmacher:
> > ```
> > * to me it is unclear what Notifier configuration parameters are needed, so I have made all configuration parameters mandatory. This means that if the user enables Notifier, then all configuration parameters need to have sensible values
> > ```
> 
> To match the ORT data model, `mail` and `jira` should be optional so that you could configure only one of them. However, in our case I believe we always set both options anyway, so the PR should also work for us as is.

So I checked our setup, and what we would require is that it is possible to enable the notifier job without providing any configuration, because in our case the validation script then fills in the required values. Is that currently possible?

Etsija:
I did also try to work with `undefined`, to leave empty configs out altogether, but the back-end data model didn't let me do it (or, at least not easily).

mnonnenmacher:
> Providing an empty config results in this, would that work for you?

Yes, I think so, thank you.

mnonnenmacher (approved):

mmurto:
Should we then just leave the mail server and Jira config out? I'd say no-one will be filling them in run-by-run, so leave it up to the validation script for now and replace with a proper config later.

mmurto:
> Should we then just leave the mail server and Jira config out? I'd say no-one will be filling them in run-by-run, so leave it up to the validation script for now and replace with a proper config later. 

@mnonnenmacher @Etsija do you have thoughts?

mnonnenmacher:
> > Should we then just leave the mail server and Jira config out? I'd say no-one will be filling them in run-by-run, so leave it up to the validation script for now and replace with a proper config later.
> 
> @mnonnenmacher @Etsija do you have thoughts?

For testing it would still be useful to be able to fill in the data manually.

Etsija:
I'd appreciate some guidelines for this, as I'd like to move on to other issues tomorrow.

mmurto:
Sure, let's keep it for testing now if you see it being used. Though configuring SMTP server that ORT Server is close in nature to configuring Postgres params, so let's configure it somewhere else later and take it out of it.

PR #663, opened 2024-07-11
Migrate remaining list endpoints to provide the total entity count

mnonnenmacher:

### DIFF
--- a/dao/src/main/kotlin/utils/Extensions.kt
+++ b/dao/src/main/kotlin/utils/Extensions.kt
@@ -25,10 +25,14 @@ import kotlinx.datetime.minus
 
 import org.eclipse.apoapsis.ortserver.dao.QueryParametersException
 import org.eclipse.apoapsis.ortserver.model.util.ListQueryParameters
+import org.eclipse.apoapsis.ortserver.model.util.ListQueryResult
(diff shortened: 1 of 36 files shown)


Etsija:
LGTM, but I'd like @lamppu to also review, as I think this involves you.

lamppu:
> LGTM, but I'd like @lamppu to also review, as I think this involves you.

LGTM2

dao/src/main/kotlin/utils/Extensions.kt
```diff
@@ -37,6 +41,22 @@ import org.jetbrains.exposed.sql.SortOrder
  */
 fun Instant.toDatabasePrecision() = minus(nanosecondsOfSecond, DateTimeUnit.NANOSECOND)
 
+/**
+ * Run the provided [query] with the given [parameters] to create a [ListQueryResult]. The entities are mapped to the
+ * corresponding model objects using the provided [entityMapper].
+ */
+internal fun <T : LongEntity, E> SortableEntityClass<T>.listQuery(
```
oheger-bosch, line 48: Nit: I find the names of the type parameters a bit confusing: `T` is used for an entity, and `E` for a model object. Would it be better to name them `E` for the entity and `M` for the model object?

PR #668, opened 2024-07-12
ui: Show the creation time in the ort runs table and run details page

### LINKED ISSUE #656: Show the creation time in the ORT runs table
Currently, the ORT runs table only shows the date when a run was created. Also show the time, because the date alone is not sufficient information when many runs were started on the same day.

image

Etsija:
Use browser language to decide on the exact format. Examples:

English:
Capture_English

Finnish:
Capture_Finnish

Run details, English:
Capture_details_English

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/index.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/index.tsx
@@ -81,6 +81,7 @@ const RepoComponent = () => {
   const params = Route.useParams();
   const navigate = useNavigate();
   const { toast } = useToast();
+  const locale = navigator.language;
 
   const [{ data: repo }, { data: runs }] = useSuspenseQueries({
     queries: [
@@ -249,7 +250,7 @@ const RepoComponent = () => {
                       </div>
                     </TableCell>
                     <TableCell>
-                      {new Date(run.createdAt).toLocaleString().split(',')[0]}
+                      {new Date(run.createdAt).toLocaleString(locale)}
                     </TableCell>
                     <TableCell>
                       <Badge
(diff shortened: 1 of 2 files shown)


sschuberth:
In the run details, we separate date and time with a comma:

image

For consistency and readability I believe we should do so here, too.

Etsija:
> In the run details, we separate date and time with a comma:
> 
> image
> 
> For consistency and readability I believe we should do so here, too.

We do! Please look at the run list, English example. The code to produce the date-time fields is exactly the same.

sschuberth:
> Resolves #656.

This should go to the message of the commit that concludes the resolution.

sschuberth:
> We do! Please look at the run list, English example. The code to produce the date-time fields is exactly the same.

Oh, sorry. So using a comma or not is also defined by the locale? Weird!

Etsija:
> > We do! Please look at the run list, English example. The code to produce the date-time fields is exactly the same.
> 
> Oh, sorry. So using a comma or not is also defined by the locale? Weird!

Not really, as each country/language uses their own specific way to represent date and time. I could've chosen Hindi as my example as well, I'm sure it would look very different 😄

sschuberth (approved):

sschuberth:
> Not really, as each country/language uses their own specific way to represent date and time. I could've chosen Hindi as my example as well, I'm sure it would look very different 😄

Ok, let me rephrase 😉 : Weird that Finns don't separate date and time by a comma, esp. as each date and time is separated by dots, making it hard to distinguish visually.

mmurto (approved):

mmurto:
> > Not really, as each country/language uses their own specific way to represent date and time. I could've chosen Hindi as my example as well, I'm sure it would look very different 😄
> 
> Ok, let me rephrase 😉 : Weird that Finns don't separate date and time by a comma, esp. as each date and time is separated by dots, making it hard to distinguish visually.

There's a space in it, IMO it's the best separator. :D

Etsija:
> > > Not really, as each country/language uses their own specific way to represent date and time. I could've chosen Hindi as my example as well, I'm sure it would look very different 😄
> > 
> > 
> > Ok, let me rephrase 😉 : Weird that Finns don't separate date and time by a comma, esp. as each date and time is separated by dots, making it hard to distinguish visually.
> 
> There's a space in it, IMO it's the best separator. :D

Oh don't get me started on the greatness of a DOUBLE SPACE after each sentence! 😆

sschuberth:
> There's a space in it, IMO it's the best separator. :D

At least you did not mention tabs... 👿

PR #682, opened 2024-07-15
fix(ui): Wrong base default value given in Notifier config

Etsija:
This is a critical fix for the UI, so a quick review would be very appreciated.

Please see the commit message for details.

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/create-run.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/create-run.tsx
@@ -188,7 +188,7 @@ const CreateRunPage = () => {
         notifierRules: '',
         resolutionsFile: '',
         mail: {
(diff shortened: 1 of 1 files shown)


ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/create-run.tsx
```diff
@@ -188,7 +188,7 @@ const CreateRunPage = () => {
         notifierRules: '',
         resolutionsFile: '',
         mail: {
-          recipientAddresses: [{}],
+          recipientAddresses: [{ email: '' }],
```
mmurto, line 191: Unlike mentioned in the commit message, the original commit is not empty string array, but empty object array with empty objects. Would this rather work with empty array like `[]`, rather than initializing (and always submitting?) a notifier request to `email: ''`?

mmurto (changes requested):

mmurto (approved):

PR #687, opened 2024-07-15
Upgrade ktor-swagger-ui

mnonnenmacher:
Upgrade ktor-swagger-ui to v3. As the new version now correctly marks properties as nullable, also do some necessary API refactorings. Please see the commit messages for details.

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/OptionalValue.kt
+++ b/api/v1/model/src/commonMain/kotlin/OptionalValue.kt
@@ -25,6 +25,8 @@ import kotlinx.serialization.descriptors.SerialDescriptor
 import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
 
+import org.eclipse.apoapsis.ortserver.api.v1.model.OptionalValue.Present
+
 /**
  * A property type that can be used in PATCH requests. It represents two different states:
  * * [OptionalValue.Present]
@@ -62,15 +64,6 @@ sealed interface OptionalValue<out T> {
     object Absent : OptionalValue<Nothing>
 
     /**
-     * Return the [value][Present.value] if this [OptionalValue] is [Present], otherwise throw an
-     * [IllegalArgumentException].
-     */
-    val valueOrThrow: T get() {
-        require(this is Present)
-        return value
-    }
-
-    /**
      * Execute [function] if this value is [Present].
      */
     fun ifPresent(function: (T) -> Unit) {
@@ -95,6 +88,15 @@ sealed interface OptionalValue<out T> {
             is Present -> Present(transform(value))
             else -> Absent
         }
+}
+
+/**
+ * Return the [value][Present.value] if this [OptionalValue] is [Present], otherwise throw an
+ * [IllegalArgumentException].
+ */
+val <T> OptionalValue<T>.valueOrThrow: T get() {
+    require(this is Present)
+    return value
 }
 
 class OptionalValueSerializer<T>(private val valueSerializer: KSerializer<T>) : KSerializer<OptionalValue<T>> {
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -81,6 +81,7 @@ dependencies {
     implementation(libs.ktorValidation)
     implementation(libs.micrometerRegistryGraphite)
     implementation(libs.ortUtils)
+    implementation(libs.bundles.schemaKenerator)
 
     runtimeOnly(projects.config.secretFile)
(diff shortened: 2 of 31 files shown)


mnonnenmacher:
@Etsija Please have a look at the last commit which contains some small changes to the UI code.

api/v1/model/src/commonMain/kotlin/OptionalValue.kt
```diff
@@ -97,6 +90,15 @@ sealed interface OptionalValue<out T> {
         }
 }
 
+/**
+ * Return the [value][Present.value] if this [OptionalValue] is [Present], otherwise throw an
+ * [IllegalArgumentException].
+ */
+val <T> OptionalValue<T>.valueOrThrow: T get() {
```
sschuberth, line 97: Would the [`@Transient` annotation](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#transient-properties) maybe work as well?

model/src/commonMain/kotlin/JobConfigurations.kt
```diff
@@ -80,7 +80,7 @@ data class AnalyzerJobConfiguration(
     /**
      * The list of package curation providers to use.
      */
-    val packageCurationProviders: List<ProviderPluginConfiguration> = emptyList(),
+    val packageCurationProviders: List<ProviderPluginConfiguration>? = null,
```
oheger-bosch, line 83: I find it strange that we need to add quirks to our data model, just to make the Swagger generation work. If this is really required, should this change then be limited to the API module? There is no need for the property in `model` to be nullable, too.

model/src/commonMain/kotlin/JobConfigurations.kt
```diff
@@ -191,7 +191,7 @@ data class EvaluatorJobConfiguration(
     /**
      * The list of package configuration providers to use.
      */
-    val packageConfigurationProviders: List<ProviderPluginConfiguration> = emptyList(),
+    val packageConfigurationProviders: List<ProviderPluginConfiguration>? = null,
```
oheger-bosch, line 194: Analogously to my other comment.

core/build.gradle.kts
```diff
@@ -81,6 +81,7 @@ dependencies {
     implementation(libs.ktorValidation)
```
oheger-bosch, line 81: Commit message: Is this correct "Nullable properties are [...] marked as required"? Should this be "Non-nullable properties" or "marked as optional"?

model/src/commonMain/kotlin/JobConfigurations.kt
mnonnenmacher: I thought about that too, but I wanted to be consistent, and so far all properties have the same types and default values in both models. We could change it for lists, but then we should change it for all of them.

oheger-bosch: It was the original purpose of this mapping to separate the internal model from the API. IMHO, there can well be differences if this helps us. Avoiding unnecessary null checks would fall into this category.

Etsija:
> @Etsija Please have a look at the last commit which contains some small changes to the UI code.

LGTM!

model/src/commonMain/kotlin/JobConfigurations.kt
mnonnenmacher: I took a look into making more properties non-nullable in the model module, but that lead me to lot of questions how to handle some inconsistencies we already have in the job configurations. I don't want to go down this rabbit hole right now, so I'll keep the model properties non-nullable in this PR, even if it makes the current state of the mapping even more inconsistent. At least I could make some cleanups during the investigation for which I will make a separate PR after this is merged.

mnonnenmacher: @oheger-bosch Done.

api/v1/model/src/commonMain/kotlin/OptionalValue.kt
mnonnenmacher: The new JSON schema generator uses reflection by default:
https://github.com/SMILEY4/schema-kenerator/tree/develop/schema-kenerator-reflection

We could try switching to the kotlinx.serialization implementation, but I don't want to do it in this PR:
https://github.com/SMILEY4/schema-kenerator/tree/develop/schema-kenerator-serialization

sschuberth: > The new JSON schema generator uses reflection by default:

Ouch 🙄

PR #697, opened 2024-07-17
fix(ui): Guard against possible undefined context in header

Etsija:
This PR fixes an undefined context error when refreshing the browser on index pages:

Capture

### DIFF
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -128,7 +128,7 @@ export const Header = () => {
         </Sheet>
         <Breadcrumb>
           <BreadcrumbList>
-            {organizationMatch && (
+            {organizationMatch?.context && (
               <BreadcrumbItem>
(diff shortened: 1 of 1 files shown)


ui/src/components/header.tsx
```diff
@@ -128,7 +128,7 @@ export const Header = () => {
         </Sheet>
         <Breadcrumb>
           <BreadcrumbList>
-            {organizationMatch && (
+            {organizationMatch && organizationMatch.context && (
```
sschuberth, line 131: Could we instead use [chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html), like just `(organizationMatch?.context && (`?

Etsija: We obviously can. 😄

sschuberth (approved):

PR #699, opened 2024-07-17
ui: Implement pagination for lists

lamppu:
Please see individual commit messages for details.

As discussed in the frontend alignment meeting, the lists are rendered as before.

Näyttökuva (94)

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -52,6 +52,7 @@ const router = createRouter({
     },
     auth: undefined!,
   },
+  defaultPreload: 'intent',
 });
 
 // Register the router instance for type safety
--- a/ui/src/components/data-table/data-table-pagination.tsx
+++ b/ui/src/components/data-table/data-table-pagination.tsx
@@ -1,0 +1,167 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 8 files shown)


Etsija (commented):
Checked the code and tested the implementation. The only nit I have is, could the pagination component be made to appear only when there are actual items in the list to paginate?
Capture

sschuberth:
Also some visual glitch that jumps to my eyes: The spinners in the "Items per page" and "Page" fields look different. Can they be aligned? (Preferably to the visuals of the former.)

*Edit:* Ah, or is this because the former is not actually a spinner, but a drop-down with preset number (I have not checked this, sorry).

If they cannot be aligned I'd also be fine with having no spinner at all in the "Page" field.

Oh, and more more thing, at least in @Etsija's screenshot the total page count is wrong, it should be 1 instead of 0 even for 0 entries.

lamppu:
Yeah the items per page is a select menu like this

Näyttökuva (95)

lamppu:
The number input spinner styling seems to be browser specific, I'll look into how to change it/hide it

lamppu:
> Oh, and more more thing, at least in @Etsija's screenshot the total page count is wrong, it should be 1 instead of 0 even for 0 entries.

Yeah, that was because the page count is counted like this: `pageCount: Math.ceil(data.pagination.totalCount / pageSize),` and the totalCount was zero. However if the toolbar will be hidden then, there's probably no need to change that..

sschuberth:
> The number input spinner styling seems to be browser specific, I'll look into how to change it/hide it

Let's not spend too much time for that now. The root cause was basically my misunderstanding of popup vs spinner; so it actually makes sense that they look different (though IMO it's a bit odd to have those double arrows for the popup).

sschuberth:
> However if the toolbar will be hidden then, there's probably no need to change that..

Agreed.

lamppu:
> IMO it's a bit odd to have those double arrows for the popup

Btw, the double arrows for the popup are easier to replace as it is an icon used in the [shadcn select component](https://github.com/eclipse-apoapsis/ort-server/blob/ffa51dc32e04a8d7a888a8929df149b4a2ab45d6/ui/src/components/ui/select.tsx#L41), so it could be replaced with some other icon.

PR #700, opened 2024-07-17
feat(orchestrator): Add an `Issue` to the `OrtRun` in case of a failure

bs-ondem:
The orchestrator adds an `Issue` to the current `OrtRun` whenever it receives an error message from a worker. This helps to track any failures that occur.

### DIFF
--- a/orchestrator/src/main/kotlin/Orchestrator.kt
+++ b/orchestrator/src/main/kotlin/Orchestrator.kt
@@ -49,6 +49,7 @@ import org.eclipse.apoapsis.ortserver.model.orchestrator.WorkerError
 import org.eclipse.apoapsis.ortserver.model.orchestrator.WorkerMessage
 import org.eclipse.apoapsis.ortserver.model.repositories.OrtRunRepository
 import org.eclipse.apoapsis.ortserver.model.repositories.RepositoryRepository
+import org.eclipse.apoapsis.ortserver.model.runs.Issue
 import org.eclipse.apoapsis.ortserver.model.util.OptionalValue
 import org.eclipse.apoapsis.ortserver.model.util.asPresent
 import org.eclipse.apoapsis.ortserver.transport.AdvisorEndpoint
@@ -127,7 +128,8 @@ class Orchestrator(
         db.blockingQueryCatching(transactionIsolation = isolationLevel) {
             ortRunRepository.update(
                 id = configWorkerError.ortRunId,
-                status = OrtRunStatus.FAILED.asPresent()
+                status = OrtRunStatus.FAILED.asPresent(),
+                issues = listOf(ConfigEndpoint.createErrorIssue()).asPresent()
             )
         }.onFailure {
             log.warn("Failed to handle 'ConfigWorkerError' message.", it)
@@ -258,19 +260,21 @@ class Orchestrator(
      * Handle the given [error] message with the given [header] from a worker of the given [endpoint].
      */
     private fun handleWorkerError(endpoint: Endpoint<*>, header: MessageHeader, error: WorkerMessage) {
-        handleCompletedJob(endpoint, header, error, JobStatus.FAILED)
+        handleCompletedJob(endpoint, header, error, JobStatus.FAILED, listOf(endpoint.createErrorIssue()))
     }
 
     /**
      * Handle a [message] with the given [header] about a worker job for the given [endpoint] that completed in the
-     * given [status]. This is the central scheduling function. It determines the current job execution state of the
-     * affected ORT run, schedules the next job(s) if possible, or decides that the ORT run is now finished.
+     * given [status] with optional [issues]. This is the central scheduling function. It determines the current job
+     * execution state of the affected ORT run, schedules the next job(s) if possible, or decides that the ORT run is
+     * now finished.
      */
     private fun handleCompletedJob(
         endpoint: Endpoint<*>,
         header: MessageHeader,
         message: WorkerMessage,
-        status: JobStatus
+        status: JobStatus,
+        issues: List<Issue> = emptyList()
     ) {
         log.info(
             "Job {} for endpoint '{}' completed in status '{}'.",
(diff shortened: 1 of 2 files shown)


orchestrator/src/main/kotlin/Orchestrator.kt
```diff
@@ -49,6 +49,7 @@ import org.eclipse.apoapsis.ortserver.model.orchestrator.WorkerError
 import org.eclipse.apoapsis.ortserver.model.orchestrator.WorkerMessage
 import org.eclipse.apoapsis.ortserver.model.repositories.OrtRunRepository
 import org.eclipse.apoapsis.ortserver.model.repositories.RepositoryRepository
+import org.eclipse.apoapsis.ortserver.model.runs.Issue
```
sschuberth, line 52: > The orchestrator adds an `Issue` to the current `OrtRun`

This sounds a bit as if you'd describe the current state  (before your change), and not the behavior that your change introduces. Maybe better say "Make the orchestrator add an `Issue`...".

orchestrator/src/main/kotlin/Orchestrator.kt
```diff
@@ -270,7 +272,8 @@ class Orchestrator(
         endpoint: Endpoint<*>,
         header: MessageHeader,
         message: WorkerMessage,
-        status: JobStatus
+        status: JobStatus,
+        issues: List<Issue>? = null
```
sschuberth, line 276: As I guess we'd also not want to add an empty list of issues, I wonder whether we could instead have `issues: List<Issue> = emptyList()` here, and below do

    if (issues.isNotEmpty()) ortRunRepository.update(job.ortRunId, issues = issues.asPresent())

oheger-bosch, line 276: Please update the doc comment for the method for the new `issues' parameter.

orchestrator/src/main/kotlin/Orchestrator.kt
```diff
@@ -404,3 +408,13 @@ class Orchestrator(
...
  */
 typealias CreatedJobs = List<JobScheduleFunc>
+
+/**
+ * Create an [Issue] object representing an error that occurred in any [Endpoint].
+ */
+fun <T : Any> Endpoint<T>.createErrorIssue(): Issue = Issue(
+    timestamp = Clock.System.now(),
+    source = configPrefix,
+    message = "An unexpected failure occurred in the $configPrefix worker.",
```
oheger-bosch, line 418: Maybe the message should be more drastic, e.g. something like "The $configPrefix worker failed due to an unexpected error." to make clear that this worker failed completely.

orchestrator/src/test/kotlin/OrchestratorTest.kt
```diff
@@ -380,8 +381,12 @@ class OrchestratorTest : WordSpec() {
 
                 verify(exactly = 1) {
                     ortRunRepository.update(
-                        id = withArg { it shouldBe advisorJob.ortRunId },
-                        status = withArg { it.verifyOptionalValue(OrtRunStatus.FAILED) }
+                        id = withArg { it shouldBe configWorkerError.ortRunId },
+                        status = withArg { it.verifyOptionalValue(OrtRunStatus.FAILED) },
+                        issues = withArg {
+                            val issues = it.shouldBeTypeOf<OptionalValue.Present<List<Issue>>>().value
+                            issues.size shouldBe 1
```
oheger-bosch, line 388: It would be good to check some properties of the issue itself, maybe the severity and the source. To not repeat this over and over, would it be possible to extract this logic to a helper function?

orchestrator/src/main/kotlin/Orchestrator.kt
bs-ondem: It shouldn't be an issue to add empty lists since we always add empty issues if there are none (see [here](https://github.com/eclipse-apoapsis/ort-server/blob/main/dao/src/main/kotlin/repositories/DaoOrtRunRepository.kt#L77) and [here](https://github.com/eclipse-apoapsis/ort-server/blob/main/dao/src/main/kotlin/Database.kt#L212-L213)). So, it's a matter of preference whether using `issues: List<Issue>? = null` or `List<Issue> = emptyList()`.

orchestrator/src/test/kotlin/OrchestratorTest.kt
bs-ondem: Hmm, the helper function for verifying these properties is specific to worker error issues and I couldn't make it completely generic. What about something like this, tailored to worker error issues:

```
fun verifyIssues(
    issues: OptionalValue<Collection<Issue>>,
    validate: Collection<Issue>.() -> Unit
) {
    val issuesList = issues.shouldBeTypeOf<OptionalValue.Present<Collection<Issue>>>().value
    issuesList.validate()
}

fun Collection<Issue>.verifyWorkerErrorIssues(severity: String, source: String) {
    shouldBeSingleton { issue ->
        issue.severity shouldBe severity
        issue.source shouldBe source
    }
}

```

oheger-bosch: Yes, that's okay. I did not mean a fully generic verification function for issues. Just one for this specific test class that covers the duplicate lines and checks some more properties.

orchestrator/src/test/kotlin/OrchestratorTest.kt
```diff
@@ -650,7 +655,9 @@ class OrchestratorTest : WordSpec() {
                     every { complete(analyzerJob.id, any(), any()) } returns analyzerJob
                 }
                 val repositoryRepository = mockk<RepositoryRepository>()
-                val ortRunRepository = createOrtRunRepository()
+                val ortRunRepository = createOrtRunRepository {
+                        every { update(ortRun.id, issues = any()) } returns mockk()
```
sschuberth, line 659: This line looks like indented too much.

oheger-bosch (approved):

PR #706, opened 2024-07-18
ui: First skeleton version of new ORT run details page

Etsija:
This PR adds a temporary route `.../runs2/` that will be used for pages and (upcoming) subpages for viewing ORT run details results. The PR touches existing implementation very minimally:
- root layout has one added Tailwind class
- header has an additional (temporary) breadcrumb item, to make the breadcrumb show identical information as for the current run details page

As there already are ongoing discussions about the philosophy for showing the ORT run results, the purpose of this PR is to be more of a basis for discussion, rather than a polished and finalized way to show the results. (Think of it more like a Figma skeleton of a part of the UI.)

No links in the current UI lead directly to the new page, so should the users want to see it, it can be done by replacing `.../runs/...` in the URL with `.../runs2/...`.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -29,6 +29,7 @@
     "@radix-ui/react-separator": "^1.0.3",
     "@radix-ui/react-slot": "^1.0.2",
     "@radix-ui/react-switch": "^1.0.3",
+    "@radix-ui/react-tabs": "^1.1.0",
     "@radix-ui/react-toast": "^1.1.5",
     "@radix-ui/react-tooltip": "^1.0.7",
     "@tanstack/react-query": "^5.29.2",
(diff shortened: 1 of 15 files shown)


lamppu:
Could the sidebar component maybe go to the -components folder instead for now? I'm currently making a generic sidebar component, although of course I could just refactor that too

Etsija:
> Could the sidebar component maybe go to the -components folder instead for now? I'm currently making a generic sidebar component, although of course I could just refactor that too

If at all possible, could be best to reuse your component later on, so sure, I'll move mine, as it is for now specific to this route.

lamppu (approved):

sschuberth (approved):

PR #736, opened 2024-07-25
ui: Implement editing and deleting secrets

lamppu:

Näyttökuva (103)
Näyttökuva (104)
Näyttökuva (105)

### DIFF
--- a/ui/src/components/delete-dialog.tsx
+++ b/ui/src/components/delete-dialog.tsx
@@ -1,0 +1,94 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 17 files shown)


Etsija:
Design question: editing is handled with dialogs, right? What's the philosophy behind choosing those instead of edit routes as is done for organizations, products and repositories?

lamppu:
> > Design question: editing is handled with dialogs, right? What's the philosophy behind choosing those instead of edit routes as is done for organizations, products and repositories?
> 
> Well, it felt maybe a bit excessive _in this case_ to create a `secrets/$secretName` route for that reason only, and to redirect to a separate page. I felt it gave a more cohesive feel for both the actions for the buttons in the end of the row to result in a similar flow and view (although of course you could argue that it's more cohesive for all edits to behave the same way throughout the application). I also thought this might be ok, as the table itself already breaks the pattern, as the items in it don't have their own page where to perform the edit and delete actions. But yeah, feel free to disagree, and we can discuss it more when you get back if needed.

But actually now that I'm working on the infrastructure services, I realize it would probably be best to create an edit route instead, because the infrastructure service refers to a couple of secrets, and it would be good to be able to link to an edit page for those secrets straight from the infrastructure service details.

Etsija (approved):

PR #753, opened 2024-07-30
feat(api): Add endpoints for user management

wkl3nk:
Add endpoints to add or remove users to groups on the level of an _Organization_ to assign the respective
roles and permissions to them. Available groups are _readers_, _writers_ and _admins_.

### DIFF
--- a/clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
+++ b/clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
@@ -383,6 +383,43 @@ class DefaultKeycloakClient(
         }.getOrElse {
             throw KeycloakClientException("Failed to load client roles for user '${id.value}'.", it)
         }.body()
+
+    override suspend fun addUserToGroup(username: UserName, groupName: GroupName) {
+        val user = getUser(username)
+        val group = getGroup(groupName)
+
+        runCatching {
+            httpClient.put("$apiUrl/users/${user.id.value}/groups/${group.id.value}")
+        }.onFailure {
+            throw KeycloakClientException(
+                "Failed to add user '{$username.value}' to group '{$groupName}'.",
+                it
+            )
+        }
+    }
+
+    override suspend fun removeUserFromGroup(username: UserName, groupName: GroupName) {
+        val user = getUser(username)
+        val group = getGroup(groupName)
+
+        runCatching {
+            httpClient.delete("$apiUrl/users/${user.id.value}/groups/${group.id.value}")
+        }.onFailure {
+            throw KeycloakClientException(
+                "Failed to remove user '${username.value}' from group {'$groupName.value}'.",
+                it
+            )
+        }
+    }
+
+    override suspend fun getGroupMembers(groupName: GroupName): Set<User> {
+        val group = getGroup(groupName)
+        return runCatching {
+            httpClient.get("$apiUrl/groups/${group.id.value}/members")
+        }.getOrElse {
+            throw KeycloakClientException("Failed to get members of group '${groupName.value}'.", it)
+        }.body()
+    }
 }
 
 /**
--- a/core/src/main/kotlin/api/OrganizationsRoute.kt
+++ b/core/src/main/kotlin/api/OrganizationsRoute.kt
@@ -23,6 +23,7 @@ import io.github.smiley4.ktorswaggerui.dsl.routing.delete
 import io.github.smiley4.ktorswaggerui.dsl.routing.get
 import io.github.smiley4.ktorswaggerui.dsl.routing.patch
 import io.github.smiley4.ktorswaggerui.dsl.routing.post
+import io.github.smiley4.ktorswaggerui.dsl.routing.put
 
 import io.ktor.http.HttpStatusCode
 import io.ktor.server.application.call
@@ -37,6 +38,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.CreateInfrastructureService
 import org.eclipse.apoapsis.ortserver.api.v1.model.CreateOrganization
 import org.eclipse.apoapsis.ortserver.api.v1.model.CreateProduct
 import org.eclipse.apoapsis.ortserver.api.v1.model.CreateSecret
+import org.eclipse.apoapsis.ortserver.api.v1.model.IdentifyUser
 import org.eclipse.apoapsis.ortserver.api.v1.model.PagedResponse
 import org.eclipse.apoapsis.ortserver.api.v1.model.SortDirection
 import org.eclipse.apoapsis.ortserver.api.v1.model.SortProperty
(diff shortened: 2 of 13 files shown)


core/src/main/kotlin/api/OrganizationsRoute.kt
```diff
@@ -319,5 +325,65 @@ fun Route.organizations() = route("organizations") {
                 }
             }
         }
+
+        route("groups") {
```
oheger-bosch, line 329: There are different ways to structure these endpoints, and I am not sure which is the best one from a usability PoV. You are approaching it from the groups. One alternative could be having a /users endpoint. In order to add a user with a specific permission, one would send a PUT to this endpoint with a body containing the user ID and the permission. One advantage I would see here is that a GET on users would directly return all relevant information about users.

core/src/main/kotlin/api/OrganizationsRoute.kt
```diff
@@ -319,5 +325,65 @@ fun Route.organizations() = route("organizations") {
...
+                    requirePermission(OrganizationPermission.WRITE)
+
+                    val user = call.receive<User>()
+                    val organizationId = call.requireIdParameter("organizationId")
+                    val groupId = call.requireParameter("groupId")
+
+                    try {
+                        organizationService.addUserToGroup(user.username, organizationId, groupId)
+                        call.respond(HttpStatusCode.NoContent)
+                    } catch (e: ResourceNotFoundException) {
```
oheger-bosch, line 343: The typical pattern to handle exceptions in the REST API is to map them to HTTP status codes using the _StatusPages_ plugin. Refer to `StatusPages.kt`.

core/src/main/kotlin/api/OrganizationsRoute.kt
wkl3nk: I got your point. When I think of the user management aspect from the UI viewpoint on the other hand, in future I might want to have a GET operation on _organizations/{organizationId}/groups/..._ level to get a list of users that have some kind of group membership there. So, maybe in future something like _GET organizations/{organizationId}/groups/readers_ returning a list of users in the _readers_ group on organization level.

core/src/main/kotlin/api/OrganizationsRoute.kt
wkl3nk: That's good. I know something similar in Spring Boot.

core/src/main/kotlin/api/OrganizationsRoute.kt
oheger-bosch: Unfortunately, I am no expert for UX. Probably both approaches have their pros and cons, and there are different use cases that can be handled better with either of them.

It is certainly a valid use case to list all users in a specific group. It might also be useful to search for a specific user in the whole set of users. Would it be perhaps easier for the UI to have one endpoint that returns all information and then do the filtering for different presentations locally?

wkl3nk: I would say, not "search for a specific user in the whole set of users" , but "search for a user in all the users of an Organization", because we have multi-tenancy and there are rare (system admin) cases where all users shall be shown.

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
```diff
@@ -383,6 +383,51 @@ class DefaultKeycloakClient(
         }.getOrElse {
             throw KeycloakClientException("Failed to load client roles for user '${id.value}'.", it)
         }.body()
+
+    override suspend fun addUserToGroup(username: UserName, groupName: GroupName) {
+        // FIXME: Unfortunately, no way to distinguish if there was a technical problem or the user was not found.
+        val user = try { getUser(username) } catch (e: KeycloakClientException) {
+            throw UserNotFoundException("User with username '${username.value}' not found.", e)
+        }
+        val group = getGroup(groupName)
```
oheger-bosch, line 392: What if a non-existing group name is specified, would this throw a Keycloak exception as well? So, does this require some error handling, too?

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
```diff
@@ -383,6 +383,51 @@ class DefaultKeycloakClient(
         }.getOrElse {
             throw KeycloakClientException("Failed to load client roles for user '${id.value}'.", it)
         }.body()
+
+    override suspend fun addUserToGroup(username: UserName, groupName: GroupName) {
+        // FIXME: Unfortunately, no way to distinguish if there was a technical problem or the user was not found.
```
oheger-bosch, line 388: In this class, typically `runCatching` is used for exception handling. For reasons of consistency, it would make sense to use this here as well.

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
```diff
@@ -383,6 +383,51 @@ class DefaultKeycloakClient(
...
+        }.onFailure {
+            throw KeycloakClientException(
+                "Failed to add user '{$username.value}' to group '{$groupName}'.",
+                it
+            )
+        }
+    }
+
+    override suspend fun removeUserFromGroup(username: UserName, groupName: GroupName) {
+        // FIXME: Unfortunately, no way to distinguish if there was a technical problem or the user was not found.
```
oheger-bosch, line 405: Maybe the code for obtaining a user and a group by name could be extracted into helper methods. This will be needed for the other endpoints as well. If the methods returned a `Result`, they could be nicely integrated with further logic that is using `runCatching` (modulo Kotlin's limited supported for monads of course).

services/authorization/src/main/kotlin/DefaultAuthorizationService.kt
```diff
@@ -731,4 +732,22 @@ class DefaultAuthorizationService(
             keycloakClient.removeGroupClientRole(group.id, it)
         }
     }
+
+    override suspend fun addUserToGroup(
+        username: String,
+        organizationId: Long,
+        groupName: String
+    ) {
+        val group = keycloakGroupPrefix + groupName // Allow multiple ORT instances to share the same Keycloak realm
```
oheger-bosch, line 741: Nit: This should probably mean "Allow multiple ORT *Server* instances..."

oheger-bosch (commented):
IIUC, there is not yet a final decision regarding the exact organization of the endpoint (by group or by users). Nevertheless, I think we can go ahead with the current approach and change it later if needed.

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
wkl3nk: It was not a good idea to introduce this UserNotFound exception and try to fix the shortcomings of the AuthorizationService and KeycloakService APIs, because they don't make a difference between business exceptions and technical exceptions. I actually don't want to fix these interfaces within this feature task, so whenever something goes wrong, a KeycloakClientException will be thrown and propagated through the layers.

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
wkl3nk: I did simplify exception handling by letting KeycloakClientExceptions pass through the layers.

oheger-bosch (approved):

oheger-bosch (approved):

PR #719, opened 2024-07-22
ui: Add creating and viewing secrets

lamppu:
Please see individual commits for details.

PR #760, opened 2024-08-02
ui: Use secrets as defaults in ORT run creation form

Etsija:
With #719 and #736 we now have full CRUD support for handling secrets on organization, product and repository levels. This PR adds support for organization, product and repository admins to set up (and possibly enforce) some default values to ORT run creation form.

### DIFF
--- a/ui/src/helpers/defaults-helpers.ts
+++ b/ui/src/helpers/defaults-helpers.ts
@@ -1,0 +1,192 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import {
+  ZodArray,
+  ZodBoolean,
+  ZodEnum,
+  ZodNullable,
+  ZodNumber,
+  ZodObject,
+  ZodOptional,
+  ZodString,
+  ZodType,
+} from 'zod';
+
+type PropertyPathWithType = {
+  path: string;
+  type: string;
+};
+
+/**
+ * Check if a path is excluded based on the provided excluded paths.
+ *
+ * @param path The path to check.
+ * @param excludedPaths The array of paths to be excluded.
+ * @returns True if the path is excluded, otherwise false.
+ */
+const isPathExcluded = (path: string, excludedPaths: string[]): boolean => {
+  return excludedPaths.some(
+    (excludedPath) =>
+      path === excludedPath || path.startsWith(`${excludedPath}.`)
+  );
+};
+
+/**
+ * Get the type of a Zod schema
+ *
+ * @param schema The Zod schema.
+ * @returns The type as a string.
+ */
+const getType = (schema: ZodType): string => {
+  if (schema instanceof ZodString) {
+    return 'string';
+  } else if (schema instanceof ZodNumber) {
+    return 'number';
+  } else if (schema instanceof ZodBoolean) {
+    return 'boolean';
+  } else if (schema instanceof ZodEnum) {
+    return 'enum';
+  } else if (schema instanceof ZodArray) {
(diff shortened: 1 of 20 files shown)


oheger-bosch:
This is indeed not the typical use case for secrets.

What would be your plans for the long run? Some kind of templates for ORT run trigger requests? I guess this would require a completely new endpoint and new structures in the database. (Probably not quite easy ones to store the various configuration options.)

Etsija:
> This is indeed not the typical use case for secrets.
> 
> What would be your plans for the long run? Some kind of templates for ORT run trigger requests? I guess this would require a completely new endpoint and new structures in the database. (Probably not quite easy ones to store the various configuration options.)

I think we are aiming to use templates, yes, but the details are not clear yet, and the feature in this PR was required to (mis)use the secrets to implement some default behaviour soon. As described in the PR and its commits, we are not using values of the secrets at all. Moreover, when trying to specify a default value that is out-of-context with regards to the ORT run creation form, it is silently ignored, so this feature is still safe to use, although, as said, misuses the concept of secrets.

Etsija:
It would also be fairly easy to extend the functionality for the admins that define these "special secrets". One example of such would be for a repository admin to specify a secret
```
Name: jobConfigContext DEFAULT LOCK
Description: specific-branch-for-this-repository
```
in effect enforcing all ORT runs for that repository to use a preset branch of the configuration repository - in effect disabling the `Job configuration context` field in the run creation form.

ui/src/helpers/get-defaults.ts
```diff
@@ -0,0 +1,159 @@
+/*
```
sschuberth, line 1: > When secrets are used for setting default values, the precendence is as
> follows: organization defaults > product defaults > repository defaults.

Typo in "precendence". Also, as discussed internally, the precedence should be inverted so that default values that are "closer" to the "leaf" of the hierarchy, i.e. repository secrets, should have a higher priority that "root" default values.

ui/src/helpers/get-defaults.ts
```diff
@@ -0,0 +1,159 @@
...
+  orgSecrets: PagedResponse_Secret_,
+  prodSecrets: PagedResponse_Secret_,
+  repoSecrets: PagedResponse_Secret_
+): { [key: string]: string } {
+  const defaultValuesMap: { [key: string]: string } = {};
+
+  // Helper function to add secrets to the map
+  const addSecrets = (secrets: PagedResponse_Secret_) => {
+    secrets.data.forEach((secret) => {
+      if (secret.name.endsWith('DEFAULT') && secret.description) {
```
sschuberth, line 48: To recap what @haikoschol proposed during our internal discussion:
- any secret names that have special semantic should start with a "_" as it's a syntactically valid character that hardly any real secret name would start with (in our discussion we were talking about a dash instead, but Scaleway does not seem to allow that as secret names need to match the regex `^[_a-zA-Z0-9]([-_.a-zA-Z0-9]*[_a-zA-Z0-9])?$`)
- instead of using "DEFAULT" as a suffix, we should use it as a prefix following the leading "_"
- maybe we should also introduce an infix that further describes *what for* the default is, like "PARAMETER" or "PARAM"

So as an example, the name of a secret whose value contains the default value for the parameter identified by the `jobConfig.evaluator.ruleSet` JSON path would be named `_DEFAULT_PARAM-jobConfig.evaluator.ruleSet`.

ui/src/helpers/defaults-helpers.ts
```diff
@@ -82,3 +82,61 @@ export const getPropertyPaths = (
   // For primitive types and other unsupported types, return an empty array
   return [];
 };
+
+/**
+ * Encode the property path based on the locked status.
```
sschuberth, line 87: What *is* a "lock status" and what is it required for?

ui/src/helpers/get-defaults.ts
Etsija: Back-end only allows letters, numbers, hyphens and spaces in secret names, no underscores.

Etsija:
@sschuberth I've changed this PR into draft because it is undergoing big changes so maybe not worth while commenting it now.

ui/src/helpers/get-defaults.ts
sschuberth: > no underscores.

Let's double-check on that as I'm not aware of any generic check for underscores in the code. IIRC any restrictions should be secret-provider-specific, and at least Scaleway explicitly allowed underscores when I implemented support for it.

sschuberth: Ah, it's this validation:

https://github.com/eclipse-apoapsis/ort-server/blob/a211137352b46dc3658fa954f18483233db95694/api/v1/model/src/commonMain/kotlin/Secret.kt#L55-L57

ui/src/helpers/defaults-helpers.ts
sschuberth: Note to myself: "Locked" default values are defaults that cannot be overridden, as they will cause the corresponding run form field to become disabled.

Etsija:
This feature is best to handle with the validation script for now, so the PR can be closed.

PR #770, opened 2024-08-05
ui: Create page for downloadable reports of ORT Run

Etsija:
This PR is for the temporary `.../runs2/...` route, which holds the implementation of the new ORT run details pages. It implements a sub-page, which is used for downloading all reports from the run.

Capture


lamppu (approved):

sschuberth:
I'm trusting @lamppu to have checked the actual implementation, I'm just approving based on the screenshot 😓

PR #772, opened 2024-08-06
fix(Analyzer): Change the default user and group ID in Docker file

oheger-bosch:
The Docker build for Analyzer images was failing when it tried to add the non-privileged user and the associated group. The exit code '4' seems to indicate a conflict with an already existing user and group. So, use another user ID to circumvent the conflict.

In our CI the build was failing as well. I could verify that this fix actually solves the problem.


wkl3nk (approved):

MarcelBochtler (approved):

sschuberth:
Wouldn't it make sense to do this for the other Dockerfiles as well for consistency, even if it might not be strictly required?

PR #775, opened 2024-08-06
Use suspend functions in Transport implementations

oheger-bosch:
Change the entry points into ORT Server components to use suspend functions, so that further usage of coroutines does not require additional effort.

### DIFF
--- a/dao/src/testFixtures/kotlin/MockDatabaseModule.kt
+++ b/dao/src/testFixtures/kotlin/MockDatabaseModule.kt
@@ -83,9 +83,10 @@ fun KoinTest.verifyDatabaseModuleIncluded() {
 
 /**
  * Run [block] with a [mocked database module][mockDatabaseModule]. Afterward do cleanup by calling
- * [unmockDatabaseModule].
+ * [unmockDatabaseModule]. Note that this is an *inline* function which makes it possible to invoke *suspend* functions
+ * in [block].
  */
-fun withMockDatabaseModule(block: () -> Unit) {
+inline fun withMockDatabaseModule(block: () -> Unit) {
     mockDatabaseModule()
     try {
         block()
--- a/transport/spi/src/testFixtures/kotlin/Utils.kt
+++ b/transport/spi/src/testFixtures/kotlin/Utils.kt
@@ -28,6 +28,8 @@ import java.util.concurrent.BlockingQueue
 import java.util.concurrent.LinkedBlockingQueue
 import java.util.concurrent.TimeUnit
 
+import kotlinx.coroutines.runBlocking
+
 import org.eclipse.apoapsis.ortserver.config.ConfigManager
 import org.eclipse.apoapsis.ortserver.model.orchestrator.OrchestratorMessage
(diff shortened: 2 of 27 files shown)


transport/spi/src/testFixtures/kotlin/Utils.kt
```diff
@@ -69,7 +71,9 @@ fun startReceiver(configManager: ConfigManager): LinkedBlockingQueue<Message<Orc
     }
 
     Thread {
-        MessageReceiverFactory.createReceiver(OrchestratorEndpoint, configManager, ::handler)
+        runBlocking {
+            MessageReceiverFactory.createReceiver(OrchestratorEndpoint, configManager, ::handler)
+        }
     }.start()
```
sschuberth, line 77: Could this be replaced with `runBlocking(Dispatchers.Default) {}` which implicitly creates a new thread?

dao/src/testFixtures/kotlin/MockDatabaseModule.kt
```diff
@@ -85,7 +85,7 @@ fun KoinTest.verifyDatabaseModuleIncluded() {
  * Run [block] with a [mocked database module][mockDatabaseModule]. Afterward do cleanup by calling
  * [unmockDatabaseModule].
  */
-fun withMockDatabaseModule(block: () -> Unit) {
+inline fun withMockDatabaseModule(block: () -> Unit) {
```
sschuberth, line 88: Maybe it's worth spending a sentence in the commit message on why you decided to inline these instead of making them `suspend`, too?

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
```diff
@@ -52,7 +52,7 @@ class SqsMessageReceiverFactory : MessageReceiverFactory {
         from: Endpoint<T>,
         configManager: ConfigManager,
         handler: EndpointHandler<T>
-    ) = runBlocking {
+    ) = coroutineScope {
```
sschuberth, line 55: Couldn't this simply be omitted now, and below change `isActive` to `coroutineContext.isActive`?

transport/spi/src/testFixtures/kotlin/Utils.kt
oheger-bosch: I don't think that this will work, since `runBlocking` waits for the termination of the passed block. The whole action here has to be executed in parallel to the test code.

transport/sqs/src/main/kotlin/SqsMessageReceiverFactory.kt
oheger-bosch: Yes, I assume this is equivalent. Changed accordingly.

dao/src/testFixtures/kotlin/MockDatabaseModule.kt
oheger-bosch: Extended the commit message and also the comment for these functions.

PR #804, opened 2024-08-13
deps: Downgrade Vite to 5.3.5

mnonnenmacher:
With Vite 5.4.0 the `VITE_` configuration variables do not appear anymore in the built JS file. For example, instead of:

    Pe.BASE=nd.VITE_API_URL||"http://localhost:8080";

the output file now contains only:

    Pe.BASE="http://localhost:8080";

This breaks the entrypoint script for the UI Docker image which makes it impossible to use the image with other than the default values.

Downgrade Vite to 5.3.5 until a proper solution is found.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -70,7 +70,7 @@
     "tailwindcss": "^3.4.1",
     "tailwindcss-animate": "^1.0.7",
     "typescript": "^5.2.2",
-    "vite": "^5.1.6",
+    "vite": "5.3.5",
     "vitest": "^2.0.2"
   }
 }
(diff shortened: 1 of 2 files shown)


sschuberth:
Nit: Maybe use 4 instead of 2 spaces for indentation in the commit message so that the code renders nicely on GitHub?

ui/package.json
```diff
@@ -70,7 +70,7 @@
     "tailwindcss": "^3.4.1",
     "tailwindcss-animate": "^1.0.7",
     "typescript": "^5.2.2",
-    "vite": "^5.1.6",
+    "vite": "5.3.5",
```
sschuberth, line 73: Nit: Should this better be `~5.3.5` to allow for patch level updates?

mmurto (approved):

ui/package.json
mnonnenmacher: I'd rather not do any automatic updates until we know what caused the issue. Also I think the Vite team does not maintain old minor versions anyway.

PR #806, opened 2024-08-13
Revert "refactor(config): Migrate from `constructorArgs` to `providedProperties`"

sschuberth:
This reverts commit 43767ba until [1] is resolved.

[1]: https://youtrack.jetbrains.com/issue/KTIJ-30059


workers/config/src/main/kotlin/ConfigValidator.kt
```diff
@@ -20,12 +20,11 @@
 package org.eclipse.apoapsis.ortserver.workers.config
```
oheger-bosch, line 20: Maybe add a few more details to the commit message, so that it is not necessary to look up the reference to understand the reason behind this change, e.g. something like "... to enable proper syntax highlighting in the validation script".
But I leave this up to you.

oheger-bosch (approved):

PR #808, opened 2024-08-14
Job Monitor: Detect and delete jobs that exceed a timeout

oheger-bosch:
This PR adds a new "long-running jobs finder" component to Kubernetes Job Monitor. The component detects worker jobs that are running longer than a configured timeout and deletes them. This allows dealing with hanging jobs automatically. Different timeouts can be configured for the different worker types.

### DIFF
--- a/transport/kubernetes-jobmonitor/src/main/kotlin/JobHandler.kt
+++ b/transport/kubernetes-jobmonitor/src/main/kotlin/JobHandler.kt
@@ -91,6 +91,13 @@ internal class JobHandler(
                     status?.conditions.orEmpty().any { it.type in COMPLETED_CONDITIONS }
 
         /**
+         * Return a flag whether this job has run into a timeout according to the given [threshold]. This means that
+         * the job was started before the given date and is not yet completed.
+         */
+        fun V1Job.isTimeout(threshold: OffsetDateTime): Boolean =
+            !isCompleted() && status?.startTime?.isBefore(threshold) == true
+
+        /**
          * Obtain the ID of the ORT run from this job from the label used for this purpose. If the label is not set,
          * return *null*.
          */
--- a/transport/kubernetes-jobmonitor/src/main/kotlin/JobHandler.kt
+++ b/transport/kubernetes-jobmonitor/src/main/kotlin/JobHandler.kt
@@ -27,9 +27,6 @@ import io.kubernetes.client.openapi.models.V1Pod
 import java.time.OffsetDateTime
 import java.util.TreeMap
 
-import kotlin.time.Duration
-import kotlin.time.Duration.Companion.seconds
-
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.datetime.Clock
@@ -52,14 +49,8 @@ internal class JobHandler(
     /** The object to send notifications about failed jobs. */
     private val notifier: FailedJobNotifier,
 
-    /** The namespace that contains the objects of interest. */
-    private val namespace: String,
-
-    /**
-     * A duration that defines an interval in which a job should be considered as recently processed. This component
-     * keeps a list of the jobs that have been deleted during this interval to prevent duplicate processing of jobs.
-     */
-    private val recentlyProcessedInterval: Duration = 60.seconds,
+    /** The configuration. */
+    private val config: MonitorConfig,
 ) {
     companion object {
         private val logger = LoggerFactory.getLogger(JobHandler::class.java)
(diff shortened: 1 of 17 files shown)


sschuberth:
Probably @haikoschol could also have a look?

haikoschol (approved):

transport/kubernetes-jobmonitor/src/main/kotlin/JobHandler.kt
```diff
@@ -194,7 +185,19 @@ internal class JobHandler(
     private fun findPodsForJob(jobName: String): List<V1Pod> {
         val selector = "job-name=$jobName"
 
-        return api.listNamespacedPod(namespace, null, null, null, null, selector, null, null, null, null, false).items
+        return api.listNamespacedPod(
+            config.namespace,
```
nnobelis, line 196: Can you put name parameters ? (and in other locations)
@mnonnenmacher always requests that :)

transport/kubernetes-jobmonitor/src/main/kotlin/MonitorConfig.kt
```diff
@@ -0,0 +1,127 @@
...
+     */
+    val lostJobsMinAge: Duration,
+
+    /**
+     * A duration that defines an interval in which a job should be considered as recently processed. The job handler
+     * keeps a list of the jobs that have been deleted during this interval to prevent duplicate processing of jobs.
+     */
+    val recentlyProcessedInterval: Duration,
+
+    /** Flag whether the watching component should be active. */
```
nnobelis, line 96: In the log message above, it is "the watcher component".
It should be consistent.

(more cases underneath)

transport/kubernetes-jobmonitor/src/main/kotlin/Reaper.kt
```diff
@@ -34,8 +32,8 @@ internal class Reaper(
     /** The object to query and manipulate jobs. */
     private val jobHandler: JobHandler,
 
-    /** The maximum age of a completed job before it gets removed. */
-    private val maxJobAge: Duration,
+    /** The configuration. */
```
nnobelis, line 35: IMO such doc is useless :)

transport/kubernetes-jobmonitor/src/test/kotlin/ReaperTest.kt
```diff
@@ -63,23 +63,23 @@ class ReaperTest : WordSpec({
...
             val zoneOffset = ZoneOffset.ofHours(2)
             val date1 = OffsetDateTime.of(
-                2023,
-                Month.MAY.value,
-                3,
-                12,
-                8,
-                11,
-                0,
+                /* year = */ 2023,
```
nnobelis, line 66: Why the named parameters are commented ?

sschuberth: Because that's a Java function being called, which does not support named parameters natively.

transport/kubernetes-jobmonitor/src/main/kotlin/JobHandler.kt
oheger-bosch: We would here have to use commented parameter names as well, but I would rather not do this. I am not a big fan of the Kubernetes API with its numerous parameters. In these cases, however, it is quite clear which information is passed, since we have variables with speaking names between the *null* parameters.

transport/kubernetes-jobmonitor/src/main/kotlin/MonitorConfig.kt
oheger-bosch: I have replaced "watching component" by "watcher component" in two places. I do not want to change the names of configuration properties though.

transport/kubernetes-jobmonitor/src/main/kotlin/Reaper.kt
oheger-bosch: But they do no harm.

transport/kubernetes-jobmonitor/src/main/kotlin/JobHandler.kt
nnobelis: Fine with me. I was thinking about the last `false` though.

transport/kubernetes-jobmonitor/src/test/kotlin/ReaperTest.kt
nnobelis: TIL

nnobelis (approved):

PR #813, opened 2024-08-14
feat: Log an error if unable to connect to database server

wkl3nk:
In rare cases, the database server may stop accepting incoming connection requests, causing the datasource module to throw an exception. This update adds an error log message to ensure the issue is clearly visible in logs and can be detected by monitoring systems.

### DIFF
--- a/dao/src/main/kotlin/Database.kt
+++ b/dao/src/main/kotlin/Database.kt
@@ -49,6 +49,10 @@ import org.jetbrains.exposed.sql.transactions.transactionManager
 import org.koin.core.module.Module
 import org.koin.dsl.module
 
+import org.slf4j.LoggerFactory
+
+private val logger = LoggerFactory.getLogger(DataSource::class.java)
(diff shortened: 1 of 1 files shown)


dao/src/main/kotlin/Database.kt
```diff
@@ -17,7 +17,7 @@
  * License-Filename: LICENSE
  */
 
-@file:Suppress("TooManyFunctions")
+@file:Suppress("TooManyFunctions", "TooGenericExceptionCaught")
```
sschuberth, line 20: To avoid this, maybe below use

    return runCatching {
        HikariDataSource(dataSourceConfig)
    }.onFailure {
        logger.error("Failed to create data source.", it)
    }.getOrThrow()

wkl3nk: Good idea.

sschuberth (approved):

PR #818, opened 2024-08-14
fix(workers): Use ORT's function to fix the `user.home` system property

mnonnenmacher:
Under some scenarios the property can be set to "?". Use ORT's helper function to fix that based on environment variables. See [1] for some background information.

[1]: https://bugs.openjdk.java.net/browse/JDK-8193433


haikoschol (approved):
This solves a problem for me where `ConfigFileBuilder` created files in `?/.netrc` (running on Scaleway managed Kubernetes). They are now generated in `/home/ort/.netrc`.

sschuberth (approved):

PR #824, opened 2024-08-15
Fix some issues in generated OpenAPI spec

mnonnenmacher:

### DIFF
--- a/core/src/main/kotlin/plugins/OpenApi.kt
+++ b/core/src/main/kotlin/plugins/OpenApi.kt
@@ -43,6 +43,9 @@ import io.ktor.server.routing.routing
 
 import kotlinx.datetime.Instant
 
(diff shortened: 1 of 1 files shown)


mnonnenmacher:
@mmurto I have verified that this fixes the issues in #797.

nnobelis (approved):

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -116,7 +117,9 @@ fun Application.configureOpenApi() {
                     }
                     .connectSubTypes()
                     .handleNameAnnotation()
-                    .generateSwaggerSchema()
+                    .generateSwaggerSchema {
+                        optionalHandling = OptionalHandling.NON_REQUIRED
```
sschuberth, line 133: Weird that it is not the default to mark optional properties as not required... 😕

PR #827, opened 2024-08-16
Address remaining Detekt issues

sschuberth:


storage/spi/src/testFixtures/kotlin/StorageProviderFactoryForTesting.kt
```diff
@@ -49,6 +49,7 @@ class StorageProviderFactoryForTesting : StorageProviderFactory {
          * Stores the map with data of the latest provider instance that has been created. This can be used to query
          * or set test data directly.
          */
+        @Suppress("DoubleMutabilityForCollection")
```
sschuberth, line 52: @oheger-bosch maybe as a follow-up you could think about a nicer way to address this?

mnonnenmacher (approved):

PR #828, opened 2024-08-16
Non secret env variables

Juli0q:
Add option to define non-secret Env Variables in .ort.yml

### DIFF
--- a/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
+++ b/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
@@ -464,7 +464,7 @@ fun InfrastructureServiceDeclaration.mapToApi() =
 
 fun ApiEnvironmentVariableDeclaration.mapToModel() = EnvironmentVariableDeclaration(name, secretName)
 
-fun EnvironmentVariableDeclaration.mapToApi() = ApiEnvironmentVariableDeclaration(name, secretName)
+fun EnvironmentVariableDeclaration.mapToApi() = ApiEnvironmentVariableDeclaration(name, secretName, value)
 
 fun EnvironmentConfig.mapToApi() =
     ApiEnvironmentConfig(
--- a/api/v1/model/src/commonMain/kotlin/EnvironmentVariableDeclaration.kt
+++ b/api/v1/model/src/commonMain/kotlin/EnvironmentVariableDeclaration.kt
@@ -26,7 +26,8 @@ import kotlinx.serialization.Serializable
  *
  * This class can be used to describe environment variables that are referenced during an ORT run. When triggering an
  * ORT run the list of required variables can be specified. During the Analyzer step, these variables are then
- * defined. The variable values are obtained from secrets, which are referenced by their names.
+ * defined. The variable values are obtained from secrets, which are referenced by their names. Either a value or a
+ * secret name can be provided.
  */
 @Serializable
 data class EnvironmentVariableDeclaration(
@@ -34,5 +35,8 @@ data class EnvironmentVariableDeclaration(
     val name: String,
 
     /** The name of the secret that contains the value of the variable. */
-    val secretName: String
+    val secretName: String? = null,
+
+    /** The value of the environment variable. */
+    val value: String? = null
 )
--- a/workers/analyzer/src/test/kotlin/AnalyzerRunnerTest.kt
+++ b/workers/analyzer/src/test/kotlin/AnalyzerRunnerTest.kt
@@ -57,7 +57,8 @@ import org.eclipse.apoapsis.ortserver.model.runs.PackageManagerConfiguration
 import org.eclipse.apoapsis.ortserver.workers.common.context.WorkerContext
 import org.eclipse.apoapsis.ortserver.workers.common.context.WorkerOrtConfig
 import org.eclipse.apoapsis.ortserver.workers.common.env.config.ResolvedEnvironmentConfig
-import org.eclipse.apoapsis.ortserver.workers.common.env.definition.EnvironmentVariableDefinition
+import org.eclipse.apoapsis.ortserver.workers.common.env.definition.SecretVariableDefinition
+import org.eclipse.apoapsis.ortserver.workers.common.env.definition.SimpleVariableDefinition
 import org.eclipse.apoapsis.ortserver.workers.common.mapToOrt
 
 import org.ossreviewtoolkit.model.AnalyzerResult
@@ -308,8 +309,8 @@ class AnalyzerRunnerTest : WordSpec({
             val secret2 = mockk<Secret>()
             val environmentConfig = ResolvedEnvironmentConfig(
                 environmentVariables = setOf(
-                    EnvironmentVariableDefinition("MY_ENV_VAR", secret1),
-                    EnvironmentVariableDefinition("ANOTHER_ENV_VAR", secret2)
+                    SecretVariableDefinition("MY_ENV_VAR", secret1),
+                    SecretVariableDefinition("ANOTHER_ENV_VAR", secret2)
                 )
             )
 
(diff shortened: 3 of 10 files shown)


mmurto (dismissed):
Do I understand correctly that this allows a repository's `.ort.yml` to contain environment variables that are then used by ORT Server somewhere? Some documentation that environment variables can be added to `.ort.yml`, and how they are then used (Core probably doesn't get them? Are they for Workers?) would be great.

mnonnenmacher:
> Do I understand correctly that this allows a repository's `.ort.yml` to contain environment variables that are then used by ORT Server somewhere? Some documentation that environment variables can be added to `.ort.yml`, and how they are then used (Core probably doesn't get them? Are they for Workers?) would be great.

Almost correct, the definitions need to be in a `.ort.env.yml` file because we cannot extend the `.ort.yml` model. There is currently only a bit of documentation in the code, but an example for defining environment variables in that file can be found here:

https://github.com/eclipse-apoapsis/ort-server/blob/main/workers/common/src/test/resources/.ort.env.variables.yml

```yaml
environmentVariables:
- name: "variable1"
  secretName: "testSecret1"
- name: "variable2"
  secretName: "testSecret2"
```

Currently the values can only be defined by secrets, with this PR they can also be set directly in the file:

```yaml
environmentVariables:
- name: "variable1"
  value: "testSecret1"
- name: "variable2"
  value: "testSecret2"
```

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
mnonnenmacher, line 1: Commit message:
* "With this commit" is redundant, better say "Add the option to set...".
* It's `.ort.env.yml`, not `.ort.yml`.

api/v1/model/src/commonMain/kotlin/EnvironmentVariableDeclaration.kt
```diff
@@ -34,5 +34,8 @@ data class EnvironmentVariableDeclaration(
     val name: String,
 
     /** The name of the secret that contains the value of the variable. */
-    val secretName: String
+    val secretName: String? = null,
+
+    /** The value of the environment variable. */
```
mnonnenmacher, line 39: The docs should also describe that it is an error to set both `secretName` and `value`. Same in the file below.

mnonnenmacher (changes requested):

workers/analyzer/src/test/kotlin/AnalyzerRunnerTest.kt
```diff
@@ -565,8 +565,8 @@ class AnalyzerRunnerTest : WordSpec({
             val secret2 = mockk<Secret>()
             val environmentConfig = ResolvedEnvironmentConfig(
                 environmentVariables = setOf(
-                    EnvironmentVariableDefinition("MY_ENV_VAR", secret1),
-                    EnvironmentVariableDefinition("ANOTHER_ENV_VAR", secret2)
+                    SecretVariableDefinition("MY_ENV_VAR", secret1),
+                    SecretVariableDefinition("ANOTHER_ENV_VAR", secret2)
```
oheger-bosch, line 569: There should be a test case as well for a `SimpleVariableDefinition`.

workers/common/src/main/kotlin/common/env/definition/EnvironmentVariableDefinition.kt
```diff
@@ -21,16 +21,28 @@ package org.eclipse.apoapsis.ortserver.workers.common.env.definition
 
 import org.eclipse.apoapsis.ortserver.model.Secret
 
+interface EnvironmentVariableDefinition {
```
oheger-bosch, line 24: Please update documentation. The interface and the implementing classes should have correct documentation.
Also, this should probably be a `sealed` interface.

workers/common/src/test/kotlin/common/env/config/EnvironmentConfigLoaderTest.kt
```diff
@@ -206,6 +207,19 @@ class EnvironmentConfigLoaderTest : StringSpec() {
             config.shouldContainDefinition<MavenDefinition>(service) { it.id == "repo1" }
         }
 
+        "Environment definitions value are processed" {
```
oheger-bosch, line 210: I do not understand this change. The newly added test case does not seem to be a replacement for the deleted test case "Environment variable definitions are processed". Could you please explain?

workers/common/src/test/resources/.ort.env.direct-variables.yml
```diff
@@ -0,0 +1,5 @@
+environmentVariables:
+- name: "variable1"
+  value: "testSecret1"
```
oheger-bosch, line 3: Nit: Maybe do not call the values "testSecret". The point of direct variable values is that they are not sensitive.

mmurto:
> Almost correct, the definitions need to be in a `.ort.env.yml` file because we cannot extend the `.ort.yml` model. There is currently only a bit of documentation in the code, but an example for defining environment variables in that file can be found here:

Ah, so the functionality was already there, but this extends it to literal values, understood. It would be great to have some documentation (be it in code or commit or wherever) on what the environment variables do. What can be set with them, probably not `DATABASE_URL`? Unless there of course already exists such for the existing functionality, so please point me at it if it exists. :)

workers/analyzer/src/test/kotlin/AnalyzerRunnerTest.kt
Juli0q: I have added SimpleVariableDefinition to the already existing Text "create a process builder with the correct environment". Because imo it would just duplicate unnecessary code to create a new test. Is that fine?

workers/common/src/test/kotlin/common/env/config/EnvironmentConfigLoaderTest.kt
Juli0q: I think I made a mistake while copying the test as a template for the new test. I've reversed my changes.

mnonnenmacher:
> > Almost correct, the definitions need to be in a `.ort.env.yml` file because we cannot extend the `.ort.yml` model. There is currently only a bit of documentation in the code, but an example for defining environment variables in that file can be found here:
> 
> Ah, so the functionality was already there, but this extends it to literal values, understood. It would be great to have some documentation (be it in code or commit or wherever) on what the environment variables do. What can be set with them, probably not `DATABASE_URL`? Unless there of course already exists such for the existing functionality, so please point me at it if it exists. :)

We still need to write the documentation for that feature, but it's best if you look at the example files from the tests.
The `.ort.yml` is mainly used by the analyzer worker to setup the environment. This includes env variables, a `.netrc` file and package manager configuration files like a Maven `settings.xml`.  It is also used by the scanner worker to create a `.netrc` file so that source code or archives can be downloaded from servers that require credentials.

workers/common/src/main/kotlin/common/env/definition/EnvironmentVariableDefinition.kt
```diff
@@ -21,16 +21,31 @@ package org.eclipse.apoapsis.ortserver.workers.common.env.definition
...
+}
+
 /**
  * A class defining an environment variable which needs to be present when analyzing a repository.
  *
  * This is typically used to provide credentials or other values to external tools called during the analysis. The
  * variable can have an arbitrary name. Its value is obtained from a [Secret].
  */
-data class EnvironmentVariableDefinition(
+data class SecretVariableDefinition(
```
oheger-bosch, line 37: The documentation is still not ideal. Parts of the documentation from `SecretVariableDefinition` should be moved to the new interface, especially regarding the purpose. The concrete implementation classes should then have specific documentation how they obtain their value. `SimpleVariableDefinition` is still missing documentation.

oheger-bosch (approved):

PR #830, opened 2024-08-16
fix(dao): Do not load unnecessary data when creating ORT run

mnonnenmacher:
The code to determine the next run index when creating a new ORT run was unnecessarily loading the complete list of ORT runs for the repository and mapping it to the business model. Therefore a lot of unnecessary data was loaded not only from the `ort_runs` table, but also from associated tables like `labels` or `issues`.

To fix this, replace the call to `listForRepository` with a custom query that loads only the last index. This increases performance especially for repositories with a lot of runs and makes it also a lot less likely that the transaction has to be reattempted when runs are created concurrently.

### DIFF
--- a/dao/src/main/kotlin/repositories/DaoOrtRunRepository.kt
+++ b/dao/src/main/kotlin/repositories/DaoOrtRunRepository.kt
@@ -47,6 +47,7 @@ import org.jetbrains.exposed.sql.SizedCollection
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
 import org.jetbrains.exposed.sql.and
 import org.jetbrains.exposed.sql.deleteWhere
+import org.jetbrains.exposed.sql.max
 
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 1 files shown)


sschuberth (dismissed):

dao/src/main/kotlin/repositories/DaoOrtRunRepository.kt
```diff
@@ -63,7 +64,15 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
         issues: Collection<Issue>,
         traceId: String?
     ): OrtRun = db.blockingQuery {
-        val nextIndex = (listForRepository(repositoryId).data.maxByOrNull { it.index }?.index ?: 0) + 1
+        val lastIndex = OrtRunsTable
+            .select(OrtRunsTable.index)
```
oheger-bosch, line 68: An option could be to use the  `max` function instead of the combination of `order by` and `limit`. According to [this article](https://medium.com/@hamida.meknassi/min-max-vs-order-by-limit-performance-dfb6420701ee), this should provide better performance.

MarcelBochtler (approved):

oheger-bosch (approved):

PR #848, opened 2024-08-19
ui: Switch to using a new toast component

Etsija:
This PR addresses the problem of error messages from the API disappearing too fast, by switching to a new toast component which has a specific button to close the error toast.


ui/package.json
```diff
@@ -41,11 +41,13 @@
     "clsx": "^2.1.0",
     "cmdk": "^1.0.0",
     "lucide-react": "^0.428.0",
+    "next-themes": "^0.3.0",
```
sschuberth, line 43: Why is this also needed as a top-level component (and not e.g. a dependency of sonner)?

ui/src/routes/_layout/create-organization.tsx
```diff
@@ -66,10 +64,13 @@ const CreateOrganizationPage = () => {
...
     },
     onError(error: ApiError) {
-      toast({
-        title: error.message,
+      toast.error(error.message, {
         description: <ToastError error={error} />,
-        variant: 'destructive',
+        duration: Infinity,
+        cancel: {
+          label: 'Cancel',
```
sschuberth, line 71: Is this what the user sees / needs to click in order to get rid of the toast? If so, I'd prefer "Dismiss" over "Cancel".

mmurto:
Do you think it would make sense to use `toaster` instead of `sonner` as the file name? I think this is more descriptive and would keep the import the same as before the change. It could also be clearer to re-export `toast` from `sonner` from somewhere to have that as an "internal api".

ui/src/components/ui/sonner.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+
+import { useTheme } from 'next-themes';
+import { Toaster as Sonner } from 'sonner';
+
+type ToasterProps = React.ComponentProps<typeof Sonner>;
+
+const Toaster = ({ ...props }: ToasterProps) => {
+  const { theme = 'system' } = useTheme();
```
mmurto, line 16: Does this do anything without the `ThemeProvider` from `next-themes`?

mmurto (changes requested):

ui/src/components/ui/sonner.tsx
Etsija: Probably not, at least for now, I've just taken the component as-it-is with the install command from shadcn-ui. I would think that once/if we take the light and dark modes into use, this will have an effect.

ui/src/components/providers.tsx
```diff
@@ -39,7 +39,17 @@ export const Providers = ({ children }: { children: ReactNode }) => {
   return (
     <AuthProvider {...oidcConfig}>
       <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
-      <Toaster />
+      <Toaster
```
mmurto, line 42: I believe currently the UI is broken after the first commit, as `Toaster` has been migrated to the toasts are still executed with `useToast`, right?

I would squash all three commits to one.

ui/src/components/ui/toaster.tsx
```diff
@@ -7,36 +7,32 @@
...
-import { useToast } from '@/components/ui/use-toast';
+import { useTheme } from 'next-themes';
+import { Toaster as Sonner } from 'sonner';
 
-export function Toaster() {
-  const { toasts } = useToast();
+type ToasterProps = React.ComponentProps<typeof Sonner>;
+
+const Toaster = ({ ...props }: ToasterProps) => {
+  const { theme = 'system' } = useTheme();
```
mmurto, line 16: I'd remove this for as long as themes are not supported.

mmurto (changes requested):

ui/src/components/ui/toaster.tsx
Etsija: There are also [Vite specific theming instructions](https://ui.shadcn.com/docs/dark-mode/vite) on shadcn-ui, which don't imply importing any external theme libraries.

mmurto (approved):

ui/src/components/ui/toaster.tsx
sschuberth: :+1: for dark theme 😁

PR #861, opened 2024-08-21
feat(ui): Beautify the look of version information a bit

sschuberth:
Make it look a bit similar to [1].

[1]: https://v0.dev/t/LoBns0Spg7G


sschuberth:
This makes it look like:

image

mmurto (commented):
We haven't been very strict with conventional commits, so maybe more of a nit, but `style` is [more traditionally](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type) a change of _code style_, not graphical style, so this should maybe be a feat.

mmurto (approved):

PR #862, opened 2024-08-21
feat(ui): Download logs from an ORT Run

Etsija:


ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
```diff
@@ -94,11 +117,15 @@ const RunComponent = () => {
...
-    });
+  async function handleDownload(runId: number, filename?: string) {
+    if (!filename) {
+      await downloadLogs(runId);
+    } else {
+      await downloadReports({
+        runId: runId,
+        fileName: filename,
+      });
+    }
```
mmurto, line 128: I think using one function here adds some complexity without providing much value. I directly using `downloadLogs` and `downloadReports` in the `onClick`s should work without wrapping them in this function, or doesn't it?

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
```diff
@@ -185,6 +212,19 @@ const RunComponent = () => {
...
+              <TableCell>Logs</TableCell>
+              <TableCell>
+                <Link onClick={() => handleDownload(ortRun.id)}>
+                  <Button
+                    variant='outline'
+                    className='font-semibold text-blue-400'
+                  >
+                    Download ZIP file
+                  </Button>
+                </Link>
```
mmurto, line 225: Does the download logic work if this is not a `Link` but a `Button` with the same `onClick`? If it works, this should probably be a `Button`.

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
```diff
@@ -61,7 +61,30 @@ const RunComponent = () => {
       }),
   });
 
-  const downloadZipFile = async ({
+  const downloadLogs = async (runId: number) => {
```
sschuberth, line 64: Commit message nits:
- "zipfile" -> "Zip file" (or "ZIP file")
- Please use passive voice, so no mention of "we" etc. as in "so we use default values of"
- IIUC there is no Scaleway-specific logic in this commit, so Scaleway should also not be mentioned as part of the commit message, to not create the false impression that this would only work on a Scaleway deployment.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
```diff
@@ -61,7 +61,30 @@ const RunComponent = () => {
...
+    try {
+      // TODO: use a readymade query from TanStack query client for this?
```
sschuberth, line 66: Nit: Start capitalized after a colon 😉

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
```diff
@@ -185,6 +212,19 @@ const RunComponent = () => {
...
             )}
+            <TableRow>
+              <TableCell>Logs</TableCell>
+              <TableCell>
+                <Link onClick={() => handleDownload(ortRun.id)}>
+                  <Button
+                    variant='outline'
+                    className='font-semibold text-blue-400'
+                  >
+                    Download ZIP file
```
sschuberth, line 223: IMO we should say *what* is being downloaded (and the fact that the archive is in ZIP format is less important), so I propose "Download log archive".

sschuberth (changes requested):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
Etsija: IIUC there actually is a Scaleway-specific issue with this.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
Etsija: Using Button only stretches the buttons along the whole column, but that should be fixable.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
sschuberth: Like what?

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
Etsija: Really, in English? In Finnish, it's lower-case.

mmurto: TanStack Query should in fact not be used for this, so the whole comment should be removed, which resolves this. 😎

sschuberth: > Really, in English?

In American English, yes, esp. if the following sentence is a full sentence.

mmurto (approved):

PR #865, opened 2024-08-21
feat(ui): Include the global run ID into the log archive download

sschuberth:
This avoids conflicts when downloading multiple archives.

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
@@ -75,7 +75,7 @@ const RunComponent = () => {
       const url = window.URL.createObjectURL(blob);
       const a = document.createElement('a');
       a.href = url;
-      a.download = 'logs.zip';
+      a.download = `logs-global-run-${runId}.zip`;
       a.click();
       window.URL.revokeObjectURL(url);
     } catch (error) {


sschuberth:
> Adding some combination of organization name, product name, repo name, local run index and run time would probably be even better.

I was tempted to do some of that as well, but then I 1) wasn't sure whether it's always fine to expose organization etc. names as part of the archive if you're passing the file on, and 2) the API works with the *global* run ID underneath, and I felt it's good to reflect that in the file name to have a 1:1 match, e.g. to enter it in the top-right search field.

mmurto:
> > Adding some combination of organization name, product name, repo name, local run index and run time would probably be even better.
> 
> I was tempted to do some of that as well, but then I 1) wasn't sure whether it's always fine to expose organization etc. names as part of the archive if you're passing the file on, and 2) the API works with the _global_ run ID underneath, and I felt it's good to reflect that in the file name to have a 1:1 match, e.g. to enter it in the top-right search field.

Clear! My guess (based on personal experience :D) is that the bigger problem will be quickly identifying the correct zip from dozens with incrementing IDs, but this can surely be adjusted later if the need arises.

Etsija:
What about the report files? Could they include the run ID as well?

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
```diff
@@ -75,7 +75,7 @@ const RunComponent = () => {
       const url = window.URL.createObjectURL(blob);
       const a = document.createElement('a');
       a.href = url;
-      a.download = 'logs.zip';
+      a.download = `logs-global-run-${runId}.zip`;
```
sschuberth, line 78: Hmm. Apparently the backend already provides a proper name here:

https://github.com/eclipse-apoapsis/ort-server/blob/03f5738d78aacb1ef4814b3d3eb657cf641cbee8/core/src/main/kotlin/api/RunsRoute.kt#L98

So, wouldn't it be possible to use that name from the response?

sschuberth: See https://github.com/eclipse-apoapsis/ort-server/pull/867.

PR #867, opened 2024-08-21
feat(ui): Prefer the backend-provided log archive name

sschuberth:
Otherwise align with [1] as a fallback.

[1]: https://github.com/eclipse-apoapsis/ort-server/blob/03f5738d78aacb1ef4814b3d3eb657cf641cbee8/core/src/main/kotlin/api/RunsRoute.kt#L98


mmurto (approved):

sschuberth:
> I believe links cannot either anymore exceed the column 70 in commit messages. I've had to use URL shorteners.

It should be fine for footnote links. Commitlint [passes](https://github.com/eclipse-apoapsis/ort-server/actions/runs/10493231020/job/29066681819?pr=867).

PR #871, opened 2024-08-22
ui: Choose the log level for ORT Run logs

Etsija:
Capture1

Capture2

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
@@ -17,6 +17,7 @@
  * License-Filename: LICENSE
  */
 
+import { ChevronDownIcon } from '@radix-ui/react-icons';
 import { useSuspenseQuery } from '@tanstack/react-query';
 import { createFileRoute } from '@tanstack/react-router';
 
@@ -32,6 +33,12 @@ import {
 import { Badge } from '@/components/ui/badge';
 import { Button } from '@/components/ui/button';
 import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuTrigger,
+} from '@/components/ui/dropdown-menu';
 import { Separator } from '@/components/ui/separator';
 import {
   Table,
@@ -126,6 +133,13 @@ const RunComponent = () => {
     }
   };
 
+  const handleLevelChoice = (level: string) => {
+    // Get the list of steps that were run to only request logs for those steps. The `config` step is
+    // always run implicitly and not part of the jobs, so add that manually.
+    const jobNames = Object.keys(ortRun.jobs).concat('config');
+    downloadLogs(ortRun.id, level, jobNames.join(','));
+  };
+
   return (
     <Card className='mx-auto w-full max-w-4xl'>
       <CardHeader className='flex flex-row items-start'>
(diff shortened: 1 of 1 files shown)


ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
```diff
@@ -223,13 +235,44 @@ const RunComponent = () => {
...
 
-                    downloadLogs(ortRun.id, '', jobNames.join(','));
+                    downloadLogs(ortRun.id, level, jobNames.join(','));
                   }}
                   variant='outline'
-                  className='font-semibold text-blue-400'
+                  className={'rounded-r-none font-semibold text-blue-400'}
                 >
-                  Download log archive
+                  Download log archive, level={level}
```
sschuberth, line 243: I propose to shorten this a bit (and make it read less technical) by saying `Download {level} log archive`.

sschuberth: And just to confirm how this is working: Selecting a level from the drop-down does not start the download immediately, but updates the text of the button, and a user still needs to click the main button to start the download, correct?

sschuberth (changes requested):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
Etsija: Correct, that's intentional.

mmurto: I'd make this work the way that clicking a level from the dropdown immediately starts the download. I think that if someone selects a level, they always want to download it immediately after, so the second click is not needed.

mmurto (changes requested):

Etsija:
@sschuberth also requested the workings of this, so which behaviour do I implement?

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
sschuberth: > I'd make this work the way that clicking a level from the dropdown immediately starts the download.

I don't have a *very* strong opinion here, but *if* doing that IMO the "Choose the log level" title in the drop down should go away completely, and the individual entries need to be full titles instead of just the level name.

mmurto: > > I'd make this work the way that clicking a level from the dropdown immediately starts the download.
> 
> I don't have a _very_ strong opinion here, but _if_ doing that IMO the "Choose the log level" title in the drop down should go away completely, and the individual entries need to be full titles instead of just the level name.

Yeah, that would be ideal IMO.

- Download INFO log archive
  - Download ERROR log archive
  - Download WARN log archive
  - etc.

mmurto: It also should probably be somehow communicated that selecting a level downloads them for that level and all above. Not sure what the best way is.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
```diff
@@ -218,18 +235,44 @@ const RunComponent = () => {
...
+                      <DropdownMenuRadioItem value='WARN'>
+                        Download WARN log archive
+                      </DropdownMenuRadioItem>
+                      <DropdownMenuRadioItem value='INFO'>
+                        Download INFO log archive
+                      </DropdownMenuRadioItem>
+                      <DropdownMenuRadioItem value='DEBUG'>
+                        Download DEBUG log archive
+                      </DropdownMenuRadioItem>
+                    </DropdownMenuRadioGroup>
```
mmurto, line 273: As this is not really a radio, a better way could be something like:

```typescript
<DropdownMenuItem
  onClick={() => {
    downloadLogs(ortRun.id, "ERROR", jobNames.join(','));
  }
>
  Download ERROR log archive
</DropdownMenuItem>
```

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
Etsija: Umm what? I totally think this is a radio, as radio items indicate what value was chosen before, which I think is useful if the user wants to save logs for different levels from the same run. And isn't your example quite repetitive?

mmurto: [Radio menu](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio) sets a selected value, that can then be used in a form. This is just a list of buttons with differenc actions executed immediately on click.

Etsija: > [Radio menu](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio) sets a selected value, that can then be used in a form. This is just a list of buttons with differenc actions executed immediately on click.

I'd at least group them then with `<DropdownMenuGroup>`, to lift the state handling and function calling to the group component, instead of repeating almost exactly the same code for each item.

mmurto: I don't think there's need for state handling with `onClick`s, or what state do you need? As for repetition, it's so little and abstraction brings more complexity, so I'd just repeat it for the levels as they're so few and simple.

Etsija:
I retained the little handleLevelChoice() function, to not repeat what Sebastian did and wrote for every single item.

mmurto:
> I retained the little handleLevelChoice() function, to not repeat what Sebastian did and wrote for every single item.

That logic could be direclty added inside of `downloadLogs` function, but feel free to do as you see best.

mmurto (approved):

Etsija:
> > I retained the little handleLevelChoice() function, to not repeat what Sebastian did and wrote for every single item.
> 
> That logic could be direclty added inside of `downloadLogs` function, but feel free to do as you see best.

I feel like I'd want it to stay this way, in anticipation of future PR, where for example the steps are also chosen in the UI, to retain an "interception function" for the `jobNames`.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
sschuberth: Maybe instead of e.g. "Download INFO *log* archive" say "Download INFO *level* archive" to emphasize it's a level that includes less verbose levels, too?

sschuberth (approved):

PR #880, opened 2024-08-22
Deduplication migration

oheger-bosch:
This PR adds a migration script to remove the duplicated packages and projects from the DB.

This is still WIP, but I want to share my current state before my vacations. The migration test does not yet test projects. It is currently failing for packages, since after the migration, there are more packages left than expected. I think, there is some copy & paste error in the script.

From the script, you can get the basic idea. Feel free to ignore if you come up with a better solution.


oheger-bosch:
Replaced by #1073.

PR #886, opened 2024-08-22
Fix role sync and jobmonitor issues

mnonnenmacher:

### DIFF
--- a/core/src/main/kotlin/plugins/Lifecycle.kt
+++ b/core/src/main/kotlin/plugins/Lifecycle.kt
@@ -21,6 +21,8 @@ package org.eclipse.apoapsis.ortserver.core.plugins
 
 import io.ktor.server.application.Application
 
+import kotlin.concurrent.thread
+
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 
@@ -42,8 +44,10 @@ fun Application.configureLifecycle() {
     environment.monitor.subscribe(DatabaseReady) {
         val authorizationService by inject<AuthorizationService>()
 
-        runBlocking(Dispatchers.IO) {
-            syncRoles(authorizationService)
+        thread {
+            runBlocking(Dispatchers.IO) {
+                syncRoles(authorizationService)
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/plugins/Lifecycle.kt
```diff
@@ -42,8 +44,10 @@ fun Application.configureLifecycle() {
     environment.monitor.subscribe(DatabaseReady) {
         val authorizationService by inject<AuthorizationService>()
 
-        runBlocking(Dispatchers.IO) {
-            syncRoles(authorizationService)
+        thread {
```
sschuberth, line 47: Obviously, it would make more sense to turn the calling function into a suspend function so just `runBlocking` is dropped, but I guess that's not easily possible for some reason? Could you provide a brief rationale in the commit message?

sschuberth, line 47: Maybe getting a bit off-topic, but could you further explain (here, not in the ocmmit message) about the different to Java's `Thread {}` (with upper-case "T") as used e.g. at

https://github.com/eclipse-apoapsis/ort-server/blob/cff36a377033e4c81752cc7e2212a71639beceb1/transport/spi/src/testFixtures/kotlin/Utils.kt#L74

mnonnenmacher: `Thread {}` is just calling the `Thread` constructor with a runnable argument and you have to start it on your own. Kotlin's `thread` factory function also starts the thread automatically and offers some configuration options which are not used here.

sschuberth: Ok, so I guess that confirms you're intentionally using `thread {}`, not `Thread {}`, here.

And that we should probably align on the former in existing code?

sschuberth (approved):

core/src/main/kotlin/plugins/Lifecycle.kt
mnonnenmacher: Actually, I never thought about using `Thread {}`. And yes, as `thread {}` avoid the start call I would prefer that.

PR #887, opened 2024-08-23
Show the ongoing duration for active runs

### LINKED ISSUE #833: The run overview page should show the ongoing duration for active runs
Currently, the "Duration" column just show the total duration for finished runs. However, it'd also be interesting for how long an active runs has been running until now. Probably that "dynamic" time could displayed in the same column, but somehow visually different than the "static" total runtime once finished.

On a somewhat related note, also the tool tips for the individual "Job Status" circles could, besides the name and status, also display the current / final duration. Of course, data in a tool tip is static, so the time for a running analyzer state would not update while the tool tip it displayed, but once if the tool tip is re-displayed the next time.

Etsija:
Please have a look at the individual commits.

Resolves #833.

### DIFF
--- a/ui/src/components/ort-run-job-status.tsx
+++ b/ui/src/components/ort-run-job-status.tsx
@@ -28,12 +28,17 @@ import {
   getStatusBackgroundColor,
   getStatusClass,
 } from '@/helpers/get-status-colors';
+import { RunDuration } from './run-duration';
 
 type OrtRunJobStatusProps = {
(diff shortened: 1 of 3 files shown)


mmurto (approved):

sschuberth:
> Resolves #833.

Please remember to put this into the message of the respective commit that concludes the implementation, not in the PR discussion. But let's keep it as-is for this PR, no need to amend the commit message.

PR #904, opened 2024-08-26
ui: Refactor UI config

### LINKED ISSUE #860: Refactor UI config
UI code currently directly refers to environment variables in multiple places. We should initialize config from env variables in one place, and then refer to that configuration wherever it is needed. Could also consider using https://env.t3.gg/ for typesafe env variables in addition.

Etsija:

Resolves #860.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -32,6 +32,7 @@
     "@radix-ui/react-tabs": "^1.1.0",
     "@radix-ui/react-toast": "^1.1.5",
     "@radix-ui/react-tooltip": "^1.0.7",
+    "@t3-oss/env-core": "^0.11.1",
     "@tanstack/react-query": "^5.29.2",
     "@tanstack/react-query-devtools": "^5.32.0",
     "@tanstack/react-router": "^1.33.7",
--- a/ui/src/config.ts
+++ b/ui/src/config.ts
@@ -19,6 +19,8 @@
 
 import { AuthProviderProps } from 'react-oidc-context';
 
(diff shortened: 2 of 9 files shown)


ui/package.json
```diff
@@ -32,6 +32,7 @@
     "@radix-ui/react-tabs": "^1.1.0",
     "@radix-ui/react-toast": "^1.1.5",
     "@radix-ui/react-tooltip": "^1.0.7",
+    "@t3-oss/env-core": "^0.11.1",
```
mnonnenmacher, line 35: Could you shortly summarize the benefits of using T3 in the commit message?

mnonnenmacher (changes requested):

ui/src/config.ts
```diff
@@ -0,0 +1,70 @@
+/*
```
lamppu, line 1: Commit message typo: Inizialise -> Initialize

Etsija:
I'd really appreciate an opinion from @mmurto as well to this PR.

ui/package.json
sschuberth: Commit message nit: There is no point in using `[1]` if you don't refer to it 😉 I propose to change this to:

```
T3 Env [1] ...

...

[1]: https://env.t3.gg/docs/core
```

So refer to the footnote in the first sentence, and drop "See " from the footnote.

mmurto (changes requested):

mmurto (approved):

PR #908, opened 2024-08-26
ci(renovate): Configure custom manager for `libs.versions.toml`

### LINKED ISSUE #801: Configure custom Renovate manager to update Docker image in libs.versions.toml
The `eclipseTemurin` version in `libs.versions.toml` should be kept in sync with the version used in Dockerfiles. To automate this, configure a [custom manager](https://docs.renovatebot.com/modules/manager/regex/) to update the version in `libs.versions.toml` as well. For an example see: https://github.com/renovatebot/renovate/discussions/25354

mnonnenmacher:
Configure a custom manager for the Eclipse Temurin version defined in `libs.versions.toml` so that it gets updated automatically. This ensures that the version does not deviate from the version used in Dockerfiles.

Fixes #801.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -5,6 +5,17 @@
     ":semanticCommitScopeDisabled",
     ":semanticCommitTypeAll(deps)"
   ],
+  "customManagers": [
+    {
+      "description": "Update Eclipse Temurin in libs.versions.toml",
+      "customType": "regex",
+      "datasourceTemplate": "docker",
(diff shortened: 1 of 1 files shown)


mnonnenmacher:
I couldn't easily test this locally (https://docs.renovatebot.com/modules/platform/local/) because my Git and Node versions are not compatible with Renovate. I would just merge this as is and hope that it works, and then debug only if it doesn't.

PR #912, opened 2024-08-27
ui: Implement log downloading for the runs2 route

Etsija:
Capture

### DIFF
--- a/ui/src/components/sidebar.tsx
+++ b/ui/src/components/sidebar.tsx
@@ -28,7 +28,7 @@ interface SidebarNavProps {
   sections: {
     label?: string;
     items: {
-      to: string;
+      to?: string;
       params?: Record<string, string | undefined>;
       title: string;
       visible?: boolean; // default: true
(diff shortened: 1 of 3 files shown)


ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/route.tsx
```diff
@@ -27,10 +27,30 @@ const Layout = () => {
...
+      to: '/organizations/$orgId/products/$productId/repositories/$repoId/runs2/$runIndex',
+    },
+    {
+      title: 'License Findings',
+      to: '/organizations/$orgId/products/$productId/repositories/$repoId/runs2/$runIndex',
+    },
+    {
+      title: 'Rule Violations',
+      to: '/organizations/$orgId/products/$productId/repositories/$repoId/runs2/$runIndex',
+    },
```
mmurto, line 45: If non-existing items are added, I'd somehow indicate for the buttons/links that they're currently disabled.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs2/$runIndex/logs/index.tsx
mmurto, line 1: Does this need a separate page, as it only contains one button?

mmurto (changes requested):

ui/src/components/sidebar.tsx
```diff
@@ -32,6 +32,7 @@ interface SidebarNavProps {
       params?: Record<string, string | undefined>;
       title: string;
       visible?: boolean; // default: true
+      disabled?: boolean; // default: false
```
mmurto, line 35: Would this be a little simpler if the `to` param was optional, and the link is disabled if it is not given? It may confuse someone that they have to give an URL even if the link is disabled.

mmurto (changes requested):

ui/src/components/sidebar.tsx
mmurto: The change to Sidebar params should also probably be a separate commit.

Etsija: Oh, of course, was going to do that but I forgot.

Etsija: @mmurto let me still fix this so the "disabled" doesn't need to be given.

mmurto (approved):

PR #922, opened 2024-08-28
feat: Include the log level into the archive download

sschuberth:
This helps to avoid archives being overwritten when different levels are downloaded for the same run.


Etsija:
UI part LGTM.

mnonnenmacher (approved):

PR #930, opened 2024-08-29
Add endpoint to retrieve Vulnerabilities for an ORT run

MarcelBochtler:


services/hierarchy/src/main/kotlin/VulnerabilityService.kt
```diff
@@ -0,0 +1,86 @@
...
+                    ).select(
+                    VulnerabilitiesTable.id,
+                    VulnerabilitiesTable.externalId,
+                    VulnerabilitiesTable.summary,
+                    VulnerabilitiesTable.description,
+                    IdentifiersTable.type,
+                    IdentifiersTable.namespace,
+                    IdentifiersTable.name,
+                    IdentifiersTable.version
+                ).where { AdvisorJobsTable.ortRunId eq ortRunId }
```
sschuberth, line 66: Indentation looks weird here. Is it correct?

services/hierarchy/src/test/kotlin/VulnerabilityServiceTest.kt
```diff
@@ -0,0 +1,189 @@
...
+                val vulnerabilities = createVulnerabilities(
+                    Identifier("Maven", "org.apache.logging.log4j", "log4j-core", "2.14.0") to "CVE-2021-45046",
+                    Identifier("Maven", "com.fasterxml.jackson.core", "jackson-databind", "2.9.6") to "CVE-2018-14721"
+                )
+                val advisorResult = createAdvisorResults(vulnerabilities)
+                val ortRun = createVulnerabilityEntries(advisorResult)
+
+                val results = service.listForOrtRunId(ortRun.id).data
+
+                results.size shouldBe 2
```
sschuberth, line 67: Prefer `shouldHaveSize`.

services/hierarchy/src/test/kotlin/VulnerabilityServiceTest.kt
```diff
@@ -0,0 +1,189 @@
...
+            }
+
+            "return an empty list if an ORT run has no vulnerable packages" {
+                val service = VulnerabilityService(db)
+
+                val ortRun = createVulnerabilityEntries(emptyMap())
+
+                val results = service.listForOrtRunId(ortRun.id).data
+
+                results shouldBe emptyList()
```
sschuberth, line 107: Prefer `should beEmpty()`.

services/hierarchy/src/main/kotlin/VulnerabilityService.kt
MarcelBochtler: This is what my IntelliJ is doing with autoformatting.
As detekt is not complaining, I would guess that it is correct?
And yes, it also looks weird and wrong to my eye.

MarcelBochtler:
Converting to draft, as I decided to include references into the same PR.

services/hierarchy/src/main/kotlin/VulnerabilityService.kt
```diff
@@ -0,0 +1,146 @@
...
+            result.add(
+                VulnerabilityWithIdentifier(
+                    vulnerability = vulnerability.copy(
+                        references = mutableListOf(vulnsWithIdAndReference.reference)
+                    ),
+                    identifier = identifier
+                )
+            )
+        }
+    }
```
mnonnenmacher, line 113: Maybe I don't understand the algorithm correctly, but does this not change the size of the result set? For example, if the query has a limit of 20 and the first two vulnerabilities have each 10 references, then only 2 items are returned? And the first vulnerability has more than 20 references, than only the first 20 are returned?

MarcelBochtler: Good catch. The implementation was incorrect.
Changed it, and added a test for it.

dao/src/main/kotlin/utils/Extensions.kt
```diff
@@ -57,6 +59,22 @@ internal fun <E : LongEntity, M> SortableEntityClass<E>.listQuery(
     return ListQueryResult(data, parameters, totalCount)
 }
 
+/**
+ * Run the [query] using the [parameters] to create a [ListQueryResult]. The entities are mapped to the  corresponding
+ * model objects using the provided [entityMapper].
```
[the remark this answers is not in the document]
MarcelBochtler: > Does it define the sortable columns?

Yes.

api/v1/model/src/commonMain/kotlin/Vulnerability.kt
```diff
@@ -0,0 +1,30 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.api.v1.model
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class Vulnerability(
```
mnonnenmacher, line 25: In the API model we could also add the identifier directly to this class to avoid the need for the separate `VulnerabilityWithIdentifier` class.

mnonnenmacher (approved):

Etsija:
I've tried to add this to merge queue twice today, as I would need an updated API for my newest upcoming PR, but the bot removes the PR from merge queue due to failing tests, although this page shows all checks pass.

sschuberth:
> Am I blocked from adding an approved PR for back-end to merge queue?

No.

> That would explain it, but could someone else (@sschuberth ?) try it?

What usually happens is this: First, tests are only run on the PR branch itself. If you're lucky and all tests succeed despite being flaky (😬), then you're allowed to add the PR to the merge queue. When the PR arrives at the head of the queue, the PR branch is rebased against current `main`, and all tests are run again. Due to the flakyness, some tests might fail now, which causes the PR to be removed from the queue. The PR overview, though, still shows the test status from the first PR test run, where everything is green.

TL;DR just try again to add the PR to merge queue, which I'll do now.

PR #936, opened 2024-08-30
ui: Include traffic lights in ORT Run results page

Etsija:
Capture

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/index.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/index.tsx
@@ -24,6 +24,7 @@ import { Repeat } from 'lucide-react';
 import { useRepositoriesServiceGetOrtRunByIndexKey } from '@/api/queries';
 import { RepositoriesService } from '@/api/requests';
 import { LoadingIndicator } from '@/components/loading-indicator';
+import { OrtRunJobStatus } from '@/components/ort-run-job-status';
 import { Badge } from '@/components/ui/badge';
 import { Button } from '@/components/ui/button';
 import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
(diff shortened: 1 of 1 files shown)


ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/index.tsx
```diff
@@ -46,6 +49,7 @@ const RunComponent = () => {
         repositoryId: Number.parseInt(params.repoId),
         ortRunIndex: Number.parseInt(params.runIndex),
       }),
+    refetchInterval: pollInterval,
```
mmurto, line 52: It would probably make sense to disable refetch after the run is finished, as it should not change after that.

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/index.tsx
Etsija: Polling / stop polling has been now tested to work with FINISHED and ACTIVE runs, respectively.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/index.tsx
```diff
@@ -46,6 +49,14 @@ const RunComponent = () => {
         repositoryId: Number.parseInt(params.repoId),
         ortRunIndex: Number.parseInt(params.runIndex),
       }),
+    refetchInterval: (run) => {
+      if (
+        run.state.data?.status === 'FINISHED' ||
+        run.state.data?.status === 'FAILED'
+      )
+        return false;
```
mmurto, line 57: Shouldn't this have `{}`?

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/index.tsx
mmurto: Apparently works without. Are they auto-formatted away if you add them?

mmurto: I've never faced it before, only in one liners like `if (condition) return true` but not separate lines and indicated by indentation. IMO it's clearer with braces, but as the formatter doesn't care, feel free to do as you see best.

mmurto (approved):

PR #938, opened 2024-08-30
ci(docker): Use tag suffix also for the base image cache

mnonnenmacher:
If the `imageVariant` is configured in the build matrix, it is used as a suffix for the image tag. Also use the suffix for the cache image tag, otherwise the caches for different image variants overwrite each other.

This is a fixup for ab3657e.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -175,8 +175,8 @@ jobs:
         push: true
         tags: ${{ steps.meta-base.outputs.tags }}
         labels: ${{ steps.meta-base.outputs.labels }}
-        cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ github.repository_owner }}/ort-server-${{ matrix.docker.image }}:cache
-        cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ github.repository_owner }}/ort-server-${{ matrix.docker.image }}:cache,mode=max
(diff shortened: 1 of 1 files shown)


nnobelis (approved):

bs-ondem (approved):

.github/workflows/docker-build.yml
```diff
@@ -175,8 +175,8 @@ jobs:
         push: true
         tags: ${{ steps.meta-base.outputs.tags }}
         labels: ${{ steps.meta-base.outputs.labels }}
-        cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ github.repository_owner }}/ort-server-${{ matrix.docker.image }}:cache
-        cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ github.repository_owner }}/ort-server-${{ matrix.docker.image }}:cache,mode=max
+        cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ github.repository_owner }}/ort-server-${{ matrix.docker.image }}:cache${{ steps.compute-tag-suffix.outputs.suffix }}
```
sschuberth, line 178: Just double-checking: Isn't there some separator missing between "cache" and "$", like a dash or a colon?

mnonnenmacher: The separator is part of the suffix, see the `compute-tag-suffix` step.

PR #943, opened 2024-08-30
docs(ui): Align form descriptions

sschuberth:
Align the wording to have fewer different strings. Use punctuation even for single sentences.

### DIFF
--- a/ui/src/routes/_layout/create-organization.tsx
+++ b/ui/src/routes/_layout/create-organization.tsx
@@ -36,7 +36,6 @@ import {
 import {
   Form,
   FormControl,
-  FormDescription,
   FormField,
   FormItem,
   FormLabel,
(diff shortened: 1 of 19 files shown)


ui/src/routes/_layout/create-organization.tsx
```diff
@@ -104,7 +104,7 @@ const CreateOrganizationPage = () => {
                     <Input {...field} autoFocus />
                   </FormControl>
                   <FormDescription>
-                    Enter the name of your organization
+                    The name of the organization.
```
mmurto, line 107: While adjusting the descriptions, it would make sense to remove redundant ones. If the field in org creation form has a visible title of `name`, description like this is totally redundant.

IMO [this](https://primer.style/ui-patterns/forms/overview#caption) is quite a reasonable guideline.

mmurto (changes requested):

ui/src/routes/_layout/create-organization.tsx
sschuberth: Done in a new commit as it was easier to filter these out after the "normalization" of descriptions.

mmurto (approved):

PR #945, opened 2024-08-30
New job status `FINISHED_WITH_ISSUES`

nnobelis:

The threshold upon which the issues and the rule violations will trigger this new status is hardcoded to a severity of `WARNING`.
In a future PR, it will be made configurable.

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/JobStatus.kt
+++ b/api/v1/model/src/commonMain/kotlin/JobStatus.kt
@@ -39,5 +39,8 @@ enum class JobStatus {
     FAILED,
 
     /** The job was processed successfully. */
-    FINISHED
+    FINISHED,
+
+    /** The job has finished but there were some issues over the threshold. */
+    FINISHED_WITH_ISSUES
 }
--- a/api/v1/model/src/commonMain/kotlin/OrtRun.kt
+++ b/api/v1/model/src/commonMain/kotlin/OrtRun.kt
@@ -160,7 +160,8 @@ enum class OrtRunStatus {
     CREATED,
     ACTIVE,
     FINISHED,
-    FAILED
+    FAILED,
+    FINISHED_WITH_ISSUES
 }
 
 @Serializable
--- a/model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
+++ b/model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
@@ -142,7 +142,9 @@ data class EvaluatorWorkerError(
 @Serializable
 data class ReporterWorkerResult(
     /** The ID of the Reporter job, as it is stored in the database. */
-    override val jobId: Long
+    override val jobId: Long,
+    /** if `true`, the result has issues over the threshold. */
+    val hasIssues: Boolean = false
 ) : OrchestratorMessage(), WorkerMessage
 
 /**
--- a/model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
+++ b/model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
@@ -122,7 +122,9 @@ data class ScannerWorkerError(
 @Serializable
 data class EvaluatorWorkerResult(
     /** The ID of the Evaluator job, as it is stored in the database. */
-    override val jobId: Long
+    override val jobId: Long,
+    /** if `true`, the result has issues over the threshold. */
+    val hasIssues: Boolean = false
 ) : OrchestratorMessage(), WorkerMessage
 
 /**
--- a/model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
+++ b/model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
@@ -82,7 +82,9 @@ data class AnalyzerWorkerError(
 @Serializable
 data class AdvisorWorkerResult(
     /** The ID of the Advisor job, as it is stored in the database. */
-    override val jobId: Long
+    override val jobId: Long,
+    /** if `true`, the result has issues over the threshold. */
+    val hasIssues: Boolean = false
 ) : OrchestratorMessage(), WorkerMessage
 
 /**
--- a/model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
+++ b/model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
@@ -62,7 +62,9 @@ interface WorkerMessage {
 @Serializable
 data class AnalyzerWorkerResult(
     /** The ID of the Analyzer job, as it is stored in the database. */
-    override val jobId: Long
+    override val jobId: Long,
+    /** if `true`, the result has issues over the threshold. */
+    val hasIssues: Boolean = false
 ) : OrchestratorMessage(), WorkerMessage
 
 /**
--- a/model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
+++ b/model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
@@ -102,7 +102,9 @@ data class AdvisorWorkerError(
 @Serializable
 data class ScannerWorkerResult(
     /** The ID of the Scanner job, as it is stored in the database. */
-    override val jobId: Long
+    override val jobId: Long,
+    /** if `true`, the result has issues over the threshold. */
+    val hasIssues: Boolean = false
 ) : OrchestratorMessage(), WorkerMessage
 
 /**
--- a/orchestrator/src/main/kotlin/Orchestrator.kt
+++ b/orchestrator/src/main/kotlin/Orchestrator.kt
@@ -195,7 +195,11 @@ class Orchestrator(
      * Handle messages of the type [EvaluatorWorkerResult].
      */
     fun handleEvaluatorWorkerResult(header: MessageHeader, evaluatorWorkerResult: EvaluatorWorkerResult) {
-        handleWorkerResult(EvaluatorEndpoint, header, evaluatorWorkerResult)
+        if (!evaluatorWorkerResult.hasIssues) {
+            handleWorkerResult(EvaluatorEndpoint, header, evaluatorWorkerResult)
+        } else {
+            handleWorkerResultWithIssues(EvaluatorEndpoint, header, evaluatorWorkerResult)
+        }
     }
 
     /**
--- a/orchestrator/src/main/kotlin/Orchestrator.kt
+++ b/orchestrator/src/main/kotlin/Orchestrator.kt
@@ -213,7 +213,11 @@ class Orchestrator(
      * Handle messages of the type [ReporterWorkerResult].
      */
     fun handleReporterWorkerResult(header: MessageHeader, reporterWorkerResult: ReporterWorkerResult) {
-        handleWorkerResult(ReporterEndpoint, header, reporterWorkerResult)
+        if (!reporterWorkerResult.hasIssues) {
+            handleWorkerResult(ReporterEndpoint, header, reporterWorkerResult)
+        } else {
+            handleWorkerResultWithIssues(ReporterEndpoint, header, reporterWorkerResult)
+        }
     }
 
     /**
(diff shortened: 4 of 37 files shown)


api/v1/model/src/commonMain/kotlin/JobStatus.kt
```diff
@@ -39,5 +39,8 @@ enum class JobStatus {
     FAILED,
 
     /** The job was processed successfully. */
-    FINISHED
+    FINISHED,
+
+    /** The job has finished but there were some severe issues. */
```
sschuberth, line 44: IMO "severe" is relative here, esp. if the threshold becomes configurable later, and probably should be omitted.

api/v1/model/src/commonMain/kotlin/JobStatus.kt
```diff
@@ -39,5 +39,8 @@ enum class JobStatus {
...
+    FINISHED_WITH_ERRORS
```
sschuberth, line 45: To align with docs and ORT terminology, better name this `FINISHED_WITH_ISSUES`? Similar for several other places / class names that use "error".

api/v1/model/src/commonMain/kotlin/OrtRun.kt
```diff
@@ -160,7 +160,8 @@ enum class OrtRunStatus {
     CREATED,
     ACTIVE,
     FINISHED,
-    FAILED
+    FAILED,
```
sschuberth, line 163: As a side note, having e.g. test execution statuses in mind, I find "failed" already confusing here. For tests, it means that the test *did* run but the assertions failed. Whereas "error" means that a test did not run.

I'd actually propose to use the same terminology here:
- `FAILED` -> `ERROR`
- Introduce a new `FAILED` or `FAILED_WITH_ISSUES`.

mnonnenmacher: I don't think test terminology is a good fit here, it is better to look at other pipeline implementations. For example:

* Jenkins: SUCCESS, UNSTABLE, FAILURE, NOT_BUILT or ABORTED
* GitHub Workflows: completed, action_required, cancelled, failure, neutral, skipped, stale, success, timed_out, in_progress, queued, requested, waiting, pending 
* Azure Pipelines: Canceled, Failed, Succeeded, SucceededWithIssues, Skipped

What they have in common is the use of "success/succeeded" for a completely successful run and "failure/failed" for a completely failed run. This matches with our current terminology.

For the new status introduced in this commit Jenkins has "UNSTABLE" and Azure Pipelines has "SucceededWithIssues". The problem is to find a name that is not overloaded:
* "ERROR" is also the name of a severity, so does `FINISHED_WITH_ERRORS` mean that an issue or rule violation had `ERROR` severity? Actually it doesn't, because the threshold should be configurable.
* "ISSUES" is also used for the `Issue` class, so does `FINISHED_WITH_ISSUES` mean that there are issues with any severity? Does it include rule violations or vulnerabilities?

A not overloaded name could be `FINISHED_WITH_PROBLEMS`.

Etsija:
Note: there are more places in the UI where the `FINISHED` and `FAILED` statuses are handled in the current code, and one of the latest changes was [this](https://github.com/eclipse-apoapsis/ort-server/blob/ef54c2d38cd75aa7a0a5e04454a93b5b17724eaa/ui/src/routes/_layout/organizations/%24orgId/products/%24productId/repositories/%24repoId/_layout/runs/%24runIndex/index.tsx#L52-L56), which needs to also handle the new `FAILED_WITH_ISSUES` handling.

I would suggest that as the basis of the UI part of the PR, you search the `ui/` path for a string `FINISHED` to find all files that involve it. (I'm sure you already did, so this is just a heads up that there has been some changes in the UI since you made this PR).

mmurto:
What's the rationale for the new status? It duplicates already existing information, as we have the issues (/rule violations etc.) related to a run stored in the database as separate entities, so we could always just return for example the count of issues above some specified threshold.

mnonnenmacher:
> What's the rationale for the new status? It duplicates already existing information, as we have the issues (/rule violations etc.) related to a run stored in the database as separate entities, so we could always just return for example the count of issues above some specified threshold.

It should be possible to determine if a run or job is "red", "yellow", or "green" from a single property. Dynamically determining this has several problems:
* The run summaries returned by the endpoint that lists all runs for a repository do not contain individual issues, rule violations, or vulnerabilities, so the required information is not available.
* The severity threshold will be configurable like in ORT itself (e.g., hints usually do not make a run yellow, and some might want to ignore warnings).
* Every API consumer would have to reimplement that logic, this includes the UI, CI integrations, and custom logic from users.
* There might be other features built on top, like aborting a pipeline if a job is "yellow" to provide immediate feedback in CI.

mmurto:
> > What's the rationale for the new status? It duplicates already existing information, as we have the issues (/rule violations etc.) related to a run stored in the database as separate entities, so we could always just return for example the count of issues above some specified threshold.
> 
> It should be possible to determine if a run or job is "red", "yellow", or "green" from a single property. Dynamically determining this has several problems:
> 
> * The run summaries returned by the endpoint that lists all runs for a repository do not contain individual issues, rule violations, or vulnerabilities, so the required information is not available.
> * The severity threshold will be configurable like in ORT itself (e.g., hints usually do not make a run yellow, and some might want to ignore warnings).
> * Every API consumer would have to reimplement that logic, this includes the UI, CI integrations, and custom logic from users.
> * There might be other features built on top, like aborting a pipeline if a job is "yellow" to provide immediate feedback in CI.

Understood. IMO the original implementation could provide a clearer separation between ORT Server administration issues, where the analysis fails due to some configuration or system being unavailable, and analysis results, where the analysis consumer is required to do something. I find that kind of separation helpful.

Of the four points, I think others but the severity threshold could be solved by returning a dynamic `hasIssues` or similar from the run summary endpoint.

api/v1/model/src/commonMain/kotlin/OrtRun.kt
nnobelis: @sschuberth  are you fine with `FINISHED_WITH_PROBLEMS` ?

sschuberth: > it is better to look at other pipeline implementations.

Agreed.

> What they have in common is the use of "success/succeeded" for a completely successful run and "failure/failed" for a completely failed run. This matches with our current terminology.

Good.

> @sschuberth are you fine with `FINISHED_WITH_PROBLEMS` ?

I'm not sure yet. While I agree that the term "problems" is not overloaded / not used yet in ORT, it somehow sounds a bit too strong like, well, a problem, in the sense of something bad. But actually, e.g. for rule violations that might turn out to be false-positives, we don't know yet if really something bad happened.

What all of `Issue`, `RuleViolation` and `VulnerabilityReference` have in common is a `severity` property. So maybe call this `FINISHED_WITH_SEVERITY`? I'm not fully convinced yet myself, I'm just throwing in ideas.

mnonnenmacher: After thinking about the options my preferred name would be `FINISHED_WITH_ISSUES`. Even if it could be confused with the issues class the name itself is still very clear: the run could finished, but it found some issues.
`FAILED` should be reserved for runs that could not create a result.

nnobelis: Yes sorry I meant to type `FINISHED_WITH_ISSUES` to :sweat_smile:

sschuberth: Either `FINISHED_WITH_ISSUES` or `FINISHED_WITH_SEVERITY` are fine with me. Please remember the align the class names in this PR accordingly.

nnobelis:
> Note: there are more places in the UI where the `FINISHED` and `FAILED` statuses are handled in the current code, and one of the latest changes was [this](https://github.com/eclipse-apoapsis/ort-server/blob/ef54c2d38cd75aa7a0a5e04454a93b5b17724eaa/ui/src/routes/_layout/organizations/%24orgId/products/%24productId/repositories/%24repoId/_layout/runs/%24runIndex/index.tsx#L52-L56), which needs to also handle the new `FINISHED_WITH_ISSUES` state.
> 
> I would suggest that as the basis of the UI part of the PR, you search the `ui/` path for a string `FINISHED` to find all files that involve it. (I'm sure you already did, so this is just a heads up that there has been some changes in the UI since you made this PR).

@Etsija I changed another location. Please tell me I there is still something to change.

Etsija:
> @Etsija I changed another location. Please tell me I there is still something to change.
> 
> I searched the usages as your recommended but all the other ones were in generated code.

I believe you found all usages.

workers/analyzer/src/main/kotlin/analyzer/AnalyzerWorker.kt
```diff
@@ -100,7 +102,11 @@ internal class AnalyzerWorker(
             ortRunService.storeAnalyzerRun(analyzerRun.mapToModel(jobId))
         }
 
-        RunResult.Success
+        if (analyzerRun.result.issues.values.flatten().any { it.severity >= Severity.WARNING }) {
```
mnonnenmacher, line 105: Call `analyzerRun.result.getAllIssues()` instead, the current call does not get issues from the dependency tree.
This should appply issue resolutions to ignore resolved issues, but we can do that in a follow-up PR, because I think it requires some more changes and would make this PR too large.

mnonnenmacher (changes requested):

workers/advisor/src/main/kotlin/advisor/AdvisorComponent.kt
```diff
@@ -64,6 +64,12 @@ class AdvisorComponent : EndpointComponent<AdvisorRequest>(AdvisorEndpoint) {
                     Message(message.header, AdvisorWorkerResult(advisorJobId))
                 }
 
+                is RunResult.FinishedWithIssues -> {
+                    logger.warn("Advisor job '$advisorJobId' finished with issues.")
+                    // TODO: Send the right message type.
+                    Message(message.header, AdvisorWorkerError(advisorJobId))
```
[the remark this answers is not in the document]
nnobelis: I know this is confusing, but this change is just for the project to build (hence the `TODO`). In the following commits, the correct `RunResult`s are set for each worker.

orchestrator/src/main/kotlin/Orchestrator.kt
```diff
@@ -257,6 +257,15 @@ class Orchestrator(
...
     }
 
+    /**
+     * Handle the given [result] message with the given [header] from a worker of the given [endpoint]. The run finished
+     * on the worker, but with issues over the threshold.
+     */
+    @Suppress("UnusedPrivateMember")
+    private fun handleWorkerResultWithIssues(endpoint: Endpoint<*>, header: MessageHeader, result: WorkerMessage) {
+        handleCompletedJob(endpoint, header, result, JobStatus.FINISHED_WITH_ISSUES)
+    }
```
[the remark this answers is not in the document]
nnobelis: I moved it to the next commit.

model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
```diff
@@ -74,6 +74,16 @@ data class AnalyzerWorkerError(
...
 
+/**
+ * A message notifying the Orchestrator about a result produced by the Analyzer Worker. This result has issues over the
+ * threshold.
+ */
+@Serializable
+data class AnalyzerWorkerResultWithIssues(
+    /** The ID of the Analyzer job, as it is stored in the database. */
+    override val jobId: Long
+) : OrchestratorMessage(), WorkerMessage
```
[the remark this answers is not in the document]
nnobelis: I did the change.

For the future, I wonder if we could change the `OrchestratorMessage` with e.g.:

```
data class OrchestratorMessage(
    val jobId: Long,
    /** Analyzer, Scanner, ... */
    val type: OcherstratorMessageType
    /** Finished, FinishedWithErrors, ...*/
    val result: Status
)
```

and no subclass, to simplify even further the messaging and the amount of boilerplate code needed to be written.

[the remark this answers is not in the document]
mnonnenmacher: For structures like that sealed classes are often convenient, because when you handle them with a `when` block the compiler can already guarantee that all cases are handled. Actually, that would be an argument against my proposal.
But as we discussed we offline, we will have to look into this topic anyway outside of this PR.

mnonnenmacher (approved):

mnonnenmacher (approved):

PR #958, opened 2024-09-02
test(transport): Increase the timeout when querying the blocking queue 

nnobelis:
This is to try to limit the flakiness of the CI tests, e.g.:

```
org.eclipse.apoapsis.ortserver.transport.sqs.SqsMessageReceiverFactoryTest > Messages can have empty trace IDs FAILED
    java.lang.AssertionError: Expected value to not be null, but was null.
        at org.eclipse.apoapsis.ortserver.transport.testing.UtilsKt.checkMessage(Utils.kt:88)
        at org.eclipse.apoapsis.ortserver.transport.sqs.SqsMessageReceiverFactoryTest$1$5.invokeSuspend(SqsMessageReceiverFactoryTest.kt:110)

        Caused by:
        java.lang.AssertionError: Expected value to not be null, but was null.
            at org.eclipse.apoapsis.ortserver.transport.testing.UtilsKt.checkMessage(Utils.kt:88)
            ... 1 more
```

### DIFF
--- a/transport/spi/src/testFixtures/kotlin/Utils.kt
+++ b/transport/spi/src/testFixtures/kotlin/Utils.kt
@@ -28,6 +28,8 @@ import java.util.concurrent.BlockingQueue
 import java.util.concurrent.LinkedBlockingQueue
 import java.util.concurrent.TimeUnit
 
+import kotlin.concurrent.thread
+
 import kotlinx.coroutines.runBlocking
 
 import org.eclipse.apoapsis.ortserver.config.ConfigManager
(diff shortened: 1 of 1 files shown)


transport/spi/src/testFixtures/kotlin/Utils.kt
```diff
@@ -71,12 +73,12 @@ fun startReceiver(configManager: ConfigManager): LinkedBlockingQueue<Message<Orc
         queue.offer(message)
     }
 
-    Thread {
+    thread {
```
sschuberth, line 76: Cross-linking [this discussion](https://github.com/eclipse-apoapsis/ort-server/pull/886#discussion_r1728707065) for reference.

sschuberth (approved):

PR #965, opened 2024-09-02
ui: Edit sidebar and main content alignment

lamppu:
Please see individual commits and pictures and explanation below for details.

Näyttökuva (112)

To get the sidebar to always be in the same place on the page and not jump around, the main contents have to have the same width everywhere (unless there can be more or less empty space on the right side than on the left side of the sidebar+main-content combined div, i.e. the combined div won't be centered). The contents everywhere else would look quite stupid with a wider width, so the table in the Infrastructure Services page now has a horizontal scroll bar. This page could possibly have an exception to the rule, but the issue is, that if the width is declared on the individual pages themselves, the sidebar will flash on the left side before rendering in the correct spot, as it is declared on the route.tsx files, and I guess will do a first render before the other contents are ready to be rendered. When the surrounding div with the fixed width is declared in the route.tsx files, this flashing effect goes away. If we want to have the Infrastructure Services page wider, perhaps the surrounding div class name could be evaluated based on the current route, but then of course the sidebar would jump when visiting this page.

Näyttökuva (113)

Näyttökuva (114)

### DIFF
--- a/ui/src/components/page-layout.tsx
+++ b/ui/src/components/page-layout.tsx
@@ -27,7 +27,7 @@ interface PageLayoutProps {
 
 export const PageLayout = ({ sections, children }: PageLayoutProps) => {
   return (
-    <Page className='flex flex-col gap-4 md:flex-row'>
+    <Page className='flex flex-col justify-center gap-4 md:flex-row'>
       {sections && (
         <Pane className='w-full md:w-52'>
           <Sidebar sections={sections} />
--- a/ui/src/components/page-layout.tsx
+++ b/ui/src/components/page-layout.tsx
@@ -33,7 +33,7 @@ export const PageLayout = ({ sections, children }: PageLayoutProps) => {
           <Sidebar sections={sections} />
         </Pane>
       )}
-      <Content>{children}</Content>
+      <Content className='w-full md:max-w-4xl'>{children}</Content>
     </Page>
   );
 };
--- a/ui/src/components/page-layout.tsx
+++ b/ui/src/components/page-layout.tsx
@@ -1,0 +1,39 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import { Sidebar, SidebarNavProps } from '@/components/sidebar';
+import { Content, Page, Pane } from '@/components/ui/layout';
+
+interface PageLayoutProps {
+  sections: SidebarNavProps['sections'] | undefined;
+  children: React.ReactNode;
+}
+
+export const PageLayout = ({ sections, children }: PageLayoutProps) => {
+  return (
+    <Page className='flex flex-col gap-4 md:flex-row'>
+      {sections && (
+        <Pane className='w-full md:w-52'>
+          <Sidebar sections={sections} />
+        </Pane>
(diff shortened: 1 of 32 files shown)


ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/route.tsx
```diff
@@ -74,10 +74,12 @@ const Layout = () => {
...
   return (
-    <div className='flex h-full w-full gap-2'>
+    <>
       <Sidebar sections={sections} />
-      <Outlet />
-    </div>
+      <div className='w-full md:max-w-4xl'>
+        <Outlet />
+      </div>
+    </>
```
mmurto, line 82: This sort of pattern is not repated in many `route.tsx` files. It could make sense to extract it to a component. [Here's](https://primer.style/components/page-layout/react/alpha) an example of what the API could look like.

mmurto (changes requested):

Etsija:
Apart from the obvious little jump, when page contents require vertical scrollbar in the browser, the result looks good.

mmurto:
> To get the sidebar to always be in the same place on the page and not jump around, the main contents have to have the same width everywhere

This could be avoided by implementing the local navigation in the app header, [similar to how GitHub does it](https://primer.style/foundations/layout#anatomy-of-a-page).

Etsija:
> > To get the sidebar to always be in the same place on the page and not jump around, the main contents have to have the same width everywhere
> 
> This could be avoided by implementing the local navigation in the app header, [similar to how GitHub does it](https://primer.style/foundations/layout#anatomy-of-a-page).

But that navigation would be restricted to quite few options, before it overflows from the right. I'm not sure even the current options of run results sidebar would fit and I don't know what else needs to be added to that in future.

Etsija:
> GitHub seems to have 8 items there, which takes it to approximately middle of the page on 1920w, so a few more should fit. I'd imagine we could design the routes with 10 primary pages under each hierarchy level, and we could use dropdowns if we want more.

It is suggested by @mnonnenmacher and @sschuberth that some, if not all of the sidebar items also include updating badges after the name: for example, number of vulnerabilities of each severity (color coded), number of dependencies, an indicator that the job is still running, to name a few. That will increase the total size of each sidebar/navbar item, which needs to be taken into account when deciding, whether we implement this with a sidebar or a local navbar.

I would suggest that we merge this PR and gather some user experience, then decide if we should do it with a local navbar instead.

mmurto:
> I would suggest that we merge this PR and gather some user experience, then decide if we should do it with a local navbar instead. I do agree that a navbar would simplify the DOM of the "main area".

Totally, no need to go for it in this PR, it was just an observation related to the comment.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/route.tsx
lamppu: Okay, now I've added a suggestion of one way of doing this

lamppu:
The components in layout.tsx are basically just named divs, but I kind of like how they make the PageLayout component quite readable.

Etsija:
UI checked, still looking good.
I like the refactoring!

ui/src/components/page-layout.tsx
```diff
@@ -0,0 +1,42 @@
...
+export const PageLayout = ({ sections }: PageLayoutProps) => {
+  return (
+    <Page className='flex flex-col justify-center gap-4 md:flex-row'>
+      {sections && (
+        <Pane className='w-full md:w-52'>
+          <Sidebar sections={sections} />
+        </Pane>
+      )}
+      <Content className='w-full md:max-w-4xl'>
+        <Outlet />
```
mmurto, line 38: I'd take `children` in and pass `Outlet` there from routes. Outlet is so integral part of the the routing and I think it doesn't work outside of `route.tsx`, so it would be good if they were visible in route files.

mmurto (changes requested):

mmurto:
> The components in layout.tsx are basically just named divs, but I kind of like how they make the PageLayout component quite readable.

Good improvement! As a follow-up at some point, we could improve the components to be modular in a way that they support multiple layouts, for example using other components as Panes than the sidebar, and perhaps having a Header component etc. But ready to merge after the `Outlet` change in the review comments.

ui/src/components/page-layout.tsx
Etsija: Out of curiosity, where would it be seen if they don't work from inside `<PageLayout>`? TanStack Router devtools? Because I didn't notice anything when clicking away in the UI.

mmurto (approved):

PR #968, opened 2024-09-03
ui: Display the (user)name of the currently logged in user

### LINKED ISSUE #645: Display the name of the currently logged in user
Currently, the named of the logged-in user is not displayed anywhere. As a simple solution, we should probably just add a read-only top-level entry in the profile menu that names the user.

In the mid-term, we probably want to have a dedicated profile settings page, that not only displays the name, but also the user's roles, allows to customize the user icon ([Gravatar](https://gravatar.com/) support! 😁), etc.

Etsija:
Capture1

Capture2

Resolves #645.

### DIFF
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -24,6 +24,7 @@ import { useForm } from 'react-hook-form';
 import { z } from 'zod';
 
 import { ModeToggle } from '@/components/mode-toggle';
+import { Avatar, AvatarFallback } from '@/components/ui/avatar';
 import { Form, FormControl, FormField, FormItem } from '@/components/ui/form';
 import { useUser } from '@/hooks/use-user';
 import {
@@ -213,6 +214,15 @@ export const Header = () => {
             </Button>
           </DropdownMenuTrigger>
           <DropdownMenuContent align='end'>
+            <DropdownMenuItem className='flex gap-2' disabled>
+              <Avatar className='h-8 w-8'>
+                <AvatarFallback className='h-8 w-8 bg-red-400'>
+                  {user.username?.slice(0, 2).toUpperCase()}
+                </AvatarFallback>
+              </Avatar>
+              <span className='font-semibold'>{user.username}</span>
+            </DropdownMenuItem>
+            <DropdownMenuSeparator />
             <Link to='/about'>
               <DropdownMenuItem>About</DropdownMenuItem>
             </Link>
--- a/ui/package.json
+++ b/ui/package.json
@@ -19,6 +19,7 @@
     "@hookform/resolvers": "^3.3.4",
     "@radix-ui/react-accordion": "^1.1.2",
     "@radix-ui/react-alert-dialog": "^1.0.5",
+    "@radix-ui/react-avatar": "^1.1.0",
     "@radix-ui/react-checkbox": "^1.0.4",
     "@radix-ui/react-dialog": "^1.0.5",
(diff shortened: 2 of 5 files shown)


sschuberth:
Thanks! Seeing the screenshot, I'm inclined for a more compact representation. Either just put the username after the existing "Logout" so it reads "Logout admin" (where the "admin" part could still be dimmed / italic / whatever, but does not have to be), or have a separate entry at the top that just shows the username ("admin" in this case) plus a separator to separate from the following "About" entry.

Once more looking at GitHub, the latter is what they do, as they combine that entry with a "Switch account" functionality, which is something we might also want to add in the future:

image

@mmurto any opinion as well?

ui/src/components/header.tsx
```diff
@@ -222,6 +222,10 @@ export const Header = () => {
               </Link>
             )}
             <DropdownMenuSeparator />
+            <DropdownMenuItem className='flex gap-1' disabled>
+              <div>Logged in as:</div>
+              <div className='font-semibold'>{user.username}</div>
+            </DropdownMenuItem>
```
mmurto, line 228:
```suggestion
            <DropdownMenuItem disabled>
              Logged in as: <span className='font-semibold'>{user.username}</span>
            </DropdownMenuItem>
```

mmurto: Something like this would probably be a better HTML.

mmurto: The current "Logged in as: username" also takes quite a lot more space than the other items, streching the menu. Would it work to display it in less space, maybe just the username, or put it in two rows? Check how other apps do it.

mmurto (changes requested):

mmurto:
> @mmurto any opinion as well?

Agree. Maybe only the username seems to be the most common, or you could add an avatar with initials next to it. It may feel strange with admin user as there's also an admin link, but it's such an edge case so I would consider what it looks to like for real users.

ui/src/components/header.tsx
Etsija: I'd rather go for @sschuberth 's suggestion and just put the username to the top row ("header") of the menu.

Etsija:
> > @mmurto any opinion as well?
> 
> Agree. Maybe only the username seems to be the most common, or you could add an avatar with initials next to it. It may feel strange with admin user as there's also an admin link, but it's such an edge case so I would consider what it looks to like for real users.

Do we need an avatar now (I don't think we have a library for that yet), or do we add it later?

mmurto:
> > > @mmurto any opinion as well?
> 
> > 
> 
> > Agree. Maybe only the username seems to be the most common, or you could add an avatar with initials next to it. It may feel strange with admin user as there's also an admin link, but it's such an edge case so I would consider what it looks to like for real users.
> 
> 
> 
> Do we need an avatar now (I don't think we have a library for that yet), or do we add it later?
> 
> Capture
> 
> 
> 
> 

We could use https://ui.shadcn.com/docs/components/avatar with just the fallback for initials, but feel free to do as you see fit.

Etsija:
> We could use https://ui.shadcn.com/docs/components/avatar with just the fallback for initials, but feel free to do as you see fit.

I have free time, so maybe I'll go for the avatar 😄 as Johanna already did it for the other UI.

sschuberth:
As a follow-up: Should the standard profile icon now also match the new initials-based icon from the menu, like GitHub does it?

PR #971, opened 2024-09-03
Fix issues in OpenAPI spec

mnonnenmacher:
Apply some fixes to make the generated OpenAPI pass validation with https://apitools.dev/swagger-parser/online/. See the commit messages for details.

### DIFF
--- a/core/src/main/kotlin/plugins/OpenApi.kt
+++ b/core/src/main/kotlin/plugins/OpenApi.kt
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2022 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ * Copyright 2024 SMILEY4
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,22 +25,31 @@ import io.github.smiley4.ktorswaggerui.data.AuthType
 import io.github.smiley4.ktorswaggerui.routing.openApiSpec
 import io.github.smiley4.ktorswaggerui.routing.swaggerUI
 import io.github.smiley4.schemakenerator.core.connectSubTypes
+import io.github.smiley4.schemakenerator.core.data.BaseTypeData
+import io.github.smiley4.schemakenerator.core.data.Bundle
+import io.github.smiley4.schemakenerator.core.data.TypeId
+import io.github.smiley4.schemakenerator.core.data.flatten
 import io.github.smiley4.schemakenerator.core.handleNameAnnotation
 import io.github.smiley4.schemakenerator.reflection.collectSubTypes
 import io.github.smiley4.schemakenerator.reflection.processReflection
 import io.github.smiley4.schemakenerator.swagger.OptionalHandling
-import io.github.smiley4.schemakenerator.swagger.compileReferencingRoot
-import io.github.smiley4.schemakenerator.swagger.data.RefType
-import io.github.smiley4.schemakenerator.swagger.data.TitleType
+import io.github.smiley4.schemakenerator.swagger.data.CompiledSwaggerSchema
+import io.github.smiley4.schemakenerator.swagger.data.SwaggerSchema
 import io.github.smiley4.schemakenerator.swagger.generateSwaggerSchema
 import io.github.smiley4.schemakenerator.swagger.handleCoreAnnotations
-import io.github.smiley4.schemakenerator.swagger.withAutoTitle
+import io.github.smiley4.schemakenerator.swagger.steps.SwaggerSchemaCompileUtils.merge
+import io.github.smiley4.schemakenerator.swagger.steps.SwaggerSchemaCompileUtils.resolveReferences
+import io.github.smiley4.schemakenerator.swagger.steps.SwaggerSchemaCompileUtils.shouldReference
+import io.github.smiley4.schemakenerator.swagger.steps.SwaggerSchemaUtils
+import io.github.smiley4.schemakenerator.swagger.steps.buildTypeDataMap
 
 import io.ktor.server.application.Application
 import io.ktor.server.application.install
 import io.ktor.server.config.ApplicationConfig
 import io.ktor.server.routing.route
 import io.ktor.server.routing.routing
+
+import io.swagger.v3.oas.models.media.Schema
 
 import kotlinx.datetime.Instant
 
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -150,3 +159,163 @@ fun Application.configureOpenApi() {
...
+            if (typeData.typeParameters.isNotEmpty()) {
+                val paramString = typeData.typeParameters
+                    .map { (_, param) -> determineTitle(typeDataMap[param.type]!!, typeDataMap) }
+                    .joinToString("-")
+                "${it}_$paramString"
+            } else {
+                it
+            }
+        }.let {
+            it + (typeData.id.additionalId?.let { a -> "#$a" } ?: "")
```
sschuberth: Prefer `.orEmpty()` over ` ?: ""`.

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -150,3 +159,163 @@ fun Application.configureOpenApi() {
...
+        }
+    }
+}
+
+private fun Bundle<SwaggerSchema>.customCompileReferencingRoot(): CompiledSwaggerSchema {
+    return CustomCompileReferenceRootStep().compile(this)
```
sschuberth, line 207: Maybe turn into an expression?

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -150,3 +159,163 @@ fun Application.configureOpenApi() {
...
+        if (typeData.typeParameters.isNotEmpty()) {
+            val paramString = typeData.typeParameters
+                .map { (_, param) -> getRefPath(typeDataMap[param.type]!!, typeDataMap) }
+                .joinToString("-")
+            "${it}_$paramString"
+        } else {
+            it
+        }
+    }.let {
+        it + (typeData.id.additionalId?.let { a -> "#$a" } ?: "")
```
sschuberth: Prefer `.orEmpty()` over ` ?: ""`.

sschuberth: Isn't this basically the same as `determineTitle()` can can be de-duplicated?

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -150,3 +159,163 @@ fun Application.configureOpenApi() {
...
+     */
+    fun compile(bundle: Bundle<SwaggerSchema>): CompiledSwaggerSchema {
+        val result = CustomCompileReferenceStep().compile(bundle)
+        if (shouldReference(result.swagger)) {
+            val refPath = getRefPath(result.typeData, bundle.buildTypeDataMap())
+            return CompiledSwaggerSchema(
+                typeData = result.typeData,
+                swagger = schemaUtils.referenceSchema(refPath, true),
+                componentSchemas = buildMap {
+                    this.putAll(result.componentSchemas)
```
sschuberth, line 228: Drop `this.`.

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -150,3 +159,163 @@ fun Application.configureOpenApi() {
...
+                    this[refPath] = result.swagger
```
sschuberth, line 229: Maybe use `put(refPath, result.swagger)` to align with `putAll` above?

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -150,3 +159,163 @@ fun Application.configureOpenApi() {
...
+                }
+            )
+        } else {
+            return CompiledSwaggerSchema(
```
sschuberth, line 233: Pull out `return` here and above to before `if`.

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -150,3 +159,163 @@ fun Application.configureOpenApi() {
...
+            }
+        } else {
+            refObj
+        }
+    }
+
+    private fun placeholder() = Schema<Any>()
+
+    private fun Collection<SwaggerSchema>.find(id: TypeId): SwaggerSchema? {
+        return this.find { it.typeData.id == id }
```
sschuberth, line 296: Turn into an expression?

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -150,3 +159,163 @@ fun Application.configureOpenApi() {
...
+
+/**
+ * An adapted version of [io.github.smiley4.schemakenerator.swagger.steps.SwaggerSchemaCompileUtils.getRefPath] that
+ * generates valid references for the generated schemas. For example, the type `GenericType<Type1, Type2>` is
+ * transformed to `GenericType_Type1-Type2` instead of `GenericType<Type1,Type2>`.
+ *
+ * According to the [specification](https://swagger.io/specification/#components-object), schema keys must match the
+ * regular expression `^[a-zA-Z0-9\.\-_]+$`.
+ */
+private fun getRefPath(typeData: BaseTypeData, typeDataMap: Map<TypeId, BaseTypeData>): String {
```
sschuberth, line 308: This seems to be mostly the same as `determineTitle()`, can the code be de-duplicated?

sschuberth (changes requested):

core/src/main/kotlin/plugins/OpenApi.kt
mnonnenmacher: I didn't want to optimize the copied code, in case we have to integrate any upstream changes. But I also hope that the upstream issue will get fixed an we can delete it again.

sschuberth: Sorry, I missed that this *all* was copied code. We can keep it as-is then IMO, in the the hope to soon remove it again.

sschuberth: Should we then temporarily add "Copyright 2024 SMILEY4" to the header of this file?

mnonnenmacher: Yes, done.

sschuberth: Of course, this now fails the check for the expected license header 🙄

mnonnenmacher: I have changed the rule to allow that additional copyright statement and added a note to the commit message.

sschuberth (approved):

PR #981, opened 2024-09-05
ui: Use custom hooks for queries

### LINKED ISSUE #976: Deduplicate UI query options
Many components specify diplicate query options in the component and in the loader, like:

https://github.com/eclipse-apoapsis/ort-server/blob/be0c8973d76871f0f90c0d4ed9ec02b005948a87/ui/src/routes/_layout/organizations/%24orgId/products/%24productId/repositories/%24repoId/_layout/runs/%24runIndex/index.tsx#L47-L66

https://github.com/eclipse-apoapsis/ort-server/blob/be0c8973d76871f0f90c0d4ed9ec02b005948a87/ui/src/routes/_layout/organizations/%24orgId/products/%24productId/repositories/%24repoId/_layout/runs/%24runIndex/index.tsx#L220-L231

To reduce repetition and possibility for errors, these should be replaced with TanStack Query's `queryOptions()` like [this](https://tanstack.com/router/latest/docs/framework/react/guide/external-data-loading#a-more-realistic-example-using-tanstack-query).

Etsija:

Resolves #976.

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId/index.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/index.tsx
@@ -17,7 +17,6 @@
  * License-Filename: LICENSE
  */
 
-import { useSuspenseQueries } from '@tanstack/react-query';
 import { createFileRoute, Link, useNavigate } from '@tanstack/react-router';
 import {
   ColumnDef,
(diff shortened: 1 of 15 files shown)


ui/src/routes/_layout/organizations/$orgId/index.tsx
```diff
@@ -159,6 +143,14 @@ const OrganizationComponent = () => {
...
   });
 
+  if (orgError || prodError) {
+    return orgError || prodError;
+  }
+
+  if (!organization || !products) {
+    return <LoadingIndicator />;
+  }
+
```
mmurto, line 153: Would checking for `isError` and `isPending` be enough to assume that the data is available, as described [here](https://tanstack.com/query/latest/docs/framework/react/guides/queries#query-basics)?

ui/src/routes/_layout/organizations/$orgId/index.tsx
```diff
@@ -148,7 +132,7 @@ const OrganizationComponent = () => {
   const table = useReactTable({
     data: products?.data || [],
     columns,
-    pageCount: Math.ceil(products.pagination.totalCount / pageSize),
+    pageCount: Math.ceil(products?.pagination.totalCount ?? 0 / pageSize),
```
mmurto, line 135: As a follow-up I'd consider moving the table to a separate component, so the table hook is only executed when the data is available.

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId/index.tsx
Etsija: Could this be done in a follow-up PR? That check is needed there for now, until the tables are separated.

mmurto: Sure!

mmurto (approved):

PR #1002, opened 2024-09-09
ui: Fix pagination on tables

lamppu:

### DIFF
--- a/ui/src/components/toast-error.tsx
+++ b/ui/src/components/toast-error.tsx
@@ -20,7 +20,7 @@
 import { ApiError } from '@/api/requests';
 
 type ToastErrorProps = {
-  error: ApiError;
+  error: unknown;
 };
 
 type ErrorBody = {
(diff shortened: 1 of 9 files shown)


ui/src/components/toast-error.tsx
```diff
@@ -20,7 +20,7 @@
 import { ApiError } from '@/api/requests';
 
 type ToastErrorProps = {
-  error: ApiError;
+  error: unknown;
```
Etsija, line 23: It seems like in v5, TanStack Query has changed the default error type [from unknown to Error](https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#typescript-error-is-now-the-default-type-for-errors-instead-of-unknown), so just a heads up, this might change once the codegen library is updated.

Etsija:
As a general comment (and I sincerely recommend reading this excellent blog about TanStack Query): error reporting with localized toasts for all components which use queries leads to quite a lot of boilerplate code. We could maybe at some point simplify this with Error Boundaries and [global callbacks](https://tkdodo.eu/blog/react-query-error-handling#the-global-callbacks).

Etsija (changes requested):
Didn't you say you found a bug from the `DataTablePagination` component? Applying the fix for nullish coalescent operator still doesn't fix the step fwd/bwd buttons; for this page, both are still enabled, and the `<` button leads to page 0 (which doesn't do anything), and the `>` button to page 2, which is empty (as there are no more vulnerabilities). Referring to my picture, I think all pagination buttons should be disabled.

Capture1

lamppu:
> Didn't you say you found a bug from the `DataTablePagination` component? Applying the fix for nullish coalescent operator (as that's the only change in this PR that involves my page) still doesn't fix the step fwd/bwd buttons; for this page, both are still enabled, and the `<` button leads to page 0 (which doesn't do anything), and the `>` button to page 2, which is empty (as there are no more vulnerabilities). Referring to my picture, I think all pagination buttons should be disabled.
> 
> Related to this, maybe the mouse pointers for disabled pagination buttons could also be changed to the default one, instead of the "finger pointing to something which should do something".
> 

I did, but this will be fixed in https://github.com/eclipse-apoapsis/ort-server/pull/999 (the problem was that the Link component needs to be disabled as well). The mouse pointer problem will also go away when the Link component is disabled too.

ui/src/components/toast-error.tsx
lamppu: I don't think that that will be a problem, I think this will accept an error of type Error too.

Etsija (approved):

PR #1007, opened 2024-09-10
ui: Add basis for progress information of an ORT Run

Etsija:
Capture

### DIFF
--- a/ui/src/components/statistics-card.tsx
+++ b/ui/src/components/statistics-card.tsx
@@ -17,65 +17,37 @@
  * License-Filename: LICENSE
  */
 
-import { Link, useParams } from '@tanstack/react-router';
-
-import {
-  AdvisorJob,
-  AnalyzerJob,
-  EvaluatorJob,
-  ReporterJob,
-  ScannerJob,
-} from '@/api/requests';
 import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
 
 type StatisticsCardProps = {
-  job?:
-    | AnalyzerJob
-    | AdvisorJob
-    | ScannerJob
-    | EvaluatorJob
-    | ReporterJob
-    | undefined;
   title: string;
   icon?: React.ComponentType<{ className?: string }>;
   value?: number | string;
-  resultsPath: 'issues' | 'packages' | 'vulnerabilities' | 'rule-violations';
-  pollInterval: number;
+
+  className?: string;
 };
 
 export const StatisticsCard = ({
   title,
   icon: Icon,
   value,
-  resultsPath,
+  className,
 }: StatisticsCardProps) => {
-  const { orgId, productId, repoId, runIndex } = useParams({ strict: false });
   return (
-    <Link
-      from='/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex'
-      to={resultsPath}
-      params={{
-        orgId: orgId || '',
-        productId: productId || '',
-        repoId: repoId || '',
-        runIndex: runIndex || '',
-      }}
-    >
-      <Card className='hover:bg-muted/50'>
-        <CardHeader>
-          <CardTitle>
-            <div className='flex items-center justify-between'>
-              <span className='text-sm font-semibold'>{title}</span>
-              {Icon && <Icon />}
-            </div>
-          </CardTitle>
-        </CardHeader>
-        <CardContent className='text-sm'>
+    <Card className={className}>
+      <CardHeader>
+        <CardTitle>
           <div className='flex items-center justify-between'>
-            <div className='text-2xl font-bold'>{value ? value : '-'}</div>
+            <span className='text-sm font-semibold'>{title}</span>
+            {Icon && <Icon />}
           </div>
-        </CardContent>
-      </Card>
-    </Link>
+        </CardTitle>
+      </CardHeader>
+      <CardContent className='text-sm'>
+        <div className='flex items-center justify-between'>
+          <div className='text-2xl font-bold'>{value ? value : '-'}</div>
+        </div>
+      </CardContent>
+    </Card>
   );
 };
--- a/ui/src/components/statistics-card.tsx
+++ b/ui/src/components/statistics-card.tsx
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import { Link, useParams } from '@tanstack/react-router';
+
+import {
+  AdvisorJob,
+  AnalyzerJob,
+  EvaluatorJob,
+  ReporterJob,
+  ScannerJob,
+} from '@/api/requests';
+import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
+
+type StatisticsCardProps = {
+  job?:
+    | AnalyzerJob
+    | AdvisorJob
+    | ScannerJob
+    | EvaluatorJob
+    | ReporterJob
+    | undefined;
+  title: string;
+  icon?: React.ComponentType<{ className?: string }>;
+  value?: number | string;
+  resultsPath: 'issues' | 'packages' | 'vulnerabilities' | 'rule-violations';
+  pollInterval: number;
+};
+
+export const StatisticsCard = ({
+  title,
+  icon: Icon,
+  value,
+  resultsPath,
+}: StatisticsCardProps) => {
+  const { orgId, productId, repoId, runIndex } = useParams({ strict: false });
+  return (
+    <Link
+      from='/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex'
+      to={resultsPath}
+      params={{
+        orgId: orgId || '',
+        productId: productId || '',
+        repoId: repoId || '',
+        runIndex: runIndex || '',
+      }}
+    >
+      <Card className='hover:bg-muted/50'>
+        <CardHeader>
+          <CardTitle>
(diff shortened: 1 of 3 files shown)


mmurto:
I'd also remove the traffic lights from the top-left card if it's duplicated in the individual cards.

mmurto:
> I like that picture. Is the pulse icon there an indicator of the underlying job running or something else? Could be the traffic light, which I think the current users know already?

The pulse icon is from the linked example, I just changed the text in browser. Either the traffic light or the icon from the sidebar would probably make most sense.

mnonnenmacher:
> I'd also remove the traffic lights from the top-left card if it's duplicated in the individual cards.

Not directly related to this PR, but I am missing some basic information about the run when viewing other subpages, like reports or logs. How about moving the content of the top-left card with global id, timestamp and traffic lights to the top of the sidebar?

mmurto:
> > I'd also remove the traffic lights from the top-left card if it's duplicated in the individual cards.
> 
> Not directly related to this PR, but I am missing some basic information about the run when viewing other subpages, like reports or logs. How about moving the content of the top-left card with global id, timestamp and traffic lights to the top of the sidebar?

IMO the idea would be that the navigation bar's `repo / index` would provide enough context on what item the user is currently interacting with. Can you elaborate a little on what makes you miss the information you mentioned, what do you need them for?

mmurto:
A few comments on the visuals:

- To me the cards look a little tight and they look off as the padding doesn't match other cards - I'd got with a matching padding. There's nothing preventing from spreading the cards to two rows, they don't need to fit side-by-side.
- Would it look good if instead of the colored dot in the card, you'd change the icon color based on the status?
- I think the blue `Details ->` is too eye-catching for what it does. Could either try with the whole card being a link with a chevron icon or something in the corner, or the title having an icon next to it to tell that it's a link.

Etsija:
>     * To me the cards look a little tight and they look off as the padding doesn't match other cards - I'd got with a matching padding. There's nothing preventing from spreading the cards to two rows, they don't need to fit side-by-side.

OK that's easy enough, I was a little afraid that the cards weren't "tight" enough as-is.

>     * Would it look good if instead of the colored dot in the card, you'd change the icon color based on the status?

I'll have to try this out - might be that the chosen icons don't look so good with all status colors in dark and light mode. But it's a nice idea to declutter the cards a bit.

>     * I think the blue `Details ->` is too eye-catching for what it does. Could either try with the whole card being a link with a chevron icon or something in the corner, or the title having an icon next to it to tell that it's a link.

And do you agree that the card would only serve as a link for results that are available and non-zero? My thought was, no sense in tricking the user to click a card for which there are no results yet.

mmurto:
> And do you agree that the card would only serve as a link for results that are available and non-zero? My thought was, no sense in tricking the user to click a card for which there are no results yet.

I'm not sure. I think it may be safer to just have the link always lead there, as if it's 0, the user unlikely wants to click it, but if the user for some reason still wants to go there, they're annoyed that the link now doesn't work for some reason.

Etsija:
@mmurto pushed again, see what you think.

mmurto:
Prompted by there being a card for both Packages and Dependencies, I think we should really think what the user wants to see here. IMO the most important are rule violations, issues and vulnerabilities, so basically "how many items were found that require some action". The total number of dependencies is probably interesting information so it could be provided as well, total number of licenses found IMO not so much.

So I'd rework this to show those 4 cards. That of course means that they're not each linked to individual step of ORT run.

sschuberth:
> The total number of dependencies is probably interesting information so it could be provided as well

Agreed, though that should probably rather be the number of unique dependencies, i.e. the size of the de-duplicated list of packages.

> total number of licenses found IMO not so much.

Hmm, well, it depends I'd say. A high number of maybe > 100 different licenses could be seen as a quick indicator that you're probably scanning a license list for which you should configure an exclude.

sschuberth:
And a remark regarding the icons in the cards: The menu shows the icons to the left of the title, and I believe it should be the same for the cards. The chain / link icon I would omit completely in favor of making the most prominent / bug number of the card clickable.

mmurto:
> > The total number of dependencies is probably interesting information so it could be provided as well
> 
> Agreed, though that should probably rather be the number of unique dependencies, i.e. the size of the de-duplicated list of packages.

I agree.

> > total number of licenses found IMO not so much.
> 
> Hmm, well, it depends I'd say. A high number of maybe > 100 different licenses could be seen as a quick indicator that you're probably scanning a license list for which you should configure an exclude.

That's a use-case, but not sure if prominently showing the total amount of licenses found is the best way to do it. I'd rather create a rule that creates a warning if more than 10 licenses are detected in one file.

Etsija:
Issues icon (used a "bug" as in WebApp report) is now always red, and its corresponding "finding time" is not shown, as it's not related to an individual job.

Can we agree that this could be tweaked more on a later follow-up PR? I need to debug why the evaluator jobs are all failing now.

mmurto:
> Can we agree that the "graphical" part of this PR could be tweaked more on a later follow-up PR? I need to debug why the evaluator jobs are all failing now.

As the data is not there yet due to endpoints not existing there's really no rush to merge this PR, so you could just leave this as is as a draft.

Etsija:
> If you have multiple tabs open to compare different runs the only information you have to identify the run is the index. It would help with keeping the overview to have a little more context information visible on all screens, like the timestamp and the job statuses. The global run id is also useful when having the Loki dashboard open in another tab, at least as long as we don't have a log viewer in the UI.

Can you make a suggestion issue about this, describing exactly what you would want and where? Then we can review with the team and plan actions correspondingly.

Etsija:
I think I've tried to answer to all the layout change requests. Would it be too bad if we merge this now, as it's serving as a basis for quickly add status/results support for the (hopefully soon) upcoming new endpoints (issues, packages)?

mmurto:
I think the layout looks nicer if the card titles like "Issues" are left like in the other cards, even if that makes the icon be places differently than the sidebar.

ui/src/components/ort-run-info.tsx
```diff
@@ -0,0 +1,113 @@
...
+};
+
+export const OrtRunInfo = ({
+  job,
+  title,
+  icon: Icon,
+  nTotal,
+  resultsPath,
+  pollInterval,
+}: OrtRunInfoProps) => {
```
mmurto, line 59: I would refactor this component into a more generic StatisticsCard or something that takes in a title, icon, value and maybe something else if required.

mmurto (changes requested):

Etsija:
> I think the layout looks nicer if the card titles like "Issues" are left like in the other cards, even if that makes the icon be places differently than the sidebar.

They were, but I already changed it as per Sebastian's request.

ui/src/components/ort-run-info.tsx
Etsija: Maybe later, but I was hoping we could merge this now, without more time-consuming changes, so that I can start working on the groups admin issue.

mmurto: Not just renaming, it shouldn't take in job or anything that specific.

mnonnenmacher:
> Can you make a suggestion issue about this, describing exactly what you would want and where? Then we can review with the team and plan actions correspondingly.

Done: #1039

ui/src/components/statistics-card.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
+
+type StatisticsCardProps = {
+  job?:
+    | AnalyzerJob
+    | AdvisorJob
+    | ScannerJob
+    | EvaluatorJob
+    | ReporterJob
+    | undefined;
```
mmurto, line 38: IMO this should be a generic statistics card that does not take a job, resultpath or pollInterval in.

ui/src/components/statistics-card.tsx
```diff
@@ -0,0 +1,81 @@
...
+  pollInterval: number;
+};
+
+export const StatisticsCard = ({
+  title,
+  icon: Icon,
+  value,
+  resultsPath,
+}: StatisticsCardProps) => {
+  const { orgId, productId, repoId, runIndex } = useParams({ strict: false });
```
mmurto, line 52: Nor should it use useParams, to allow using it inside any route.

mmurto (changes requested):

ui/src/components/statistics-card.tsx
Etsija: You mean it should just take in the `to` as an absolute path?

mmurto: I think it should be a generic card that may or may not be a link, so I'd move the link outside of the component.

Etsija: I want to keep the link inside the component for now, to keep the DOM consistent; otherwise, also Card needs to be taken out of the component to retain the hovering action for link and non-link cards.  

If we need non-link statistics cards like this later on, we could change this.

mmurto: > I want to keep the link inside the component for now, to keep the DOM consistent; otherwise, also Card needs to be taken out of the component to retain the hovering action for link and non-link cards.

You can take in `className` as props like [other components](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/components/ui/card.tsx#L14-L27) do, and then use the `hover:bg-color` at call site.

mmurto: > I want to keep the link inside the component for now, to keep the DOM consistent; otherwise, also Card needs to be taken out of the component to retain the hovering action for link and non-link cards.

You can take in `className` as props like [other components](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/components/ui/card.tsx#L14-L27) do, and then use the `hover:bg-color` at call site.

ui/src/components/statistics-card.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export const StatisticsCard = ({
+  title,
+  icon: Icon,
+  value,
+  resultsPath,
+}: StatisticsCardProps) => {
+  return (
+    <Link to={resultsPath} className={resultsPath ? '' : 'cursor-default'}>
+      <Card className={resultsPath ? 'hover:bg-muted/50' : ''}>
```
mmurto, line 39: I'd do it with the `className` as props approach I suggested, as this renders a link even if `resultsPath` is not given.

mmurto (changes requested):

ui/src/components/statistics-card.tsx
Etsija: So to recap: move Link to parent page, then use className inside the component to cover for hovering action when it's a link?

mmurto: That's how I'd do it, but feel free to do it the other way if that makes more sense to you. If the component works as a generic link, it should expose `params` and `search` in addition to `to`.

Etsija: @mmurto redid it with link removed, would this now be ready to merge?

mmurto (approved):

PR #1008, opened 2024-09-10
fix(logaccess): Prevent a potential endless loop

wkl3nk:
Given Loki is used as log aggregation system, when multiple log statements are generated in quick succession (burst), these log statements potentially all carry the same timestamp. When the Loki HTTP API is used to fetch the log statements for a given time interval using chunks defined by a start timestamp, this leads to an endless loop in the ORT server if the chunk size is smaller than the number of log statements with identical timestamps. This bugfix prevents this endless loop by making sure that each chunk is fetched at most a single time.

### DIFF
--- a/logaccess/loki/src/main/kotlin/LokiLogFileProvider.kt
+++ b/logaccess/loki/src/main/kotlin/LokiLogFileProvider.kt
@@ -104,7 +104,18 @@ class LokiLogFileProvider(
                 }
 
                 if (statements.size >= config.limit && deDuplicatedStatements.isNotEmpty()) {
-                    downloadChunk(statements.last().timestamp, deDuplicatedStatements)
+                    val lastTimestamp = statements.last().timestamp
+
+                    if (lastTimestamp != from) {
+                        downloadChunk(lastTimestamp, deDuplicatedStatements)
+                    } else {
+                        logger.error(
+                            "Possible loss of log data for run '{}' and source '{}' " +
+                            "due to number of identical timestamps exceeds chunk size limit.",
+                            ortRunId,
+                            source
+                        )
+                    }
                 }
             }
 
--- a/logaccess/loki/src/test/kotlin/LokiLogFileProviderTest.kt
+++ b/logaccess/loki/src/test/kotlin/LokiLogFileProviderTest.kt
@@ -47,9 +47,12 @@ import io.mockk.mockk
 import java.io.File
 import java.util.EnumSet
 import java.util.Locale
+import java.util.UUID
 
 import kotlin.time.Duration.Companion.milliseconds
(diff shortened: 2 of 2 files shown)


logaccess/loki/src/main/kotlin/LokiLogFileProvider.kt
```diff
@@ -104,7 +104,18 @@ class LokiLogFileProvider(
                 }
 
                 if (statements.size >= config.limit && deDuplicatedStatements.isNotEmpty()) {
-                    downloadChunk(statements.last().timestamp, deDuplicatedStatements)
+                    val lastTimestamp = statements.last().timestamp
```
mnonnenmacher, line 107: Commit message nit: Typo in "**t**his leads"

logaccess/loki/src/main/kotlin/LokiLogFileProvider.kt
```diff
@@ -104,7 +104,18 @@ class LokiLogFileProvider(
...
+                    if (lastTimestamp != from) {
+                        downloadChunk(lastTimestamp, deDuplicatedStatements)
+                    } else {
+                        logger.warn(
+                            "Possible loss of log data for run '{}' and source '{}' " +
+                            "due to number of identical timestamps exceeds chunk size limit.",
+                            ortRunId,
+                            source
+                        )
+                    }
```
mnonnenmacher, line 118: As a quick fix this is good enough, but I think we also need a proper solution that does not have the risk of losing log data. An inefficient solution could be to reload the chunk with a larger size until the end timestamp does not equal the start timestamp anymore.
But I also wonder why chunking was implemented in the first place instead of just downloading the full logs.

logaccess/loki/src/main/kotlin/LokiLogFileProvider.kt
```diff
@@ -104,7 +104,18 @@ class LokiLogFileProvider(
                 }
 
                 if (statements.size >= config.limit && deDuplicatedStatements.isNotEmpty()) {
-                    downloadChunk(statements.last().timestamp, deDuplicatedStatements)
+                    val lastTimestamp = statements.last().timestamp
+
+                    if (lastTimestamp != from) {
+                        downloadChunk(lastTimestamp, deDuplicatedStatements)
+                    } else {
+                        logger.warn(
```
mnonnenmacher, line 112: Should we make this an error because it is much more likely that log lines got lost than that the lines with the same timestamp match the chunk size exactly?

wkl3nk: Yes, I now made it an error in the log.

mnonnenmacher (approved):

PR #1021, opened 2024-09-11
Add a workflow to test the website build

mnonnenmacher:

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -54,28 +54,17 @@ jobs:
     - name: Generate OpenAPI spec
       run: ./gradlew --stacktrace :core:generateOpenApiSpec
 
+    - name: Install pnpm
+      uses: pnpm/action-setup@v4
+      with:
+        version: 9
+
     - name: Install Node
       uses: actions/setup-node@v4
       with:
         node-version: 20
-
-    - uses: pnpm/action-setup@v4
-      name: Install pnpm
-      with:
-        version: 9
-
-    - name: Get pnpm store directory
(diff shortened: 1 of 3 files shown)


.github/workflows/website-test.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Install pnpm
+      uses: pnpm/action-setup@v4
+      with:
+        version: 9
+
+    - name: Install Node
+      uses: actions/setup-node@v4
+      with:
+        node-version: 20
+        cache: pnpm
```
sschuberth, line 24: Shouldn't "Install Node" go before "Install pnpm" to make the cache work?

.github/workflows/build-and-test.yml
```diff
@@ -54,15 +54,17 @@ jobs:
     - name: Generate OpenAPI spec
       run: ./gradlew --stacktrace :core:generateOpenApiSpec
 
+    - name: Install pnpm
```
sschuberth, line 57: Interesting, this is doing the opposite to what I was mentioning in the previous commit. Maybe better link to https://github.com/actions/setup-node?tab=readme-ov-file#usage, quoting "Package manager should be pre-installed" in the commit message?

mmurto (changes requested):
The UI workflows currently handle caching a few steps later:

https://github.com/eclipse-apoapsis/ort-server/blob/3a1eb258cad79922cf6acf26fd2d2c5db09a9e7d/.github/workflows/build-and-test.yml#L67-L78

If the caching is handled by the setup-node action, this should probably be removed.

PR #1024, opened 2024-09-12
fix(ui): Fix line breaks in vulnerability descriptions

Etsija:
Capture

@mnonnenmacher this is an easy & quick first fix for showing the vulnerability descriptions.

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/vulnerabilities/index.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/vulnerabilities/index.tsx
@@ -151,7 +151,9 @@ const renderSubComponent = ({
     <div className='flex flex-col gap-2'>
       <div className='italic text-muted-foreground'>
         <div className='font-semibold'>Description:</div>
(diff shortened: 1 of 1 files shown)


mmurto:
A small suggestion while modifying the component: IMO it would make sense for the references to be visible when you open the details, as it already takes a lot of spaces and I don't think there's any reason to have to click twice to see everything.

mmurto (dismissed):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/vulnerabilities/index.tsx
```diff
@@ -151,9 +151,16 @@ const renderSubComponent = ({
...
+          {row.original.vulnerability.description}
+        </div>
       </div>
-      <Accordion type='single' collapsible className='ml-4'>
+      <Accordion
+        type='single'
+        collapsible
+        className='ml-4'
+        defaultValue='references'
+      >
```
mmurto, line 163: It should probably not be an accordion if it's open by default, just show them if the main thing is open. Should also be a separate commit.

mmurto (changes requested):

mmurto (approved):

PR #1030, opened 2024-09-13
ui: Get rid of violating props passed to DOM elements

### LINKED ISSUE #1028: React does not recognize the X prop on a DOM element
In [this commit](https://github.com/eclipse-apoapsis/ort-server/commit/bf74b0b65722321ec37a0b9e40e4038c86ed90a3), `DataTablePagination` was changed to use props instead of table context, to provide type safety for links. This creates a lot of React warnings in dev console for every page:

_Warning: React does not recognize the `X` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `X` instead. If you accidentally passed it from a parent component, remove it from the DOM element._
```
a
Link<@http://localhost:5173/node_modules/.vite/deps/chunk-535XYBFZ.js?v=1d5fd888:3583:35
div
div
DataTablePagination@http://localhost:5173/src/components/data-table/data-table-pagination.tsx:34:36
div
div
DataTable@http://localhost:5173/src/components/data-table/data-table.tsx:31:26
div
```

@mmurto any idea to fix this?  [This seems to be related](https://stackoverflow.com/questions/54468535/how-to-solve-warning-react-does-not-recognize-the-x-prop-on-a-dom-element), and the props in the warnings list router provider's props as well as the `setPageSizeOptions` and `setCurrentPageOptions` of the `DataTablePagination` component. Seems as if the router provider's props are "leaking" down to the pagination component.

Etsija:

Resolves #1028.

### DIFF
--- a/ui/src/components/data-table/data-table.tsx
+++ b/ui/src/components/data-table/data-table.tsx
@@ -49,6 +49,8 @@ export function DataTable<TData>({
   renderSubComponent,
   children,
   className,
+  setCurrentPageOptions,
+  setPageSizeOptions,
   ...props
 }: DataTableProps<TData>) {
   const pagination = table.getState().pagination;
@@ -137,8 +139,8 @@ export function DataTable<TData>({
             currentPage={pagination.pageIndex + 1}
             pageSize={pagination.pageSize}
             totalPages={totalPages}
-            setCurrentPageOptions={props.setCurrentPageOptions}
-            setPageSizeOptions={props.setPageSizeOptions}
+            setCurrentPageOptions={setCurrentPageOptions}
+            setPageSizeOptions={setPageSizeOptions}
           />
         </div>
       )}
--- a/ui/src/routes/_layout/organizations/$orgId/index.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/index.tsx
@@ -227,13 +227,13 @@ const OrganizationComponent = () => {
             table={table}
             setCurrentPageOptions={(currentPage) => {
               return {
-                ...Route,
+                to: Route.to,
                 search: { ...search, page: currentPage },
               };
             }}
             setPageSizeOptions={(size) => {
               return {
-                ...Route,
+                to: Route.to,
                 search: { ...search, pageSize: size },
               };
             }}
--- a/ui/src/routes/_layout/index.tsx
+++ b/ui/src/routes/_layout/index.tsx
@@ -123,13 +123,13 @@ export const IndexPage = () => {
             table={table}
             setCurrentPageOptions={(currentPage) => {
               return {
-                ...Route,
+                to: Route.to,
                 search: { ...search, page: currentPage },
               };
             }}
             setPageSizeOptions={(size) => {
               return {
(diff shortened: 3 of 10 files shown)


ui/src/components/data-table/data-table.tsx
```diff
@@ -53,11 +53,15 @@ export function DataTable<TData>({
 }: DataTableProps<TData>) {
   const pagination = table.getState().pagination;
   const totalPages = table.getPageCount();
+  // We need to extract the violating props from the props object
```
sschuberth, line 56: ORT projects usually follow the convention to use passive voice in comments, commit messages etc. so you shoud better say "Extract violating properties ...". Also in pain text, "props" should be spelled out as "properties" unless it refers directly to the variable name.

ui/src/components/data-table/data-table.tsx
```diff
@@ -53,11 +53,15 @@ export function DataTable<TData>({
...
+  // without using them in the component.
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { setPageSizeOptions, setCurrentPageOptions, ...restProps } = props;
```
sschuberth, line 59: As "rest" sounds more like "sleep", maybe better "remainingProps"?

ui/src/components/data-table/data-table.tsx
```diff
@@ -49,16 +49,14 @@ export function DataTable<TData>({
...
 }: DataTableProps<TData>) {
   const pagination = table.getState().pagination;
   const totalPages = table.getPageCount();
 
   return (
-    <div
-      className={cn('w-full space-y-2.5 overflow-auto', className)}
-      {...props}
-    >
+    <div className={cn('w-full space-y-2.5 overflow-auto', className)}>
```
mmurto, line 59: You should keep the {...props} here and above to allow for all div props to be given to the wrapping div.

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId/index.tsx
```diff
@@ -227,13 +227,13 @@ const OrganizationComponent = () => {
             table={table}
             setCurrentPageOptions={(currentPage) => {
               return {
-                ...Route,
+                to: Route.to,
```
mmurto, line 230: Just to make sure, you did try that this works on all pages without passing `params`? I have no doubt that it doesn't work on the organization page at `/`, but does it work for products, repos and runs as well, as those pages do have params?

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId/index.tsx
Etsija: Yes, I tried it on all pages, it works without params.

mmurto (approved):

ui/src/components/data-table/data-table.tsx
sschuberth: Ok, then let's stick with that...

PR #1034, opened 2024-09-13
Deduplicate packages in the database

mnonnenmacher:
Add a maintenance job to deduplicate packages in the database. See the commit messages for details.


services/maintenance/src/main/kotlin/MaintenanceService.kt
```diff
@@ -0,0 +1,194 @@
...
+
+            logger.info("All maintenance jobs have been completed.")
+        }
+    }
+
+    /**
+     * Create entries in the database for all jobs that are not yet present.
+     */
+    private fun createJobs() {
+        db.blockingQuery(transactionIsolation = Connection.TRANSACTION_SERIALIZABLE) {
```
oheger-bosch, line 126: `createJobs` accesses `jobs`, and therefore also needs to use the mutex.

services/maintenance/src/main/kotlin/MaintenanceJob.kt
```diff
@@ -0,0 +1,43 @@
...
+
+package org.eclipse.apoapsis.ortserver.services.maintenance
+
+import org.eclipse.apoapsis.ortserver.model.MaintenanceJobData
+import org.eclipse.apoapsis.ortserver.utils.logging.withMdcContext
+
+abstract class MaintenanceJob {
+    abstract val name: String
+
+    var active = false
```
oheger-bosch, line 28: I doubt that it is thread-safe to access the field like that without further synchronization.

core/src/main/kotlin/plugins/Lifecycle.kt
```diff
@@ -50,6 +54,14 @@ fun Application.configureLifecycle() {
                 syncRoles(authorizationService)
             }
         }
+
+        thread {
+            MDC.setContextMap(mdcContext)
```
oheger-bosch, line 59: Question: Will this be removed again when the job is done?

oheger-bosch (commented):
I guess, this will work for the problem at hand (although the deduplication of projects is not addressed).
However, I am a bit uneasy with introducing this new mechanism in such a way without a clear design. We should be prepared that the whole code might be thrown away in the future.

core/src/main/kotlin/plugins/Lifecycle.kt
mnonnenmacher: It's stored in a `ThreadLocal` so it will die with the thread.

oheger-bosch: Sorry, what I meant was the whole block to add the job to the maintenance service. When the migration is complete it is not needed any more. If the job is started, it will probably still do some (minor) background activity.
I guess, the underlying problem here is that the minimum maintenance service implementation does not provide a mechanism to add/remove jobs dynamically; therefore, starting of jobs has to be hard-coded.

oheger-bosch (dismissed):

sschuberth:
> Will be replaced by https://github.com/eclipse-apoapsis/ort-server/pull/1073

So should this be closed now, @mnonnenmacher?

mnonnenmacher:
> > Will be replaced by #1073
> 
> So should this be closed now, @mnonnenmacher?

I would like to keep this open to act as a template, because we will have to implement some other maintenance tasks, for example, to move the Keycloak sync out of core.

PR #1049, opened 2024-09-16
Add PUT to the allowed methods in CORS config

mnonnenmacher:


sschuberth (approved):

Etsija:
I also verified locally that this indeed solves the problem.

Issue #1058, opened 2024-09-17
Access token becomes too long when too many roles added to a user


Etsija:
When testing the Groups endpoints, by setting several (5) groups rights to one user, I noticed organizations endpoint returns a 400 Bad Request. Investigating further, I noticed the authorization header cannot be sent anymore.

1. Adding only one groups right to a user: auth header is sent, connection OK, UI works OK.

![Capture2](https://github.com/user-attachments/assets/cc0b07f7-94ae-4393-9e84-2c56cd877778)
![Capture1](https://github.com/user-attachments/assets/ecef037f-b553-44ee-9e95-7b47e6124635)

2. Adding five different groups rights to a user: auth header is _empty_, 400 Bad Request.

![Capture6](https://github.com/user-attachments/assets/4dae59db-5f9b-450b-908d-7e2dfcf49af6)
![Capture3](https://github.com/user-attachments/assets/bbcb62e2-7d4c-45ea-953e-4af1b3e6e245)
![Capture4](https://github.com/user-attachments/assets/810194f0-9793-4d0e-81cc-09529a178aca)
![Capture5](https://github.com/user-attachments/assets/cc3c0f5d-2232-4df1-823c-a4c6f7737820)

@mnonnenmacher Johanna recalls that you were saying you've found (and possibly fixed) this problem internally, so could you please advise? My hunch is, if the groups roles are part of the auth token, it becomes too long and fails to be included in the requests.

We have currently no other means to overcome this problem but 
- delete all but one user's groups rights from Keycloak
- delete cookies and session storage from browser
- then the user can log in again

As to the limit of groups the user can currently be attached to, I don't have the exact answer; I only know one group is OK, and five is not.

Etsija:
To clarify: the question is _not_ how many groups you grant access to; it is rather how many roles are attached to those groups. I'm using a seeded repository in Docker Compose environment, which has a lot of products randomly seeded for each organization, and many repositories under those products.

Example: by just giving READERS access rights to three organizations, Keycloak has 270 separate roles attached to those three groups. This is probably the reason why so low number of groups access rights makes the access token too long.

mnonnenmacher:
@Etsija Let me have a look at the Keycloak config in Compose. In general, we should refactor the way Keycloak is used (see #984) but that is not going to happen soon.

mnonnenmacher:
@Etsija Could you try with the changes from #1064? With this the roles from the ort-server client are not mapped to the access token anymore.

[...]

PR #1064, opened 2024-09-17
Compose: Do not map client roles to access token

mnonnenmacher:

### DIFF
--- a/scripts/docker/keycloak/master-realm.json
+++ b/scripts/docker/keycloak/master-realm.json
@@ -868,6 +868,54 @@
       }
     } ]
   }, {
+    "id" : "61da6eac-c833-49a8-ab87-cc0ddf9662f2",
+    "name" : "roles",
+    "description" : "OpenID Connect scope for add user roles to the access token",
+    "protocol" : "openid-connect",
+    "attributes" : {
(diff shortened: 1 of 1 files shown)


Etsija:
I can verify that this resolves the issue #1058.

Verification steps:

1. As the PR only touches `master-realm.json`, I checked out the PR but used existing Docker images: `docker compose pull && docker compose up -d`.

2. I verified in Keycloak that the `userinfo roles` are not added to the access token:

Capture4

3. I created a new user in Keycloak and attached it to all 11 organizations in my seeded local db as a reader:

Capture1

4. This created a whopping number of ~1600 different roles to the user:

Capture2

5. I verified that the authorization header is still sent, and the UI works:

Capture5
Capture6

As a side note, this does seem to slow down the UI a bit: the request for all organizations in the main index page takes about 1.5 seconds - earlier on, it was very fast. It must be noted however that my seeded local db is quite large, as it includes 11 organizations, (randomly chosen) 0-15 products per organization, (randomly chosen) 0-15 repositories per product, and (randomly chosen) 0-15 secrets per entity. As we are using efficient query caching in front-end, this initial delay in fetching the items is quite acceptable.

mmurto (approved):

PR #1072, opened 2024-09-19
Add endpoint to get packages found in an ORT run

lamppu:
Please see the individual commit messages for details.

### DIFF
--- a/scripts/requests/runs.http
+++ b/scripts/requests/runs.http
@@ -1,3 +1,7 @@
 ### Get all vulnerabilities for an ORT run
 GET {{host}}/runs/1/vulnerabilities
 Authorization: Bearer {{$auth.token("keycloak")}}
+
+### Get all packages for an ORT run
+GET {{host}}/runs/1/packages
+Authorization: Bearer {{$auth.token("keycloak")}}
--- a/services/hierarchy/src/main/kotlin/PackageService.kt
+++ b/services/hierarchy/src/main/kotlin/PackageService.kt
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.services
+
+import org.eclipse.apoapsis.ortserver.dao.dbQuery
+import org.eclipse.apoapsis.ortserver.dao.tables.AnalyzerJobsTable
+import org.eclipse.apoapsis.ortserver.dao.tables.runs.analyzer.AnalyzerRunsTable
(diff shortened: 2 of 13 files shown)


services/hierarchy/src/main/kotlin/PackageService.kt
```diff
@@ -0,0 +1,341 @@
...
+val vcsProcessedTable = VcsInfoTable.alias("vcs2")
+
+/**
+ * A service to interact with packages.
+ */
+class PackageService(private val db: Database) {
+    suspend fun listForOrtRunId(
+        ortRunId: Long,
+        parameters: ListQueryParameters = ListQueryParameters.DEFAULT
+    ): ListQueryResult<Package> = db.dbQuery {
```
mnonnenmacher, line 71: I haven't tested it, but I think the implementation could be simplified by using the DAO API:

```kotlin
val query = PackagesTable
    .innerJoin(PackagesAnalyzerRunsTable)
    .innerJoin(AnalyzerRunsTable)
    .innerJoin(AnalyzerJobsTable)
    .select(PackagesTable.columns)
    .where { AnalyzerJobsTable.ortRunId eq ortRunId }
    .withDistinct()

val packages = PackageDao.wrapRows(query).map { it.mapToModel() }
```

The `mapToModel()` function already has the code to load all related data.

The only thing missing here is applying the pagination and sorting to the query. The code for that you could take from the `SizedIterable<T>.apply` helper function in `utils/Extensions.kt` in the `dao` module.

This is probably a little less efficient as, for example, the data from the `remote_artifacts` table is loaded in separate queries, but as long as the `limit' is not very large this should not matter (and you already do it for the authors and processed declared license anyway).

Now that I think of it, the code for the vulnerabilities endpoint could probably be simplified in the same way.

services/hierarchy/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -0,0 +1,510 @@
...
+                                "path"
+                            ),
+                            isMetadataOnly = false,
+                            isModified = false
+                        )
+                    ),
+                ).id
+
+                val results = service.listForOrtRunId(ortRunId).data
+
```
mnonnenmacher, line 148: You could define variables for the two packages used above and the just compare the result with `service.listForOrtRunId(ortRunId).data should containExactlyInAnyOrder(pkg1, pkg2)`. It will be harder to read the diff if the test fails, but it requires a lot less code.

services/hierarchy/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -0,0 +1,510 @@
...
+        val analyzerJob = fixtures.createAnalyzerJob(
+            ortRunId = ortRun.id,
+            configuration = AnalyzerJobConfiguration(),
+        )
+
+        fixtures.analyzerRunRepository.create(
+            analyzerJobId = analyzerJob.id,
+            startTime = Clock.System.now().toDatabasePrecision(),
+            endTime = Clock.System.now().toDatabasePrecision(),
```
mnonnenmacher, line 491: You should not have to call `toDatabasePrecision()` here.

scripts/requests/runs.http
```diff
@@ -1,3 +1,7 @@
 ### Get all vulnerabilities for an ORT run
 GET {{host}}/runs/1/vulnerabilities
 Authorization: Bearer {{$auth.token("keycloak")}}
+
+### Get all packages for an ORT run
+GET {{host}}/runs/1/packages
+Authorization: Bearer {{$auth.token("keycloak")}}
```
mnonnenmacher, line 7: Add the missing line break at the end.

mnonnenmacher (changes requested):

services/hierarchy/src/main/kotlin/PackageService.kt
mnonnenmacher: I have applied this approach to the `VulnerabilityService` in #1078 and could remove a lot of code.

lamppu: Ok good thanks, I'll try similar changes here.

mnonnenmacher (approved):

PR #1073, opened 2024-09-19
fix(dao): Add a migration to drop duplicate packages and projects

oheger-bosch:


dao/src/test/kotlin/tables/PackageDeduplicationMigrationTest.kt
```diff
@@ -0,0 +1,385 @@
...
+import org.eclipse.apoapsis.ortserver.model.runs.RemoteArtifact
+import org.eclipse.apoapsis.ortserver.model.runs.VcsInfo
+
+import org.jetbrains.exposed.sql.insert
+
+class PackageDeduplicationMigrationTest : StringSpec() {
+    val extension = extension(DatabaseMigrationTestExtension("71", "72"))
+
+    init {
+        "package migration should deduplicate packages" {
```
mnonnenmacher, line 72: I find the test hard to understand because all situations are tested in a single test case. This makes it difficult to figure out which scenarios are actually tested (e.g., compared to https://github.com/boschglobal/ort-server/blob/40ac248888f421b66bc4f84a5aeec66a120f8e94/services/maintenance/src/test/kotlin/jobs/DeduplicatePackagesJobTest.kt#L147).

Also, migration tests should be in the `migration` package and be named like the migration to easily match them (see the other tests in the package). And they should not use any of the existing Exposed tables or DAOs. This is because this code will change with future migration, but has to stay constant for the migration test. Instead, they should be copied like in the other migration tests.

However, I would be fine with addressing any of those test changes in a follow-up PR.

dao/src/main/resources/db/migration/V72__deduplicatePackagesAndProjects.sql
```diff
@@ -0,0 +1,610 @@
...
+  NOT EXISTS (
+  SELECT
+    1
+  FROM
+    project_scopes ps
+  WHERE
+    ps.project_id = p.id
+);
+
+-- A view that assigns each project a has value computed from its properties and relations.
```
mnonnenmacher, line 332: Typo: "hash value"

dao/src/main/resources/db/migration/V72__deduplicatePackagesAndProjects.sql
```diff
@@ -0,0 +1,610 @@
...
+  SELECT
+    1
+  FROM
+    projects_authors pa
+  WHERE
+    pa.project_id = p.id
+);
+
+-- A view with the scopes assigned to a project as a single string.
+CREATE VIEW projects_with_scopes AS
```
mnonnenmacher, line 308: Just to double check, the dependency graphs are stored as part of the analyzer result and only loosely associated to the projects via the scope name, so they can be ignored here, right?

dao/src/main/resources/db/migration/V72__deduplicatePackagesAndProjects.sql
```diff
@@ -0,0 +1,610 @@
...
+  NOT EXISTS (
+  SELECT
+    1
+  FROM
+    packages_authors pa
+  WHERE
+    pa.package_id = p.id
+);
+
+-- A view that assigns each package a has value computed from its properties and relations.
```
mnonnenmacher, line 184: Typo: "hash value"

mnonnenmacher (changes requested):
Could you include the commit https://github.com/eclipse-apoapsis/ort-server/pull/1034/commits/24c038477b6f55b96edb117024c624b8f7e5c6b4 in this PR and change the version of your migration to V73?
Adding the index is good anyway, and this would help with updating our dev environment where the migration was already applied.

dao/src/main/resources/db/migration/V72__deduplicatePackagesAndProjects.sql
mnonnenmacher, line 1: Commit message: Add a bit more context (you could copy some information from the commit message of https://github.com/eclipse-apoapsis/ort-server/pull/1034/commits/cb62da8925b1ed855db37f81b04b461fbb4e495b) and maybe also mention that this migration while take a while to run for larger databases and therefore any liveness probes might have to be adapted.

dao/src/test/kotlin/tables/PackageDeduplicationMigrationTest.kt
oheger-bosch: Renamed the class and moved it into the `migration` package.

dao/src/main/resources/db/migration/V72__deduplicatePackagesAndProjects.sql
oheger-bosch: Yes, also a dependency graph contains all projects in the repository; so it cannot be associated with a single project.

dao/src/main/resources/db/migration/V73__deduplicatePackagesAndProjects.sql
mnonnenmacher, line 1: Commit message nit: Linebreaks are odd.

mnonnenmacher (changes requested):

dao/src/main/resources/db/migration/V73__deduplicatePackagesAndProjects.sql
oheger-bosch: Sorry, still a bit fighting with the new IntelliJ UI.

mnonnenmacher (approved):

PR #1080, opened 2024-09-20
Store skipExcluded in the database

mnonnenmacher:


Etsija:
Could you hold on to this for a while, as I need to check one theory about the `repositoryConfigPath`?

mnonnenmacher:
> Could you hold on to this for a while, as I need to check one theory about the `repositoryConfigPath`?

Do you think is related to this change? This fix is necessary anyway, it only affects the way results are stored, not any logic that actually applies `skipExcluded`.

sschuberth (dismissed):

Etsija:
Tests fail constantly due to 
```
org.jetbrains.exposed.sql.statements.BatchDataInconsistentException: Can't add a new batch because columns: analyzer_configurations.skip_excluded don't have default values. DB defaults are not supported in batch inserts
```

sschuberth:
Now `V73__deduplicatePackagesAndProjectsTest > package migration should deduplicate packages` [fails](https://github.com/eclipse-apoapsis/ort-server/actions/runs/10989889069/job/30509052198?pr=1080#step:5:1994) with

    java.sql.BatchUpdateException: Batch entry 0 INSERT INTO analyzer_configurations (allow_dynamic_versions, analyzer_run_id, skip_excluded) VALUES (('FALSE'::boolean), ('3'::int8), ('FALSE'::boolean))
    RETURNING * was aborted: ERROR: column "skip_excluded" of relation "analyzer_configurations" does not exist

dao/src/main/kotlin/tables/runs/scanner/ScannerConfigurationsTable.kt
```diff
@@ -33,7 +33,7 @@ import org.jetbrains.exposed.dao.id.LongIdTable
 object ScannerConfigurationsTable : LongIdTable("scanner_configurations") {
     val scannerRunId = reference("scanner_run_id", ScannerRunsTable)
 
-    val skipConcluded = bool("skip_concluded").default(false)
+    val skipConcluded = bool("skip_concluded")
```
sschuberth, line 36: Commit message:
- Typo in "unnnecessary".
- Maybe say "Remove unnecessary `default` *calls*"

sschuberth (changes requested):

sschuberth (approved):

dao/src/test/kotlin/migrations/V73__deduplicatePackagesAndProjectsTest.kt
```diff
@@ -70,8 +70,12 @@ import org.jetbrains.exposed.sql.insert
 class V73__deduplicatePackagesAndProjectsTest : StringSpec() {
     val extension = extension(DatabaseMigrationTestExtension("72", "73"))
 
+    // TODO: The tests in this class are disabled because they are incompatible with a later migration. To enable them
```
oheger-bosch, line 73: We could even consider to fully delete this test class. It is highly unlikely that it will be reworked. It has fulfilled its purpose and is no longer needed - keeping it will only cause maintenance effort.

Maybe this could be a general pattern for dealing with tests for migrations. Forcing them to not use any DAO classes rather complicates creating such tests.

PR #1086, opened 2024-09-23
fix(reporter): Do not crash the reporter if there is no analyzer result

nnobelis:
If the Analyzer worker has failed, the Reporter fails as well because it tries to obtain VCS information which is not available. The generated notifications then state that the run failed in the phases Analyzer and Reporter. Because this is misleading, the Reporter worker should handle this case gracefully, i.e. go to the `FINISHED_WITH_ISSUES` state instead.

### DIFF
--- a/workers/reporter/src/test/kotlin/reporter/ReporterWorkerTest.kt
+++ b/workers/reporter/src/test/kotlin/reporter/ReporterWorkerTest.kt
@@ -64,6 +64,7 @@ import org.eclipse.apoapsis.ortserver.workers.common.env.EnvironmentService
 import org.eclipse.apoapsis.ortserver.workers.common.mapToOrt
 
 import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.Repository
 
 private const val ORT_SERVER_MAPPINGS_FILE = "org.eclipse.apoapsis.ortserver.workers.common.OrtServerMappingsKt"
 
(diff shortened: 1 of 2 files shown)


sschuberth:
Commit message: The title "Reporter should not crash if there is no Analyzer result" sounds like an issue description for a bug that is yet to be fixed. Better say "Do not crash the reporter if there is no analyzer result".

workers/reporter/src/test/kotlin/reporter/ReporterWorkerTest.kt
```diff
@@ -343,4 +346,45 @@ class ReporterWorkerTest : StringSpec({
...
+        mockkTransaction {
+            val result = worker.run(REPORTER_JOB_ID, TRACE_ID)
+
+            result shouldBe RunResult.FinishedWithIssues
+        }
+
+        val slotReporterRun = slot<ReporterRun>()
+        coVerify {
+            ortRunService.storeReporterRun(capture(slotReporterRun))
+        }
```
oheger-bosch, line 388: Was it missed to do some checks on the reporter run captured by the slot? Otherwise, it does not make sense to use `capture` here.

nnobelis: Fixed, it was a copy paste error.

oheger-bosch (approved):

PR #1102, opened 2024-09-25
ci: Add a workflow to run ORT

mnonnenmacher:
Add a workflow to run ORT on the ORT Server. This will later be included in the release workflow to generate an SBOM and probably also in PR checks.

To be able to manually trigger a workflow, it must first be merged to the main branch. Therefore, only include the basics in this commit to download the image and run the requirements command. This makes it easier to prepare another PR which adds the execution of the ORT tools and the required configuration.

### DIFF
--- a/.github/workflows/ort.yml
+++ b/.github/workflows/ort.yml
@@ -1,0 +1,21 @@
+name: ORT
+
+on:
+  workflow_dispatch:
+
+env:
+  ORT_IMAGE: ghcr.io/oss-review-toolkit/ort-minimal
+
+jobs:
+  ort:
+    name: Run ORT
+    runs-on: ubuntu-22.04
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v4
+
+    - name: Pull ORT Docker Image
+      run: docker pull ${{ env.ORT_IMAGE }}
+
+    - name: Check ORT Requirements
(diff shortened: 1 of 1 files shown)


.github/workflows/ort.yml
nnobelis, line 1: Commit message type _required_

nnobelis (approved):

.github/workflows/ort.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  ort:
+    name: Run ORT
+    runs-on: ubuntu-22.04
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v4
+
+    - name: Pull ORT Docker Image
+      run: docker pull ${{ env.ORT_IMAGE }}
```
sschuberth, line 18: Is it really necessary to download the massive Docker image? Can't we just download the distribution archive instead, and install eventually missing tools via setup actions?

mnonnenmacher: Maybe I'll give it a try, for simplicity I wanted to start with the Docker image.

mnonnenmacher: I takes about a minute to pull the minimal image, I'll compare that to the other approach which should be faster.

PR #1104, opened 2024-09-25
fix(dao): Add missing indexes for foreign keys columns

nnobelis:
Most of the columns receiving these indexes are already part of some composite indexes, although not in the first position. When a column is in second position or more of a composite index, the latter is not used for lookups (see [1] and [2]).
All missing indexes are added by this commit except on 'snippet.artifact_id' and 'snippet.vcs_id' as this table is usually big.

[1]: https://dba.stackexchange.com/a/6426
[2]: https://www.cybertec-postgresql.com/en/index-your-foreign-key/


dao/src/main/resources/db/migration/V75__addMissingIndexes.sql
```diff
@@ -0,0 +1,219 @@
...
+ *  * Unless required by applicable law or agreed to in writing, software
+ *  * distributed under the License is distributed on an "AS IS" BASIS,
+ *  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  * See the License for the specific language governing permissions and
+ *  * limitations under the License.
+ *  *
+ *  * SPDX-License-Identifier: Apache-2.0
+ *  * License-Filename: LICENSE
+ *
+ */
```
mnonnenmacher, line 20: We don't have license headers in any of the other migration files.

mnonnenmacher (changes requested):

mnonnenmacher:
@nnobelis Somehow this change makes the `RepositoriesRouteIntegrationTest` fail.

nnobelis:
> @nnobelis Somehow this change makes the `RepositoriesRouteIntegrationTest` fail.

This is this index
`CREATE INDEX ort_runs_repository_id ON ort_runs (repository_id);`
Causing:

```
Caused by: java.sql.BatchUpdateException: Batch entry 0 INSERT INTO ort_runs (created_at, "index", job_configs, repository_id, revision, status, trace_id) VALUES (('2024-09-26 10:00:06+02'), ('8'::int8), ('{"analyzer":{"allowDynamicVersions":true,"recursiveCheckout":true,"packageCurationProviders":[]},"parameters":{}}'), ('1'::int8), ('branch-45'), ('CREATED'), ('85bc0447-1f6d-4530-9632-fe7a5db0e905'))
RETURNING * was aborted: ERROR: could not serialize access due to read/write dependencies among transactions
  Detail: Reason code: Canceled on identification as a pivot, during write.
```

Now this change makes me nervous. Any idea why we start getting those errors ?

mnonnenmacher:
> > @nnobelis Somehow this change makes the `RepositoriesRouteIntegrationTest` fail.
> 
> This is this index `CREATE INDEX ort_runs_repository_id ON ort_runs (repository_id);` Causing:
> 
> ```
> Caused by: java.sql.BatchUpdateException: Batch entry 0 INSERT INTO ort_runs (created_at, "index", job_configs, repository_id, revision, status, trace_id) VALUES (('2024-09-26 10:00:06+02'), ('8'::int8), ('{"analyzer":{"allowDynamicVersions":true,"recursiveCheckout":true,"packageCurationProviders":[]},"parameters":{}}'), ('1'::int8), ('branch-45'), ('CREATED'), ('85bc0447-1f6d-4530-9632-fe7a5db0e905'))
> RETURNING * was aborted: ERROR: could not serialize access due to read/write dependencies among transactions
>   Detail: Reason code: Canceled on identification as a pivot, during write.
> ```
> 
> Now this change makes me nervous. Any idea why we start getting those errors ?
> 
> EDIT: I just that [5007a34](https://github.com/eclipse-apoapsis/ort-server/commit/5007a34082370eb8ceab8ecf3c1da891c749ec0a) Looks connected IMO

This issue only occurred because the query to get the last index was highly inefficient and was later fixed by 4b78007. So it seems that the index makes the insertion slower which is expected, but I would have expected that this is much less relevant than the duration of the old query.

nnobelis:
Closing this PR for now has there could be some unforeseen problems with the new indexes.

Maybe it will be reopened with a smaller scope (subset of the indexes) in the future.

PR #1105, opened 2024-09-25
Configure ORT workflow

mnonnenmacher:
Configure the ORT workflow to acutally run ORT. See the commit messages for details.
Note that the workflow is still only used for testing, proper exit code handling will have to be done when integrating it in automatically triggered workflows. Also several findings still need to be addressed.
The results of a test run can be found here:
https://github.com/eclipse-apoapsis/ort-server/actions/runs/11036682245

### DIFF
--- a/.github/workflows/ort.yml
+++ b/.github/workflows/ort.yml
@@ -3,6 +3,9 @@ name: ORT
 on:
   workflow_dispatch:
 
+env:
+  ORT_CONFIG_DIR: ${{ github.workspace }}/ort-server/.github/ort
+
 jobs:
   ort:
     name: Run ORT
--- a/.github/workflows/ort.yml
+++ b/.github/workflows/ort.yml
@@ -3,9 +3,6 @@ name: ORT
 on:
   workflow_dispatch:
 
-env:
-  ORT_IMAGE: ghcr.io/oss-review-toolkit/ort-minimal
-
 jobs:
   ort:
     name: Run ORT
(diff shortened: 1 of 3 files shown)


.github/workflows/ort.yml
```diff
@@ -3,19 +3,91 @@ name: ORT
...
+
+    - name: Setup pnpm
+      uses: pnpm/action-setup@v4
+      with:
+        version: 9
+
+    - name: Setup Node
+      uses: actions/setup-node@v4
+      with:
+        node-version: 20
```
sschuberth, line 24: Should we also have `cache: pnpm` here?

.github/workflows/ort.yml
```diff
@@ -3,19 +3,91 @@ name: ORT
...
+
+    - name: Get latest ORT version
+      run: |
+        echo "ORT_VERSION=$(curl -s https://api.github.com/repos/oss-review-toolkit/ort/releases/latest | jq -r .tag_name)" >> $GITHUB_ENV
+
+    - name: Install ORT
+      run: |
+        curl -L -o ort.tar.gz https://github.com/oss-review-toolkit/ort/releases/download/${{ env.ORT_VERSION }}/ort-${{ env.ORT_VERSION }}.tgz
+        tar xfz ort.tar.gz
+        rm ort.tar.gz
```
sschuberth, line 34: You could pipe `curl` to `tar` in order to avoid dealing with the file names, and deleting the file in the end.

.github/workflows/ort.yml
```diff
@@ -3,19 +3,91 @@ name: ORT
...
+        echo "ort-${{ env.ORT_VERSION }}/bin" >> $GITHUB_PATH
+
+    - name: Cache ORT Cache Directory
+      uses: actions/cache@v4
+      with:
+        path: ~/.ort/cache
+        key: ${{ runner.os }}-ort-cache
```
sschuberth, line 46: Nit: Maybe rather use `ort-cache` as a prefix than suffix?

.github/workflows/ort.yml
```diff
@@ -3,19 +3,91 @@ name: ORT
...
+
+    - name: Run ORT Analyzer
```
sschuberth, line 48: Should we run `ort requirements --list=commands` beforehand for informational purposes?

.github/workflows/ort.yml
```diff
@@ -3,19 +3,91 @@ name: ORT
...
+      run: |
+        set +e
```
sschuberth, line 50: Maybe set this globally via `BASHOPTS` instead?

.github/ort.config.yml
```diff
@@ -0,0 +1,20 @@
...
+  analyzer:
+    skipExcluded: true
+  advisor:
+    skipExcluded: true
+  reporter:
+    config:
+      CycloneDx:
+        output.file.formats: JSON
+      SpdxDocument:
+        creationInfo.comment: SPDX document created by the ORT.
```
sschuberth, line 15: This information is already present inside `creators` in the correct "Tool:" syntax, so do we really need this additionally?

.github/workflows/ort.yml
```diff
@@ -34,6 +34,11 @@ jobs:
         rm ort.tar.gz
         echo "ort-${{ env.ORT_VERSION }}/bin" >> $GITHUB_PATH
 
+    - name: Copy ORT Config
+      run: |
+        mkdir -p $HOME/.ort/config
+        cp ort-server/.github/ort.config.yml $HOME/.ort/config/config.yml
```
sschuberth, line 40: Alternatively, commit `ort.config.yml` as `config.yml` to a directory that you can make the `ORT_CONFIG_DIR` environment variable point to.

sschuberth (changes requested):

.github/workflows/ort.yml
mnonnenmacher: I wasn't sure if this would have an effect, but as ORT calls `pnpm install` it should actualy help. I have now also added the setup-gradle action for the same reason, but we will only see if this has an effect after the merge, because AFAIK both actions only write to the cache from the main branch.

.github/workflows/ort.yml
mnonnenmacher: In the docs they usually use this pattern, but looking at our cache it seems that most actions use the pattern as you proposed, so I have changed it.

.github/workflows/ort.yml
mnonnenmacher: I removed it because I didn't see value in it.

.github/workflows/ort.yml
mnonnenmacher: There are also steps that should not continue on error and I don't think this is the final error handling anyway.

.github/workflows/ort.yml
sschuberth: I agree that it probably doesn't help with ORT Server specifically. But in general it helps to understand which ORT requirements are already fulfilled by the default action runners, and which additional setup actions would be needed to fulfill all of ORT's requirements.

But I'm fine with not doing that here.

.github/workflows/ort.yml
```diff
@@ -3,19 +3,94 @@ name: ORT
...
+        cache: pnpm
+        cache-dependency-path: ui/pnpm-lock.yaml
+
+    - name: Get latest ORT version
+      run: |
+        echo "ORT_VERSION=$(curl -s https://api.github.com/repos/oss-review-toolkit/ort/releases/latest | jq -r .tag_name)" >> $GITHUB_ENV
+
+    - name: Install ORT
+      run: |
+        curl -L -o ort.tar.gz https://github.com/oss-review-toolkit/ort/releases/download/${{ env.ORT_VERSION }}/ort-${{ env.ORT_VERSION }}.tgz | tar xfz -
```
sschuberth, line 40: This should omit `-o ort.tar.gz` now due to piping.

sschuberth (changes requested):

mnonnenmacher:
New test run: https://github.com/eclipse-apoapsis/ort-server/actions/runs/11050683116

sschuberth (approved):

PR #1114, opened 2024-09-27
Add an endpoint for creating new users

### LINKED ISSUE #1088: Add user endpoint (and the corr. UI) needed
Needs for this endpoint - probably for "superuser's" usage only - have been expressed.  If there's no-one working with this at the moment, I could volunteer. Later on, it could be expanded to a full CRUD for users.

I've already verified how user addition is done on Keycloak REST API, and seems like there are readymade functions [here](https://github.com/eclipse-apoapsis/ort-server/blob/f34f99d591116430a57a0f63f5c38249772c9046/clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt#L305-L378).

Etsija:
A subsequent PR will implement a minimal admin UI for creating new users.

Resolves #1088.

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/User.kt
+++ b/api/v1/model/src/commonMain/kotlin/User.kt
@@ -27,7 +27,16 @@ import kotlinx.serialization.Serializable
 @Serializable
 data class User(
     val id: String,
-    val username: String
+    val username: String,
+)
+
+@Serializable
+data class CreateUser(
+    val username: String,
+    val password: String? = null,
+
+    /** Specifies whether the password is for one-time use only */
+    val temporary: Boolean = true
 )
 
 /**
--- a/clients/keycloak/src/main/kotlin/KeycloakClient.kt
+++ b/clients/keycloak/src/main/kotlin/KeycloakClient.kt
@@ -127,14 +127,22 @@ interface KeycloakClient {
     suspend fun getUser(username: UserName): User
 
     /**
-     * Create a new [user][User] in the Keycloak realm with the given [username], [firstName], [lastName] and [email].
+     * Create a new [user][User] in the Keycloak realm with the given [username], [firstName], [lastName], [email],
+     * [password] and [temporary].
      */
     suspend fun createUser(
         username: UserName,
         firstName: String? = null,
         lastName: String? = null,
-        email: String? = null
+        email: String? = null,
+        password: String? = null,
+        temporary: Boolean = false
     )
+
+    /**
+     * Check whether the [user][User] has credentials in the Keycloak realm.
+     */
+    suspend fun getUserHasCredentials(username: UserName): Boolean
 
     /**
      * Update the [user][User] with the given [id] within the Keycloak realm with the new [username], [firstName],
--- a/core/src/main/kotlin/apiDocs/AdminDocs.kt
+++ b/core/src/main/kotlin/apiDocs/AdminDocs.kt
@@ -23,6 +23,8 @@ import io.github.smiley4.ktorswaggerui.dsl.routes.OpenApiRoute
 
 import io.ktor.http.HttpStatusCode
 
+import org.eclipse.apoapsis.ortserver.api.v1.model.CreateUser
+
 val runPermissionsSync: OpenApiRoute.() -> Unit = {
     operationId = "runPermissionsSync"
     summary = ""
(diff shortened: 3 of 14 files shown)


api/v1/model/src/commonMain/kotlin/User.kt
```diff
@@ -27,7 +27,14 @@ import kotlinx.serialization.Serializable
...
     val id: String,
-    val username: String
+    val username: String,
+)
+
+@Serializable
+data class CreateUser(
+    val username: String,
+    val password: String? = null,
+    val temporary: Boolean = true
```
mmurto, line 37: This and OpenAPI should document what temporary does.

mmurto (changes requested):

clients/keycloak/src/main/kotlin/KeycloakClient.kt
```diff
@@ -127,15 +127,23 @@ interface KeycloakClient {
...
+     * [password] and [temporary].
      */
     suspend fun createUser(
         username: UserName,
         firstName: String? = null,
         lastName: String? = null,
-        email: String? = null
+        email: String? = null,
+        password: String? = null,
+        temporary: Boolean = false
```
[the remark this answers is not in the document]
sschuberth: While it'd be possible to model that with a dedicated `Password` class or some `require` checks, personally I'd be fine with just documenting that `temporary` is ignores with `password` is `null`.

api/v1/model/src/commonMain/kotlin/User.kt
Etsija: @murto I couldn't find any examples from `ktorswaggerui` where the descriptions could be given at the field level, only at body level.

clients/keycloak/src/main/kotlin/KeycloakClient.kt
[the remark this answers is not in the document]
Etsija: Done.

Etsija:
@mnonnenmacher @sschuberth @mmurto could I kindly ask for a review decision on this PR? I'd like to know whether it still needs rework, or whether I can move on to something else (like, add a very minimal UI for it or start working on UI implementation for the new endpoints).

api/v1/model/src/commonMain/kotlin/User.kt
mnonnenmacher: I just noticed that my answer above was stuck in pending state since yesterday...

Etsija: @mnonnenmacher I'm sorry but I lack the knowledge of how to use that. Is it a library that needs to be imported from somewhere?  The example you linked is lacking on some crucial info (at least to me), and as such, IntelliJ doesn't recognize it, so obviously the `User.kt` file is lacking an import.

mnonnenmacher:
> @mnonnenmacher @sschuberth @mmurto could I kindly ask for a review decision on this PR? I'd like to know whether it still needs rework, or whether I can move on to something else (like, add a very minimal UI for it or start working on UI implementation for the new endpoints).

After looking at the PR again I just wonder, why did you not add the new routes to the existing admin route?

Etsija:
> > @mnonnenmacher @sschuberth @mmurto could I kindly ask for a review decision on this PR? I'd like to know whether it still needs rework, or whether I can move on to something else (like, add a very minimal UI for it or start working on UI implementation for the new endpoints).
> 
> After looking at the PR again I just wonder, why did you not add the new routes to the existing admin route?

Can't think of a reason why not. We had an internal team review and wanted to get some comments from you as to that.  Another thing is, I'm not sure `UserService` belongs under `hierarchy`, either.

api/v1/model/src/commonMain/kotlin/User.kt
mnonnenmacher: To me that's not mandatory, we do not use that anywhere else.

mnonnenmacher (changes requested):

core/src/main/kotlin/apiDocs/AdminDocs.kt
```diff
@@ -41,3 +43,29 @@ val runPermissionsSync: OpenApiRoute.() -> Unit = {
...
+            }
+        }
+    }
+
+    response {
+        HttpStatusCode.NoContent to {
+            description = "Successfully created the user."
+        }
+
+        HttpStatusCode.Conflict to {
```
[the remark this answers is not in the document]
Etsija: Actually, as was mentioned in the PR where the organization group handling was introduced, Keycloak doesn't distinguish between technical and logical errors, so a general "500" is propagated through the layers in error cases. This is now fixed and documented.

PR #1125, opened 2024-09-30
ui: Show some basic information about found packages in the results main page

Etsija:

Screenshot from 2024-10-01 07-25-08

### DIFF
--- a/ui/src/components/statistics-card.tsx
+++ b/ui/src/components/statistics-card.tsx
@@ -23,7 +23,7 @@ type StatisticsCardProps = {
   title: string;
   icon?: React.ComponentType<{ className?: string }>;
   value?: number | string;
-
+  description?: string;
   className?: string;
 };
 
@@ -31,6 +31,7 @@ export const StatisticsCard = ({
   title,
   icon: Icon,
   value,
+  description,
   className,
 }: StatisticsCardProps) => {
   return (
@@ -44,8 +45,9 @@ export const StatisticsCard = ({
         </CardTitle>
       </CardHeader>
       <CardContent className='text-sm'>
-        <div className='flex items-center justify-between'>
+        <div className='flex flex-col'>
           <div className='text-2xl font-bold'>{value ? value : '-'}</div>
+          <div className='text-xs'>{description}</div>
         </div>
       </CardContent>
     </Card>
--- a/ui/src/components/statistics-card.tsx
+++ b/ui/src/components/statistics-card.tsx
@@ -22,7 +22,7 @@ import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
 type StatisticsCardProps = {
   title: string;
   icon?: React.ComponentType<{ className?: string }>;
-  value?: number | string;
+  value?: React.ReactNode;
   description?: string;
   className?: string;
 };
--- a/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/-components/packages-statistics-card.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/-components/packages-statistics-card.tsx
@@ -67,7 +67,16 @@ export const PackagesStatisticsCard = ({
   );
 
   if (packagesIsPending) {
-    return <LoadingIndicator />;
+    return (
+      <StatisticsCard
+        title='Packages'
+        icon={() => (
+          <ListTree className={`h-4 w-4 ${getStatusFontColor(status)}`} />
+        )}
+        value={<LoadingIndicator />}
+        className='h-full hover:bg-muted/50'
+      />
+    );
   }
 
   if (packagesIsError) {
--- a/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/-components/packages-statistics-card.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/-components/packages-statistics-card.tsx
@@ -1,0 +1,102 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import { ListTree } from 'lucide-react';
+import { useMemo } from 'react';
+
+import { usePackagesServiceGetPackagesByRunId } from '@/api/queries';
+import { JobStatus } from '@/api/requests';
+import { LoadingIndicator } from '@/components/loading-indicator';
+import { StatisticsCard } from '@/components/statistics-card';
+import { ToastError } from '@/components/toast-error';
+import { getStatusFontColor } from '@/helpers/get-status-class';
+import { toast } from '@/lib/toast';
+
+type PackagesStatisticsCardProps = {
+  status: JobStatus | undefined;
+  runId: number;
+};
+
+export const PackagesStatisticsCard = ({
+  status,
+  runId,
+}: PackagesStatisticsCardProps) => {
+  const {
+    data: packages,
+    isPending: packagesIsPending,
+    isError: packagesIsError,
+    error: packagesError,
+  } = usePackagesServiceGetPackagesByRunId({
+    runId: runId,
+    // Use a large page size for packages request to try to use all packages
+    // for finding de-duplicated package types.
+    // Another option would be to do two packages queries: first to find the
+    // total number of packages, second to fetch using the total as the page
+    // size.
+    // This can be simplified once the ORT Run statistics query is implemented.
+    limit: 100000,
+  });
+
+  // Find de-duplicated package types in packages data and sort alphabetically.
+  // Packages data can be quite large, so use memoization to avoid unnecessary
+  // component rerenders.
+  const ecoSystems = useMemo(
+    () =>
+      [
+        ...new Set(
+          packages?.data.map((item) => item.identifier?.type).filter(Boolean)
+        ),
+      ].sort(),
+    [packages?.data]
+  );
+
+  if (packagesIsPending) {
+    return <LoadingIndicator />;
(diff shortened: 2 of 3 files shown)


ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/index.tsx
```diff
@@ -97,7 +109,34 @@ const RunComponent = () => {
...
+  }
+
+  const packagesTotal = packages.pagination.totalCount;
+
+  // Find deduplicated package types in packages data.
+  const detectedPackageManagers = [
+    ...new Set(
+      packages.data.map((item) => item.identifier?.type).filter(Boolean)
+    ),
+  ];
```
mmurto, line 137: This likely doesn't work as the query is paginated, so `packages` only contains the first page with the default size.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/index.tsx
```diff
@@ -97,7 +109,34 @@ const RunComponent = () => {
...
+
+  console.log(detectedPackageManagers);
```
mmurto, line 139: Should be removed.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/index.tsx
```diff
@@ -265,8 +304,9 @@ const RunComponent = () => {
                   className={`h-4 w-4 ${getStatusFontColor(ortRun.jobs.analyzer?.status)}`}
                 />
               )}
-              value={ortRun.jobs.analyzer ? '-' : 'N/A'}
-              className='hover:bg-muted/50'
+              value={ortRun.jobs.analyzer ? packagesTotal : 'N/A'}
+              description={`from ${detectedPackageManagers.length} package managers (${detectedPackageManagers.join(`,`)})`}
```
mmurto, line 308: IMO this would look better with a badge for each package manager instead of a sentence.

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/index.tsx
sschuberth: I think a sentence is fine for now. For monorepos with several package managers badges would probably take up too much screen space.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/index.tsx
mmurto: Should this rather be changed to describe that the list is of the package sources, or something else that would describe it better? I believe to type to just be [ORT Identifier's type](https://github.com/oss-review-toolkit/ort/blob/a03471c7c5981bb1950d76f708bd151ca196dfbd/model/src/main/kotlin/Identifier.kt#L32-L37), so data to do what you're suggesting is not available here. The same is probably true for NPM/Yarn/pnpm all using NPM packages.

sschuberth: > I believe to type to just be [ORT Identifier's type](https://github.com/oss-review-toolkit/ort/blob/a03471c7c5981bb1950d76f708bd151ca196dfbd/model/src/main/kotlin/Identifier.kt#L32-L37), so data to do what you're suggesting is not available here.

The information about whether an `Identifier` belongs to a package or project is not available in the `Identifier` itself. What the code *should* do is to collect the types of the identifiers of all root projects in an ORT result instead of iterating over packages.

mmurto: > > I believe to type to just be [ORT Identifier's type](https://github.com/oss-review-toolkit/ort/blob/a03471c7c5981bb1950d76f708bd151ca196dfbd/model/src/main/kotlin/Identifier.kt#L32-L37), so data to do what you're suggesting is not available here.
> 
> The information about whether an `Identifier` belongs to a package or project is not available in the `Identifier` itself. What the code _should_ do is to collect the types of the identifiers of all root projects in an ORT result instead of iterating over packages.

IMO there can be value for both, showing what package managers are used to manage the project and what type of dependencies the project uses.

mmurto: So maybe for this part it would make sense to reword the text to something like "100 packages from 2 _registries_ (Npm, Maven)". The package managers used by the project can be added here or somewhere else when the data is available, but I wouldn't hold this PR to add that.

sschuberth: > IMO there can be value for both

Maybe, but for practical matters I've always found it to be much more useful to know which package managers / project types are used instead of which ecosystems packages stem from.

> So maybe for this part it would make sense to reword the text

I agree that this initial implementation could "cheat" by adjusting the text to the current implementation instead of making the implementation match my more complex ask. However, I'd propose a different wording, and say "100 packages of 2 types (Maven, NPM) found". Oh, and the list of package types should be sorted alphabetically to be deterministic.

mmurto: > Oh, and the list of package types should be sorted alphabetically to be deterministic.

Sorted by package count per type could be good and also deterministic.

sschuberth: Agreed, but I'd also be fine with not having per-type counts in this first iteration. At least not in the card; that's probably too detailed information that could go to a header on the "Packages" tab.

mnonnenmacher: > "100 packages of 2 types (Maven, NPM) found"

How about "100 packages from 2 ecosystems (Maven, NPM) found" instead?

sschuberth: I was at first thinking about almost exactly the same as I was using "ecosystems" myself [above](https://github.com/eclipse-apoapsis/ort-server/pull/1125#discussion_r1781198645), but then I didn't want to coin yet another new term. On the other hand, we should have the liberty to use "new" terms in the UI if we believe the (non-ORT) user is familiar with them.

TL;DR  "100 packages from 2 ecosystems (Maven, NPM)" would be ok with me.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/index.tsx
```diff
@@ -97,7 +118,32 @@ const RunComponent = () => {
...
+  }
+
+  const packagesTotal = packages.pagination.totalCount;
+
+  // Find de-duplicated package types in packages data and sort alphabetically.
+  const ecoSystems = [
+    ...new Set(
+      packages.data.map((item) => item.identifier?.type).filter(Boolean)
+    ),
+  ].sort();
```
mmurto, line 146: This should maybe be memoized as the array can be quite large.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/-components/packages-statistics-card.tsx
```diff
@@ -0,0 +1,102 @@
...
+        ...new Set(
+          packages?.data.map((item) => item.identifier?.type).filter(Boolean)
+        ),
+      ].sort(),
+    [packages?.data]
+  );
+
+  if (packagesIsPending) {
+    return <LoadingIndicator />;
+  }
```
mmurto, line 71: This would look better if the card including the title and icon etc was there while the data is pending. I would maybe try [skeleton](https://ui.shadcn.com/docs/components/skeleton) at first.

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/-components/packages-statistics-card.tsx
mmurto: Or just a loading replacing the package count, skeleton doesn't look too good on first try.

Etsija: I replaced all "real" data with a loading indicator, because there's no point in showing "<spinner> from ??? ecosystems (???)". So during loading and processing, only the title and icon are shown.

ui/src/components/statistics-card.tsx
```diff
@@ -44,8 +46,11 @@ export const StatisticsCard = ({
         </CardTitle>
       </CardHeader>
       <CardContent className='text-sm'>
-        <div className='flex items-center justify-between'>
-          <div className='text-2xl font-bold'>{value ? value : '-'}</div>
+        <div className='flex flex-col'>
+          <div className='text-2xl font-bold'>
+            {value ? value === 'loading' ? <LoadingIndicator /> : value : '-'}
+          </div>
```
mmurto, line 52: This should not be based on matching a magic string like "loading". The simplest solution could be to type value as `React.ReactNode`.

mmurto (changes requested):

mmurto:
There may be a bug in the packages endpoint - please hold on merging this until I've verified it.

Etsija:
> There may be a bug in the packages endpoint - please hold on merging this until I've verified it.

I disabled auto-merge.

sschuberth:
> > There may be a bug in the packages endpoint - please hold on merging this until I've verified it.
> 
> I disabled auto-merge.

So, this is about https://github.com/eclipse-apoapsis/ort-server/issues/1137, right? But does that issue actually block the merge of this PR?

sschuberth:
> If this is merged before fixing #1137, the UI will call the endpoint and show some errors to the end users.

Now that #1137 is fixed, can we proceed with this?

mmurto:
> > If this is merged before fixing #1137, the UI will call the endpoint and show some errors to the end users.
> 
> Now that #1137 is fixed, can we proceed with this?

No objections from me.

sschuberth:
> No objections from me.

So, mind re-reviewing to resolve your "1 change requested" status, @mmurto?

mmurto (dismissed):

sschuberth:
Looks like there are build errors now. @Etsija can you please rebase and investigate?

mmurto (approved):

PR #1126, opened 2024-09-30
ci: Consistently use Ubuntu 24.04

mnonnenmacher:
Use `ubuntu-24.04` for all GitHub workflows. Previously, some workflows used `ubuntu-latest` while others used `ubuntu-22.04`. Using a specific version avoids unwanted upgrades which could break CI.

Upgrading the Docker build to `ubuntu-24.04` requires updating the `free-disk-space` action as the ubuntu-24.04 runners do not have pre-installed Docker images anymore. Also the list of tools to delete needs to be updated.

### DIFF
--- a/.github/actions/free-disk-space/action.yml
+++ b/.github/actions/free-disk-space/action.yml
@@ -6,54 +6,37 @@ runs:
   using: "composite"
 
   steps:
-    - name: Print Disk Space
-      shell: bash
-      run: df -h
-    - name: List Docker Images
-      if: ${{ false }} # Can be enabled if the 'Remove Unneeded Docker Images' step below needs to be updated.
-      shell: bash
-      run: docker images
-    - name: Remove Unneeded Docker Images
-      shell: bash
-      run: |
-        docker image rm \
-          node:16 \
-          node:16-alpine \
-          node:18 \
-          node:18-alpine \
-          node:20 \
-          node:20-alpine \
-          debian:10 \
-          debian:11 \
-          ubuntu:20.04 \
-          ubuntu:22.04
-    - name: Print Disk Space
-      shell: bash
-      run: df -h
-    - name: Get Size of Installed Tools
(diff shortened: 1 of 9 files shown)


sschuberth:
Unfortunately, at least the REUSE failure seems to be a real one.

mmurto:
> Unfortunately, at least the REUSE failure seems to be a real one.

Installing reuse with `pipx install reuse` instead of `pip` should fix it.

sschuberth:
> > Unfortunately, at least the REUSE failure seems to be a real one.
> 
> Installing reuse with `pipx install reuse` instead of `pip` should fix it.

Using [pipx](https://github.com/pypa/pipx) anyway is a good idea!

mnonnenmacher:
> > > Unfortunately, at least the REUSE failure seems to be a real one.
> > 
> > 
> > Installing reuse with `pipx install reuse` instead of `pip` should fix it.
> 
> Using [pipx](https://github.com/pypa/pipx) anyway is a good idea!

I had to do the same change in the ORT CI for installing ScanCode, but I was wondering why I didn't have to do it for reuse there as well. It seems the reason is that it's using the setup-python action:

```yaml
  reuse-tool:
    runs-on: ubuntu-24.04
    steps:
    - name: Checkout Repository
      uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
    - name: Setup Python
      uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5
      with:
        python-version: "3.10"
        cache: pip
    - name: Check REUSE Compliance
      run: |
        pip install --user reuse
        ~/.local/bin/reuse lint
```

@sschuberth I would align the job in ORT with the on here as it is simpler, if you agree.

sschuberth:
> @sschuberth I would align the job in ORT with the on here as it is simpler, if you agree.

Yes, fine with me.

mnonnenmacher:
Anyone mind approving this?

sschuberth (approved):

Issue #1074, opened 2024-09-19
Add an endpoint to get currently active runs in server instance


lamppu:
This would allow to get the data needed for the admin view described in https://github.com/eclipse-apoapsis/ort-server/issues/837.

lamppu:
I could take this on, if it's ok. I wonder though if there will be a need to get runs by other statuses as well, so should this endpoint just be /runs, and it could be filtered with a query parameter like status=active for the use case in the UI issue.

PR #1139, opened 2024-10-02
Add runs endpoint

lamppu:

Resolves https://github.com/eclipse-apoapsis/ort-server/issues/1074.

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/FilterOperatorAndValue.kt
+++ b/api/v1/model/src/commonMain/kotlin/FilterOperatorAndValue.kt
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.api.v1.model
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+/**
+ * Object representing the operator and value by which the result is filtered with.
+ */
+data class FilterOperatorAndValue<T>(
+    val operator: ComparisonOperator,
+    val value: T
+)
+
+@Serializable
+/**
+ * An enumeration of filter operators.
+ */
+enum class ComparisonOperator {
+    EQUALS,
+    NOT_EQUALS,
+    IN,
+    NOT_IN,
+    LIKE,
+    GREATER_THAN,
+    LESS_THAN,
+    GREATER_OR_EQUAL,
+    LESS_OR_EQUAL
+}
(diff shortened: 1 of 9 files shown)


Etsija:
If this resolves #1074, it should be mentioned in the description.

sschuberth:
Please rebase to resolve a merge conflict.

sschuberth:
> If this resolves #1074, it should be mentioned in the description.

Actually not in the PR description, but in the commit message.

Etsija:
> > If this resolves #1074, it should be mentioned in the description.
> 
> Actually not in the PR description, but in the commit message.

Really? This PR consists of five commits, which together resolve the issue. In these cases, I've always put the "Resolves..." text to the PR description, because no individual commit of those five is responsible for resolving the issue. Can you elaborate why we would tag only one commit (and which? The latest one?) for resolving the issue?

sschuberth:
> Really?

Yes, really. 

> Can you elaborate why we would tag only one commit, and which one should it be?

Assume that all but the last commit were already there, maybe created by someone else. Then logically it would still be the last commit that concludes the resolution of the issue.

Also, it does not make sense to sometimes resolve issues in commits and other times in PRs. To have one easy to remember rule, we align on the former. 

> I understand that PR descriptions are not seen in the commit history, so it would make sense to refer to issues being resolved in commit messages, but when looking at commit history, I see varying approaches to this, so maybe it should be discussed with the team.

This is not a team decision, but a project lead decision, IMO. So @mnonnenmacher has the last say. I tried to point it out in all PRs that I reviewed if the resolution was done in the PR instead of the commit. Could be that I missed some occurrences. But if it was not done in the commit, it's a mistake, IMO.

mnonnenmacher:
> > Really?
> 
> Yes, really.
> 
> > Can you elaborate why we would tag only one commit, and which one should it be?
> 
> Assume that all but the last commit were already there, maybe created by someone else. Then logically it would still be the last commit that concludes the resolution of the issue.
> 
> Also, it does not make sense to sometimes resolve issues in commits and other times in PRs. To have one easy to remember rule, we align on the former.

In my experience it's sufficient to only link the commit that does the final change to fix an issue, like the last one in this PR, this provides enough context already and it's easy to find the PR from there. However, you could also link the previous commits by using "Relates to #12345" if you think that is important, I sometimes do this when I only partly fix an issue. To me adding the link to the PR description is optional but adding it to the commit is mandatory.

> > I understand that PR descriptions are not seen in the commit history, so it would make sense to refer to issues being resolved in commit messages, but when looking at commit history, I see varying approaches to this, so maybe it should be discussed with the team.
> 
> This is not a team decision, but a project lead decision, IMO. So @mnonnenmacher has the last say. I tried to point it out in all PRs that I reviewed if the resolution was done in the PR instead of the commit. Could be that I missed some occurrences. But if it was not done in the commit, it's a mistake, IMO.

The definition of the role is mainly about organizational responsibilities, but a project lead is not automatically a technical lead:
https://www.eclipse.org/projects/handbook/#roles-pl
It specifically says:

> A project lead is not necessary a technical lead. We have no formal designation of technical lead; this tends to occur organically.

This matches with my experience that the people who are most active in contributing and reviewing code automatically have more influence on technical decisions.

api/v1/model/src/commonMain/kotlin/FilterOperatorAndValue.kt
```diff
@@ -0,0 +1,47 @@
...
+data class FilterOperatorAndValue<T>(
+    val operator: SqlOperator,
+    val value: T
+)
+
+@Serializable
+/**
+ * An enumeration of SQL operators.
+ */
+enum class SqlOperator {
```
mnonnenmacher, line 37: I think I would prefer a more generic name like `ComparisonOperator`, at least in the API module. Mentioning SQL here exposes an implementation detail and they might in future also be used for requests that do get their data from the database, like the user endpoints.

mnonnenmacher (changes requested):

sschuberth:
I believe you might need to rebase once more to make the new prettier checks work property.

mnonnenmacher (approved):
Very nice PR!
@lamppu  You need to rebase once more because there are conflicts.

sschuberth (approved):

PR #1140, opened 2024-10-02
Issues refactoring (part 1)

oheger-bosch:
This PR changes the way how issues created by Analyzer and Advisor are linked to an ORT run with the purpose to simplify corresponding queries, for instance in the Issues endpoint.

What is missing, are the issues generated by Scanner. They will be addressed in a follow-up PR.


dao/src/main/kotlin/tables/runs/advisor/AdvisorResultsTable.kt
```diff
@@ -55,15 +55,23 @@ class AdvisorResultDao(id: EntityID<Long>) : LongEntity(id) {
...
     )
+
+    /**
+     * Filter the given [identifierIssues] for the issues belonging to this result. This is done based on the
+     * assumption that the source of the issues contains the advisor name. This solution is a bit hacky, but it
+     * should work if the advisors follow default conventions. Note that the primary use case is to fetch the issues
+     * for a whole ORT run and not necessarily drilling this down to single advisor results.
+     */
+    private fun filterResultIssues(identifierIssues: List<Issue>): List<Issue> =
+        identifierIssues.filter { it.source.lowercase().contains(advisorName.lowercase()) }
```
sschuberth, line 76: Prefer Kotlin's `in` operator over Java's `contains()`.

mnonnenmacher: Or use `it.source.contains(advisorName, ignoreCase = true)` to avoid the `lowercase()` calls.

sschuberth: Yes, that's even better.

dao/src/main/resources/db/migration/V76__issuesOrtRun.sql
```diff
@@ -0,0 +1,40 @@
...
+  ori.ort_run_id,
+  ori.issue_id,
+  NULL,
+  CASE WHEN LOWER(i.source) = 'analyzer' THEN 'analyzer'
+       WHEN LOWER(i.source) = 'advisor' THEN 'advisor'
+       WHEN LOWER(i.source) = 'scanner' THEN 'scanner'
+       WHEN LOWER(i.source) = 'evaluator' THEN 'evaluator'
+       WHEN LOWER(i.source) = 'reporter' THEN 'reporter'
+       WHEN LOWER(i.source) = 'notifier' THEN 'notifier'
+       ELSE 'config'
```
mnonnenmacher, line 31: I would rather use `NULL` as the fallback. ORT mostly uses the name of the external tool as issue source, like "FossId" or "NPM" (check our DB for reference). The vast majority of issues will be handled by this else case, and it's better to not assign them to any worker than to the wrong one.

mnonnenmacher (changes requested):

dao/src/main/resources/db/migration/V76__issuesOrtRun.sql
oheger-bosch: This statement only applies to the issues directly linked to an ORT run which are created by ORT Server and not by ORT. Here the majority is created by the Config worker, so it should be safe to use this as default.

mnonnenmacher (approved):

PR #1155, opened 2024-10-04
Website improvements

mnonnenmacher:


mmurto (approved):

sschuberth:
Something seems to be wrong with the `prettier` check being marked as required already, although it does not seem to start yet.

mmurto:
> Something seems to be wrong with the `prettier` check being marked as required already, although it does not seem to start yet.

Old prettier was renamed to prettier-ui and prettier-website was created.

PR #1156, opened 2024-10-04
refactor(github): Make `cleanupRatio` deterministic

mnonnenmacher:
It is not clear why it was implemented with a probabilistic approach in the first place, so change the implementation to deterministically perform the cleanup at the configured ratio. This improves test stability as it was not possible to guarantee that the tests would succeed with the previous approach.

### DIFF
--- a/config/github/src/main/kotlin/GitHubConfigFileCache.kt
+++ b/config/github/src/main/kotlin/GitHubConfigFileCache.kt
@@ -28,6 +28,7 @@ import java.io.FileInputStream
 import java.io.InputStream
 import java.io.RandomAccessFile
 import java.nio.channels.FileLock
+import java.util.concurrent.atomic.AtomicInteger
 
 import kotlin.time.Duration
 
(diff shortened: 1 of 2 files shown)


config/github/src/main/kotlin/GitHubConfigFileCache.kt
```diff
@@ -38,6 +38,7 @@ import kotlinx.datetime.Clock
 import kotlinx.datetime.Instant
 
 import org.slf4j.LoggerFactory
+import java.util.concurrent.atomic.AtomicInteger
```
sschuberth, line 41: Wrong import grouping.

config/github/src/main/kotlin/GitHubConfigFileCache.kt
```diff
@@ -94,6 +95,8 @@ internal class GitHubConfigFileCache(
      */
     private val cleanupMaxAge: Duration
 ) : GitHubConfigCache {
+    private var cleanupCounter = AtomicInteger(0)
```
sschuberth, line 98: Should be a `val`.

sschuberth (changes requested):

sschuberth (approved):

PR #1159, opened 2024-10-04
docs(CONTRIBUTING): Improve the commit message documentation

### LINKED ISSUE #319: Document the commit convention of the project
ORT Server project currently uses a commit convention that looks like [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), but the convention does not semantically adhere to Conventional Commits. As an example, [`feat` types are used for development environment commits](https://github.com/eclipse-apoapsis/ort-server/commit/33504d47c5802662eed33215a7d8d5039e28dcdc), which would lead to a `MINOR` version change in Semantic Versioning if using the tooling around Conventional Commits/Semantic Versioning, and this does not make sense, as nothing in the application changed in the example commit.

It would be good to decide and document the commit convention.

mnonnenmacher:
Add more details about the conventional commit types, the handling of breaking changes, and the versioning of the project.

Fixes #319.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,17 +43,101 @@ Commits should also be self-contained, meaning that they should pass all tests a
 
 ### Commit messages
 
-We use [conventional commits](https://www.conventionalcommits.org/) for our commit messages.
-This means that each commit message should follow the format `<type>(<scope>): <description>`.
-The valid types are defined in the [Commitlint configuration file](.commitlintrc.yml).
-The scopes should represent the part of the project that was changed, e.g., `analyzer`, `core`, or `ui`.
-If a commit touches multiple parts of the project, and it is not possible to split it into multiple commits, the scope can be omitted.
-
 Commit messages should be written in the imperative mood.
 For example, use "Add feature" instead of "Added feature".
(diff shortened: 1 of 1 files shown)


sschuberth (approved):

CONTRIBUTING.md
```diff
@@ -43,18 +43,102 @@ Commits should also be self-contained, meaning that they should pass all tests a
...
+
 The scopes should represent the part of the project that was changed, e.g., `analyzer`, `core`, or `ui`.
+This usually matches the name of the Gradle module that was changed.
 If a commit touches multiple parts of the project, and it is not possible to split it into multiple commits, the scope can be omitted.
 
-Commit messages should be written in the imperative mood.
-For example, use "Add feature" instead of "Added feature".
+#### Breaking changes
+
+Breaking changes should be indicated by either adding a `BREAKING CHANGE:` section to the commit message body or by adding a `!` after the type, for example:
```
mmurto, line 84: We probably should communicate what is considered a breaking change, so is the REST API only thing we consider for versioning. Also, if we plan to do API versioning (`/api/v1/` etc.) what exactly is a breaking change if we strictly follow that?

mnonnenmacher: @mmurto I have created a separate issue for this: #1172

PR #1173, opened 2024-10-07
Implement sorting for rule violations endpoint

Etsija:
Please have a look at the commit for details.

### DIFF
--- a/core/src/main/kotlin/api/RunsRoute.kt
+++ b/core/src/main/kotlin/api/RunsRoute.kt
@@ -161,7 +161,7 @@ fun Route.runs() = route("runs/{runId}") {
             call.forRun(ortRunRepository) { ortRun ->
                 requirePermission(RepositoryPermission.READ_ORT_RUNS.roleName(ortRun.repositoryId))
 
-                val pagingOptions = call.pagingOptions(SortProperty("external_id", SortDirection.ASCENDING))
+                val pagingOptions = call.pagingOptions(SortProperty("rule", SortDirection.ASCENDING))
 
                 val ruleViolationsForOrtRun =
                     ruleViolationService.listForOrtRunId(ortRun.id, pagingOptions.mapToModel())
--- a/core/src/main/kotlin/apiDocs/RunsDocs.kt
+++ b/core/src/main/kotlin/apiDocs/RunsDocs.kt
@@ -253,7 +253,7 @@ val getVulnerabilitiesByRunId: OpenApiRoute.() -> Unit = {
 
 val getRuleViolationsByRunId: OpenApiRoute.() -> Unit = {
     operationId = "GetRuleViolationsByRunId"
-    summary = "Get the rules violations found in an ORT run."
+    summary = "Get the rule violations found in an ORT run."
     tags = listOf("RuleViolations")
 
     request {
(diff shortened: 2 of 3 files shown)


core/src/main/kotlin/api/RunsRoute.kt
```diff
@@ -161,7 +161,7 @@ fun Route.runs() = route("runs/{runId}") {
             call.forRun(ortRunRepository) { ortRun ->
                 requirePermission(RepositoryPermission.READ_ORT_RUNS.roleName(ortRun.repositoryId))
 
-                val pagingOptions = call.pagingOptions(SortProperty("external_id", SortDirection.ASCENDING))
```
sschuberth, line 164: I wonder whether this should be moved to a preceding fixup commit as it seems to have been wrong in any case, because `RuleViolationsTable` has no `external_id` property. Any opinion @mnonnenmacher?

Etsija: The code shows that the endpoint was copy-pasted from violations, as I just noticed apiDocs is also wrong: it refers to vulnerabilities in endpoint text.

sschuberth: Yeah, so that's a good indication that this copy & paste mistake should be fixed separately.

dao/src/main/kotlin/tables/runs/evaluator/RuleViolationsTable.kt
```diff
@@ -34,11 +34,11 @@ import org.jetbrains.exposed.sql.and
  * A table to represent a rule violation.
  */
 object RuleViolationsTable : SortableTable("rule_violations") {
-    val rule = text("rule")
-    val packageIdentifierId = reference("package_identifier_id", IdentifiersTable).nullable()
+    val rule = text("rule").sortable()
+    val packageIdentifierId = reference("package_identifier_id", IdentifiersTable).sortable().nullable()
```
mnonnenmacher, line 38: Not sure if it makes sense to make this sortable, IIUC it would just sort by the numeric id, not the string representation of the identifier.

core/src/main/kotlin/api/RunsRoute.kt
mnonnenmacher: But fixing the default sort property requires making another one sortable, so that's an argument for doing it in the same commit.

core/src/main/kotlin/apiDocs/RunsDocs.kt
```diff
@@ -309,7 +309,7 @@ val getRuleViolationsByRunId: OpenApiRoute.() -> Unit = {
                             limit = 20,
                             offset = 0,
                             totalCount = 1,
-                            sortProperties = listOf(SortProperty("packageUrl", SortDirection.ASCENDING))
+                            sortProperties = listOf(SortProperty("rule", SortDirection.ASCENDING))
```
sschuberth, line 312: I'm confused. This clearly is not a doc-only change. Shouldn't the changes in this file and the changes in `core/src/main/kotlin/api/RunsRoute.kt` from the previous commit be combined in one commit, @mnonnenmacher?

Etsija: This _is_ a doc-only change; it only affects the examples in Swagger documentation.

mnonnenmacher: To me it's fine the way it is.

mnonnenmacher (approved):

PR #1176, opened 2024-10-07
Improve publishing configuration

mnonnenmacher:
Improve the publishing configuration as a preparation to create the release workflow for Maven Central.

### DIFF
--- a/buildSrc/src/main/kotlin/ort-server-publication-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/ort-server-publication-conventions.gradle.kts
@@ -69,4 +69,6 @@ mavenPublishing {
             }
         }
     }
+
+    publishToMavenCentral()
 }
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/ort-server-publication-conventions.gradle.kts
```diff
@@ -69,4 +71,6 @@ mavenPublishing {
             }
         }
     }
+
+    publishToMavenCentral(SonatypeHost.DEFAULT)
```
sschuberth, line 75: We should be able to omit the argument as [it's the default](https://github.com/vanniktech/gradle-maven-publish-plugin/blob/a40d9943fafb363f1a5c024f5b7b5ceda89d52b4/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt#L58). I'll make the same change for ORT.

sschuberth (changes requested):

sschuberth (approved):

PR #1185, opened 2024-10-08
build: Customize the logic to determine the version

mnonnenmacher:
The previous implementation had the problem that pre-releases always got the revision appended. Also, the logic to remove the pre-release identifier did not work with versioned pre-releases like "RC1".

Fix this by completely customizing the logic to determine the version:

- Releases only contain the semantic version, like "1.2.3".
- Pre-releases additionally contain the pre release version, like "1.2.3-RC1".
- Commits ahead of the last release tag get the commit count and revision appended, like "1.2.3-RC1.001.sha.0123456".

The implementation is more verbose than before but makes it easier to understand how the version is constructed.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -36,18 +36,26 @@ semver {
     // Do not create an empty release commit when running the "releaseVersion" task.
     createReleaseCommit = false
 
+    // Use "RC" instead of "SNAPSHOT" as the default pre-release identifier.
+    defaultPreRelease = "RC"
+
     // Do not let untracked files bump the version or add a "-SNAPSHOT" suffix.
     noDirtyCheck = true
 }
 
 // Only override a default version (which usually is "unspecified"), but not a custom version.
 if (version == Project.DEFAULT_VERSION) {
-    version = semver.semVersion.takeIf { it.isPreRelease }
-        // To get rid of a build part's "+" prefix because Docker tags do not support it, use only the original "build"
-        // part as the "pre-release" part.
-        ?.toString()?.replace("${semver.defaultPreRelease}+", "")
-        // Fall back to a plain version without pre-release or build parts.
-        ?: semver.version
+    val semVersion = semver.semVersion
+
+    // Set the version based on the following rules:
+    // - If the current commit is tagged as a release (e.g., 1.2.3) or pre-release (e.g., 1.2.3-RC1), the version equals
+    //   the tag.
+    // - If the current commit is ahead of the last release or pre-release tag, the version is the tag plus the
+    //   pre-release identifier, the commit count, and the SHA. For example, "1.2.3-RC1.001.sha.0123456". Note that
+    //   semantic versions would use a "+" instead of a "." in front of the commit count, but this version is also used
+    //   as a Docker tag and Docker does not allow "+" in tags.
+    val shaLength = if (semVersion.commitCount > 0) 7 else 0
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -36,18 +36,41 @@ semver {
...
-        ?: semver.version
+    val semVersion = semver.semVersion
+
+    version = buildString {
+        // Use the major, minor, and patch version as the version string, for example, "1.2.3".
+        append(semVersion.major)
+        append(".")
+        append(semVersion.minor)
+        append(".")
+        append(semVersion.patch)
```
sschuberth, line 56: FYI, this should be the same as `append(semVersion.toVersionString(v2 = false))`.

build.gradle.kts
```diff
@@ -36,18 +36,41 @@ semver {
...
+
+            // If the current commit is ahead of the last tag, append the commit count and the SHA, for example,
+            // "1.2.3-RC1.001.sha.0123456".
+            if (semVersion.commitCount > 0) {
+                append(".")
+                append("%03d".format(semVersion.commitCount))
+                append(".sha.")
+                append(semVersion.sha.take(7))
+            }
+        }
```
sschuberth, line 72: In general, I wonder whether you could achieve the same with just

    semVersion.toInfoVersionString(shaLength = 7).replace('+', '.')

However, from a SemVer perspective, I believe it's [correct to have a '+' before the `<build>` part](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions).

sschuberth (changes requested):

build.gradle.kts
mnonnenmacher: This would work with a small adaption:
```kotlin
val shaLength = if (semVersion.commitCount > 0) 7 else 0
version = semVersion.toInfoVersionString(shaLength = shaLength).replace('+', '-')
```

I found it difficult to understand how the existing functions behave in all possible scenarios, that's why I made the verbose implementation.

>I believe it's [correct to have a '+' before the <build> part](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions).

I know, but we replaced it with '-' also before because '+' cannot be used in Docker tags. I initially implemented it so that the '+' is only removed for the Docker tag but kept for the app version itself, but you changed that in 6fdc420.

sschuberth: > that's why I made the verbose implementation.

Ok, so I guess you prefer to keep it as-is.

There's still the [optional commit message nit](https://github.com/eclipse-apoapsis/ort-server/pull/1185#discussion_r1792882282), though.

sschuberth (approved):

build.gradle.kts
mnonnenmacher: I first wanted to clarify if we should move the '+' replacement back to the Docker build job before applying any changes.
As I did the manual testing for the simplified code above now as well I would be fine with using it, hoping that the library is stable there.

mnonnenmacher: > I first wanted to clarify if we should move the '+' replacement back to the Docker build job before applying any changes.

On the other hand I'm not even sure if we should continue publishing snapshots of the Docker images once we have releases.

mnonnenmacher: I have now switched to the simpler implementation but kept the "+" replacement in Gradle.

sschuberth (approved):

PR #1189, opened 2024-10-09
Fix release artifact signing

mnonnenmacher:
The release workflow was using the wrong key id, see the commit messages for details.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,7 +45,7 @@ jobs:
         # https://keyserver.ubuntu.com/pks/lookup?search=apoapsis-dev&fingerprint=on&op=index
         ORG_GRADLE_PROJECT_signingInMemoryKeyId: 437c0602
         ORG_GRADLE_PROJECT_signingInMemoryKeyPassword: ${{ secrets.ORG_GPG_PASSPHRASE }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -41,7 +41,9 @@ jobs:
         ORG_GRADLE_PROJECT_mavenCentralPassword: ${{ secrets.ORG_OSSRH_PASSWORD }}
         ORG_GRADLE_PROJECT_RELEASE_SIGNING_ENABLED: true
         ORG_GRADLE_PROJECT_signingInMemoryKey: ${{ secrets.ORG_GPG_PRIVATE_KEY }}
-        ORG_GRADLE_PROJECT_signingInMemoryKeyId: ${{ secrets.ORG_GPG_KEY_ID }}
+        # secrets.ORG_GPG_KEY_ID contains the wrong value, so hardcode the key id as found on:
```
sschuberth, line 44: I guess this is only a work-around until EF sets the `ORG_GPG_KEY_ID` variable correctly?

sschuberth (approved):

.github/workflows/release.yml
mnonnenmacher: Yes, I want to first make sure that the publishing actually works with that value and then ask them to fix it. I have only with `publishToMavenLocal` so far.

PR #1191, opened 2024-10-09
Release improvements

mnonnenmacher:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,4 +54,7 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
-        gh release create $ORT_SERVER_VERSION --notes-file RELEASE_NOTES.md
+        if [[ "$ORT_SERVER_VERSION" =~ -RC[0-9]+ ]]; then
+          PRERELEASE_ARG="--prerelease"
+        fi
+        gh release create $ORT_SERVER_VERSION --notes-file RELEASE_NOTES.md $PRERELEASE_ARG
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,9 +41,7 @@ jobs:
         ORG_GRADLE_PROJECT_mavenCentralPassword: ${{ secrets.ORG_OSSRH_PASSWORD }}
         ORG_GRADLE_PROJECT_RELEASE_SIGNING_ENABLED: true
         ORG_GRADLE_PROJECT_signingInMemoryKey: ${{ secrets.ORG_GPG_PRIVATE_KEY }}
-        # secrets.ORG_GPG_KEY_ID contains the wrong value, so hardcode the key id as found on:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -54,4 +54,8 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
-        gh release create $ORT_SERVER_VERSION --notes-file RELEASE_NOTES.md
+        PRERELEASE_ARG=""
```
sschuberth, line 57: Nit: You could even omit this, as interpolating a non-existing variable will also result in the empty string.

.github/workflows/release.yml
```diff
@@ -54,4 +54,8 @@ jobs:
...
+        if [[ "$ORT_SERVER_VERSION" == *RC* ]]; then
```
sschuberth, line 58: Should the pattern be `-RC` for safety? Or even `[[ "$ORT_SERVER_VERSION" =~ "-RC[0-9]+" ]]` to align a bit with line 7?

.github/workflows/release.yml
```diff
@@ -48,13 +48,14 @@ jobs:
       run: ./gradlew --stacktrace --no-configuration-cache publishAndReleaseToMavenCentral
 
     - name: Generate Release Notes
-      # Temporary implementation for the very first release. The release notes for this will be written by hand as they
-      # would otherwise contain the complete commit history.
-      run: echo "Initial release!" > RELEASE_NOTES.md
-      #run: ./gradlew -q printChangeLog > RELEASE_NOTES.md
+      run: ./gradlew -q printChangeLog > RELEASE_NOTES.md
```
mmurto, line 51: This seems to print the commits in groups, right? We could consider something more sophisticated like https://keepachangelog.com/en/1.0.0/ as well. I'd say especially as we are quite strict with the commits, using all commits for changelog is not that nice.

sschuberth: > This seems to print the commits in groups, right?

Yes, it's the same mechanism as at https://github.com/oss-review-toolkit/ort/releases.

mnonnenmacher: I believe that proper user-facing release notes have to be handwritten, but for now I'm happy to at least have a fully automated solution.

.github/workflows/release.yml
```diff
@@ -54,4 +54,7 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
-        gh release create $ORT_SERVER_VERSION --notes-file RELEASE_NOTES.md
+        if [[ "$ORT_SERVER_VERSION" =~ "-RC[0-9]+" ]]; then
```
sschuberth, line 55: In my tests, the regex must not be quoted to match, so omit `"` around `-RC[0-9]+`.

sschuberth (changes requested):

sschuberth (approved):

PR #1198, opened 2024-10-10
Allow to specify an alternative ort.env.yml file path 

nnobelis:

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/OrtRun.kt
+++ b/api/v1/model/src/commonMain/kotlin/OrtRun.kt
@@ -117,6 +117,12 @@ data class OrtRun(
     val resolvedJobConfigContext: String? = null,
 
     /**
+     * The optional path to an environment configuration file. If this is not defined, the environment configuration is
+     * read from the default location `.ort.env.yml`.
+     */
+    val environmentConfigPath: String? = null,
+
+    /**
      * The trace ID that is assigned to this run. This is generated when the run is created. It can be used to
      * correlate the logs from different components that are taking part in processing of the run.
      */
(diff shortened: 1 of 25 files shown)


api/v1/model/src/commonMain/kotlin/OrtRun.kt
```diff
@@ -116,6 +116,12 @@ data class OrtRun(
      */
     val resolvedJobConfigContext: String? = null,
 
+    /**
+     * The optional path to a environment configuration file. If this is not defined, the environment configuration is
```
oheger-bosch, line 120: Nit: *an* environment configuration file.

core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
```diff
@@ -1114,7 +1126,8 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
                     JobConfigurations(),
                     "jobConfigContext",
                     labelsMap,
-                    traceId = "t3"
+                    traceId = "t3",
+                    null
```
oheger-bosch, line 1130: Shouldn't there be at least one occurrence where a value is set for the new property to test the mappings?

workers/analyzer/src/test/kotlin/AnalyzerEndpointTest.kt
```diff
@@ -352,7 +352,8 @@ class AnalyzerEndpointTest : KoinTest, StringSpec() {
                     emptyList(),
                     null,
                     null,
-                    "trace-id"
+                    null,
+                    "trace-id",
```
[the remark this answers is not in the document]
nnobelis: image
I think it is ok.

The order is different between the constructor of `OrtRun` and the function `DaoOrtRunRepository.create`.

oheger-bosch (approved):

PR #1207, opened 2024-10-11
ui: Show job statuses in product list

lamppu:
If a product of an organization only has one repository, show the status of the latest run in that repository in the product list in the organization details page to get a quick overview of runs in the organization.

Screenshot from 2024-10-11 11-03-15


mnonnenmacher:
As an idea, I think showing the date of the last run would also be very helpful to get an overview.

Etsija:
> As an idea, I think showing the date of the last run would also be very helpful to get an overview.

Maybe just a date?

lamppu:
> > As an idea, I think showing the date of the last run would also be very helpful to get an overview.
> 
> Maybe just a date of creation after the run status? Or, date of creation for ongoing run, date of completion for a finished one?

Something like this maybe?

Product details:
Screenshot from 2024-10-14 11-09-52

Organization details:
Screenshot from 2024-10-14 11-10-39

sschuberth:
> Something like this maybe?

Personally, I'd also be fine with simply dropping the "Started / Completed At:" lines as it's clear that an active run cannot have completed yet, so the timestamp cannot refer to that state either. In exchange, maybe do like we do for the duration of ongoing runs and print the timestamp in italics to highlight its slightly different meaning?

lamppu:
Screenshot from 2024-10-14 12-58-19

mmurto (approved):

mmurto:
@mnonnenmacher, after you've deployed this, can you try out the performance on your infra as well? This makes a lot of requests to the API, so let's keep an eye on whether it's too much.

PR #1219, opened 2024-10-14
feat(storage): Use suspend functions in the `StorageProvider`

nnobelis:
The `StorageProvider` interface from the storage abstraction defines some functions for reading and writing unstructured data. Some of the existing implementations implement those using HTTP requests, which are done using the suspendable functions of the HTTP client. This commit changes the API of the `StorageProvider` to make the functions suspendable. Please note that some of the Exposed transaction calls done by the `DatabaseStorageProvider.kt` have therefore to be adapted (see [1]).

[1]: https://jetbrains.github.io/Exposed/transactions.html#working-with-coroutines

### DIFF
--- a/storage/s3/src/main/kotlin/S3StorageProvider.kt
+++ b/storage/s3/src/main/kotlin/S3StorageProvider.kt
@@ -36,7 +36,6 @@ import kotlin.io.path.outputStream
 import org.eclipse.apoapsis.ortserver.storage.Key
 import org.eclipse.apoapsis.ortserver.storage.StorageEntry
 import org.eclipse.apoapsis.ortserver.storage.StorageProvider
-import org.eclipse.apoapsis.ortserver.utils.logging.runBlocking
 
 /**
  * Implementation of the [StorageProvider] interface that is backed by AWS S3 storage.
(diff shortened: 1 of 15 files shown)


storage/s3/src/main/kotlin/S3StorageProvider.kt
```diff
@@ -51,33 +50,31 @@ class S3StorageProvider(
...
-        ) { response ->
-            val data = checkNotNull(response.body) { "No data found for ${key.key}." }
-            val contentType = response.contentType
-            val tempFile = createTempFile()
-
-            runCatching {
-                data.writeToFile(tempFile)
-                StorageEntry.create(tempFile.toFile(), contentType)
-            }.onFailure { tempFile.deleteExisting() }.getOrThrow()
+    override suspend fun read(key: Key): StorageEntry = s3Client.getObject(
```
sschuberth, line 53: Not sure about the formatting: I think for lambdas it was just ok to start on this line, but I'm unsure for function calls. Any guideline here, @mnonnenmacher?

storage/s3/src/main/kotlin/S3StorageProvider.kt
```diff
@@ -96,30 +93,26 @@ class S3StorageProvider(
...
-            runCatching {
-                s3Client.deleteObject {
-                    this.bucket = bucketName
-                    this.key = key.key
-                }
-            }.isSuccess
-        } else {
-            false
-        }
+    override suspend fun delete(key: Key): Boolean = if (contains(key)) {
```
sschuberth, line 108: See above about formatting.

storage/spi/src/main/kotlin/Storage.kt
```diff
@@ -84,17 +87,18 @@ class Storage(
     /**
      * Return the [StorageEntry] associated with the given [key]. Throw a [StorageEntry] if this operation fails.
      */
-    fun read(key: Key): StorageEntry =
-        wrapException {
+    suspend fun read(key: Key): StorageEntry {
+        return wrapException {
```
sschuberth, line 91: This should stay an extension function.

workers/common/src/main/kotlin/common/OrtServerFileArchiveStorage.kt
```diff
@@ -58,15 +59,16 @@ class OrtServerFileArchiveStorage(
             }
     }
 
-    override fun getData(provenance: KnownProvenance): InputStream? {
+    override fun getData(provenance: KnownProvenance): InputStream? = runBlocking {
```
oheger-bosch, line 62: Not sure whether using `runBlocking` here is a good solution. From what I have read, this function should only be used in tests and for the main entry point.
Would it be an alternative to create a new `CoroutineScope` in this class and use it to launch the suspend functions?

workers/scanner/src/main/kotlin/scanner/OrtServerFileListStorage.kt
```diff
@@ -58,15 +59,16 @@ class OrtServerFileListStorage(
             }
     }
 
-    override fun getData(provenance: KnownProvenance): InputStream? {
+    override fun getData(provenance: KnownProvenance): InputStream? = runBlocking {
```
oheger-bosch, line 62: Same as for `OrtServerFileArchiveStorage` regarding usage of `runBlocking`.

workers/common/src/main/kotlin/common/OrtServerFileArchiveStorage.kt
nnobelis: The problem with `launch` is that you get a `Job`. Should we then `await` on it ?

oheger-bosch: You are right, it is probably not worth the effort. I was reading [this article](https://discuss.kotlinlang.org/t/globalscope-and-runblocking-should-not-be-used-so-how-do-you-start-a-coroutine/24004/3), and while statements are a bit contradicting IMHO, it could be interpreted that `runBlocking` was appropriate for our use case.

oheger-bosch (approved):

workers/common/src/main/kotlin/common/OrtServerFileArchiveStorage.kt
nnobelis: I will remember that from your link:

> runBlocking should not be used from inside the coroutine, but it is a normal way to enter the coroutine world.

storage/s3/src/main/kotlin/S3StorageProvider.kt
nnobelis: @sschuberth Are you fine with merging the PR ?

sschuberth: Sure, I've just put it into the merge queue.

PR #1229, opened 2024-10-15
api: Add endpoints for listing and deleting users

Etsija:
Screenshot from 2024-10-15 11-57-46

Screenshot from 2024-10-15 11-57-59

This PR will be followed with a corresponding UI implementation PR.


model/src/commonMain/kotlin/User.kt
```diff
@@ -19,19 +19,16 @@
 
 package org.eclipse.apoapsis.ortserver.model
 
+import kotlinx.serialization.Serializable
+
 /**
  * A data class representing a user.
  */
+@Serializable
```
mnonnenmacher, line 27: Can you explain why the class now needs to be serializable and why the removed properties are not required anymore?

core/src/main/kotlin/api/AdminRoute.kt
mnonnenmacher, line 1: Having a separate commit for the move (refactor) would have made the review easier.

core/src/main/kotlin/api/AdminRoute.kt
```diff
@@ -64,6 +68,13 @@ fun Route.admin() = route("admin") {
     route("users") {
         val userService by inject<UserService>()
 
+        get(getUsers) {
+            requireSuperuser()
+
+            val users = userService.getUsers()
+            call.respond(users)
```
mnonnenmacher, line 75: Endpoints should only return classes from the `:api:v1:model` module.

mnonnenmacher (changes requested):

core/src/main/kotlin/api/AdminRoute.kt
Etsija: @mnonnenmacher can you explain how to avoid this from detekt? Do I have to rewrite the mapping to map `Set<User>`(from model) -> into a `Set<User>` (from API) instead of mapping the classes directly?
```
> Task :core:detektMain FAILED
/home/runner/work/ort-server/ort-server/core/src/main/kotlin/api/AdminRoute.kt:75:52: Name shadowed: implicit lambda parameter 'it' [NoNameShadowing]
```

model/src/commonMain/kotlin/User.kt
Etsija: I think the model class is now used properly, and the reason for omitting data was explained in my commit message.

core/src/test/kotlin/api/AdminRouteIntegrationTest.kt
```diff
@@ -42,4 +61,74 @@ class AdminRouteIntegrationTest : AbstractIntegrationTest({
...
+                    parameter("username", TEST_USER.username.value)
+                }
+                val response = superuserClient.delete("/api/v1/admin/users") {
+                    parameter("username", TEST_USER.username.value)
+                }
+
+                response shouldHaveStatus HttpStatusCode.InternalServerError
+            }
+        }
+    }
```
mnonnenmacher, line 133: There should also be a test for each endpoint to verify that the superuser role is required (like for `GET /admin/sync-roles`).

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #1232, opened 2024-10-15
Remove Renovate

mnonnenmacher:
The Renovate action is now run in https://github.com/eclipse-apoapsis/renovate.


sschuberth (approved):
Nit: All of [these](https://docs.renovatebot.com/configuration-options/) seem to be equally valid "default" locations. However, I agree that for discoverability the root location is a good choice.

mnonnenmacher:
> Nit: All of [these](https://docs.renovatebot.com/configuration-options/) seem to be equally valid "default" locations. However, I agree that for discoverability the root location is a good choice.

You are right, I meant more like the "usual" default location 😀

PR #1233, opened 2024-10-15
website: Add Docker Compose docs

mnonnenmacher:
This is the beginning of migrating the documentation to the website.


website/docs/getting-started/docker-compose.md
```diff
@@ -0,0 +1,192 @@
...
+docker compose up
+```
+
+This will use the ORT Server images [published on GitHub](https://github.com/orgs/eclipse-apoapsis/packages?ecosystem=container).
+By default, the `main` tag is used.
+Those images are built from the `main` branch of the repository.
+To use a different tag, you can set the `ORT_SERVER_IMAGE_TAG` environment variable, for example:
+
+```shell
+ORT_SERVER_IMAGE_TAG=0.1.0-SNAPSHOT-001.sha.aa4d3fa docker compose up
```
sschuberth, line 31: Didn't we swap out "SNAPSHOT" for "RC"? In general, I believe something that really exists should be used here, even if just an example.

website/docs/getting-started/docker-compose.md
```diff
@@ -0,0 +1,192 @@
...
+By default, the ORT Server API is exposed on port 8080.
+If this port is already in use, it can be changed using the `ORT_SERVER_CORE_PORT` environment variable:
+
+```shell
+ORT_SERVER_CORE_PORT=8090 docker compose up
+```
+
+When using a different port, please make sure that it is not used by [another service](#accessing-the-services).
+
+### Update the images
```
sschuberth, line 43: Nit: You seem to be inconsistently using title-case for headers in this document; e.g. here, the "Image" noun should probably be capitalized.

website/docs/getting-started/docker-compose.md
```diff
@@ -0,0 +1,192 @@
...
+
+Then execute the ORT Server in IntelliJ with the run configuration ["Run ORT Server"](https://github.com/eclipse-apoapsis/ort-server/blob/main/.run/Run%20ORT%20Server.run.xml).
+
+## Troubleshooting
+
+### Flyway migration errors
+
+When starting the ORT Server, you can run into the following error:
+
+> Exception in thread "main" org.flywaydb.core.api.exception.FlywayValidateException: Validate failed: Migrations have failed validation`
```
sschuberth, line 132: This seems to contain an unmatched trailing backtick.

website/docs/getting-started/docker-compose.md
```diff
@@ -0,0 +1,192 @@
+# Docker Compose
+
+The easiest way to run the ORT Server for testing is to use [Docker Compose](https://docs.docker.com/compose/).
+For a proper deployment to Kubernetes, the project will later provide a Helm chart.
+
+:::warning
+Do not use the Docker Compose setup in production as it uses multiple insecure defaults, like providing Keycloak without TLS.
+:::
+
+To use Docker Composer, you need to have [Docker](https://docs.docker.com/get-docker/) and [Docker Compose](https://docs.docker.com/compose/install/) installed.
```
oheger-bosch, line 10: Typo: "Docker Composer" -> "Docker Compose"

website/docs/getting-started/docker-compose.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+This will build the images with the `latest` tag.
+Then you can start Docker Compose using the local images by setting the image prefix to an empty string and the tag to `latest`:
+
+```shell
+ORT_SERVER_IMAGE_PREFIX= ORT_SERVER_IMAGE_TAG=latest docker compose up
+```
+
+### Build specific images
```
oheger-bosch, line 78: Suggestion: I am not sure whether the information how to build images should be under the Docker Compose topic - at least, I would not have expected to find it there. Should there be a dedicated page explaining how to build the project? The compose documentation could then link to this.

website/docs/getting-started/docker-compose.md
mnonnenmacher: I read that several style guides recommend using title case only for the first two levels. However,  I did some research again and this seems not to be up-to-date. For example, Microsoft style guide recommends to only use sentence case (https://learn.microsoft.com/en-us/style-guide/scannable-content/headings) and the APA switched from using title case only for the first two levels to using it for all headings (https://academicguides.waldenu.edu/writingcenter/apa/other/headinglevels). So for simplicity I'll go with the APA recommendation.

sschuberth: > So for simplicity I'll go with the APA recommendation.

Fine with me as long as it's consistently applied.

sschuberth (approved):

PR #1234, opened 2024-10-15
deps: Update schemakenerator and ktor-swagger-ui

mnonnenmacher:
Update schmemakenerator to v1.4.3 and ktor-swagger-ui to v3.5.1. The libraries have to be updated together to create a valid `openapi.json`.

With this update, nullable properties are properly modeled in the OpenAPI spec, so also update the UI to handle those nullable properties.


mnonnenmacher:
@mmurto @lamppu @Etsija I'm not sure if I handled the now nullable properties correctly, would be good if one of you had a look. I also considered using `z.union([z.string(), z.null()]).optional()` instead of `?? undefined` in some places, but I wasn't sure if it even makes sense to model the difference betwenn `null` and `undefined` for the forms.

mmurto:
> I wasn't sure if it even makes sense to model the difference betwenn `null` and `undefined` for the forms.

This is IMO the important part regarding edit forms. The difference is important in that (at least often) `undefined` in edit form means that we don't want to edit the field, while `null` means that we want to set the field as null. Does it work like that for the API?

Let's use the following organization as an example:

```typescript
// This is the type returned by API.
export type Organization = {
  description?: string | null;
  id: number;
  name: string;
};

const currentOrganization = {
  id: 1,
  name: "Organization One",
  description: "Example"
}

// To change only the name, we can send this
const editPayload = {
  name: "New name"
}

// To remove description, we can send this
const editPayload2 = {
  description: null
}
```

If we want to be able to set nullable fields to null, we need the differentiation between null and undefined.

Regarding other types, `JobSummary` in `types.gen.ts` is now generated as:

```typescript
export type JobSummary = {
  createdAt: string;
  finishedAt?: string | null;
  id: number;
  startedAt?: string | null;
  status: JobStatus;
};
```

Does this make sense, and what does it mean? What difference does the API convey with `finishedAt` or `startedAt` being `undefined` or `null?`

mnonnenmacher:
> > I wasn't sure if it even makes sense to model the difference betwenn `null` and `undefined` for the forms.
> 
> This is IMO the important part regarding edit forms. The difference is important in that (at least often) `undefined` in edit form means that we don't want to edit the field, while `null` means that we want to set the field as null. Does it work like that for the API?

Yes, exactly. However, I'm not sure if there is any need to set those values to null, at least for the forms we currently have.

> Regarding other types, `JobSummary` in `types.gen.ts` is now generated as:
> 
> ```ts
> export type JobSummary = {
>   createdAt: string;
>   finishedAt?: string | null;
>   id: number;
>   startedAt?: string | null;
>   status: JobStatus;
> };
> ```
> 
> Does this make sense, and what does it mean? What difference does the API convey with `finishedAt` or `startedAt` being `undefined` or `null?`

In this case the representation is wrong, in Kotlin the `finishedAt` property is defined as `val finishedAt: Instant? = null`, so the correct representation in TypeScript should be `finishedAt: string | null` because it cannot be undefined, right?

In general, undefined values are only relevant for objects sent to the API. I need to check why the `JobSummary` is represented like that.

Regarding the code changes in this PR, does it look good to you so that we can merge it and address the other topics later? Or do you see any issues with it?

ui/src/helpers/calc-overall-vulnerability.ts
```diff
@@ -29,7 +29,7 @@ import { VulnerabilityRating } from '@/helpers/get-status-class';
 // Now that we have scores and vectors available, we could do better with
 // calculation of the overall vulnerability rating.
 export function calcOverallVulnerability(
-  ratings: (number | undefined)[]
+  ratings: (number | null | undefined)[]
```
mmurto, line 32: The `null` should be taken into account in the below filter. The `rating is null` means that, as TypeScript wasn't able to narrow the type on it's own, we told is that if we filter undefineds away, all of the ratings are numbers. After adding nulls as another option, this is no longer the case.

Luckily TypeScript recent release made this better, so the whole `is number` can be removed. The filter should be changed to:

```typescript
  const filteredRatings = ratings.filter(
    (rating) => rating !== undefined && rating !== null
  );
```

mmurto (changes requested):

mmurto:
Other than the one thing no issues. There can be some [uncontrolled/controlled input errors](https://stackoverflow.com/questions/47012169/a-component-is-changing-an-uncontrolled-input-of-type-text-to-be-controlled-erro) in the console, but they're harmless and were probably there before, so we'll address them separately.

mmurto (approved):

PR #1235, opened 2024-10-15
fix(ui): Do not show the ecosystem overview for no packages

sschuberth:


ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/-components/packages-statistics-card.tsx
```diff
@@ -100,11 +100,16 @@ export const PackagesStatisticsCard = ({
...
+      description={(() => {
+        switch (ecoSystems.length) {
+          case 0:
+            return '';
+          case 1:
+            return `from 1 ecosystem (${ecoSystems})`;
+          default:
+            return `from ${ecoSystems.length} ecosystems (${ecoSystems.join(`,`)})`;
+        }
+      })()}
```
mmurto, line 112: This would be a little more idiomatic without the function definition and call as:

```typescript
ecoSystems.length
  ? ecoSystems.length > 1
    ? `from ${ecoSystems.length} ecosystems (${ecoSystems.join(`,`)})`
    : `from 1 ecosystem (${ecoSystems})`
  : ''"
```

The number [0 is Falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), so the ternary will handle others in the truthy branch and it in the falsy.

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/-components/packages-statistics-card.tsx
mmurto: Another option could be to change the `description` prop of `StatisticsCard` to be `ReactNode`, in which case a function could be passed to it as is without the need of `()()`.

sschuberth: Yeah, I was reading about the nested ternary operator approach over [here](https://dev.to/gvergnaud/bringing-pattern-matching-to-typescript-introducing-ts-pattern-v3-0-o1k), but deliberately decided against it (due to the hard to read nested). However, as you say it's more idiomatic, and it's fewer lines as well, I'll change it.

mmurto: Imo the most non-idiomatic part is the immediate call of the arrow function as `(() => {})**()**`, but that should be fixed by changing `description` to [`ReactNode`](https://reactnative.dev/docs/react-node), which would allow just passing the function without calling it here. As `string` is also `ReactNode`, the change should not require changes elsewhere where the component is used.

sschuberth: I sticked to your code as it's more compact.

mmurto (approved):

PR #1237, opened 2024-10-16
Refactor Scanner issues

oheger-bosch:
This PR associates the issues created by the Scanner with the current ORT run, so that they can be fetched directly with other issues related to this run.


dao/src/main/resources/db/migration/V82__scannerIssuesOrtRun.sql
```diff
@@ -0,0 +1,70 @@
+-- This migration script integrates the issues generated by the Scanner into the ORT run issues table.
```
sschuberth, line 1: Commit message:
- "which are independent o*f* ORT runs"

workers/common/src/main/kotlin/common/OrtRunService.kt
```diff
@@ -118,6 +119,10 @@ class OrtRunService(
             config = config,
             scanners = scannerRun.scanners
         )
+
+        getScannerJob(scannerRun.scannerJobId)?.let { job ->
```
sschuberth, line 123: I once more propose to use `also` instead of `let` as it better fits the intended semantics.

dao/src/main/resources/db/migration/V82__scannerIssuesOrtRun.sql
```diff
@@ -0,0 +1,70 @@
...
+INNER JOIN scan_summaries_issues ssi on ssi.scan_summary_id = ss.id
+INNER JOIN issues i on ssi.issue_id = i.id
+INNER JOIN (
+  SELECT
+    sr.id AS scan_result_id,
+    artifact_provenances.identifier_id,
+    artifact_provenances.artifact_id
+  FROM
+    scan_results sr,
+    (SELECT *
```
nnobelis, line 29: Could you use named columns instead of `*`, to ease the readability of the query ?
Same underneath.

oheger-bosch: Hm, probably yes. But what is done here is actually creating a temporary view that joins the `package_provenances` table with the `remote_artifacts` table. Then, in the outer SELECT, some fields from this view are selected. When using named columns, this would mean to either list all the columns of the involved tables (which is quite a number) or only the subset that is selected (which is somehow redundant to the outer `SELECT` clause). I find the current variant indeed more readable.

nnobelis: I thought about listing only the columns you need e.g. `url`, `hash_value` and `hash_algorithm` (and 4 in the second case). This would be a bit faster too.

However since it's code that is run only one time, I am fine with the way it is.

nnobelis (approved):

PR #1238, opened 2024-10-16
ui: Add runs overview page for admins

lamppu:
Add runs overview page for superusers to get a quick overview of the runs on the server instance.

Screenshot from 2024-10-21 11-09-30

Screenshot from 2024-10-21 11-09-59

Screenshot from 2024-10-16 09-33-17

### DIFF
--- a/ui/src/routes/_layout/admin/runs/index.tsx
+++ b/ui/src/routes/_layout/admin/runs/index.tsx
@@ -1,0 +1,326 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import { createFileRoute, Link } from '@tanstack/react-router';
+import {
+  createColumnHelper,
+  getCoreRowModel,
+  useReactTable,
+} from '@tanstack/react-table';
+import { View } from 'lucide-react';
+
+import {
+  useOrganizationsServiceGetOrganizationById,
+  useProductsServiceGetProductById,
+  useRepositoriesServiceGetRepositoryById,
+  useRunsServiceGetOrtRuns,
+} from '@/api/queries';
+import { prefetchUseRunsServiceGetOrtRuns } from '@/api/queries/prefetch';
+import { OrtRunSummary } from '@/api/requests';
+import { DataTable } from '@/components/data-table/data-table';
+import { DataTableToolbar } from '@/components/data-table/data-table-toolbar';
+import { FilterMultiSelect } from '@/components/data-table/filter-multi-select';
+import { LoadingIndicator } from '@/components/loading-indicator';
+import { OrtRunJobStatus } from '@/components/ort-run-job-status';
+import { RunDuration } from '@/components/run-duration';
+import { ToastError } from '@/components/toast-error';
+import { Badge } from '@/components/ui/badge';
+import { Button } from '@/components/ui/button';
+import {
+  Card,
+  CardContent,
+  CardDescription,
+  CardHeader,
+  CardTitle,
+} from '@/components/ui/card';
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from '@/components/ui/tooltip';
+import { config } from '@/config';
+import { getStatusBackgroundColor } from '@/helpers/get-status-class';
+import { toast } from '@/lib/toast';
+import { getOrtDateTimeString } from '@/lib/utils';
+import { ortRunStatus, paginationSchema, statusSchema } from '@/schemas';
+
+const defaultPageSize = 10;
+const pollInterval = config.pollInterval;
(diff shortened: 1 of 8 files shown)


mmurto (changes requested):

ui/src/routes/_layout/runs/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+  },
+  {
+    accessorKey: 'createdAt',
+    header: 'Created At',
+    cell: ({ row }) => (
+      <div>
+        {new Date(row.original.createdAt).toLocaleString(navigator.language)}
+      </div>
+    ),
+    size: 100,
```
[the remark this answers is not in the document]
lamppu: The default min-width for a column is 150 px so it lowers that, so that some of the columns are a bit narrower and the columns fit on the current width that we have on the layout. I still changed it for a couple of the columns cause I noticed the table was actually still overflowing even though it looked ok.

ui/src/routes/_layout/create-organization.tsx
```diff
@@ -139,5 +142,12 @@ const CreateOrganizationPage = () => {
 };
 
 export const Route = createFileRoute('/_layout/create-organization')({
-  component: CreateOrganizationPage,
+  component: function Layout() {
+    const user = useUser();
+    return (
+      <PageLayout sections={rootSections(user.hasRole(['superuser']))}>
+        <CreateOrganizationPage />
+      </PageLayout>
```
mmurto, line 150: Generally this sort of layout sharing should be done by creating a `_adminSidebarLayout` etc. directory in the current `_layout` directory, add a `route.tsx` there with the shared PageLayout with the same sections and move `create-organization.tsx` and `index.tsx` there (or handling it with file names like `_adminSidebarLayout.route.tsx`, `_adminSidebarLayout.index.tsx` and `_adminSidebarLayout.create-organization.tsx` directly in _layout directory). That ensures that all shared layouts are handled the same way, so no need to go through and process what `rootSections()` etc. do.

Here I however wonder whether we could and should just omit the sidebar sections from create-organization page? I think with the current amount of links in the sidebar, it's not really important to be able to select from those in there. As `runs/index` is also available only for superusers, should it maybe be under the admn route and have a link to it there?

mmurto (changes requested):

ui/src/routes/_layout/create-organization.tsx
lamppu: Ah ok, I understood it wrong then. Well, the thing is, if you omit it from the create organization page, then there will be this jump in the layout when accessing that page, so that's mainly why I would add it there.

>As runs/index is also available only for superusers, should it maybe be under the admin route and have a link to it there?

So to move it to the admin dashboard? I figured this might be something that would be good to be more easily accessible from the front page, but I'm not sure. Or did you mean just change the link?

mmurto: > Ah ok, I understood it wrong then. Well, the thing is, if you omit it from the create organization page, then there will be this jump in the layout when accessing that page, so that's mainly why I would add it there.
> 
> > As runs/index is also available only for superusers, should it maybe be under the admin route and have a link to it there?
> 
> So to move it to the admin dashboard? I figured this might be something that would be good to be more easily accessible from the front page, but I'm not sure. Or did you mean just change the link?

Yeah exactly, move it to the admin dashboard, probably as a sub-page there. I think as it's a superuser functionality it doesn't really need to be that easily accessible from the front page.

Some layout jumping is totally fine if the pages have different layout. I think it also maybe doesn't make sense to show the sidebar on the front-page, as normal users see only one item there for the page that they're already on, so it provides no value for the regular user.

lamppu: > Yeah exactly, move it to the admin dashboard, probably as a sub-page there. 

Alright, I guess I'll change this to a draft for a bit and do some refactoring on the admin dashboard layout first to align it with the rest of the ui before adding this there.

ui/src/routes/_layout/admin/runs/index.tsx
```diff
@@ -0,0 +1,326 @@
...
+    cell: ({ row }) =>
+      row.original.finishedAt ? (
+        <>{getOrtDateTimeString(row.original.finishedAt)}</>
+      ) : (
+        <span className='italic'>Not finished yet</span>
+      ),
+    size: 95,
+  }),
+  columnHelper.accessor('status', {
+    header: () => <div>Run Status</div>,
```
mmurto, line 141: This should be a bare string.

ui/src/routes/_layout/admin/runs/index.tsx
```diff
@@ -0,0 +1,326 @@
...
+      <Badge
+        className={`border ${getStatusBackgroundColor(row.original.status)} hover:bg-${getStatusBackgroundColor(row.original.status)}`}
+      >
+        {row.original.status}
+      </Badge>
+    ),
+  }),
+  columnHelper.display({
+    id: 'jobStatuses',
+    header: () => <div>Job Status</div>,
```
mmurto, line 152: This should be a bare string.

ui/src/routes/_layout/admin/runs/index.tsx
```diff
@@ -0,0 +1,326 @@
...
+        productId={row.original.productId.toString()}
+        repoId={row.original.repositoryId.toString()}
+        runIndex={row.original.index.toString()}
+      />
+    ),
+    size: 100,
+  }),
+  columnHelper.display({
+    id: 'duration',
+    header: () => <div>Duration</div>,
```
mmurto, line 167: This should be a bare string.

ui/src/routes/_layout/admin/runs/index.tsx
```diff
@@ -0,0 +1,326 @@
...
+        createdAt={row.original.createdAt}
+        finishedAt={row.original.finishedAt ?? undefined}
+        pollInterval={pollInterval}
+      />
+    ),
+    size: 100,
+  }),
+  columnHelper.display({
+    id: 'actions',
+    header: () => <div>Actions</div>,
```
mmurto, line 179: This should be a bare string.

mmurto (changes requested):

mmurto (approved):

Issue #1220, opened 2024-10-14
Link menu is not updated correctly when opening a new run via global run id search function


hanna-modica:
If I am viewing a run in the UI and then go to the search function in the top right "Enter global run ID" and enter a different id, I get a view of the new run, however the links in the menu (overview, packages, etc.) are still pointing to the previously shown run.

![image](https://github.com/user-attachments/assets/2f84da7c-4ddf-41b7-b058-633c8f5f8774)

PR #1242, opened 2024-10-17
fix(ui): Ensure path params are updated in links

lamppu:
Call `useParams` so that the correct path parameters are used in the sidebar links.

Resolves https://github.com/eclipse-apoapsis/ort-server/issues/1220.


mmurto (approved):
This fixes the problem, but as a related issue, we should at some point make the navigation props for sidebar type-safe - they currently just take in strings. That should help us catch this sort of bugs by not allowing `to` if `params` is also required.

PR #1245, opened 2024-10-17
ui: Align admin dashboard layout with the rest of the UI

lamppu:

Screenshot from 2024-10-18 12-47-31
Screenshot from 2024-10-18 12-47-52
Screenshot from 2024-10-18 12-48-06

### DIFF
--- a/ui/src/routes/_layout/admin/index.tsx
+++ b/ui/src/routes/_layout/admin/index.tsx
@@ -17,7 +17,6 @@
  * License-Filename: LICENSE
  */
 
-import { zodResolver } from '@hookform/resolvers/zod';
 import { createFileRoute, Link } from '@tanstack/react-router';
 import {
   ColumnDef,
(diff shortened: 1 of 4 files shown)


lamppu:
Now that I'm refactoring the admin page layout, I wonder what should be done about the organizations table in the admin page, should it stay there, and should I refactor it to use the `DataTable` component to match the style of the other tables, or should I just leave it as it is now, because it might remain completely redundant and could be removed at some point, so perhaps not worth putting time on that? Any thoughts on this @mmurto @sschuberth @mnonnenmacher, here's also the link to the original PR with some discussion on the table https://github.com/eclipse-apoapsis/ort-server/pull/358.

Screenshot from 2024-10-17 16-20-45

mnonnenmacher:
If it does not offer any functionality that the other organization view does not have, I see no reason to keep it.

mmurto (approved):

PR #1264, opened 2024-10-21
Fix format date

sschuberth:


ui/src/lib/utils.ts
```diff
@@ -24,9 +24,34 @@ export function cn(...inputs: ClassValue[]) {
...
+    expect(formatDate('2024-06-11T13:07:45Z', 'en-US')).toBe(
+      '06/11/2024, 03:07:45 PM'
+    );
+
+    expect(formatDate('2024-06-11T13:07:45Z', 'de-DE')).toBe(
+      '11.06.2024, 15:07:45'
+    );
+
+    expect(formatDate('2024-06-11T13:07:45Z', 'fi-FI')).toBe(
+      '11.06.2024 klo 15.07.45'
```
sschuberth, line 54: @lamppu, from your screenshot at https://github.com/eclipse-apoapsis/ort-server/pull/1238 I see that a Finnish date / time without "klo" is used. Are you using a different locale than `fi-FI` there? because otherwise I would have no idea how to get rid of the "klo" here.

mmurto:
I believe the formatting was already fixed in https://github.com/eclipse-apoapsis/ort-server/pull/1238.

ui/src/lib/utils.ts
lamppu: Ok, um weird, it is `fi-FI`

lamppu: Looks like it could be browser specific, for example in Firefox it doesn't add the "klo" but in Chrome it does

sschuberth: Oh, wow. Thanks for checking!

mmurto:
> > I believe the formatting was already fixed in #1238.
> 
> Oh, right, let me still add that test then.
> 
> Still the question remains whether you see Finnish time now with or without "klo", and how to eventually adjust the test accordingly.

I'd say it doesn't matter whether it's with or without klo in Finnish, so as long as the test passes, we can leave it to the browsers to fight over. :D

mmurto (approved):

PR #1269, opened 2024-10-22
ui: Add package manager options to ORT Run creation form

Etsija:
This PR adds the ability to also set options for package managers as part of the analyzer job configuration. When using the rerun functionality, the options are copied to the new run and in the analyzer configuration section of the form, all package managers for which options are set are opened by default.

Setting options for package managers:
Screenshot from 2024-10-22 12-34-57

Options are also copied to the job configurations part of ORT run results:
Screenshot from 2024-10-22 10-35-51

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/-components/package-manager-field.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/-components/package-manager-field.tsx
@@ -1,0 +1,271 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


sschuberth:
Can we keep the checkbox in the same line as the package manager name, and not align it vertically across all options?

Etsija:
> And a question: Are PM options maintained when rerunning? My expectation would be that they are.

Yes. By default no options are set, but as said in the description and the corr. commit message, they are maintained when rerunning.

Etsija:
> Can we keep the checkbox in the same line as the package manager name, and not align it vertically across all options?

I am not sure, I think I need to consult @mmurto as to how to change the DOM to do this.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/-components/package-manager-field.tsx
```diff
@@ -0,0 +1,286 @@
...
+    | 'Maven'
+    | 'NPM'
+    | 'NuGet'
+    | 'PIP'
+    | 'Pipenv'
+    | 'PNPM'
+    | 'Poetry'
+    | 'Yarn'
+    | 'Yarn2';
+};
```
mmurto, line 182: Is this the same as `PackageManagerId`? If so, it should be used.

mmurto (changes requested):

mmurto (approved):

PR #1271, opened 2024-10-22
Deduplicate issues

oheger-bosch:
This PR further refactors the database model for issues to deduplicate entities, so that issues with matching properties are reused. It also contains a migration script to remove existing duplicates. Refer to the single commits for further information.

### DIFF
--- a/core/src/main/resources/application.conf
+++ b/core/src/main/resources/application.conf
@@ -85,6 +85,7 @@ database {
   sslCert = ${?DB_SSL_CERT}
   sslKey = ${?DB_SSL_KEY}
   sslRootCert = ${?DB_SSL_ROOT_CERT}
+  initSqlStatement = ${?DB_INIT_SQL_STATEMENT}
 }
 
 keycloak {
--- a/dao/src/main/kotlin/tables/runs/shared/IssuesTable.kt
+++ b/dao/src/main/kotlin/tables/runs/shared/IssuesTable.kt
@@ -29,6 +29,7 @@ import org.eclipse.apoapsis.ortserver.model.runs.Issue
 
 import org.jetbrains.exposed.dao.LongEntity
 import org.jetbrains.exposed.dao.id.EntityID
+import org.jetbrains.exposed.sql.and
 
 /**
  * A table to represent an ort issue.
@@ -42,8 +43,19 @@ object IssuesTable : SortableTable("issues") {
 
 class IssueDao(id: EntityID<Long>) : LongEntity(id) {
     companion object : SortableEntityClass<IssueDao>(IssuesTable) {
+        fun findByIssue(issue: Issue): IssueDao? = find {
+            IssuesTable.issueSource eq issue.source and
+                    (IssuesTable.message eq issue.message) and
+                    (IssuesTable.severity eq issue.severity) and
+                    (IssuesTable.affectedPath eq issue.affectedPath)
+        }.firstOrNull()
+
+        /**
+         * Return an [IssueDao] to represent the given [issue]. If the properties of the [issue] can be matched,
+         * an existing [IssueDao] is returned, otherwise a new one is created.
+         */
         fun createByIssue(issue: Issue): IssueDao =
-            new {
+            findByIssue(issue) ?: new {
                 source = issue.source
                 message = issue.message
                 severity = issue.severity
--- a/dao/src/main/kotlin/tables/runs/shared/IssuesTable.kt
+++ b/dao/src/main/kotlin/tables/runs/shared/IssuesTable.kt
@@ -21,6 +21,7 @@ package org.eclipse.apoapsis.ortserver.dao.tables.runs.shared
 
 import kotlinx.datetime.Instant
 
+import org.eclipse.apoapsis.ortserver.dao.utils.DigestFunction
 import org.eclipse.apoapsis.ortserver.dao.utils.SortableEntityClass
 import org.eclipse.apoapsis.ortserver.dao.utils.SortableTable
 import org.eclipse.apoapsis.ortserver.model.Severity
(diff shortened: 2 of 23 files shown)


dao/src/main/resources/db/migration/V83__scannerSummariesIssuesTimestamps.sql
```diff
@@ -0,0 +1,29 @@
...
+-- Since the timestamp column is not nullable, create a new table that is populated and then renamed.
+CREATE TABLE scan_summaries_issues_with_timestamps
+(
+    id                bigserial                         PRIMARY KEY,
+    scan_summary_id   bigint REFERENCES scan_summaries  NOT NULL,
+    issue_id          bigint REFERENCES issues          NOT NULL,
+    timestamp         timestamp                         NOT NULL
+);
+
+INSERT INTO scan_summaries_issues_with_timestamps
```
nnobelis, line 14: Do we need some indexes on this table ?
Maybe also some foreign key constraints ?

dao/src/main/kotlin/tables/runs/shared/IssuesTable.kt
```diff
@@ -42,8 +43,19 @@ object IssuesTable : SortableTable("issues") {
 
 class IssueDao(id: EntityID<Long>) : LongEntity(id) {
     companion object : SortableEntityClass<IssueDao>(IssuesTable) {
+        fun findByIssue(issue: Issue): IssueDao? = find {
+            IssuesTable.issueSource eq issue.source and
+                    (IssuesTable.message eq issue.message) and
```
nnobelis: Micro optimization: Check the `message` last as it is potentially the longest column

dao/src/main/kotlin/tables/runs/shared/IssuesTable.kt
```diff
@@ -45,7 +47,7 @@ class IssueDao(id: EntityID<Long>) : LongEntity(id) {
     companion object : SortableEntityClass<IssueDao>(IssuesTable) {
         fun findByIssue(issue: Issue): IssueDao? = find {
             IssuesTable.issueSource eq issue.source and
-                    (IssuesTable.message eq issue.message) and
+                    (DigestFunction(IssuesTable.message) eq DigestFunction(stringLiteral(issue.message))) and
```
nnobelis, line 50: Same here, I would put the test with the function call as the last one.

core/src/main/resources/application.conf
nnobelis, line 1: Running any SQL from configuration sounds insecure. Don't you want to put some sanitization here ?

nnobelis, line 1: Please mention in the commit message that this set `connectionInitSql` from Hikari, and put a link to a doc, e.g.:
https://github.com/brettwooldridge/HikariCP/blob/604d7d5838e96afbbdd4e19eedab7497cf368eb6/README.md?plain=1#L312

dao/src/main/resources/db/migration/V83__scannerSummariesIssuesTimestamps.sql
oheger-bosch: Foreign key constraints are in place due to the `REFERENCES` keywords. I am not sure whether we need an index, and if yes, on which columns. Maybe we should monitor the performance in production.

dao/src/main/kotlin/tables/runs/shared/IssuesTable.kt
oheger-bosch: From this declaration, an SQL statement is generated. I would expect that the database constructs its own optimized execution plan and that the change of the order does not have any effect.

dao/src/main/kotlin/tables/runs/shared/IssuesTable.kt
oheger-bosch: As above.

dao/src/main/resources/db/migration/V83__scannerSummariesIssuesTimestamps.sql
nnobelis: With Postgres you generally need an index on the foreign key columns i.e. `scan_summary_id` and `scan_summary_id`.

However since the previous indexes caused performance problem, I agree to can wait to see how it behaves in production.

dao/src/main/kotlin/tables/runs/shared/IssuesTable.kt
nnobelis: This is wishful thinking :)

Better explained:

> If none of the conditions apply, the order the conditions are specified may affect the order of evaluation, but since the database is going to have to fetch every single record to satisfy the query, the time spent fetching will likely dwarf the time spent evaluating the conditions.

dao/src/test/kotlin/DatabaseTest.kt
```diff
@@ -63,7 +63,8 @@ class DatabaseTest : WordSpec({
                 sslMode = "myTestSSLMode",
                 sslCert = "myTestSSLCert",
                 sslKey = "myTestSSLKey",
-                sslRootCert = "myTestSSLRootCert"
+                sslRootCert = "myTestSSLRootCert",
+                initSqlStatement = "test,public"
```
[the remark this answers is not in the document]
oheger-bosch: Changed.

core/src/main/resources/application.conf
oheger-bosch: The configuration is under control of the server admins. So, they are responsible for setting correct values. This applies to every configuration setting.

nnobelis: Will there be a follow up PR in the infrastructure repository ?

nnobelis (approved):

sschuberth:
@oheger-bosch, this PR causes problems on our side:

```
ERROR: update or delete on table "issues" violates foreign key constraint "identifiers_ort_issues_ort_issue_id_fkey" on table "identifiers_issues"
  Detail: Key (id)=(96) is still referenced from table "identifiers_issues".
```

Any idea how to fix that?

mmurto:
It's unfortunate that the migration was split into two (83 and 84) but the first one already requires the changes in this PR to work. Migration 83 was executed successfully by our deployment pipeline, which rendered old deployments unable to server requests, but as migration 84 failed, our deployment pipeline hasn't deployed the changes in this PR to the application code. We probably should take good care to ensure that we rollback all migrations of a PR if some fail.

PR #1277, opened 2024-10-23
Upgrade Java version for ORT Server to 21

oheger-bosch:
Build all artifacts with Java 21.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -20,12 +20,6 @@ jobs:
     - name: Checkout Repository
       uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4
 
-    - name: Setup Java
-      uses: actions/setup-java@b36c23c0d998641eff861008f374ee103c25ac73 # v4
-      with:
-        distribution: temurin
-        java-version: 17
-
     - name: Setup Gradle
       uses: gradle/actions/setup-gradle@d156388eb19639ec20ade50009f3d199ce1e2808 # v4
 
@@ -37,12 +31,6 @@ jobs:
     steps:
     - name: Checkout Repository
       uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4
-
-    - name: Setup Java
-      uses: actions/setup-java@b36c23c0d998641eff861008f374ee103c25ac73 # v4
-      with:
-        distribution: temurin
-        java-version: 17
 
     - name: Setup Gradle
       uses: gradle/actions/setup-gradle@d156388eb19639ec20ade50009f3d199ce1e2808 # v4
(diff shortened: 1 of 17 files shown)


gradle/libs.versions.toml
```diff
@@ -18,7 +18,7 @@
 [versions]
 # Docker
 # When updating this version make sure to keep it in sync with the worker Dockerfiles.
-eclipseTemurin = "17.0.12_7-jdk-jammy@sha256:56c8b3a8f4299e0190095333cf85e647e810bdf167e59f224a1fd432e47a8f12"
+eclipseTemurin = "21.0.4_7-jdk-jammy@sha256:0472478e22da0f66043fa6acd8cd30126592349f47937adafc2340794e5bf06a"
```
sschuberth, line 21: Maybe link to https://github.com/docker-library/docs/blob/master/eclipse-temurin/README.md#supported-tags-and-respective-dockerfile-links in the commit message for reference?

.github/workflows/build-and-test.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       uses: actions/setup-java@b36c23c0d998641eff861008f374ee103c25ac73 # v4
       with:
         distribution: temurin
-        java-version: 17
+        java-version: 21
```
sschuberth, line 27: What about `java-version: '11'` in `.github/workflows/docker-build.yml`?

buildSrc/src/main/kotlin/ort-server-kotlin-jvm-conventions.gradle.kts
```diff
@@ -33,16 +35,22 @@ plugins {
     id("org.jetbrains.kotlin.jvm")
 }
 
-val javaVersion = JavaVersion.current()
-val maxKotlinJvmTarget = runCatching { JvmTarget.fromTarget(javaVersion.majorVersion) }
-    .getOrDefault(enumValues<JvmTarget>().max())
+java {
+    toolchain {
```
sschuberth, line 39: If you take a toolchain into use, then the previous commit could have simply omitted setting up Java, as explained at https://github.com/oss-review-toolkit/ort/commit/494a32434823a3f0190db4e8e128b2efb223a14a.

sschuberth: And note that also the Java version used to run the Gradle build itself needs to be configured, as the build uses ORT's Detekt rules which are also built with Java 21, see https://github.com/oss-review-toolkit/ort-workbench/commit/b800aaac9fc2c36b3246f579f1ed5a48b712ef02.

.github/workflows/build-and-test.yml
oheger-bosch: I also switched this to 21. I was not sure whether there is a specific reason why here a different Java version should be used. Maybe an oversight?

buildSrc/src/main/kotlin/ort-server-kotlin-jvm-conventions.gradle.kts
oheger-bosch: Thanks, this should simplify the workflows.

.github/workflows/build-and-test.yml
sschuberth: I guess so.

sschuberth (approved):

sschuberth (approved):

PR #1303, opened 2024-10-28
fix(storage): Don't use new context for storage

### LINKED ISSUE #1295: Scanner gets stuck on creating file lists
Commit https://github.com/eclipse-apoapsis/ort-server/commit/851212d2458e28343b0b34e27b7d0182b0be9bc4 seems to have introduced a bug in Scanner Worker. The bug can be reproduced by creating run for https://github.com/jshttp/mime-types on revision 4b3ae38 (just to have reproducible test, the run fails on other repositories and revisions as well) and running the scanner. The scanner scans everything and gets stuck with the last console log being `2024-10-25 07:58:28.902 [DefaultDispatcher-worker-13] [traceId=8d946210-b405-4d74-8170-942e3508ec8a, ortRunId=3, component=scanner-worker, scannerJobId=3] level=INFO  kotlinx.coroutines.CoroutineScope - Creating file list for provenance 64 of 366.`. The scan works on the commit before.

The bug doesn't seem to affect very small repositories, for example https://github.com/vdurmont/semver4j.git is not affected.

mmurto:
Using a separate context makes the scanner hang indefinitely when
creating file lists. Remove the context to fix this.

Fixes #1295.

### DIFF
--- a/storage/spi/src/main/kotlin/Storage.kt
+++ b/storage/spi/src/main/kotlin/Storage.kt
@@ -23,9 +23,6 @@ import java.io.ByteArrayInputStream
 import java.io.InputStream
 import java.util.ServiceLoader
 
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.withContext
-
 import org.eclipse.apoapsis.ortserver.config.ConfigManager
 import org.eclipse.apoapsis.ortserver.config.Path
 import org.eclipse.apoapsis.ortserver.utils.config.getStringOrNull
@@ -138,11 +135,7 @@ class Storage(
     private suspend fun <T> wrapException(block: suspend StorageProvider.() -> T): T =
         @Suppress("TooGenericExceptionCaught")
         try {
-            // To prevent the StorageException being suppressed by the coroutine exception handler, a new context is
-            // created here.
-            withContext(Dispatchers.IO) {
-                provider.block()
-            }
+            provider.block()
         } catch (e: Exception) {
             throw StorageException("Exception from StorageProvider.", e)
         }


storage/spi/src/main/kotlin/Storage.kt
```diff
@@ -138,11 +135,7 @@ class Storage(
     private suspend fun <T> wrapException(block: suspend StorageProvider.() -> T): T =
         @Suppress("TooGenericExceptionCaught")
         try {
-            // To prevent the StorageException being suppressed by the coroutine exception handler, a new context is
-            // created here.
-            withContext(Dispatchers.IO) {
-                provider.block()
-            }
+            provider.block()
```
sschuberth, line 138: Should we add a `TODO` here to not forget about the issue off suppressed `StorageException`?

sschuberth (approved):
Approving despite my question.

storage/spi/src/main/kotlin/Storage.kt
mmurto: I created #1304 instead.

nnobelis (changes requested):
As the comment said, this context was there for a reason.

I assume you'll get a CI test failure without it.

nnobelis:
> > As the comment said, this context was there for a reason.
> > I assume you'll get a CI test failure without it.
> 
> Yep, the test fails. Do you have ideas how to solve this with the context? If no, having the exception swallowed is preferrable to scan getting stuck.

When an exception happens in a coroutine, it suppresses all the other exceptions therefore you cannot wrap it anymore.

One of the goal of the `Storage` class was to wrap the errors of the storage provider implementations to abstract them.
When all the storage API were migrated with coroutines, this behavior didn't work anymore without context.

It tried several things but unfortunately I am no expert of coroutines.
@sschuberth  Would you have an idea ?

nnobelis:
See also https://kotlinlang.org/docs/exception-handling.html#exceptions-aggregation.

sschuberth:
> If there are no quick and easy solutions, I suggest reverting the whole commit, or what do you say?

I agree that's certainly the better option than having all scans stuck.

nnobelis:
> If there are no quick and easy solutions, I suggest reverting the whole commit, or what do you say?

And making the storage API non-suspendable again ? I will check with the team and get back to you.

sschuberth:
> It tried several things but unfortunately I am no expert of coroutines.
> @sschuberth Would you have an idea ?

I'm no expert here either. Probably @oheger-bosch or @mnonnenmacher would now more. But in general I can recommend to read https://www.lukaslechner.com/why-exception-handling-with-kotlin-coroutines-is-so-hard-and-how-to-successfully-master-it/.

mmurto:
> > If there are no quick and easy solutions, I suggest reverting the whole commit, or what do you say?
> 
> And making the storage API non-suspendable again ? I will check with the team and get back to you.

At least on our infrastructure all scans of big packages (my guess is that those with over 64 dependencies) get totally stuck until they're cleaned up after 24 hours as failed. Being non-suspendable may be unfortunate, but we should ensure that the `main` branch works first.

nnobelis:
OK then you can rollback my commit and we will (maybe) try to find another solution.

Could you please share with us which report you use for this hanging file list ? So I can try to reproduce it ?

sschuberth:
> Could you please share with us which report you use for this hanging file list ? So I can try to reproduce it ?

@nnobelis it's in the the issue description at https://github.com/eclipse-apoapsis/ort-server/issues/1295.

mmurto:
Closing this in favor of #1316.

PR #1305, opened 2024-10-28
fix(dao): Fix the migration to deduplicate issues

oheger-bosch:
This is a fix for #3ee6bb31. The `identifiers_issues` table needs to be processed by the migration as well.


sschuberth:
> This is a fix for #3ee6bb31.

Nit: Commit SHA1s should not be preceded by `#` in order to be properly highlighted by GitHub.

mmurto (approved):

PR #1307, opened 2024-10-28
ui: Issues page

Etsija:

An open question: I believe the issues are sorted by timestamp by default. It could make the issues table cleaner if the back-end sorting would be hard-coded in the UI into severity instead. Another possibility would be to do also sorting on client side, but this hasn't yet been implemented in this PR.

### DIFF
--- a/ui/src/helpers/get-status-class.ts
+++ b/ui/src/helpers/get-status-class.ts
@@ -17,27 +17,18 @@
  * License-Filename: LICENSE
  */
 
-// Status types
+import { JobStatus, OrtRunStatus, Severity } from '@/api/requests';
 
-// Statuses reported either by ORT Runs or the individual jobs within them
-type Status =
-  | 'CREATED'
-  | 'SCHEDULED'
-  | 'RUNNING'
-  | 'ACTIVE'
-  | 'FAILED'
-  | 'FINISHED'
-  | 'FINISHED_WITH_ISSUES'
-  | undefined;
+// Combine statuses reported either by ORT Runs or the individual jobs within them.
+export type Status = JobStatus | OrtRunStatus | undefined;
 
+// There is no vulnerability rating in the OpenApi spec, so define it here.
 export type VulnerabilityRating =
   | 'CRITICAL'
   | 'HIGH'
   | 'MEDIUM'
   | 'LOW'
   | 'NONE';
-
-type RuleViolationSeverity = 'ERROR' | 'WARNING' | 'HINT';
 
 // TailwindCSS classes matching to statuses
 //
@@ -89,10 +80,18 @@ const VULNERABILITY_RATING_BG_COLOR: {
 } as const;
 
 const RULE_VIOLATION_SEVERITY_BG_COLOR: {
-  [K in RuleViolationSeverity]: string;
+  [K in Severity]: string;
 } = {
   ERROR: 'bg-red-600',
-  WARNING: 'bg-orange-600',
+  WARNING: 'bg-amber-500',
+  HINT: 'bg-neutral-300',
+} as const;
+
+const ISSUE_SEVERITY_BG_COLOR: {
+  [K in Severity]: string;
+} = {
+  ERROR: 'bg-red-600',
+  WARNING: 'bg-amber-500',
   HINT: 'bg-neutral-300',
 } as const;
 
@@ -136,7 +135,12 @@ export function getVulnerabilityRatingBackgroundColor(
 
 // Get the color class for coloring the background of rule violation severities
 export function getRuleViolationSeverityBackgroundColor(
-  rating: RuleViolationSeverity
+  severity: Severity
 ): string {
-  return RULE_VIOLATION_SEVERITY_BG_COLOR[rating];
+  return RULE_VIOLATION_SEVERITY_BG_COLOR[severity];
+}
+
+// Get the color class for coloring the background of issue severities
+export function getIssueSeverityBackgroundColor(severity: Severity): string {
+  return ISSUE_SEVERITY_BG_COLOR[severity];
 }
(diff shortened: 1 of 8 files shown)


ui/src/helpers/get-status-class.ts
```diff
@@ -17,27 +17,23 @@
...
   | 'HIGH'
   | 'MEDIUM'
   | 'LOW'
   | 'NONE';
 
-type RuleViolationSeverity = 'ERROR' | 'WARNING' | 'HINT';
+// In case these differ in the future in OpenApi spec, keep them as separate types here,
+// but they could also very well be combined for simplicity.
+type RuleViolationSeverity = Severity;
+type IssueSeverity = Severity;
```
mmurto, line 36: I'd combine these as there shouldn't be any plans to separate them.

ui/src/schemas/index.ts
```diff
@@ -43,3 +43,15 @@ export const ortRunStatus: z.ZodEnum<[OrtRunStatus, ...OrtRunStatus[]]> =
...
+export const issueSeverity: z.ZodEnum<[Severity, ...Severity[]]> = z.enum([
+  'HINT',
+  'WARNING',
+  'ERROR',
+]);
+
+// Issue severity schema that is used for search parameter validation
+export const issueSeveritySchema = z.object({
+  severity: z.array(issueSeverity).optional(),
+});
```
mmurto, line 57: More idiomatic zod naming would be that `issueSeverity` is `issueSeveritySchema` and the current `issueSeveritySchema` something different.

ui/src/test-data/mock-up-issues.tsx
```diff
@@ -0,0 +1,242 @@
...
+ */
+
+// Some mock-up data to test the functionality of the Issues data table.
+// Data was created with ChatGPT, and for most part, includes made up
+// random data, not reflecting the real data that would be returned from
+// an ORT run.
+
+import { Severity } from '@/api/requests';
+
+export const mockUpIssues = {
```
mmurto, line 27: As we don't have this sort of data anywhere else, I wouldn't commit it here either.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/issues/index.tsx
```diff
@@ -18,67 +18,282 @@
...
+  row: Row<Issue>,
+  columnId: string,
+  filterValue: string[]
+): boolean => {
+  // TypeScript can't infer the type, so cast it explicitly.
+  // This is safe because the columnId is always one of the
+  // severities, which themselves are strings.
+  const rowValue = row.getValue(columnId) as string;
+  return filterValue.includes(rowValue);
+};
```
mmurto, line 81: As this is only used in one column, I would move the declaration in-line to the column definition, mostly in hope to gain better type-safety.

mmurto (changes requested):

mmurto:
> Another possibility would be to do also sorting on client side, but this hasn't yet been implemented in this PR.

Doesn't this already do sorting client-side, or did I misread?

Etsija:
> > Another possibility would be to do also sorting on client side, but this hasn't yet been implemented in this PR.
> 
> Doesn't this already do sorting client-side, or did I misread?

No sorting on client-side is implemented.

mmurto:
> > > Another possibility would be to do also sorting on client side, but this hasn't yet been implemented in this PR.
> > 
> > 
> > Doesn't this already do sorting client-side, or did I misread?
> 
> No sorting on client-side is implemented.

Ah yeah, sorting is not implemented at all, so not server-side either.

ui/src/schemas/index.ts
Etsija: I mimicked the naming from Johanna's definitions earlier in that file, so they are in line with each other.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/issues/index.tsx
Etsija: It didn't, though.

mmurto: Looking at it here, does it just filter the exact values? Would some of the built-in functions work like:

```typescript
    filterFn: 'equalsString',
```

Etsija: > Looking at it here, does it just filter the exact values? Would some of the built-in functions work like:
> 
> ```typescript
>     filterFn: 'equalsString',
> ``` 

No, it filters against an array of possible severities, like `[WARNING, ERROR]` and there are no built-in functions for that. There are three functions that do the exact opposite - filter an array of values against an option - but no built-in functions for filtering a value against an array of options. All the solutions I could find used `includes()`.

mmurto: Thanks, makes sense.

mmurto: You can make it more type-safe like this:

```typescript
    filterFn: (
      row,
      _columnId,
      filterValue
    ): boolean => {
      return filterValue.includes(row.original.severity);
    },
```

Etsija: That was a nice simplifying fix, thank you!

mmurto (approved):

PR #1309, opened 2024-10-28
fix(api): Optimize performance by reducing ORT run data fetching

wkl3nk:
Limit the fetched data for repository runs to only what is needed for the REST API response,
enhancing performance by avoiding excess loading of unused related entities.

Utilize lazy-loading in the Kotlin Exposed framework by no longer prematurely mapping of unused columns to object attributes.

Number of generated SQL statements in the BE is reduced at least by 50%.

### DIFF
--- a/core/src/main/kotlin/api/RepositoriesRoute.kt
+++ b/core/src/main/kotlin/api/RepositoriesRoute.kt
@@ -33,11 +33,9 @@ import io.ktor.server.routing.Route
 import io.ktor.server.routing.route
 
 import org.eclipse.apoapsis.ortserver.api.v1.mapping.mapToApi
-import org.eclipse.apoapsis.ortserver.api.v1.mapping.mapToApiSummary
 import org.eclipse.apoapsis.ortserver.api.v1.mapping.mapToModel
 import org.eclipse.apoapsis.ortserver.api.v1.model.CreateOrtRun
 import org.eclipse.apoapsis.ortserver.api.v1.model.CreateSecret
-import org.eclipse.apoapsis.ortserver.api.v1.model.JobSummaries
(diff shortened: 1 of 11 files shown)


dao/src/main/kotlin/tables/OrtRunsTable.kt
```diff
@@ -114,4 +116,36 @@ class OrtRunDao(id: EntityID<Long>) : LongEntity(id) {
         traceId = traceId,
         environmentConfigPath = environmentConfigPath
     )
+
+    /**
+     * Map this [OrtRunDao] to an [OrtRunSummary] which is a reduced representation of an [OrtRun].
+     * Reduce the load on the database by utilizing lazy loading (only actually required fields are loaded).
```
oheger-bosch, line 122: This comment about lazy loading confuses me, since I somehow was expecting clever tweaks on the Exposed configuration to change the loading behavior of associations or whatever. But in fact, you just access a subset of the entity's properties. So, could you just state something like "This prevents that unnecessary data is loaded from the database." ?

core/src/main/kotlin/api/RepositoriesRoute.kt
```diff
@@ -33,11 +33,9 @@ import io.ktor.server.routing.Route
 import io.ktor.server.routing.route
```
oheger-bosch, line 33: Commit message: This should probably be "excessive loading".

dao/src/main/kotlin/tables/OrtRunsTable.kt
wkl3nk: Yes, there is nothing that you need to actively do to make use of lazy loading, you just must not make the mistake to kind of "touch" attributes that would cause Exposed to load related data from other database tables.

oheger-bosch (approved):

PR #1317, opened 2024-10-29
fix(storage): Replace `withContext` by `supervisorScope`

oheger-bosch:
In order to obtain the correct exception thrown by a provider, the code was using the `withContext` function. This, however, caused problems with hanging jobs. Therefore, use `supervisorScope` instead.

### DIFF
--- a/storage/spi/src/main/kotlin/Storage.kt
+++ b/storage/spi/src/main/kotlin/Storage.kt
@@ -23,8 +23,7 @@ import java.io.ByteArrayInputStream
 import java.io.InputStream
 import java.util.ServiceLoader
 
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.withContext
+import kotlinx.coroutines.supervisorScope
 
 import org.eclipse.apoapsis.ortserver.config.ConfigManager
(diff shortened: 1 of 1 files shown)


mmurto:
A run succeeded with this fix, so I believe #1295 should be and can be considered fixed if this is merged.

oheger-bosch:
> A run succeeded with this fix, so I believe #1295 should be and can be considered fixed if this is merged.

Glad to hear this. Is a single run sufficient or do you think some more testing is needed?

mmurto:
> > A run succeeded with this fix, so I believe #1295 should be and can be considered fixed if this is merged.
> 
> Glad to hear this. Is a single run sufficient or do you think some more testing is needed?

When bisecting the issue, it consistently failed without exceptions on large repos. I nevertheless started another scan, so I'll report back after the few hours it runs for.

sschuberth (approved):

sschuberth:
> I'll report back after the few hours it runs for.

Being an optimist (🙄) I've put it into the merge queue already nonetheless.

mmurto:
My second scan succeeded as well.

PR #1319, opened 2024-10-29
feat(cli): Add initial ORT Server CLI implementation

bs-ondem:
This commit introduces the new module `cli` for the ORT Server, intended to contain all upcoming implementations related to the CLI.

### DIFF
--- a/cli/build.gradle.kts
+++ b/cli/build.gradle.kts
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import com.google.cloud.tools.jib.gradle.JibTask
+
+val dockerImagePrefix: String by project
+val dockerImageTag: String by project
+
+plugins {
+    application
+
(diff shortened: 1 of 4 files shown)


settings.gradle.kts
```diff
@@ -25,6 +25,7 @@ rootProject.name = "ort-server"
 
 include(":api:v1:mapping")
 include(":api:v1:model")
+include("cli")
```
sschuberth, line 28: This should have a leading ":" to make it "absolute".

sschuberth:
> all upcoming implementations related to the CLI.

Interesting, I wasn't aware such a thing being planned. Can you share more about the use-case and planned commands?

bs-ondem:
The CLI will simplify the integration of the ORT Server in a CI/CD environment for our customers, especially starting an ORT run in a blocking mode. To replace our proprietary solution with the ORT Server CLI, we need the following features:
- Start an ORT run
- Get details about an ORT run
- Download reports
- Download logs

cli/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+
+jib {
+    from.image = "eclipse-temurin:${libs.versions.eclipseTemurin.get()}"
+    to.image = "${dockerImagePrefix}ort-server-cli:$dockerImageTag"
+
+    container {
+        mainClass = "org.eclipse.apoapsis.ortserver.cli.OrtServerMainKt"
+        creationTime.set("USE_CURRENT_TIMESTAMP")
+    }
+}
```
sschuberth, line 58: Could the CLI be included in any Docker image? So is JIB needed after all?

bs-ondem: Basically yes, but I don't get your point to remove the Jib configuration? With the Docker image the CLI can be used from every environment (of course with Docker).

oheger-bosch: For distribution the container image is rather convenient. All CI platforms support a way to directly execute containers. A native executable would have to be installed using an alternative mechanism.
But of course, the container image does not have to be the only artifact generated by this project. The plan is further to have a client library that can be used to interact with ORT Server programmatically. The CLI is just one client of this library.

oheger-bosch (approved):

cli/build.gradle.kts
sschuberth: > The plan is further to have a client library that can be used to interact with ORT Server programmatically. The CLI is just one client of this library.

Yeah, I guess what confused me is that this CLI stub is added even before that required client library is ready.

bs-ondem: Client library will be added in the upcoming PRs. The reason to add the CLI first is just due to the order of the issues in our internal board.

PR #1320, opened 2024-10-29
ui: Tune badge colors

Etsija:
Screenshot from 2024-10-29 17-07-37

### DIFF
--- a/ui/src/routes/_layout/admin/runs/index.tsx
+++ b/ui/src/routes/_layout/admin/runs/index.tsx
@@ -141,7 +141,7 @@ const columns = [
     header: 'Run Status',
     cell: ({ row }) => (
       <Badge
-        className={`border ${getStatusBackgroundColor(row.original.status)} hover:bg-${getStatusBackgroundColor(row.original.status)}`}
+        className={`border ${getStatusBackgroundColor(row.original.status)}`}
       >
         {row.original.status}
       </Badge>
--- a/ui/src/components/ui/badge-variants.ts
+++ b/ui/src/components/ui/badge-variants.ts
@@ -14,12 +14,10 @@ export const badgeVariants = cva(
   {
     variants: {
       variant: {
-        default:
-          'border-transparent bg-primary text-primary-foreground shadow hover:bg-primary/80',
-        secondary:
-          'border-transparent bg-secondary text-secondary-foreground hover:bg-secondary/80',
+        default: 'border-transparent bg-primary text-primary-foreground shadow',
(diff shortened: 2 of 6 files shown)


sschuberth:
Can we get the usual nice screenshot? 😀

ui/src/routes/_layout/admin/runs/index.tsx
```diff
@@ -141,7 +141,7 @@ const columns = [
     header: 'Run Status',
     cell: ({ row }) => (
       <Badge
-        className={`border ${getStatusBackgroundColor(row.original.status)} hover:bg-${getStatusBackgroundColor(row.original.status)}`}
+        className={`border ${getStatusBackgroundColor(row.original.status)}`}
```
mmurto, line 144: So I would rather add `pointer-events-none` to these places where `Badge` is used and keep the colors as-is.

mmurto (changes requested):

ui/src/routes/_layout/admin/runs/index.tsx
Etsija: > So I would rather add `pointer-events-none` to these places where `Badge` is used and keep the colors as-is.

So remove them from get... color class helper functions and add them instead everywhere where a Badge is used?

mmurto: Exactly.

Etsija: > Exactly. 

So when you say "a bit more idiomatic" you really mean "blow up this tiny PR"? : 😏😏

mmurto: Sure you could include the `pointer-events-none` as a shared value in the helper function return if that makes more sense to you. The non-idiomatic part is specifying hover color when what we actually want is the color not to change.

mmurto: More options could be to create a wrapper component like `StatusBadge` that does it, or as it feels that we by-default don't want the badges to have hover color, just remove the hover colors from [here](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/components/ui/badge-variants.ts#L17-L23).

Etsija: I agree the last option would probably be best.

Etsija: Done and seems to be working.

mmurto (approved):

PR #1324, opened 2024-10-30
ui: Fix calculation of the overall vulnerability rating

Etsija:

### DIFF
--- a/ui/src/helpers/get-status-class.ts
+++ b/ui/src/helpers/get-status-class.ts
@@ -23,12 +23,13 @@ import { JobStatus, OrtRunStatus, Severity } from '@/api/requests';
 export type Status = JobStatus | OrtRunStatus | undefined;
 
 // There is no vulnerability rating in the OpenApi spec, so define it here.
-export type VulnerabilityRating =
-  | 'CRITICAL'
-  | 'HIGH'
-  | 'MEDIUM'
-  | 'LOW'
-  | 'NONE';
+export const VulnerabilityRatings = {
+  CRITICAL: 4,
+  HIGH: 3,
+  MEDIUM: 2,
+  LOW: 1,
+  NONE: 0,
+} as const;
 
 // TailwindCSS classes matching to statuses
 //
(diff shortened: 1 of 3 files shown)


ui/src/helpers/get-status-class.ts
```diff
@@ -23,12 +23,15 @@ import { JobStatus, OrtRunStatus, Severity } from '@/api/requests';
...
-  | 'HIGH'
-  | 'MEDIUM'
-  | 'LOW'
-  | 'NONE';
+export const VulnerabilityRatings = {
+  CRITICAL: { text: 'CRITICAL', value: 4 },
+  HIGH: { text: 'HIGH', value: 3 },
+  MEDIUM: { text: 'MEDIUM', value: 2 },
+  LOW: { text: 'LOW', value: 1 },
+  NONE: { text: 'NONE', value: 0 },
```
sschuberth, line 31: Is there a way in TypeScript to avoid the repetition of names and explicit definition of ordinal values? Like, in Kotlin we'd just use an `enum class` and use the `name` and `ordinal` properties for each entry.

Etsija: Yes I believe there is, let me fix.

mmurto: I didn't check if this is used in other ways, but maybe you could make it a simple const object:

```typescript
export const VulnerabilityRatings = {
  CRITICAL: 4,
  HIGH: 3,
  MEDIUM: 2,
  LOW: 1,
  NONE: 2,
} as const
```

mmurto: @Etsija, [enums are generally recommended to be avoided](https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh), so I would use the above-mentioned object instead.

Etsija: Done. Also decided to fix one DOM error while I'm at it.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/vulnerabilities/index.tsx
```diff
@@ -252,18 +252,15 @@ const VulnerabilitiesComponent = () => {
...
-          <p>
-            By clicking on "References" you can see more information about the
-            vulnerability. The overall severity rating is calculated based on
-            the highest severity rating found in the references.
-          </p>
+          <br />
+          By clicking on "References" you can see more information about the
+          vulnerability. The overall severity rating is calculated based on the
+          highest severity rating found in the references.
         </CardDescription>
```
mmurto, line 264: It makes sense to have multiple `p`s here, so looking at `CardDescription` (it's a simple `p`), I think the best solution would just to use two of them like:

```typescript
<CardDescription>
  These are the vulnerabilities...
</CardDescription>
<CardDescription>
  By clicking on...
</CardDescription>
```

mmurto (changes requested):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_layout/runs/$runIndex/vulnerabilities/index.tsx
Etsija: Oh, didn't know you can actually have several...

mmurto (approved):

PR #1336, opened 2024-10-31
ui: Use markdown renderer for for some elements

Etsija:

Screenshot from 2024-10-31 13-28-39


ui/src/components/markdown-renderer.tsx
```diff
@@ -0,0 +1,47 @@
...
+      components={{
+        h1: ({ children }) => (
+          <p className='mb-6 mt-6 text-xl font-semibold'>{children}</p>
+        ),
+        h2: ({ children }) => (
+          <p className='mb-4 mt-4 text-lg font-semibold'>{children}</p>
+        ),
+        h3: ({ children }) => (
+          <p className='mb-2 mt-2 font-semibold'>{children}</p>
+        ),
```
mmurto, line 38: Add `role` and `aria-level` for small accessibility improvements, see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role.

mmurto (changes requested):

ui/src/components/markdown-renderer.tsx
mmurto: And the headings should probably be defined up to level 6, as they can be contained in the input Markdown.

mmurto (approved):

PR #1341, opened 2024-11-01
ui: Display number of ORT runs in some views

Etsija:

Screenshot from 2024-11-01 12-32-21

Screenshot from 2024-11-01 12-32-31

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId/index.tsx
+++ b/ui/src/routes/_layout/organizations/$orgId/index.tsx
@@ -19,7 +19,7 @@
 
 import { createFileRoute, Link } from '@tanstack/react-router';
 import {
-  ColumnDef,
+  createColumnHelper,
   getCoreRowModel,
   useReactTable,
 } from '@tanstack/react-table';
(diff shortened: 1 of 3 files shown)


sschuberth:
The "Runs" column is quite wide now... can we narrow it down so that more space is left e.g. for the "Last Run date" column to not wrap its contents?

ui/src/routes/_layout/organizations/$orgId/products/$productId/-components/last-run-id.tsx
```diff
@@ -0,0 +1,64 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import { Loader2 } from 'lucide-react';
+
+import { useRepositoriesServiceGetOrtRunsByRepositoryId } from '@/api/queries';
+import { config } from '@/config';
+
+export const LastRunId = ({ repoId }: { repoId: number }) => {
```
lamppu, line 25: Why is this called LastRunId? Would something like TotalRunsForRepository or something be more descriptive of what this component returns?

Etsija: Oh, nice catch - it was LastRunId because I had initially the idea to just return the last (local) run ID. Then I realized that a far more useful is to return `totalCount` but forgot to change the name of the component :smile:

ui/src/routes/_layout/organizations/$orgId/index.tsx
```diff
@@ -58,34 +58,72 @@ import { paginationSchema } from '@/schemas';
...
+
+// In anticipation of these column definitions to be changed later, when the corresponding
+// endpoint is implemented, columnHelper.accessor() is only used when the data being
+// shown contains data from the column helper type.
+
+const columns = [
+  columnHelper.accessor(
+    ({ name, description }) => {
+      return name + description;
+    },
```
lamppu, line 75: Did you find something that made you choose to leave this function like this? I'm still a bit confused if now hypothetically you would add client side sorting to this table (I know we probably won't), would the value based on which the sorting would be done be now `name+description`. Cause then we would probably only want `name` here.

ui/src/routes/_layout/organizations/$orgId/products/$productId/index.tsx
```diff
@@ -57,47 +57,67 @@ import { paginationSchema } from '@/schemas';
...
+
+// In anticipation of these column definitions to be changed later, when the corresponding
+// endpoint is implemented, columnHelper.accessor() is only used when the data being
+// shown contains data from the column helper type.
+
+const columns = [
+  columnHelper.accessor(
+    ({ url, type }) => {
+      return url + type;
+    },
```
lamppu, line 74: And here the same question as above

ui/src/routes/_layout/organizations/$orgId/products/$productId/-components/total-runs.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <Loader2 size={16} className='mx-3 animate-spin' />
+      </>
+    );
+  }
+
+  if (runsIsError) {
+    return <span>Error loading run.</span>;
+  }
+
+  if (runs.data.length === 0) return null;
```
mmurto, line 61: Would returning 0 count rather than null make sense?

mmurto (dismissed):

ui/src/routes/_layout/organizations/$orgId/products/$productId/index.tsx
Etsija: You're probably right, we should use `name` and `url` correspondingly.

Etsija:
I've removed it from the merge queue, as there are some open questions from Johanna that I'd like you @mmurto to answer first.

ui/src/routes/_layout/organizations/$orgId/index.tsx
mmurto: For sorting it probably doesn't affect anything, as the name is the first element, so alphabetically there's no difference with unique names and name + description. Could maybe communicate the intent a little clearer with just the name, and same goes for the url + type.

Another aspect if filtering. There could be a use for having the description in the filterable value, as then you'd be able to include information in the description in the filter. But either probably works, and we can fine-tune it when filtering is in-place.

So both work IMO and are fine by me.

ui/src/routes/_layout/organizations/$orgId/products/$productId/-components/total-runs.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <Loader2 size={16} className='mx-3 animate-spin' />
+      </>
+    );
+  }
+
+  if (runsIsError) {
+    return <span>Error loading run.</span>;
+  }
+
+  if (runs.data.length === 0) return 0;
```
mmurto, line 61: Would this work without, as `totalCount` probably is 0 if there are no items?

Etsija:
Noticed there's still one logical error here, so changed into a draft.

mmurto (approved):

PR #1373, opened 2024-11-06
Get the default branch for the GitHub config provider from the API 

sschuberth:

### DIFF
--- a/config/github/src/main/kotlin/GitHubConfigFileProvider.kt
+++ b/config/github/src/main/kotlin/GitHubConfigFileProvider.kt
@@ -225,7 +225,7 @@ class GitHubConfigFileProvider(
 
     override fun resolveContext(context: Context): Context {
         val branchName = if (context == ConfigManager.EMPTY_CONTEXT) defaultBranch else context.name
-        val response = sendHttpRequest("/branches/$branchName", JSON_CONTENT_TYPE_HEADER)
+        val response = sendHttpRequest("/branches/$branchName")
 
         if (!response.isPresent()) {
             throw ConfigException("The branch '${context.name}' is not found in the config repository.", null)
@@ -256,7 +256,6 @@ class GitHubConfigFileProvider(
     override fun contains(context: Context, path: Path): Boolean {
         val response = sendHttpRequest(
             "/contents/${path.path}?ref=${context.name}",
-            JSON_CONTENT_TYPE_HEADER,
             checkSuccess = false
         )
 
@@ -279,7 +278,7 @@ class GitHubConfigFileProvider(
      */
     private fun sendHttpRequest(
         path: String,
-        contentType: String,
+        contentType: String = JSON_CONTENT_TYPE_HEADER,
         checkSuccess: Boolean = true
     ): HttpResponse = runBlocking {
         val response = sendHttpRequestWithRetry(path, contentType)
@@ -352,10 +351,7 @@ class GitHubConfigFileProvider(
      * [context]. Throw a [ConfigException] if the path does not exist or is not a directory.
      */
     private fun downloadFolderContent(context: Context, path: Path): Set<String> {
-        val response = sendHttpRequest(
-            "/contents/${path.path}?ref=${context.name}",
-            JSON_CONTENT_TYPE_HEADER
-        )
+        val response = sendHttpRequest("/contents/${path.path}?ref=${context.name}")
 
         val jsonBody = getJsonBody(response)
 
--- a/config/github/src/main/kotlin/GitHubConfigFileProvider.kt
+++ b/config/github/src/main/kotlin/GitHubConfigFileProvider.kt
@@ -45,13 +45,13 @@ import kotlinx.datetime.Instant
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonArray
 import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.contentOrNull
 import kotlinx.serialization.json.jsonArray
 import kotlinx.serialization.json.jsonObject
 import kotlinx.serialization.json.jsonPrimitive
 
 import org.eclipse.apoapsis.ortserver.config.ConfigException
 import org.eclipse.apoapsis.ortserver.config.ConfigFileProvider
-import org.eclipse.apoapsis.ortserver.config.ConfigManager
 import org.eclipse.apoapsis.ortserver.config.ConfigSecretProvider
 import org.eclipse.apoapsis.ortserver.config.Context
 import org.eclipse.apoapsis.ortserver.config.Path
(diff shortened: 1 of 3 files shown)


config/github/src/main/kotlin/GitHubConfigFileProvider.kt
mnonnenmacher, line 1: I'm not sure if the intention of the default branch property was to always map to the default branch of the repository. It could also be used to defined another branch which should be used by default. So instead of removing the property, getting the default branch could be added as a fallback if no default branch was configured.

config/github/src/main/kotlin/GitHubConfigFileProvider.kt
```diff
@@ -92,11 +89,6 @@ class GitHubConfigFileProvider(
          */
         const val REPOSITORY_NAME = "gitHubRepositoryName"
 
-        /**
-         * Configuration property that defines the default branch.
-         */
-        const val DEFAULT_BRANCH = "gitHubDefaultBranch"
-
         /**
```
mnonnenmacher, line 92: `config/github/README.md` needs to be updated as well.

mnonnenmacher (changes requested):

config/github/src/main/kotlin/GitHubConfigFileProvider.kt
sschuberth: > It could also be used to defined another branch which should be used by default.

What would be the purpose of that? Note that you can still specific any custom branch / revision to checkout via the context. Just if the context is empty, now whatever the remote has configured as the default is used. If you want to always / mostly use a branch that you do not want to specify as the context, then that branch should simply be made the default on the remote.

mnonnenmacher: If this change is not urgent to you we could wait for feedback from @oheger-bosch on Monday.

sschuberth: Fine with me.

oheger-bosch: The original intention of this property was just that: to define a default that is used if no branch name is provided when triggering an ORT run. Note that users typically will not specify a branch name, this feature is more expected to be used by developers or admins.

IIUC, you argue that the default should be configured via the branch configuration on GitHub. But is this really an advantage? You then have external configuration, and it is not immediately obvious which branch is going to be used. I would also think that the current mechanism is more flexible and supports other scenarios more easily, such as different configurations for prod and dev.

Also, querying the default branch via the GitHub API is probably counted towards the rate limit; since we had trouble with this in the past, I am a bit sensitive here.

sschuberth: > IIUC, you argue that the default should be configured via the branch configuration on GitHub. But is this really an advantage?

I think so, yes, at least when working with multiple configurations like we do. E.g. not all repositories have migrated from using `main` instead of `master` as the default Git branch yet, and switching between those repos would otherwise require changing the ORT server default branch or context all the time.

> You then have external configuration, and it is not immediately obvious which branch is going to be used.

From an end-user perspective, i.e. a user that only has access to the UIs of ORT server and GitHub, it on the contrary is now more obvious which branch is being used, and the user can manage it via the GitHub UI without fiddling with ORT server configuration.

> I would also think that the current mechanism is more flexible and supports other scenarios more easily, such as different configurations for prod and dev.

It's certainly more flexible, but IMO adds unneeded and unusual complexity. As a Git user, I would have never expected there to be another layer of configuration when telling Git which branch to clone except for 1. take any explicitly specified branch, or 2. fall back to the repositories default branch.

> Also, querying the default branch via the GitHub API is probably counted towards the rate limit;

Yes, I also believe it is.

> since we had trouble with this in the past, I am a bit sensitive here.

As the default branch barely changes, we could maybe accommodate for that by not resolving the default branch on every call to `resolveContext`, but only at construction time of the config provider.

oheger-bosch: I am confused, what do you mean by

> at least when working with multiple configurations like we do. E.g. not all repositories have migrated from using main instead of master as the default Git branch yet, and switching between those repos would otherwise require changing the ORT server default branch or context all the time.

There is only a single repository hosting the configuration of ORT Server, which is also defined in the settings of `GitHubConfigFileProvider`. So, switching to another repository would mean a change in the ORT Server configuration anyway.

And wrt end users, they typically do not care or are not even aware of that configuration information is downloaded from a repository. This is fully under the control of the ORT Server admins.

sschuberth: > So, switching to another repository would mean a change in the ORT Server configuration anyway.

Right. What I was referring to is to assist multiple users to setup their own instances of ORT server, and IMO they should not need to change ORT server configuration if their config repo uses `master` instead of `main` as the default branch.

> And wrt end users, they typically do not care or are not even aware of that configuration information is downloaded from a repository. This is fully under the control of the ORT Server admins.

When I refer to "end users", I mean both users that are happy with a hosted SaaS instance, but also users that host their own instance of the server on premise.

oheger-bosch: I am not strictly against this change, but I am also not convinced that this is an improvement. When we originally implemented this config provider, it did not even occur to us that the Git default branch could be used for this purpose; so the expectations of the target user groups may not be that clear.

The optimization you propose to query the default branch only once at construction time would probably not help us in our setup, since the provider is newly instantiated for each worker that is executed.

Maybe you could take another level of indirection and keep the configuration property for the default branch, but make it optional and use the Git default branch if it is not set?

sschuberth: > Maybe you could take another level of indirection and keep the configuration property for the default branch, but make it optional and use the Git default branch if it is not set?

That would add yet another layer of complexity that I'd like to avoid.

mnonnenmacher: So I see a benefit in adding the detection, but I also see a benefit in keeping the configuration option. Having both does not add much complexity in my opinion, it's just a few lines of code plus adding one more option to:

```kotlin
val branchName = context.name.takeUnless { it.isEmpty() } ?: getDefaultBranch() ?: DEFAULT_REPOSITORY_BRANCH
```

sschuberth: I wasn't talking about code complexity, but logic complexity, i.e. understanding all the fallbacks that this option would have then. Anyway, if that's the only way to get default branch detection in, I'll make the change.

sschuberth: But shouldn't the logic for `DEFAULT_REPOSITORY_BRANCH` be different than in your snippet? I.e. if it has a non-null value, it should override the remote's default branch (and avoid a query for that being made)?

mnonnenmacher: > But shouldn't the logic for `DEFAULT_REPOSITORY_BRANCH` be different than in your snippet? I.e. if it has a non-empty (or non-null?) value, it should override the remote's default branch (and avoid a query for that being made)?

Yes, I just quoted the existing code line, I didn't make the required changes.

mnonnenmacher (approved):

PR #1375, opened 2024-11-06
feat(ui): Add missing package managers

mnonnenmacher:
Support all package managers that are supported by ORT itself when triggering an ORT run via the UI.
As `Gradle` and `GradleInspector` are mutually exclusive, the legacy `Gradle` implementation is disabled by default.

### DIFF
--- a/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
+++ b/ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
@@ -43,6 +43,10 @@ export const packageManagers = [
     label: 'Bazel (C++, Java, and others)',
   },
   {
+    id: 'Bower',
+    label: 'Bower (JavaScript / Node.js)',
+  },
+  {
     id: 'Bundler',
     label: 'Bundler (Ruby)',
   },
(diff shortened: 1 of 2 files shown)


sschuberth:
FYI @mmurto, it's OK from my side to make all package managers available here for now, until we have a solution to programmatically determine the available package managers.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
```diff
@@ -50,17 +54,33 @@ export const packageManagers = [
...
     label: 'GoMod (Go)',
   },
+  {
+    id: 'Gradle',
+    label: 'Gradle Legacy (Java), mutually exclusive with Gradle',
+  },
   {
     id: 'GradleInspector',
-    label: 'Gradle (Java)',
+    label: 'Gradle (Java), mutually exclusive with Gradle Legacy',
```
mmurto, line 83: This is fine by me as is, but if you want to guard for Gradle and GradleInspector to not be enabled at the same time and learn some Zod while at it, it should be relatively simple with [Zod's `.refine()`](https://stackoverflow.com/a/75572374) in create-run-utils.ts for the `packageManagers` `z.object()`.

mmurto (approved):

mmurto:
I approved, but feel free to make the improvement in the review comment if you want to @mnonnenmacher

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
mnonnenmacher: I have applied the suggested change and verified that it works. Just one thing is a bit odd: When selecting both package managers the error is only shown after submitting the form. However, afterward the error message is updated just by clicking the checkboxes. So somehow it seems the validation is inactive as long as it was not tried to submit the form.

mmurto (approved):

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
mmurto: Great! The form library we use [validates the input on submit by default](https://www.react-hook-form.com/api/useform/#mode), but validates already errored fields on change. It should probably be changed on `onBlur` to validate when an input is unfocused.

sschuberth: > It should probably be changed on `onBlur` to validate when an input is unfocused.

Indeed that's generally how we should validate forms. Is this something we can easily change globally for all forms?

mmurto: I don't think it can be configured globally, but it's easy to do for each form by adding the `mode: 'onBlur'` parameter linked above to all `useForm` calls.

PR #1385, opened 2024-11-08
chore(dao): Split the `parse` function of the `EnvironmentConfigLoader`

nnobelis:
This function was doing too many things. To simplify the tests, this commit breaks it up in two subfunctions.

### DIFF
--- a/workers/common/src/main/kotlin/common/env/EnvironmentService.kt
+++ b/workers/common/src/main/kotlin/common/env/EnvironmentService.kt
@@ -101,7 +101,7 @@ class EnvironmentService(
         repositoryService: InfrastructureService?
     ): ResolvedEnvironmentConfig {
         val environmentConfigPath = context.ortRun.environmentConfigPath
-        val mergedConfig = configLoader.parse(repositoryFolder, environmentConfigPath).merge(config)
+        val mergedConfig = configLoader.resolveAndParse(repositoryFolder, environmentConfigPath).merge(config)
         val resolvedConfig = configLoader.resolve(mergedConfig, context.hierarchy)
 
         return setUpEnvironmentForConfig(context, resolvedConfig, repositoryService)
--- a/workers/common/src/main/kotlin/common/env/config/EnvironmentConfigLoader.kt
+++ b/workers/common/src/main/kotlin/common/env/config/EnvironmentConfigLoader.kt
@@ -111,8 +111,8 @@ class EnvironmentConfigLoader(
      * cause exceptions to be thrown. Semantic errors are handled according to the `strict` flag.
      */
     fun resolveAndParse(repositoryFolder: File, environmentConfigPath: String? = null): EnvironmentConfig =
-        resolveEnvironmentFile(repositoryFolder, environmentConfigPath)?.let { config ->
-            parse(config)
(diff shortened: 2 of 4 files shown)


workers/common/src/main/kotlin/common/env/config/EnvironmentConfigLoader.kt
```diff
@@ -110,7 +110,17 @@ class EnvironmentConfigLoader(
      * content. If not file path is provided, the default path `.ort.env.yml` is used. Syntactic errors in the file
      * cause exceptions to be thrown. Semantic errors are handled according to the `strict` flag.
      */
-    fun parse(repositoryFolder: File, environmentConfigPath: String? = null): EnvironmentConfig {
+    fun resolveAndParse(repositoryFolder: File, environmentConfigPath: String? = null): EnvironmentConfig =
+        resolveEnvironmentFile(repositoryFolder, environmentConfigPath)?.let { config ->
```
mnonnenmacher, line 114: A more consistent naming would help with readability, e.g. consistently use `environmentFile` or `environmentConfigFile` instead of `environmentFile`, `config`, and `environmentConfigFile` when they refer to the same thing. Also the `parse` function should be renamed accordingly to `parseEnvironment(Config)File`.

workers/common/src/test/kotlin/common/env/config/EnvironmentConfigLoaderTest.kt
```diff
@@ -383,7 +418,7 @@ class EnvironmentConfigLoaderTest : StringSpec() {
         "Simple environment variable definitions are processed" {
             val helper = TestHelper()
 
-            val config = loadConfig(".ort.env.direct-variables.yml", helper)
+            val config = parseConfig(".ort.env.direct-variables.yml", helper)
```
mnonnenmacher, line 421: The commit message says "To simplify the tests, this commit breaks it up in two subfunctions", but to be honest, I don't see in this diff how the tests became simpler?

mnonnenmacher (changes requested):

workers/common/src/test/kotlin/common/env/config/EnvironmentConfigLoaderTest.kt
nnobelis: If you look in `EnvironmentConfigLoaderTest`, you can see that I wrote some tests that could not be easily written before.

Also the tests now mainly use `parseConfig`, which does not unnecessarily resolve the parsed configuration.
Hence, introducing the functions allows to have tests doing only what needs to be tested.

mnonnenmacher: Ok, it sounded to me like also the existing tests would become simpler. Anyway, you did notice that the tests are failing?

workers/common/src/main/kotlin/common/env/EnvironmentService.kt
```diff
@@ -101,7 +101,7 @@ class EnvironmentService(
         repositoryService: InfrastructureService?
```
oheger-bosch, line 101: Commit message nit: This is not a change of the `dao` module; `workers` would be more appropriate, I think. Same for the next commit.

oheger-bosch (approved):

PR #1397, opened 2024-11-11
feat(api): Introduce initial API `client` library

bs-ondem:
Introduce a multiplatform client library with an inital `AuthService` to handle token generation and refreshing.


api/v1/client/build.gradle.kts
```diff
@@ -0,0 +1,49 @@
...
+                implementation(libs.ktorClientContentNegotiation)
+                implementation(libs.ktorKotlinxSerializationMP)
+                implementation(libs.ktorClientOkHttp)
+            }
+        }
+
+        commonTest {
+            dependencies {
+            }
+        }
```
sschuberth, line 47: Shouldn't this be removed while it's empty?

api/v1/client/build.gradle.kts
```diff
@@ -0,0 +1,49 @@
...
+kotlin {
+    jvm()
+
+    sourceSets {
+        commonMain {
+            dependencies {
+                implementation(libs.ktorClientAuth)
+                implementation(libs.ktorClientContentNegotiation)
+                implementation(libs.ktorKotlinxSerializationMP)
+                implementation(libs.ktorClientOkHttp)
```
sschuberth, line 40: Nit: Please sort these alphabetically.

api/v1/client/src/commonMain/kotlin/auth/AuthService.kt
```diff
@@ -0,0 +1,84 @@
...
+                append("grant_type", "refresh_token")
+            }
+        ).let { response ->
+            if (response.status.isSuccess()) {
+                response.body()
+            } else {
+                throw AuthenticationException(
+                    "Failed to refresh token: ${response.status.value}: ${response.bodyAsText()}."
+                )
+            }
```
sschuberth, line 77: Throwing early on `!response.status.isSuccess()` would reduce nesting / avoid the `else`.

api/v1/client/src/commonMain/kotlin/auth/TokenInfo.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class TokenInfo(
+    /**
+     * The access token.
+     */
+    @SerialName("access_token")
```
sschuberth, line 30: Instead of customizing each `@SerialName`, better provide a customized `SerialFormat` like

    val JSON = Json { namingStrategy = JsonNamingStrategy.SnakeCase }

and use that instance for serialization.

gradle/libs.versions.toml
```diff
@@ -110,6 +110,7 @@ ktorClientAuth = { module = "io.ktor:ktor-client-auth", version.ref = "ktor" }
 ktorClientContentNegotiation = { module = "io.ktor:ktor-client-content-negotiation", version.ref = "ktor" }
 ktorClientCore = { module = "io.ktor:ktor-client-core", version.ref = "ktor" }
 ktorClientOkHttp = { module = "io.ktor:ktor-client-okhttp", version.ref = "ktor" }
+ktorKotlinxSerializationMP = { module = "io.ktor:ktor-serialization-kotlinx-json", version.ref = "ktor" }
```
sschuberth, line 113: Please maintain alphabetical sorting.

sschuberth (dismissed):

api/v1/client/src/commonMain/kotlin/auth/AuthService.kt
```diff
@@ -0,0 +1,84 @@
...
+import io.ktor.client.call.body
+import io.ktor.client.request.forms.submitForm
+import io.ktor.client.statement.bodyAsText
+import io.ktor.http.Parameters
+import io.ktor.http.isSuccess
+
+/**
+ * A service to generate/refresh an OAuth token to authenticate with the ORT server.
+ */
+class AuthService(
```
oheger-bosch, line 32: Tests are missing.

api/v1/client/src/commonMain/kotlin/auth/AuthService.kt
```diff
@@ -0,0 +1,84 @@
...
+    private val client: HttpClient,
+    private val tokenUrl: String,
+    private val clientId: String
+) {
+    /**
+     * Generate a token for the given [username] and [password].
+     */
+    suspend fun generateToken(username: String, password: String): TokenInfo =
```
oheger-bosch, line 40: I am not sure whether this service should be public and directly exposed to clients. I would rather expect that there is some `OrtServerClient` class (or similar) that acts as a facade for the ORT Server API. This class would handle authentication behind the scenes.

bs-ondem: There will be an `OrtServerClient` class in the upcoming PR that will  handle authentication in its factory function. It might be better to change the visibility to internal?

oheger-bosch: Yes, this class should not be accessible from the outside.

api/v1/client/src/commonMain/kotlin/auth/TokenInfo.kt
bs-ondem: That would require configuring this when creating the HttpClient for the AuthService. I’d prefer to keep the custom names directly in the data class.

api/v1/client/build.gradle.kts
bs-ondem: Not empty anymore.

oheger-bosch (approved):

PR #1410, opened 2024-11-12
Add statistics endpoint

lamppu:

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/EcosystemStats.kt
+++ b/api/v1/model/src/commonMain/kotlin/EcosystemStats.kt
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.api.v1.model
+
+import kotlinx.serialization.Serializable
+
+/**
+ * A data class representing statistics for an ecosystem.
+ */
+@Serializable
+data class EcosystemStats(
+    /**
+     * The name of the ecosystem.
+     */
+    val name: String,
+    /**
+     * The number of packages in the ORT run that belong to this ecosystem.
+     */
+    val count: Long
+)
--- a/core/src/main/kotlin/api/RunsRoute.kt
+++ b/core/src/main/kotlin/api/RunsRoute.kt
@@ -42,12 +42,14 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.ComparisonOperator
 import org.eclipse.apoapsis.ortserver.api.v1.model.FilterOperatorAndValue
 import org.eclipse.apoapsis.ortserver.api.v1.model.JobSummaries
 import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunFilters
+import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunStatistics
 import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunStatus
 import org.eclipse.apoapsis.ortserver.api.v1.model.SortDirection
 import org.eclipse.apoapsis.ortserver.api.v1.model.SortProperty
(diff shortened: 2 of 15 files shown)


services/hierarchy/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -366,6 +366,139 @@ class PackageServiceTest : WordSpec() {
                 results should beEmpty()
             }
         }
+
+        "findEcoSystemsForOrtRun" should {
```
mnonnenmacher, line 370: The other new functions should also be tested.

core/src/main/kotlin/api/RunsRoute.kt
```diff
@@ -252,6 +256,52 @@ fun Route.runs() = route("runs") {
...
+                    requirePermission(RepositoryPermission.READ_ORT_RUNS.roleName(ortRun.repositoryId))
+
+                    val jobs = repositoryService.getJobs(ortRun.repositoryId, ortRun.index)
+
+                    val finishedStateSet: Set<JobStatus> = setOf(JobStatus.FINISHED, JobStatus.FINISHED_WITH_ISSUES)
+
+                    val analyzerJobInFinalState = jobs?.analyzer?.status in finishedStateSet.plus(JobStatus.FAILED)
+                    val analyzerJobInFinishedState = jobs?.analyzer?.status in finishedStateSet
+                    val advisorJobInFinishedState = jobs?.advisor?.status in finishedStateSet
+                    val evaluatorJobInFinishedState = jobs?.evaluator?.status in finishedStateSet
```
mnonnenmacher, line 272: If you add `== true` to the lines above you can get rid of the `takeIf`s below.

mnonnenmacher (changes requested):

core/src/main/kotlin/api/RunsRoute.kt
lamppu: I had some trouble with this one, these are already booleans, so adding the `== true` seems to be redundant. I tried to take the `takeIf`s away, but then it just does the counts either way, although maybe I did it somehow wrong, as I'm still quite new to Kotlin :smile:

lamppu: `if (analyzerJobInFinalState) issueService.countForOrtRunId(ortRun.id) else null` would look a lot more readable to me though, but I guess I mostly went for the takeIf/let approach, as I feel I've seen that mostly instead of if/else clauses.

mnonnenmacher: I read the code wrong, there are situations where an expression can return `null` or a `boolean` and then explicit comparison with `true` or `false` can make the code simpler, but this is not one of them.

I actually do prefer if/else over `takeIf` in most situations because I agree that it is a lot more readable.

lamppu: Ok yeah, I changed them now.

mnonnenmacher (approved):

core/src/main/kotlin/apiDocs/RunsDocs.kt
```diff
@@ -453,3 +455,35 @@ val getOrtRuns: OpenApiRoute.() -> Unit = {
...
+            description = "Success"
+            jsonBody<OrtRunStatistics> {
+                example("Get ORT run statistics") {
+                    value = OrtRunStatistics(
+                        issuesCount = 10,
+                        packagesCount = 200,
+                        ecosystems = listOf(
+                            EcosystemStats("Maven", 20),
+                            EcosystemStats("NPM", 145)
+                           ),
```
sschuberth, line 481: Wrong indentation.

core/src/main/kotlin/apiDocs/RunsDocs.kt
```diff
@@ -453,3 +455,35 @@ val getOrtRuns: OpenApiRoute.() -> Unit = {
...
+                        vulnerabilitiesCount = 3,
+                        ruleViolationsCount = 5,
```
sschuberth, line 483: We usually don't use trailing commas.

api/v1/model/src/commonMain/kotlin/EcosystemStats.kt
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * A data class representing statistics for an ecosystem.
+ */
+@Serializable
+data class EcosystemStats(
+    /**
+     * The name of the ecosystem.
+     */
+    val name: String,
+    /**
```
sschuberth, line 33: Use usually prefer to have empty lines before these to visually group the comment to the property. BTW, short comments can also use the more compact `/** The name of the ecosystem. */` one-line syntax.

core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
```diff
@@ -1341,6 +1343,449 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
...
+
+                val statistics = response.body<OrtRunStatistics>()
+
+                statistics.issuesCount shouldBe 3
+                statistics.packagesCount shouldBe 2
+                statistics.vulnerabilitiesCount shouldBe 1
+                statistics.ruleViolationsCount shouldBe 4
+
+                with(statistics.ecosystems) {
+                    shouldNotBeNull()
```
sschuberth, line 1618: Can be combined to `statistics.ecosystems.shouldNotBeNull {}`.

sschuberth (commented):
Just some optional comments that can be addressed as a follow-up after the merge.

Issue #123, opened 2024-04-22
ORT runs cannot be deleted


Etsija:
Related to [this PR](https://github.com/eclipse-apoapsis/ort-server/issues/100), there really should be an endpoint to delete ORT runs from the database.  Currently, not being able to delete runs leads to cascading delete failure:

We have a hierarchy Organization -> Product -> Repository -> Run. Because runs cannot be deleted, none of the items going up in hierarchy can, either.

mnonnenmacher:
@Etsija We are planning to work on that but first have to finish some other topics. I can't give you a date but this will definitely be implemented.

Etsija:
Any estimates yet when this could be implemented?

mnonnenmacher:
@Etsija Sorry, I can't give you an estimate. The item is high priority for us but we have to work on some other topics first, so better not expect it to be done within the next few weeks.

sschuberth:
> Sorry, I can't give you an estimate.

Is there an update that allows you to commit to a time-line by now @mnonnenmacher?

sschuberth:
The REST API endpoint was implemented by https://github.com/eclipse-apoapsis/ort-server/pull/1494. I've created https://github.com/eclipse-apoapsis/ort-server/issues/1505 as a follow-up.

wkl3nk:
With the 2nd of 2 commits, now the endpoints for deleting ORT runs are fully functional. See https://github.com/eclipse-apoapsis/ort-server/commit/2696a26831546cc02f413beab0de59f49164e8f1 (new endpoints and deletion in database) and https://github.com/eclipse-apoapsis/ort-server/commit/c25ebd767dfd512c10eda71d8a05a1d01f47373d (deletion of generated reports in storage)

PR #1420, opened 2024-11-13
Delete ORT runs

oheger-bosch:
This PR adds `ON DELETE CASCADE` constraints to the tables that depend on the `ort_runs` table to make sure that these entities are deleted automatically together with the associated ORT run. This is a first step to support manual and automatic removal of ORT runs.

### DIFF
--- a/dao/src/main/resources/db/migration/V86__ortRunDeleteCascade.sql
+++ b/dao/src/main/resources/db/migration/V86__ortRunDeleteCascade.sql
@@ -1,0 +1,282 @@
+-- This migration script modifies the foreign key constraints of the tables that reference ort_runs to add an
+-- ON DELETE CASCADE clause. With this setup, it is very easy to delete an ORT run and all its associated data.
+
+ALTER TABLE notifier_runs
+  DROP CONSTRAINT notifier_runs_notifier_job_id_fkey,
+  ADD CONSTRAINT notifier_runs_notifier_job_id_fkey
+    FOREIGN KEY (notifier_job_id) REFERENCES notifier_jobs (id) ON DELETE CASCADE;
+
+ALTER TABLE notifier_jobs
+  DROP CONSTRAINT notifier_jobs_ort_run_id_fkey,
+  ADD CONSTRAINT notifier_jobs_ort_run_id_fkey
+    FOREIGN KEY (ort_run_id) REFERENCES ort_runs (id) ON DELETE CASCADE;
+
+ALTER TABLE packages_analyzer_runs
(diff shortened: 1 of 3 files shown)


sschuberth:
> This is a first step to support manual and automatic removal of ORT runs.

Relates to https://github.com/eclipse-apoapsis/ort-server/issues/123.

dao/src/main/resources/db/migration/V86__ortRunDeleteCascade.sql
```diff
@@ -0,0 +1,282 @@
...
+
+ALTER TABLE notifier_jobs
+  DROP CONSTRAINT notifier_jobs_ort_run_id_fkey,
+  ADD CONSTRAINT notifier_jobs_ort_run_id_fkey
+    FOREIGN KEY (ort_run_id) REFERENCES ort_runs (id) ON DELETE CASCADE;
+
+ALTER TABLE packages_analyzer_runs
+  DROP CONSTRAINT packages_analyzer_runs_analyzer_run_id_fkey,
+  ADD CONSTRAINT packages_analyzer_runs_analyzer_run_id_fkey
+    FOREIGN KEY (analyzer_run_id) REFERENCES analyzer_runs (id) ON DELETE CASCADE;
```
mnonnenmacher, line 17: IIUC `ON DELETE CASCADE` only partly works for many to many associations. In this example it will remove the row from the `packages_analyzer_runs` join table if the analyzer run is deleted, but it will not remove the package itself. So some other mechanism is needed to clean up dangling packages which are not associated to any analyzer runs anymore.
This is not necessarily a problem, but I think it should be pointed out in the comment on top of the file and the commit message.

mnonnenmacher (changes requested):

dao/src/main/resources/db/migration/V86__ortRunDeleteCascade.sql
mmurto: One good mechanism for that would be an after delete trigger like [this](https://stackoverflow.com/a/35373676).

kamil-bielecki-bosch: Please, be careful with all kind of triggers. They can decrease performance dramatically.

mnonnenmacher: > To deal with possible orphan entries, I was thinking about a periodic task that does some cleanup.

I think I would prefer that over a trigger, having this in the application code is more accessible to the developers and also we might need more control over in which cases we want to remove orphaned entries.

mnonnenmacher (approved):

PR #1438, opened 2024-11-15
ci: Fix variable reference in Docker build

mmurto:
The pattern `${{ is_default_branch }}` attempts to get a GitHub Actions
provided named value, but the default branch status is provided by the
Docker Metadata Action as a global expression to be used with Handlebars'
template format [1].

[1]: https://github.com/marketplace/actions/docker-metadata-action#global-expressions

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -121,7 +121,7 @@ jobs:
           type=raw,value=${{ env.ORT_SERVER_VERSION }}
           type=ref,event=branch
           type=sha
-          type=raw,value=latest,enable=${{ is_default_branch }}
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-build.yml
```diff
@@ -144,7 +144,7 @@ jobs:
           type=raw,value=${{ env.ORT_SERVER_VERSION }}
           type=ref,event=branch
           type=sha
-          type=raw,value=latest,enable=${{ is_default_branch}}
+          type=raw,value=latest,enable={{ is_default_branch}}
```
sschuberth, line 147: While at it, mind adding the missing space before `}}`?

sschuberth (approved):

PR #1441, opened 2024-11-15
Inline `OptionsTransformer` in `ReporterRunner`

mnonnenmacher:
Reduce complexity by inlining the `OptionsTransformer` into its only user `ReporterRunner`.  See the commit messages for details.


workers/common/src/main/kotlin/common/Extensions.kt
```diff
@@ -36,6 +37,14 @@ import org.slf4j.LoggerFactory
 
 val logger: Logger = LoggerFactory.getLogger(ConfigManager::class.java)
 
+/**
```
sschuberth, line 40: Can you please give an outline in the commit message for what this function will be used? Because so far it's not even a real `feat` if it's just "dead code" 😉

sschuberth: Also moving this commit to directly before the commit that makes use of the new function could help.

mnonnenmacher: > Also moving this commit to directly before the commit that makes use of the new function could help.

It already is.

sschuberth (approved):

PR #1450, opened 2024-11-18
Unique constraint violations

oheger-bosch:
This PR fixes a subtle bug in the handling of unique constraint violation exceptions which could cause ORT runs to fail. Refer to the single commits for further details.


nnobelis (approved):
This is a great find !

PR #1467, opened 2024-11-21
ci: Stop publishing Maven artifacts to GitHub packages

mnonnenmacher:
Since 685f3b7 Maven artifacts are published to Maven Central, so there is no need anymore to publish them to the GitHub package registry.


sschuberth:
> Since [685f3b7](https://github.com/eclipse-apoapsis/ort-server/commit/685f3b744f1a63408f1e6898469e21e5457f26fe) Maven artifacts are published to Maven Central

But that's only true for tags, right? Is there still a need to continue publishing snapshots from `main` to GitHub? If not, maybe explicitly say in the commit message that snapshots are not needed anymore?

mnonnenmacher:
> > Since [685f3b7](https://github.com/eclipse-apoapsis/ort-server/commit/685f3b744f1a63408f1e6898469e21e5457f26fe) Maven artifacts are published to Maven Central
> 
> But that's only true for tags, right? Is there still a need to continue publishing snapshots from `main` to GitHub? If not, maybe explicitly say in the commit message that snapshots are not needed anymore?

If have added a sentence to the commit message.

sschuberth (approved):

Issue #1039, opened 2024-09-15
Make some run details permanently visible (independently of the selected menu)


mnonnenmacher:
If you have multiple tabs open to compare different runs the only context information available on all subscreens to identify the run is the index.

It would help with keeping the overview to have a little more context information visible on all screens, like the timestamp and the job statuses (as traffic lights). The global run id is also useful when having the Loki dashboard open in another tab, at least as long as we don't have a log viewer in the UI.

Also see the discussion in #1007.

Etsija:
IIRC we agreed in the alignment meeting that this would be best implemented by changing the ORT run details sub-page's layout such that
- we implement a top section containing minimal data (timestamp, global run ID, traffic lights, what else?) from the ORT run
- then change layout such that the top section is visible in all sub-pages of the ORT run
- as a result, the layout would change into: top section / sidebar on the left / main data on the right

sschuberth:
> (timestamp, global run ID, traffic lights, what else?)

The "final" list of properties agreed on with @mnonnenmacher is:
- run status (both badge and traffic lights)
- rerun button
- created at, duration
- branch name, resolved revision
- "Run ID" ([renamed from "Global id"](https://github.com/eclipse-apoapsis/ort-server/pull/2239))

PR #1470, opened 2024-11-21
ui: Redirect to latest run from repository page

mmurto:
This is waiting for #1469.

See individual commits.

image

image

image


sschuberth:
Only some nits from looking at the screenshots:
- The menu item is called "All Runs", but at least in the second screenshots the title displayed for that page is just "Runs". We should align this, and I have a tendency towards calling the menu item also just "Runs" for simplicity (unless you want to emphasize the "All" for some reason, i.e. if there also would be not-*all*-runs view).
- I like the idea to consolidate editing and deleting in a new menu entry. However, as this is not only about editing, but also deleting, maybe call the menu entry "Manage" or so instead?

mmurto:
> Only some nits from looking at the screenshots:
> 
> * The menu item is called "All Runs", but at least in the second screenshots the title displayed for that page is just "Runs". We should align this, and I have a tendency towards calling the menu item also just "Runs" for simplicity (unless you want to emphasize the "All" for some reason, i.e. if there also would be not-_all_-runs view).

I agree that it should be aligned for the repository view, so in the second screenshot the sidebar item should be "Runs" like the heading is. In the run view, so in the first screenshot, I think "All Runs" makes more sense it's a link from one run to all runs.

> * I like the idea to consolidate editing and deleting in a new menu entry. However, as this is not only about editing, but also deleting, maybe call the menu entry "Manage" or so instead?

Agree. "Settings" feels natural to me, I'll try that.

mnonnenmacher:
I very often want to see the list of runs and not the results of the last run, would it be an alternative to just have separate buttons in the repository list to either got to the list or go to the last run?
Also, having the "All Runs" button below the "Overview" breaks the hierarchy to me: The list of runs is still a logical parent of the run details, so the button should be on top and maybe separated with a divider.

sschuberth:
> I very often want to see the list of runs and not the results of the last run

Interesting, we've received pretty clear feedback from our users that they want / expected the opposite. Can you explain *why* you want to see the list of all runs, what is the information you're looking for? It it about the success / failure rate, or how the last run compares to the previous runs in that regard?

Depending on your answer, maybe we could tweaks things to not unconditionally show the last run, but the last *successful* run? Thought that logic could be more confusing to users.

sschuberth:
> The list of runs is still a logical parent of the run details

I believe this is one of the cases where we should be free to relax hierarchies implied by backend code in the UI for the sake of usability and convenience. Without knowing anything about the code and just looking at the UI, I think having an "Overview" first and then next "All Runs" to drill into more details is just fine.

sschuberth:
@mmurto how about renaming the "Settings" menu from the bottom-left of your first screenshots to "Repository" as follows:

```
Repository
  Secrets
  Users
  Settings
```

This avoids the "Repository" word duplication for the nested entries and makes more clear that no run-related stuff is in there.

mmurto:
> > I very often want to see the list of runs and not the results of the last run
> 
> Interesting, we've received pretty clear feedback from our users that they want / expected the opposite. Can you explain _why_ you want to see the list of all runs, what is the information you're looking for? It it about the success / failure rate, or how the last run compares to the previous runs in that regard?

I think this also depends on who the primary user that ends up on the repository page is. In my view that user is likely to be the most interested in the state of their repository's state, so showing the most recent run prominently makes sense to me. Now I'm mostly guessing (we should add analytics!), but I'd imagine users who are more interested in diving deep into the past runs use the UI more than other users, so they know their way around it more.

> @mmurto how about renaming the "Settings" menu from the bottom-left of your first screenshot to "Repository" as follows:
> 
> ```
> Repository
>   Secrets
>   Users
>   Settings
> ```
> 
> This avoids the "Repository" word duplication for the nested entries and makes more clear that no run-related stuff is in there.
> 
> _Edit:_ And for "Settings" I'd use https://lucide.dev/icons/settings as the icon.

Agree.

sschuberth:
> I'd imagine users who are more interested in diving deep into the past runs use the UI more than other users, so they know their way around it more.

Exactly my thoughts as well. For more technical users / admins it is reasonable to navigate in the UI a bit more.

mnonnenmacher:
>Interesting, we've received pretty clear feedback from our users that they want / expected the opposite. Can you explain why you want to see the list of all runs, what is the information you're looking for? It it about the success / failure rate, or how the last run compares to the previous runs in that regard?

I think because I'm looking at it from an admin perspective, I'm usually more interested to see how stable the results are, how often the repo is scanned, and so on.
But also from a user perspective, as long as you manually trigger all runs the last one maybe stands out, but as soon as you trigger them from CI, e.g., from each PR, the last run becomes arbitrary and you might be more interested in the last run for the main branch.

>Depending on your answer, maybe we could tweaks things to not unconditionally show the last run, but the last successful run? Thought that logic could be more confusing to users.

Yes, I think that would be confusing.

> >The list of runs is still a logical parent of the run details

>I believe this is one of the cases where we should be free to relax hierarchies implied by backend code in the UI for the sake of usability and convenience. Without knowing anything about the code and just looking at the UI, I think having an "Overview" first and then next "All Runs" to drill into more details is just fine.

I didn't mean it from a technical perspective but from how I understand the content. To me a list is a parent object of one item in the list. That's why I would prefer to have the "All runs" link on top of the sidebar, then its position would also not change after clicking on it.
And BTW, like I mentioned in #1039, I would really like if next to the "Overview" button some context information would be shown. This would also justify separating "Overview" from "All runs" with a divider.

>I think this also depends on who the primary user that ends up on the repository page is. In my view that user is likely to be the most interested in the state of their repository's state, so showing the most recent run prominently makes sense to me. Now I'm mostly guessing (we should add analytics!), but I'd imagine users who are more interested in diving deep into the past runs use the UI more than other users, so they know their way around it more.

Still, what about my proposal to offer both options? For example, the number of runs ("3" in the image below) could be clickable and lead directly to the history:
image

mmurto:
> > Interesting, we've received pretty clear feedback from our users that they want / expected the opposite. Can you explain why you want to see the list of all runs, what is the information you're looking for? It it about the success / failure rate, or how the last run compares to the previous runs in that regard?
> 
> I think because I'm looking at it from an admin perspective, I'm usually more interested to see how stable the results are, how often the repo is scanned, and so on. But also from a user perspective, as long as you manually trigger all runs the last one maybe stands out, but as soon as you trigger them from CI, e.g., from each PR, the last run becomes arbitrary and you might be more interested in the last run for the main branch.

One idea I've been toying with is that it could be nice to build the UI (with proper endpoints for it) not around ORT runs, but commits that have been analyzed. Inside commits you could then have multiple runs and the commits could also be grouped by the branch they belong to.

> Still, what about my proposal to offer both options? For example, the number of runs ("3" in the image below) could be clickable and lead directly to the history: image

Good idea, I added that! I originally tried some Actions buttons like in other tables, but the layout is starting to be to narrow for it, so we'll need to widen it.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/runs/index.tsx
```diff
@@ -47,11 +47,11 @@ import {
 } from '@/components/ui/tooltip';
 import { toast } from '@/lib/toast';
 import { paginationSearchParameterSchema } from '@/schemas';
-import { RepositoryRunsTable } from '../-components/repository-runs-table';
+import { RepositoryRunsTable } from '../../-components/repository-runs-table';
```
Etsija, line 45: Should this now actually be an absolute link `@/...` as it's not anymore "close" to the page where it's used?

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/settings.tsx
```diff
@@ -0,0 +1,275 @@
...
+      useRepositoriesServiceGetRepositoryById,
+      params.orgId,
+      params.productId,
+      params.repoId,
+    ],
+    queryFn: async () =>
+      await RepositoriesService.getRepositoryById({
+        repositoryId: Number.parseInt(params.repoId),
+      }),
+  });
```
Etsija, line 82: Is there a particular reason why a suspense hook is not used here? Also, can you explain why there are extra elements in the query key, which are not parameters of the query?

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/settings.tsx
```diff
@@ -0,0 +1,275 @@
...
+    await context.queryClient.ensureQueryData({
+      queryKey: [useRepositoriesServiceGetRepositoryByIdKey, params.repoId],
+      queryFn: () =>
+        RepositoriesService.getRepositoryById({
+          repositoryId: Number.parseInt(params.repoId),
+        }),
+    });
+  },
+  component: RepositorySettingsPage,
+});
```
Etsija, line 275: Same here: why not use the prefetch hook? Query key is also now different from the one above.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/settings.tsx
mmurto: No particular reason, as the settings page is just renamed [edit page](https://github.com/eclipse-apoapsis/ort-server/blame/main/ui/src/routes/_layout/organizations/%24orgId/products/%24productId/repositories/%24repoId/_repo-layout/edit.tsx) with the addition of the danger zone including a delete button.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/settings.tsx
mmurto: True, seems to be a bug in the old edit page, which got carried over here. I'll add a fixing comment.

ui/src/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/runs/index.tsx
mmurto: > Should this now actually be an absolute import `@/...` as it's not anymore "close" to the page where it's used?

I think it's still relatively close, as the absolute import would be `@/routes/_layout/organizations/$orgId/products/$productId/repositories/$repoId/-components/repository-runs-table`. What do you think?

Etsija: Well, yeah, probably better keep it as-is.

Etsija (approved):

PR #1472, opened 2024-11-21
refactor(reporter): Do not download all license texts at once

nnobelis:
The runner currently downloads all the license texts to a temporary directory and builds a `LicenseTextProvider` for it. This is unoptimal because not all of these texts may be needed and they can be hosted on a network share (NFS).
This commit adapts the logic to only access the requested license text files.

### DIFF
--- a/workers/reporter/src/main/kotlin/reporter/LoggingLicenseTextProvider.kt
+++ b/workers/reporter/src/main/kotlin/reporter/LoggingLicenseTextProvider.kt
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package reporter
+
+import java.io.File
+
+import kotlin.time.measureTimedValue
+
+import org.ossreviewtoolkit.reporter.DefaultLicenseTextProvider
+import org.ossreviewtoolkit.reporter.LicenseTextProvider
+
+import org.slf4j.LoggerFactory
+
+private val logger = LoggerFactory.getLogger(LoggingLicenseTextProvider::class.java)
+
+/**
+ * A [LicenseTextProvider] that delegates to a [DefaultLicenseTextProvider] and logs the time it takes to query the
+ * license text file.
+ */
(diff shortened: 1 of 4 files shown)


sschuberth:
> unoptimal

I believe this should say "non-optimal" or "less-than-ideal" or so.

workers/reporter/src/main/kotlin/reporter/LoggingLicenseTextProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * A [LicenseTextProvider] that delegates to a [DefaultLicenseTextProvider] and logs the time it takes to query the
+ * license text file.
+ */
+class LoggingLicenseTextProvider(private val licenseTextDirectory: File) : LicenseTextProvider {
+    private val delegate: DefaultLicenseTextProvider = DefaultLicenseTextProvider(listOf(licenseTextDirectory))
+
+    override fun getLicenseTextReader(licenseId: String): (() -> String)? {
+        logger.debug("Downloading custom license text of '$licenseId' from '${licenseTextDirectory.absolutePath}'.")
```
mnonnenmacher, line 41: Nit: "for" instead of "of"

workers/reporter/src/main/kotlin/reporter/LoggingLicenseTextProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+
+        val measurement = measureTimedValue {
+            delegate.getLicenseTextReader(licenseId)
+        }
+
+        logger.debug("Downloaded custom license text complete in ${measurement.duration}.")
```
mnonnenmacher, line 47: Nit: "...license text for '$licenseId' in..."

workers/reporter/src/main/kotlin/reporter/LoggingLicenseTextProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+ * license text file.
+ */
+class LoggingLicenseTextProvider(private val licenseTextDirectory: File) : LicenseTextProvider {
+    private val delegate: DefaultLicenseTextProvider = DefaultLicenseTextProvider(listOf(licenseTextDirectory))
+
+    override fun getLicenseTextReader(licenseId: String): (() -> String)? {
+        logger.debug("Downloading custom license text of '$licenseId' from '${licenseTextDirectory.absolutePath}'.")
+
+        val measurement = measureTimedValue {
+            delegate.getLicenseTextReader(licenseId)
```
mnonnenmacher, line 44: Isn't the actual download of the license text missing?

mnonnenmacher (changes requested):

workers/reporter/src/main/kotlin/reporter/LoggingLicenseTextProvider.kt
nnobelis: I changed the wording to no mention any download.

oheger-bosch:
I am afraid this will not work; or at least not solve the problem we currently have with the default license text provider. With your approach, custom license texts are never downloaded from the GitHub repository.

What is needed is a custom license text provider implementation that when asked for a specific license text, directly forwards the request to the GitHub config file provider. The latter handles caching transparently. That way only license files are accessed that are actually needed.

nnobelis:
You're right @oheger-bosch : I created https://github.com/oss-review-toolkit/ort/pull/9523 upstream to have access to the validation. When it's merged, I will rework current PR.

nnobelis:
Superseeded by https://github.com/eclipse-apoapsis/ort-server/pull/1654

PR #1474, opened 2024-11-21
Deduplicate configuration settings

oheger-bosch:
This PR reduces redundancy in configuration settings by moving shared settings to the modules that define them.

### DIFF
--- a/core/src/main/resources/application.conf
+++ b/core/src/main/resources/application.conf
@@ -87,29 +87,6 @@ orchestrator {
   }
 }
 
-secretsProvider {
-  name = "vault"
-  name = ${?SECRETS_PROVIDER_NAME}
-
-  # Properties for the Vault provider.
-  vaultUri = ${?VAULT_URI}
-  vaultRoleId = ${?VAULT_ROLE_ID}
-  vaultSecretId = ${?VAULT_SECRET_ID}
-  vaultRootPath = ${?VAULT_ROOT_PATH}
-  vaultPrefix = ${?VAULT_PREFIX}
-  vaultNamespace = ${?VAULT_NAMESPACE}
-
-  # Properties for the file-based provider.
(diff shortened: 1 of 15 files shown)


core/src/main/resources/application.conf
mnonnenmacher, line 1: How does this work, does typesafe config load all `application.conf` files on the classpath?

secrets/spi/src/main/resources/application.conf
```diff
@@ -0,0 +1,39 @@
...
+  name = "vault"
+  name = ${?SECRETS_PROVIDER_NAME}
+
+  # Properties for the Vault provider.
+  vaultUri = ${?VAULT_URI}
+  vaultRoleId = ${?VAULT_ROLE_ID}
+  vaultSecretId = ${?VAULT_SECRET_ID}
+  vaultRootPath = ${?VAULT_ROOT_PATH}
+  vaultPrefix = ${?VAULT_PREFIX}
+  vaultNamespace = ${?VAULT_NAMESPACE}
```
mnonnenmacher, line 28: Is it possible to move the settings to the respective module, like `:secrets:vault` in this case? Then the SPI would not have to know about its implementations.

core/src/main/resources/application.conf
oheger-bosch: Yes, it does and merges them to a combined configuration. See https://github.com/lightbend/config?tab=readme-ov-file#standard-behavior

secrets/spi/src/main/resources/application.conf
oheger-bosch: Yes, I have added another commit which does exactly this. In theory, such a split would be possible for other abstractions, too; but since we are currently using worker-specific properties, the configuration can only be defined in the workers modules.

mnonnenmacher (approved):

PR #1482, opened 2024-11-23
feat(storage): Add support for the Azure Blob storage

mnonnenmacher:
Add a storage implementation supporting the Azure Blob storage [1].

[1]: https://azure.microsoft.com/en-us/products/storage/blobs/


storage/azure-blob/src/main/kotlin/AzureBlobStorageProvider.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.eclipse.apoapsis.ortserver.storage.Key
+import org.eclipse.apoapsis.ortserver.storage.StorageEntry
+import org.eclipse.apoapsis.ortserver.storage.StorageProvider
+
+class AzureBlobStorageProvider(private val blobContainerClient: BlobContainerClient) : StorageProvider {
+    override suspend fun read(key: Key): StorageEntry {
+        val blobClient = blobContainerClient.getBlobClient(key.key)
+
+        return StorageEntry.create(
+            data = blobClient.openInputStream(),
```
nnobelis, line 40: Does this `InputStream` have to be closed afterwards ?

mnonnenmacher: It's documented for the `StorageEntry` class that closing the stream is in the responsibility of the caller. The test case does it for example in this line: `storage.read(Key("test")).use { ... }`

nnobelis (approved):

mnonnenmacher:
@nnobelis I had to fix some detekt issues.

nnobelis (approved):

PR #1491, opened 2024-11-25
fix(ui): Fix router getting stuck when redirecting to latest run

mmurto:
For some reason using the query directly sometimes froze the navigation.
Using the query client fixes it, but then a query invalidation is needed to
always have fresh data.

This is a fixup for aa31b4b.


sschuberth:
> This is a fixup for #aa31b4b.

Refer to commit hashes without the `#` to make it render correctly.

mmurto:
There's still a bug here.

Etsija (approved):

PR #1507, opened 2024-11-26
refactor(ui): Use JSON search params for table sorting state

mmurto:
To avoid having to manually format search params for table formatting, use
TanStack Router's ability to use JSON values in search params [1]. This
makes the logic easier to reason with, especially when implementing sorting
by multiple columns later.

While at it, align the sorting state used in search params with the data
model required by TanStack Table.


Etsija:
Commit message (and PR description) nit: missing the reference [1].

Etsija:
Otherwise, LGTM, and I'll need to rebase mine then, to align with this PR after it's merged.

PR #1514, opened 2024-11-27
style(ui): Remove the titles from actions to only have icons

sschuberth:
Create a more slick UI by removing the titles, the icons and tooltips should be enough. This makes room for more actions in the column, like deleting runs.

### DIFF
--- a/ui/src/routes/_layout/admin/runs/index.tsx
+++ b/ui/src/routes/_layout/admin/runs/index.tsx
@@ -195,8 +195,7 @@ const columns = [
                   runIndex: row.original.index.toString(),
                 }}
               >
-                View
-                <View className='ml-1 h-4 w-4' />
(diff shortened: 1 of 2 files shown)


ui/src/routes/_layout/admin/runs/index.tsx
```diff
@@ -195,7 +195,6 @@ const columns = [
                   runIndex: row.original.index.toString(),
                 }}
               >
-                View
                 <View className='ml-1 h-4 w-4' />
```
mmurto, line 198: Is the `ml-1` (margin left 1) still needed, or is the icon off-center with it?

mmurto (changes requested):

mmurto (dismissed):

mmurto (approved):

PR #1530, opened 2024-11-28
feat(core): Delete reports in storage on deletion of ORT run

wkl3nk:
On deletion of an ORT run via the REST API, also delete the associated reports from storage that were created during this ORT run.

### DIFF
--- a/model/src/commonMain/kotlin/repositories/OrtRunRepository.kt
+++ b/model/src/commonMain/kotlin/repositories/OrtRunRepository.kt
@@ -58,6 +58,13 @@ interface OrtRunRepository {
     fun getByIndex(repositoryId: Long, ortRunIndex: Long): OrtRun?
 
     /**
+     * Get the id of an ORT run by its [index][ortRunIndex] within a [repository][repositoryId].
+     * This function is more efficient than [getByIndex], as it only retrieves the ID of the ORT run or
+     * returns null if the ORT run is not found.
+     */
+    fun getIdByIndex(repositoryId: Long, ortRunIndex: Long): Long?
+
+    /**
      * List all ORT runs according to the given [parameters] and [filters].
      */
     fun list(
@@ -100,9 +107,4 @@ interface OrtRunRepository {
      * Delete an ORT run by [id].
      */
     fun delete(id: Long): Int
-
-    /**
-     * Delete an ORT run by [repositoryId] and [ortRunIndex].
-     */
-    fun deleteByRepositoryIdAndOrtRunIndex(repositoryId: Long, ortRunIndex: Long): Int
 }
--- a/services/hierarchy/build.gradle.kts
+++ b/services/hierarchy/build.gradle.kts
@@ -32,6 +32,7 @@ dependencies {
     api(libs.exposedCore)
 
     implementation(projects.dao)
+    implementation(projects.services.reportStorageService)
 
     runtimeOnly(libs.logback)
 
--- a/services/hierarchy/src/main/kotlin/RepositoryService.kt
+++ b/services/hierarchy/src/main/kotlin/RepositoryService.kt
@@ -96,6 +96,15 @@ class RepositoryService(
     }
 
     /**
+     * Get the id of an ORT run by its [index][ortRunIndex] within a [repository][repositoryId].
+     * This function is more efficient than [getOrtRun], as it only retrieves the ID of the ORT run or
+     * returns null if the ORT run is not found.
+     */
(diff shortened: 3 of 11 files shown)


model/src/commonMain/kotlin/repositories/OrtRunRepository.kt
```diff
@@ -57,6 +57,13 @@ interface OrtRunRepository {
      */
     fun getByIndex(repositoryId: Long, ortRunIndex: Long): OrtRun?
 
+    /**
+     * Gets the id of an ORT run by its [index][ortRunIndex] within a [repository][repositoryId].
```
oheger-bosch, line 61: Please use imperative voice: "Get the id...". I would not use the term "lightweight" to describe this operation; it just serves a different use case. If a client actually needs the `OrtRun` and only has the repository ID and index, there is even an additional step required.

services/hierarchy/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ group = "org.eclipse.apoapsis.ortserver.services"
 dependencies {
     api(projects.model)
     api(projects.services.authorizationService)
+    api(projects.services.reportStorageService)
```
oheger-bosch, line 31: Would it be sufficient to have this as `implementation` dependency?

services/hierarchy/src/main/kotlin/RepositoryService.kt
```diff
@@ -95,6 +95,14 @@ class RepositoryService(
         ortRunRepository.getByIndex(repositoryId, ortRunIndex)
     }
 
+    /**
+     * Lightweight alternative to [getOrtRun] that only returns the id of the ORT run or null
```
oheger-bosch, line 99: Again "Lightweight"?

services/hierarchy/src/test/kotlin/OrtRunServiceTest.kt
```diff
@@ -94,15 +107,85 @@ class OrtRunServiceTest : WordSpec() {
...
+                val jobId = createReporterJob(ortRunId)
+
+                val mockReportStorageService = mockk<ReportStorageService> {
+                    coEvery { deleteReport(any() as Long, any() as String) } just Runs
+                }
+
+                val service = createService(mockReportStorageService)
+                service.deleteOrtRun(ortRunId)
+
+                coVerify(exactly = 3) { mockReportStorageService.deleteReport(any() as Long, any() as String) }
```
oheger-bosch, line 123: Could you do a more exact verification, specifying the IDs and filenames?

services/hierarchy/src/test/kotlin/OrtRunServiceTest.kt
```diff
@@ -94,15 +107,85 @@ class OrtRunServiceTest : WordSpec() {
...
+
+                fixtures.reporterRunRepository.getByJobId(jobId) shouldBe null
+                fixtures.ortRunRepository.get(ortRunId) shouldBe null
+            }
+
+            "delete an ORT run even if some reports are not found in storage" {
+                val ortRunId = createOrtRun()
+                val jobId = createReporterJob(ortRunId)
+
+                val mockReportStorageService = mockk<ReportStorageService> {
```
oheger-bosch, line 133: In this case, it would be good to check whether other reports are deleted after an exception occured.

services/report-storage/src/main/kotlin/ReportStorageService.kt
```diff
@@ -71,6 +71,16 @@ class ReportStorageService(
...
     }
+
+    /**
+     * Deletes a report with the given [fileName] for the specified [runId].
+     */
+    suspend fun deleteReport(runId: Long, fileName: String) {
+        val key = generateKey(runId, fileName)
+        if (!reportStorage.containsKey(key)) throw ReportNotFoundException(runId, fileName)
+
+        reportStorage.delete(key)
```
oheger-bosch, line 82: `delete` returns a `Boolean` that is *false* if the report did not exist. So you could do
`if (!reportStorage.delete(key)) throw ReportNotFoundException(...)`

services/hierarchy/build.gradle.kts
wkl3nk: Yes, did not know the difference.

services/report-storage/src/main/kotlin/ReportStorageService.kt
wkl3nk: Good point. I was insecure because the implementation in the library could change over time, but the behaviour is even stated in the comment, so the risk of change is relatively low.

oheger-bosch (approved):

PR #1531, opened 2024-11-28
ui: Delete ORT runs

Etsija:

Screenshot from 2024-11-30 11-57-46

Screenshot from 2024-11-30 11-58-08

### DIFF
--- a/ui/src/components/delete-icon-button.tsx
+++ b/ui/src/components/delete-icon-button.tsx
@@ -34,7 +34,7 @@ const DeleteIconButton = forwardRef<HTMLButtonElement, DeleteIconButtonProps>(
         size='sm'
         variant='outline'
         {...props}
-        className={cn('h-9 px-2', className)}
+        className={cn('h-8 px-2', className)}
         ref={ref}
       >
         <span className='sr-only'>Delete</span>
--- a/ui/src/routes/organizations/$orgId/index.tsx
+++ b/ui/src/routes/organizations/$orgId/index.tsx
@@ -317,6 +317,7 @@ const OrganizationComponent = () => {
                 <span className='font-bold'>{organization.name}</span>?
               </>
             }
+            confirmationText={organization.name}
             trigger={<DeleteIconButton />}
           />
         </CardTitle>
--- a/ui/src/routes/organizations/$orgId/products/$productId/index.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/index.tsx
@@ -159,6 +159,7 @@ const ProductComponent = () => {
                     <span className='font-bold'>{product.name}</span>?
                   </>
                 }
+                confirmationText={product.name}
                 onDelete={handleDelete}
                 trigger={<DeleteIconButton />}
               />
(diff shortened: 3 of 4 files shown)


sschuberth:
Here's a semi-related thing I anyway wanted to bring up: When referring to runs, like in the delete confirmation dialog, I think we should have a general guideline whether to use local or global run IDs.

Here, a local run ID probably makes more sense, as that's what's visible in the row the user clicked on. But in any case I'd include the type of ID to all places in the UI that refer to IDs.

Also, for this case of the delete confirmation, I'd use a more "fluent" text without a colon saying

> Are you sure to delete the run with local ID 2?

Or actually, since this is a Yes / No question, but our choices are "Cancel" / "Delete", either rename the buttons or say

> Please confirm the deletion of the run with local ID 2.

and keep the buttons as they are. Opinions, @mmurto, @mnonnenmacher?

Etsija:
> Here's a semi-related thing I anyway wanted to bring up: When referring to runs, like in the delete confirmation dialog, I think we should have a general guideline whether to use local or global run IDs.
> 
> Here, a local run ID probably makes more sense, as that's what's visible in the row the user clicked on. But in any case I'd include the type of ID to all places in the UI that refer to IDs.
> 
> Also, for this case of the delete confirmation, I'd use a more "fluent" text without a colon saying
> 
> > Are you sure to delete the run with local ID 2?
> 
> Or actually, since this is a Yes / No question, but our choices are "Cancel" / "Delete", either rename the buttons or say
> 
> > Please confirm the deletion of the run with local ID 2.
> 
> and keep the buttons as they are. Opinions, @mmurto, @mnonnenmacher?

It's a bit difficult to change that without losing the generality of the delete dialog, which basically composes the text to show in the dialog.

mmurto:
> Here's a semi-related thing I anyway wanted to bring up: When referring to runs, like in the delete confirmation dialog, I think we should have a general guideline whether to use local or global run IDs.
> 
> Here, a local run ID probably makes more sense, as that's what's visible in the row the user clicked on. But in any case I'd include the type of ID to all places in the UI that refer to IDs.
> 
> Also, for this case of the delete confirmation, I'd use a more "fluent" text without a colon saying
> 
> > Are you sure to delete the run with local ID 2?
> 
> Or actually, since this is a Yes / No question, but our choices are "Cancel" / "Delete", either rename the buttons or say
> 
> > Please confirm the deletion of the run with local ID 2.
> 
> and keep the buttons as they are. Opinions, @mmurto, @mnonnenmacher?

IMO "local IDs" but using proper terminology for them, so "Run index". Could also mention the repository URL in the confirmation dialog to bring more context to it.

mmurto:
> The confirmation text in the delete dialog component is defined as
> 
> ```
>           <AlertDialogDescription>
>             Are you sure you want to delete this {item.descriptor}:{' '}
>             <span className='font-bold'>{item.name}</span>?
>           </AlertDialogDescription>
> ```
> 
> We wanted to do this in this way, to avoid having to write a separate confirmation text to every single delete dialog in the whole UI, because that would be more error prone. Suggestions to improve this are welcome, but if we want a tailored confirmation message in each dialog, some of the generality will be lost - and I'd need to change every dialog. That is probably outside of the scope of this PR.

This is a little different than the other deletable items as it shouldn't be identifiable by single ID, but the combination of the repository and run index. I'd either make the generic dialog more generic or create a custom component for this.

sschuberth:
> I'd either make the generic dialog more generic

I'd vote for that.

Etsija:
@mmurto I rebased after your delete dialog simplifications (which were great!), see what you think of it now.

mmurto (approved):

PR #1538, opened 2024-11-29
Deduplicate scan summaries

oheger-bosch:
This PR prevents the creation of duplicate entities for scan summaries in cases where a single scan operation yields multiple results.

### DIFF
--- a/workers/scanner/src/test/kotlin/OrtServerScanResultStorageTest.kt
+++ b/workers/scanner/src/test/kotlin/OrtServerScanResultStorageTest.kt
@@ -50,6 +50,14 @@ class OrtServerScanResultStorageTest : WordSpec() {
     private lateinit var scanResultStorage: OrtServerScanResultStorage
     private lateinit var scannerRun: ScannerRun
 
+    /**
+     * Return a [Set] with the IDs of all scan summaries assigned to a scan result.
+     */
+    private fun loadScanSummaryIds(): Set<Long> =
+        dbExtension.db.blockingQuery {
+            ScanResultDao.all().map { it.scanSummary.id.value }.toSet()
+        }
+
     init {
         beforeEach {
             scannerRun = dbExtension.fixtures.scannerRunRepository.create(dbExtension.fixtures.scannerJob.id)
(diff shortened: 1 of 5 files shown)


workers/scanner/src/test/kotlin/OrtServerScanResultStorageTest.kt
```diff
@@ -101,7 +73,7 @@ class OrtServerScanResultStorageTest : WordSpec() {
 
                 verifyAssociatedScanResults(scannerRun, scanResult)
 
-                scanResultStorage.read(provenance, scannerMatcher) shouldBe listOf(scanResult)
+                scanResultStorage.read(provenance.mapToOrt(), scannerMatcher) shouldBe listOf(scanResult)
```
nnobelis, line 84: Why these changes with `mapToOrt` ? They seem unrelated to this commit.

workers/scanner/src/main/kotlin/scanner/ScanSummaryHash.kt
```diff
@@ -0,0 +1,78 @@
+/*
```
nnobelis, line 1: Why create a full featured hash function instead of an `equals` one ?

You don't store the hash anywhere and you'd save on memory because you would not need to build a huge string.

workers/scanner/src/test/kotlin/OrtServerScanResultStorageTest.kt
oheger-bosch: `mapToOrt` is now needed, since the data types have changed. The tests have to deal with both ORT types and ORT Server types. Test data is now created using the model of ORT Server, but the scan result storage expects types from ORT's model.

workers/scanner/src/main/kotlin/scanner/ScanSummaryHash.kt
oheger-bosch: The hashes were used because again different types (ORT model and ORT Server model) are involved. But you are right, this is not the most efficient approach. Will try to come up with a better solution.

oheger-bosch: I have added a comparison function that should be more efficient.

nnobelis (approved):

PR #1540, opened 2024-11-29
Avoid some unnecessary queries

mnonnenmacher:
Avoid some unnecessary queries when assigning foreign key columns, see the commit messages for details.


oheger-bosch:
So, this means we basically have to duplicate the properties for foreign key relations, to have one object and one ID property? To be honest, I am getting more and more doubt whether Exposed is the right choice for the DB layer...

mnonnenmacher:
> So, this means we basically have to duplicate the properties for foreign key relations, to have one object and one ID property? To be honest, I am getting more and more doubt whether Exposed is the right choice for the DB layer...

I think the DSL is ok, it's the DAO layer that is making problems. But I think that all ORMs have their own set of issues because objects and database rows are just not the same thing. I would like to use something that is closer to the database, like jOOQ or SQLDelight, but the jOOQ open source version has limited compatibility with PostgreSQL versions (https://www.jooq.org/download/support-matrix) and when I tried SQLDelight two years ago before version 2 was released there were some problems with PostgreSQL support.

oheger-bosch:
If it helps to reduce the number of queries, I am fine with merging this, although I do not really like the solution. I would approve, but there are now merge conflicts.

oheger-bosch (approved):

PR #1554, opened 2024-12-02
feat(model): Add project descriptions

mnonnenmacher:
Add support for the new `description` property of projects which was introduced in ORT 42.0.0 [1].

[1]: https://github.com/oss-review-toolkit/ort/releases/tag/42.0.0

### DIFF
--- a/dao/src/main/resources/db/migration/V89__addProjectDescription.sql
+++ b/dao/src/main/resources/db/migration/V89__addProjectDescription.sql
@@ -1,0 +1,3 @@
+-- Add a description column to the projects table because project descriptions were added to ORT.
+
+ALTER TABLE projects ADD COLUMN description TEXT NOT NULL DEFAULT '';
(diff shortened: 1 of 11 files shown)


MarcelBochtler (approved):

dao/src/main/resources/db/migration/V89__addProjectDescription.sql
```diff
@@ -0,0 +1,3 @@
+-- Add a description column to the projects table because project descriptions were added to ORT.
+
+ALTER TABLE projects ADD COLUMN description TEXT NOT NULL DEFAULT '';
```
MarcelBochtler, line 3: In the upstream model, the description is optional.
Why is the description required in the DB?

mnonnenmacher: `Project.description` is not nullable in the ORT model.

PR #1555, opened 2024-12-02
feat(core): Delete ORT runs before deleting the repository

### LINKED ISSUE #100: Repository cannot be deleted because a run is linked to it
Whenever an ORT run was started for a repository, one can no longer delete this repository:

```
DELETE https://.../api/v1/repositories/123

Request uses authentication

HTTP/1.1 500 Internal Server Error
x-engine: Ktor
date: Tue, 16 Apr 2024 15:02:08 GMT
server: Ktor/2.3.8
content-length: 301
content-type: application/json

{
  "message": "Error when processing the request.",
  "cause": "org.postgresql.util.PSQLException: ERROR: update or delete on table \"repositories\" violates foreign key constraint \"ort_runs_repository_id_fkey\" on table \"ort_runs\"\n  Detail: Key (id)=(123) is still referenced from table \"ort_runs\"."
}
```

MarcelBochtler:
Enable the user to delete a repository by recursively deleting the ORT runs which are linked to this specific repository.

Resolves #100.

### DIFF
--- a/dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
+++ b/dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
@@ -176,6 +176,10 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
     override fun delete(id: Long): Int = db.blockingQuery {
         OrtRunsTable.deleteWhere { OrtRunsTable.id eq id }
     }
+
+    override fun deleteByRepository(repositoryId: Long): Int = db.blockingQuery {
+        OrtRunsTable.deleteWhere { OrtRunsTable.repositoryId eq repositoryId }
+    }
 }
 
 /**
--- a/dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
+++ b/dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
@@ -25,6 +25,8 @@ import org.eclipse.apoapsis.ortserver.dao.blockingQuery
 import org.eclipse.apoapsis.ortserver.dao.blockingQueryCatching
 import org.eclipse.apoapsis.ortserver.dao.entityQuery
 import org.eclipse.apoapsis.ortserver.dao.mapAndDeduplicate
+import org.eclipse.apoapsis.ortserver.dao.repositories.product.ProductsTable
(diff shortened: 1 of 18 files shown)


MarcelBochtler:
Re: https://github.com/eclipse-apoapsis/ort-server/issues/100#issuecomment-2060676911

> As a related question, what should happen to a run if a repository gets deleted meanwhile? Should the run be cancelled? Should the deletion be postponed until the run completes? Should the user be able to choose?

This implementation basically ignores this issue, causing the job which is running during the deletion, failing, because the result cannot be written into the database.

I don't think that we should completely block the deletion of a repository during a running job because deleting a repository is IMO an administrative task, while running / triggering a job is a task of a user and I don't think a user should be able to block an administrator.

Additionally, there is currently no concept of canceling running jobs, which might be the best approach in this scenario, but I think it will be ok, if this is implemented in an iteration of this feature.

dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
```diff
@@ -176,6 +177,12 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
     override fun delete(id: Long): Int = db.blockingQuery {
         OrtRunsTable.deleteWhere { OrtRunsTable.id eq id }
     }
+
+    override fun deleteAll(ids: List<Long>): Int = db.blockingQuery {
+        if (ids.isEmpty()) return@blockingQuery 0
+
+        OrtRunsTable.deleteWhere { OrtRunsTable.id inList ids }
+    }
```
mnonnenmacher, line 185: How about making a function to delete ORT runs by repository id instead? I believe that would make the implementation simpler on the caller side, because you don't have to fetch the list of runs first. Also, I don't know what the limitation for a list in a query is, but this implementation might fail if `ids` becomes too large.

model/src/commonMain/kotlin/repositories/OrtRunRepository.kt
```diff
@@ -112,4 +113,9 @@ interface OrtRunRepository {
      * Delete the ORT runs of a repository, specified by the [repositoryId].
      */
     fun deleteByRepository(repositoryId: Long): Int
+
+    /**
+     * Delete all ORT runs, associated to this [productId].
```
mnonnenmacher, line 118: Nit: No comma after "runs".

mnonnenmacher (approved):

PR #1557, opened 2024-12-03
ui: Truncate long report filename

Etsija:


lamppu:
Looks like the surrounding button has a class called [whitespace-nowrap](https://tailwindcss.com/docs/whitespace#no-wrap) which prevents `break-all` from working. If you'd prefer to use `break-all`, you can give the surrounding button the class `whitespace-normal`.

lamppu (approved):

PR #1559, opened 2024-12-03
feat(services): Add functions for getting run ID in various cases

lamppu:
Add functions for getting the latest ORT run ID for a repository, where a particular job has finished (or failed). These functions will be used in upcoming endpoints that will provide the information for the product overview page (which currently gathers this data from various endpoints) and an upcoming organization overview page.

### DIFF
--- a/model/src/commonMain/kotlin/JobStatus.kt
+++ b/model/src/commonMain/kotlin/JobStatus.kt
@@ -25,22 +25,36 @@ import kotlinx.serialization.Serializable
  * The status of a Job.
  */
 @Serializable
-enum class JobStatus {
+enum class JobStatus(
+    /** A flag that indicates whether the status is a final status. */
+    val final: Boolean,
+
+    /** A flag that indicates whether the status is a successful status. */
+    val successful: Boolean
+) {
     /** The job was created in the database but not yet scheduled for execution. */
-    CREATED,
+    CREATED(false, false),
 
     /** The job was scheduled for execution. */
-    SCHEDULED,
+    SCHEDULED(false, false),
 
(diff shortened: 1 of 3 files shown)


model/src/commonMain/kotlin/JobStatus.kt
```diff
@@ -25,22 +25,36 @@ import kotlinx.serialization.Serializable
...
     /** The job has finished but there were some issues over the threshold. */
-    FINISHED_WITH_ISSUES
+    FINISHED_WITH_ISSUES(true, true);
+
+    companion object {
+        /** List of final statuses. */
+        val FINAL_STATUSES: List<JobStatus> = entries.filter { it.final }
+
+        /** List of successful statuses. */
+        val SUCCESSFUL_STATUSES: List<JobStatus> = entries.filter { it.successful }
```
sschuberth, line 58: It's probably a matter of taste, but for simplicity I guess I would have hard-coded these lists with the respective enum values instead of giving each enum a `final` and `successful` property. But let's see what @mnonnenmacher says.

mnonnenmacher: I think it can be convenient in some situations to check the properties for a specific job status like `job.status.final`, at least that was the case for `OrtRunStatus`.

mnonnenmacher (approved):

model/src/commonMain/kotlin/JobStatus.kt
sschuberth: True, but with the other approach the same check would be written as `job.status in FINAL_STATUSES`. But granted, that's a bit longer.

PR #1562, opened 2024-12-03
Add function to get vulnerabilities for ORT runs

lamppu:

### DIFF
--- a/model/src/commonMain/kotlin/VulnerabilityRating.kt
+++ b/model/src/commonMain/kotlin/VulnerabilityRating.kt
@@ -1,0 +1,33 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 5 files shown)


model/src/commonMain/kotlin/VulnerabilityRating.kt
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package org.eclipse.apoapsis.ortserver.model
+
+/**
+ * The overall rating for a vulnerability. The rating is derived from the severities of the individual references of a
+ * vulnerability as the highest severity found, that matches one of the [VulnerabilityRating] values. If no such
+ * severities are found, the rating is [NONE].
+ * */
+enum class VulnerabilityRating(val score: Int) {
```
sschuberth, line 27: "score" probably is a bit of a misleading variable name here... the value seems to be more used like an ordinal value for the enum, which does not need to be stored explicitly, as it's available built-in like `CRITICAL.ordinal`. Just reorder enum entries so that `NONE` gets 0 assigned implicitly.

model/src/commonMain/kotlin/VulnerabilityRating.kt
```diff
@@ -0,0 +1,40 @@
...
+    CRITICAL(4),
+    HIGH(3),
+    MEDIUM(2),
+    LOW(1),
+    NONE(0);
+
+    companion object {
+        /** Get vulnerability rating by score. */
+        fun fromScore(score: Int?): VulnerabilityRating {
```
sschuberth, line 36: It's a bit unclear to me whether you want to create a `VulnerabilityRating` from a (actually floating point) vulnerability score value (similar to like ORT does [here](https://github.com/oss-review-toolkit/ort/blob/c7d73120c7655a796863b02a7c00219ced5fa0ac/model/src/main/kotlin/vulnerabilities/Cvss3Rating.kt#L39-L51)), or from an ordinal that represents the enum. The latter can be done like `VulnerabilityRating.values[VulnerabilityRating.CRITICAL.ordinal]` (use the ordinal as an index to he values array).

lamppu: Yes the ordinality is what I'm trying to go for here.

sschuberth:
@mnonnenmacher could you please have a look?

mnonnenmacher (approved):

PR #1584, opened 2024-12-06
feat(cli): Add `InfoCommand` to retrieve details of an `OrtRun`

bs-ondem:

### DIFF
--- a/cli/src/main/kotlin/InfoCommand.kt
+++ b/cli/src/main/kotlin/InfoCommand.kt
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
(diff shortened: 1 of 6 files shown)


cli/src/main/kotlin/InfoCommand.kt
```diff
@@ -0,0 +1,70 @@
...
+    ).convert { it.uppercase() }.groupChoice(
+        "ID" to OrtRunByIdOptions(),
+        "INDEX" to OrtRunByIndexOptions()
+    ).required()
+
+    override suspend fun run() {
+        val client = OrtServerClient.create(config.toOrtServerClientConfig())
+
+        val ortRun = when (val it = getOrtRunOption) {
+            is OrtRunByIdOptions -> {
```
oheger-bosch, line 48: General note (not related to this PR): It seems to be weird that different endpoints have to be queried for ORT runs depending on the provided parameters; maybe we have to look again at the API.

cli/src/main/kotlin/InfoCommand.kt
```diff
@@ -0,0 +1,70 @@
...
+import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.types.long
+
+import kotlinx.serialization.encodeToString
+
+import org.eclipse.apoapsis.ortserver.client.OrtServerClient
+
+class InfoCommand(private val config: OrtServerOptions) : SuspendingCliktCommand(name = "info") {
+    private val getOrtRunOption by option(
+        "--get-by",
```
oheger-bosch, line 37: Is there a way in clikt to get rid of this `get-by` parameter? It seems to be redundant for users. If an `id` (or maybe `run-id`) option is specified, it is clear that the run is to be fetched by its ID. Analogously, if there are options for the repository ID and the index, the other method is used. If there are missing or conflicting options, the command should issue an error. Is this possible?

sschuberth: While it does not seem to be possible to combine `mutuallyExclusiveOptions` and `cooccurring`, maybe [this](https://github.com/ajalt/clikt/issues/86#issuecomment-528917052) helps.

PR #1589, opened 2024-12-07
feat(secrets): Add support for Azure Key Vault

mnonnenmacher:
Add a secrets provider implementation for Azure Key Vault [1].

There is no test environment available for Azure Key Vault, therefore no integration tests have been implemented. Given that the provider is only a trivial wrapper around the Azure SDK, there is not much to test anyway.

[1]: https://azure.microsoft.com/en-us/services/key-vault/


secrets/azure-keyvault/src/main/kotlin/AzureKeyvaultProvider.kt
```diff
@@ -0,0 +1,54 @@
...
+        val secretType = when {
+            organizationId != null -> "organization-$organizationId"
+            productId != null -> "product-$productId"
+            repositoryId != null -> "repository-$repositoryId"
+            else -> throw IllegalArgumentException(
+                "Either one of organizationId, productId or repositoryId should be specified to create a path."
+            )
+        }
+
+        return Path("$secretType-$secretName")
```
sschuberth, line 52: Is the dash used here the reason why you do not rely on the default implementation, which uses an underscore?

mnonnenmacher: Yes, Azure Key Vault does not allow underscores.

sschuberth: As that's quite an interesting restriction, mind adding a code comment for that? Maybe we should even add a `check()` that secret names themselves do not contain dashes either, similar to the [checks for Scaleway](https://github.com/eclipse-apoapsis/ort-server/blob/52e1c7f73b275ae0ccdf1c725a1022d72e092b46/secrets/scaleway/src/main/kotlin/ScalewaySecretsProvider.kt#L218-L228).

PR #1607, opened 2024-12-10
fix(ui): Convert Tailwind CSS config to ESM

mnonnenmacher:
Convert the Tailwind CSS config file from CJS to ESM syntax to fix the build with Node 22.12.0 [1].

[1]: https://github.com/nodejs/node/issues/56155

### DIFF
--- a/ui/tailwind.config.js
+++ b/ui/tailwind.config.js
@@ -17,8 +17,11 @@
  * License-Filename: LICENSE
  */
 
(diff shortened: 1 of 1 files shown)


sschuberth (approved):

mnonnenmacher:
@lamppu @Etsija @mmurto According to the linked ticket another option would have been to change the file extension from `.js` to `.cjs`. If you would have preferred that for some reason, feel free to make a follow-up PR.

PR #1611, opened 2024-12-10
ci(docker): Fix calculating the ORT Server version

### LINKED ISSUE #1610: ORT Server version information is wrong
Although the latest [tag](https://github.com/eclipse-apoapsis/ort-server/tags) is 0.1.0-RC6, both the recently published images and the UI display something prefixed with "0.0.1-RC.001..."

image

image

The SHA1s seem to be recent, though.

mnonnenmacher:
Set `fetch-depth` to `0` to checkout the whole commit history including tags. This is required to correctly calculate the ORT Server version.

Fixes #1610.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -84,6 +84,8 @@ jobs:
     steps:
     - name: Checkout Repository
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      with:
+        fetch-depth: 0
 
     - name: Free Disk Space
       if: ${{ matrix.docker.freeDiskSpace }}


.github/workflows/docker-build.yml
```diff
@@ -84,6 +84,8 @@ jobs:
     steps:
     - name: Checkout Repository
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      with:
+        fetch-depth: 0
```
sschuberth, line 88: Would `fetch-tags: true` work as well?

sschuberth: I guess [not](https://github.com/actions/checkout/issues/1471).

sschuberth (approved):

.github/workflows/docker-build.yml
mnonnenmacher: I saw it as well but then found https://github.com/actions/checkout/issues/1471, so I just stick with the same config as in the release job.

PR #1619, opened 2024-12-11
Refactor run statistics count functions to take in multiple run IDs

lamppu:
Edit the count functions to count items for multiple runs to make the functions reusable for statistics endpoints on the product and organization levels. 
- The issues count is accumulated as a total sum from issues in all runs (as I wasn't sure if there should be some deduplication here)
- The vulnerabilities are counted as number of unique vulnerability+identifier pairs (could also be counted as vulnerabilities with unique external IDs, ignoring the identifiers?)
- The rule violations are counted as distinct rule violations found in runs
- The packages are counted as distinct packages found in runs, and ecosystems are counted similarly from distinct packages

Let me know if you think any of these should be counted differently.

Please see the individual commits for more details.

### DIFF
--- a/core/src/main/kotlin/api/RunsRoute.kt
+++ b/core/src/main/kotlin/api/RunsRoute.kt
@@ -277,12 +277,10 @@ fun Route.runs() = route("runs") {
 
                     val jobs = repositoryService.getJobs(ortRun.repositoryId, ortRun.index)
 
-                    val finishedStateSet: Set<JobStatus> = setOf(JobStatus.FINISHED, JobStatus.FINISHED_WITH_ISSUES)
-
-                    val analyzerJobInFinalState = jobs?.analyzer?.status in finishedStateSet.plus(JobStatus.FAILED)
-                    val analyzerJobInFinishedState = jobs?.analyzer?.status in finishedStateSet
-                    val advisorJobInFinishedState = jobs?.advisor?.status in finishedStateSet
-                    val evaluatorJobInFinishedState = jobs?.evaluator?.status in finishedStateSet
+                    val analyzerJobInFinalState = jobs?.analyzer?.status in JobStatus.FINAL_STATUSES
+                    val analyzerJobInFinishedState = jobs?.analyzer?.status in JobStatus.SUCCESSFUL_STATUSES
+                    val advisorJobInFinishedState = jobs?.advisor?.status in JobStatus.SUCCESSFUL_STATUSES
+                    val evaluatorJobInFinishedState = jobs?.evaluator?.status in JobStatus.SUCCESSFUL_STATUSES
 
                     val issuesCount = if (analyzerJobInFinalState) issueService.countForOrtRunId(ortRun.id) else null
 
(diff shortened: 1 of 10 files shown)


services/hierarchy/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -63,58 +63,12 @@ class PackageServiceTest : WordSpec() {
...
-                    vcsProcessed = VcsInfo(
-                        RepositoryType("GIT"),
-                        "https://example.com/git",
-                        "revision2",
-                        "path"
-                    ),
-                    isMetadataOnly = false,
-                    isModified = false
-                )
+                val pkg1 = generatePackage(Identifier("Maven", "com.example", "example", "1.0"))
```
sschuberth, line 66: Commit message nit: "Use *a* helper function..."

services/hierarchy/src/main/kotlin/IssueService.kt
```diff
@@ -81,10 +81,11 @@ class IssueService(private val db: Database) {
         )
     }
 
-    suspend fun countForOrtRunId(ortRunId: Long): Long = db.dbQuery {
+    /** Count issues found in provided ORT runs. */
+    suspend fun countForOrtRunIds(ortRunIds: List<Long>): Long = db.dbQuery {
```
sschuberth, line 85: To avoid the need for callers to create lists even for single IDs, consider using `vararg ortRunIds: Long` as the parameter.

sschuberth: Similar for other functions.

core/src/main/kotlin/api/RunsRoute.kt
```diff
@@ -282,24 +282,25 @@ fun Route.runs() = route("runs") {
                     val advisorJobInFinishedState = jobs?.advisor?.status in JobStatus.SUCCESSFUL_STATUSES
                     val evaluatorJobInFinishedState = jobs?.evaluator?.status in JobStatus.SUCCESSFUL_STATUSES
 
-                    val issuesCount = if (analyzerJobInFinalState) issueService.countForOrtRunId(ortRun.id) else null
+                    val issuesCount =
+                        if (analyzerJobInFinalState) issueService.countForOrtRunIds(ortRun.id) else null
 
                     val packagesCount =
-                        if (analyzerJobInFinishedState) packageService.countForOrtRunId(ortRun.id) else null
+                        if (analyzerJobInFinishedState) packageService.countForOrtRunIds(ortRun.id) else null
```
sschuberth, line 289: Nit: I haven't checked, but maybe some of these now still fit a single line again, which would be good to keep in order to reduce the diff. Similar below.

services/hierarchy/src/main/kotlin/PackageService.kt
```diff
@@ -52,20 +51,23 @@ class PackageService(private val db: Database) {
         }
     }
 
-    suspend fun countForOrtRunId(ortRunId: Long): Long = db.dbQuery {
+    /** Count packages found in provided ORT runs. */
+    suspend fun countForOrtRunIds(vararg ortRunIds: Long): Long = db.dbQuery {
         PackagesTable.joinAnalyzerTables()
             .select(PackagesTable.id)
-            .where { AnalyzerJobsTable.ortRunId eq ortRunId }
+            .where { AnalyzerJobsTable.ortRunId inList ortRunIds.toList() }
```
sschuberth, line 58: Prefer `asList()` over `toList()` here, as the former does not make a copy, but just provides another "view" on the same data. Similar in other places.

core/src/main/kotlin/api/RunsRoute.kt
lamppu: Yeah, I checked the other ones before but looks like I missed the issuesCount

sschuberth:
LGTM, but I'd like @mnonnenmacher to have a look as well.

oheger-bosch (approved):

PR #1628, opened 2024-12-12
Handle missed schedules in Orchestrator

oheger-bosch:
This PR addresses a problem with Orchestrator. Under specific conditions, it is possible that a schedule of a worker job gets lost. The affected ORT run then stays in ACTIVE state forever. Kubernetes Job Monitor was not able to detect this situation.

Refer to the individual commits for further information.

### DIFF
--- a/dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
+++ b/dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
@@ -31,6 +31,7 @@ import org.eclipse.apoapsis.ortserver.dao.tables.shared.OrtRunIssueDao
 import org.eclipse.apoapsis.ortserver.dao.utils.applyFilter
 import org.eclipse.apoapsis.ortserver.dao.utils.listQuery
 import org.eclipse.apoapsis.ortserver.dao.utils.toDatabasePrecision
+import org.eclipse.apoapsis.ortserver.model.ActiveOrtRun
 import org.eclipse.apoapsis.ortserver.model.JobConfigurations
 import org.eclipse.apoapsis.ortserver.model.OrtRun
 import org.eclipse.apoapsis.ortserver.model.OrtRunFilters
@@ -138,6 +139,12 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
             ListQueryResult(emptyList(), parameters, 0L)
         }
 
+    override fun listActiveRuns(): List<ActiveOrtRun> = db.blockingQuery {
+        OrtRunDao.find {
+            OrtRunsTable.status inList activeRunStatuses
+        }.map { ActiveOrtRun(it.id.value, it.createdAt, it.traceId) }
+    }
+
     override fun update(
         id: Long,
         status: OptionalValue<OrtRunStatus>,
@@ -199,3 +206,6 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
  */
 private fun getLabelDao(entry: Map.Entry<String, String>): LabelDao =
     LabelDao.getOrPut(entry.key, entry.value)
+
+/** A collection with the statuses in which an ORT run is considered active. */
+private val activeRunStatuses = setOf(OrtRunStatus.CREATED, OrtRunStatus.ACTIVE)
--- a/kubernetes/jobmonitor/src/main/kotlin/LostJobsFinder.kt
+++ b/kubernetes/jobmonitor/src/main/kotlin/LostJobsFinder.kt
@@ -19,7 +19,10 @@
 
 package org.eclipse.apoapsis.ortserver.kubernetes.jobmonitor
 
+import io.kubernetes.client.openapi.models.V1Job
+
 import org.eclipse.apoapsis.ortserver.kubernetes.jobmonitor.JobHandler.Companion.ortRunId
+import org.eclipse.apoapsis.ortserver.model.WorkerJob
 import org.eclipse.apoapsis.ortserver.model.repositories.AdvisorJobRepository
 import org.eclipse.apoapsis.ortserver.model.repositories.AnalyzerJobRepository
 import org.eclipse.apoapsis.ortserver.model.repositories.EvaluatorJobRepository
@@ -30,6 +33,7 @@ import org.eclipse.apoapsis.ortserver.model.repositories.ScannerJobRepository
 import org.eclipse.apoapsis.ortserver.model.repositories.WorkerJobRepository
 import org.eclipse.apoapsis.ortserver.transport.AdvisorEndpoint
 import org.eclipse.apoapsis.ortserver.transport.AnalyzerEndpoint
+import org.eclipse.apoapsis.ortserver.transport.ConfigEndpoint
 import org.eclipse.apoapsis.ortserver.transport.Endpoint
 import org.eclipse.apoapsis.ortserver.transport.EvaluatorEndpoint
 import org.eclipse.apoapsis.ortserver.transport.NotifierEndpoint
@@ -51,6 +55,10 @@ import org.slf4j.MDC
  * status in Kubernetes. It looks for jobs that should be running according to the database, but do not exist
  * (anymore) in Kubernetes. For such jobs, it sends a notification to the Orchestrator, to give it the chance to act
  * accordingly.
+ *
+ * Another problem detected by this class is missing schedules, i.e. active ORT runs for which all jobs have been
+ * completed. For such runs, a different type of notification is sent to the Orchestrator, so that it can try to resume
+ * the run.
  */
 @Suppress("LongParameterList")
 internal class LostJobsFinder(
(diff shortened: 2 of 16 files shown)


oheger-bosch:
This replaces #1586.

dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
nnobelis, line 1: Commit message: _the_ Kubernetes Job Monitor

dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
```diff
@@ -138,6 +140,13 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
             ListQueryResult(emptyList(), parameters, 0L)
         }
 
+    override fun listActiveRuns(): List<ActiveOrtRun> = db.blockingQuery {
+        OrtRunDao.find {
+            OrtRunsTable.status eq OrtRunStatus.CREATED or
```
nnobelis, line 145: _nit_ you can use `inList listOf(OrtRunStatus.CREATED, OrtRunStatus.ACTIVE)`

dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
```diff
@@ -138,6 +140,13 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
...
+                    (OrtRunsTable.status eq OrtRunStatus.ACTIVE)
+        }.map { ActiveOrtRun(it.id.value, it.createdAt, it.traceId) }
```
nnobelis, line 147: I think this is typically the case I would't use an ORM and run instead some kind of native SQL (https://github.com/JetBrains/Exposed/issues/118): because even is you say the opposite in the commit message, _the full ORT run is still loaded from the DB_ no ?

model/src/commonMain/kotlin/ActiveOrtRun.kt
```diff
@@ -0,0 +1,36 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.model
+
+import kotlinx.datetime.Instant
+
+/**
+ * A data class holding a minimum set of properties to reference an active ORT run.
```
nnobelis, line 25: s/minimal/minimum

orchestrator/src/main/kotlin/Orchestrator.kt
```diff
@@ -340,9 +332,22 @@ class Orchestrator(
...
+        return fetchNextJobs(scheduleContext)
+    }
+
+    /**
+     * Convenience function to evaluate and process this [Result] with information about the next jobs to be scheduled.
+     * If the result is successful, actually trigger the jobs. Otherwise, call the given [onError] function with the
+     * exception that occurred.
+     */
+    private fun Result<Pair<WorkerScheduleContext, List<JobScheduleFunc>>>.scheduleNextJobs(
+        onError: (Throwable) -> Unit
```
nnobelis, line 367: Don't you want to stick with `onFailure` ? Introducing the new name `onError` is unecessary complex IMO

orchestrator/src/main/kotlin/Orchestrator.kt
```diff
@@ -340,9 +332,22 @@ class Orchestrator(
 
         val ortRun = getCurrentOrtRun(ortRunId)
         val scheduleContext = createWorkerSchedulerContext(ortRun, header, workerJobs = jobs)
-        val schedules = WorkerScheduleInfo.entries.mapNotNull { it.createAndScheduleJobIfPossible(scheduleContext) }
 
-        return scheduleContext to schedules
+        return fetchNextJobs(scheduleContext)
```
nnobelis, line 358: how about `scheduleContext to fetchNextJobs(scheduleContext)` ?

Then you would not have to return a `pair` from the function.

orchestrator/src/main/kotlin/Entrypoint.kt
```diff
@@ -101,6 +102,7 @@ class OrchestratorComponent : EndpointComponent<OrchestratorMessage>(Orchestrato
             is NotifierWorkerResult -> orchestrator.handleNotifierWorkerResult(message.header, payload)
             is NotifierWorkerError -> orchestrator.handleNotifierWorkerError(message.header, payload)
             is WorkerError -> orchestrator.handleWorkerError(message.header, payload)
+            is LostSchedule -> orchestrator.handleLostSchedule(message.header, payload)
```
nnobelis, line 105: Keep alphabetical order

kubernetes/jobmonitor/src/main/kotlin/LostJobsFinder.kt
```diff
@@ -51,6 +55,10 @@ import org.slf4j.MDC
  * status in Kubernetes. It looks for jobs that should be running according to the database, but do not exist
  * (anymore) in Kubernetes. For such jobs, it sends a notification to the Orchestrator, to give it the chance to act
  * accordingly.
+ *
+ * Another problem detected by this class is missing schedules, i.e. active ORT runs for which all jobs have been
+ * completed. For such runs, a different type of notification is sent to the Orchestrator, so that it try to resume
```
nnobelis, line 60: it _tries_

kubernetes/jobmonitor/src/main/kotlin/LostJobsFinder.kt
```diff
@@ -113,15 +121,31 @@ internal class LostJobsFinder(
...
+
+        // Keep track of runs for which active jobs exist. These runs are not subject to lost schedules.
+        val ortRunsWithJobs = mutableSetOf<Long>()
 
-        jobRepositories.forEach { (endpoint, jobRepository) -> checkForLostWorkerJobs(endpoint, jobRepository) }
+        jobRepositories.forEach { (endpoint, jobRepository) ->
+            val (jobsMap, lostJobs) = gatherLostJobData(endpoint, jobRepository)
+            handleLostWorkerJobs(endpoint, lostJobs)
+
+            ortRunsWithJobs.addAll(jobsMap.keys.filterNotNull())
```
nnobelis, line 133: Can you use `+=` here, and in other places ?

kubernetes/jobmonitor/src/main/kotlin/LostJobsFinder.kt
```diff
@@ -136,6 +160,16 @@ internal class LostJobsFinder(
         val lostJobs = jobRepository.listActive(currentTime - monitorConfig.lostJobsMinAge)
```
nnobelis, line 160: So you are listing the active jobs and you are only retaining the ones that of not for this endpoint ?!

I have a hard time understanding this function and comparing it to _checkForLostSchedules_.

Maybe you should define first in its documentation what is a _lost job_ ?

kubernetes/jobmonitor/src/main/kotlin/LostJobsFinder.kt
```diff
@@ -136,6 +160,16 @@ internal class LostJobsFinder(
...
+    }
+
+    /**
+     * Perform actions to handle the given list of [lostJobs] for the given [endpoint].
+     */
+    private fun handleLostWorkerJobs(
+        endpoint: Endpoint<*>,
+        lostJobs: List<WorkerJob>
+    ) {
         if (lostJobs.isNotEmpty()) {
```
nnobelis, line 173: This function only does some logging !? If yes, why not name it `logLostWorkerJobs` ?

transport/kubernetes/src/main/kotlin/KubernetesMessageSender.kt
```diff
@@ -141,6 +147,14 @@ internal class KubernetesMessageSender<T : Any>(
         api.createNamespacedJob(msgConfig.namespace, jobBody, null, null, null, null)
     }
 
+    /**
+     * Generate a valid trace ID from the given [message]. Check the ID contained in the message. If it is empty, a
+     * unique ID is generated. This is needed because the job name is derived from the trace ID, and Kubernetes
+     * as some restrictions for such names.
+     */
+    private fun validTraceId(message: Message<T>): String =
+        message.header.traceId.takeUnless { it.isBlank() } ?: "undefined-${nameCounter.incrementAndGet()}"
```
nnobelis, line 156: Is the trace id supposed to be unique ?
Could you not get collisions if core (orcherstrator?) pod is evicted and recreated, of if the cluster restart ?

Why not use some UUIDs instead of the counter ?

orchestrator/src/main/kotlin/Orchestrator.kt
```diff
@@ -395,12 +417,15 @@ class Orchestrator(
     }
 
     /**
-     * Publish a message to the [ConfigEndpoint] and update the current ORT run to the active state.
+     * Publish a message with the given [header] to the [ConfigEndpoint] to trigger the Config worker job for the given
+     * [run]. If [updateRun] is *true*, set the status of the run to ACTIVE.
      */
-    private fun scheduleConfigWorkerJob(run: OrtRun, header: MessageHeader) {
+    private fun scheduleConfigWorkerJob(run: OrtRun, header: MessageHeader, updateRun: Boolean) {
```
nnobelis, line 423: I am a bit puzzled by this change.

Why make this a parameter ?
In your PR, it `seems` your are always calling it with `true`.

dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
oheger-bosch: In the commit message, I talk about "full `OrtRun` objects with all their relations".  This is no longer the case with this new model class; only a single `SELECT` statement is executed.

nnobelis: Understood.
But you get my point ? You are still loading 16 columns from the DB when 3 are needed.

orchestrator/src/main/kotlin/Orchestrator.kt
oheger-bosch: This makes a construct easier that is used in the next commit.

orchestrator/src/main/kotlin/Entrypoint.kt
oheger-bosch: There is no alphabetical order in this construct; the messages are ordered by worker execution.

kubernetes/jobmonitor/src/main/kotlin/LostJobsFinder.kt
oheger-bosch: Actually, it should say "it can try". Changed accordingly.

kubernetes/jobmonitor/src/main/kotlin/LostJobsFinder.kt
oheger-bosch: Isn't this defined in the comment of the `LostJobsFinder` class?

kubernetes/jobmonitor/src/main/kotlin/LostJobsFinder.kt
oheger-bosch: It is not. It sends notifications to Orchestrator via the `notifier`.

orchestrator/src/main/kotlin/Orchestrator.kt
oheger-bosch: When a lost schedule is detected, the function is invoked with `updateRun = false`. It is then not necessary to update the database; the run is already in ACTIVE state.

transport/kubernetes/src/main/kotlin/KubernetesMessageSender.kt
oheger-bosch: Switched to UUIDs.

dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
oheger-bosch: I think this is acceptable. There is currently no precedence where we execute plain SQL statements, AFAIK.

kubernetes/jobmonitor/src/main/kotlin/LostJobsFinder.kt
nnobelis: Ok, I was confused because on this commit, there is only logging inside.

kubernetes/jobmonitor/src/main/kotlin/LostJobsFinder.kt
nnobelis: Ok good enough.

nnobelis (approved):

PR #1632, opened 2024-12-12
Add endpoint to get vulnerabilities for a product

lamppu:

### DIFF
--- a/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
+++ b/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
@@ -62,6 +62,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.PagingOptions as ApiPagingOpt
 import org.eclipse.apoapsis.ortserver.api.v1.model.PluginConfiguration as ApiPluginConfiguration
 import org.eclipse.apoapsis.ortserver.api.v1.model.ProcessedDeclaredLicense as ApiProcessedDeclaredLicense
 import org.eclipse.apoapsis.ortserver.api.v1.model.Product as ApiProduct
+import org.eclipse.apoapsis.ortserver.api.v1.model.ProductVulnerability as ApiProductVulnerability
 import org.eclipse.apoapsis.ortserver.api.v1.model.ProviderPluginConfiguration as ApiProviderPluginConfiguration
 import org.eclipse.apoapsis.ortserver.api.v1.model.RemoteArtifact as ApiRemoteArtifact
 import org.eclipse.apoapsis.ortserver.api.v1.model.ReporterAsset as ApiReporterAsset
@@ -80,6 +81,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.SourceCodeOrigin as ApiSource
 import org.eclipse.apoapsis.ortserver.api.v1.model.User as ApiUser
 import org.eclipse.apoapsis.ortserver.api.v1.model.VcsInfo as ApiVcsInfo
 import org.eclipse.apoapsis.ortserver.api.v1.model.Vulnerability as ApiVulnerability
+import org.eclipse.apoapsis.ortserver.api.v1.model.VulnerabilityRating as ApiVulnerabilityRating
 import org.eclipse.apoapsis.ortserver.api.v1.model.VulnerabilityReference as ApiVulnerabilityReference
 import org.eclipse.apoapsis.ortserver.api.v1.model.VulnerabilityWithIdentifier as ApiVulnerabilityWithIdentifier
 import org.eclipse.apoapsis.ortserver.model.AdvisorJob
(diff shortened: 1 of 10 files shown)


services/hierarchy/src/main/kotlin/ProductService.kt
```diff
@@ -21,6 +21,7 @@ package org.eclipse.apoapsis.ortserver.services
 
 import org.eclipse.apoapsis.ortserver.dao.dbQuery
 import org.eclipse.apoapsis.ortserver.dao.dbQueryCatching
+import org.eclipse.apoapsis.ortserver.dao.repositories.repository.RepositoriesTable
```
mnonnenmacher, line 24: Commit message nit: "a product" -> "products" / "multiple products"
Is there actually a use case for providing multiple product ids? Otherwise the function could be simplified.

services/hierarchy/src/test/kotlin/ProductServiceTest.kt
```diff
@@ -240,4 +241,18 @@ class ProductServiceTest : WordSpec({
...
+        "return IDs for all repositories of a product" {
+            val service = ProductService(db, productRepository, repositoryRepository, ortRunRepository, mockk())
+
+            val prodId = fixtures.createProduct().id
+
+            val repo1Id = fixtures.createRepository(productId = prodId).id
+            val repo2Id = fixtures.createRepository(url = "https://example.com/repo2.git", productId = prodId).id
+            val repo3Id = fixtures.createRepository(url = "https://example.com/repo3.git", productId = prodId).id
+
+            service.getRepositoryIdsForProducts(prodId).shouldContainExactlyInAnyOrder(repo1Id, repo2Id, repo3Id)
```
mnonnenmacher, line 255: Providing multiple product IDs should also be tested.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
```diff
@@ -761,3 +766,13 @@ fun ApiSourceCodeOrigin.mapToModel() =
 fun User.mapToApi() = ApiUser(username, firstName, lastName, email)
 
 fun EcosystemStats.mapToApi() = ApiEcosystemStats(name = name, count = count)
+
+fun VulnerabilityRating.mapToApi() = ApiVulnerabilityRating.valueOf(name)
+
+fun VulnerabilityWithAccumulatedData.mapToApi() = ApiVulnerabilityWithAccumulatedData(
```
mnonnenmacher, line 772: We already have `EcosystemStats`, so maybe call this `VulnerabilityStats`?

api/v1/model/src/commonMain/kotlin/VulnerabilityWithAccumulatedData.kt
```diff
@@ -0,0 +1,49 @@
...
+
+package org.eclipse.apoapsis.ortserver.api.v1.model
+
+import kotlinx.serialization.Serializable
+
+/**
+ * A response object that gathers relevant information about a [Vulnerability].
+ */
+@Serializable
+data class VulnerabilityWithAccumulatedData(
```
mnonnenmacher, line 28: We already have `EcosystemStats`, so maybe call this `VulnerabilityStats`?

mnonnenmacher (changes requested):

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
lamppu: The only thing is, I wonder if `VulnerabilityStats` would be a bit confusing as `EcosystemStats` combines info from multiple packages, and here this provides more info about a single vulnerability (or actually a Vulnerability+Identifier pair). So maybe `VulnerabilityStats` would be better for combining info for multiple vulnerabilities. But I agree that the `VulnerabilityWithAccumulatedData` is kind of a bad name :smiley: Maybe something like `VulnerabilitySummary`?

services/hierarchy/src/main/kotlin/ProductService.kt
lamppu: Yeah I was gonna reuse this on the organization level, but now I realized that I should just get the repository ids in that case with a single query instead of getting the product ids first, and then using this. So I think I'll refactor this to only take one id.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
mnonnenmacher: "accumulated data" is very vague, but "stats" actually is as well. So how about something like `ProductVulnerability` or `ProductVulnerabilitySummary`, then the upcoming class for organizations could be named similarly.

mnonnenmacher (approved):
@lamppu You need to rebase to get the Docker fix from #1655.

mnonnenmacher:
> @lamppu You need to rebase to get the Docker fix from #1655.

Looks like you already did, I need to look into it.

PR #1636, opened 2024-12-13
Introduce an API for maintenance tasks

oheger-bosch:
This PR introduces interfaces and base classes for periodic maintenance tasks. See https://github.com/eclipse-apoapsis/ort-server/discussions/1560.

### DIFF
--- a/tasks/spi/src/main/kotlin/AbstractDITask.kt
+++ b/tasks/spi/src/main/kotlin/AbstractDITask.kt
@@ -1,0 +1,119 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.tasks.spi
+
+import com.typesafe.config.ConfigFactory
+
+import kotlin.time.measureTime
+
+import org.eclipse.apoapsis.ortserver.config.ConfigManager
+
+import org.koin.core.Koin
+import org.koin.core.KoinApplication
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.get
+import org.koin.core.module.Module
+import org.koin.dsl.koinApplication
+import org.koin.dsl.module
+
+import org.slf4j.LoggerFactory
+
+/**
+ * An abstract base class for [Task] implementations that require dependency injection.
+ *
+ * A concrete implementation must provide a list of Koin modules to be taken into account. The class then starts an
+ * isolated Koin context containing these modules. It expects that one of the objects in the context implements the
+ * [Task] interface. It obtains this object and calls its [Task.execute] method.
+ *
+ * Using this class, it is possible quite easily to get access to various components and infrastructure of ORT Server,
+ * including the database repositories, or the transport components. When setting up the context for dependency
+ * injection, a [ConfigManager] instance is included as well; so all configuration settings are available to the task.
+ */
+abstract class AbstractDITask(
+    /** A name for this task. If undefined, the simple name of the task class is used. */
+    val name: String
+) : Task {
+    companion object {
+
+        private val logger = LoggerFactory.getLogger(AbstractDITask::class.java)
+
+        /**
+         * Create a new instance of a [AbstractDITask] that depends on the given [modules] and has the given [name].
+         */
+        fun create(modules: Collection<Module>, name: String): AbstractDITask =
+            object : AbstractDITask(name) {
+                override fun requiredModules(): Collection<Module> = modules
+            }
+
+        /**
+         * Create a [Module] that contains a [ConfigManager] instance initialized with the current configuration.
+         * That way, executed tasks have access to the configuration settings. The [ConfigManager] is also needed by
+         * some modules of ORT Server.
+         */
+        private fun configModule(): Module = module {
+            single { ConfigManager.create(ConfigFactory.load()) }
+        }
+    }
+
+    override suspend fun execute() {
+        logger.debug("Setting up Koin context for task '{}'.", name)
+        val app = setUpKoinApplication()
+
+        logger.info("Executing task '{}'.", name)
+        try {
+            val executionTime = measureTime { TaskRunnerComponent(app.koin).execute() }
+            logger.info("Task '{}' finished successfully in {}.", name, executionTime)
+        } finally {
+            logger.debug("Closing Koin context after execution of task '{}'.", name)
+            app.close()
+        }
+    }
+
+    /**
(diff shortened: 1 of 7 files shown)


tasks/spi/src/main/kotlin/AbstractDITask.kt
```diff
@@ -0,0 +1,113 @@
...
+) : Task {
+    companion object {
+        private val logger = LoggerFactory.getLogger(AbstractDITask::class.java)
+
+        /** A default name to be used for tasks if no explicit name was provided. */
+        private const val DEFAULT_TASK_NAME = "Unnamed task"
+
+        /**
+         * Create a new instance of a [AbstractDITask] that depends on the given [modules]. Optionally, the task can be
+         * given a [name].
```
mnonnenmacher, line 58: Why is the task name optional?

tasks/spi/src/main/kotlin/AbstractDITask.kt
```diff
@@ -0,0 +1,113 @@
...
+        logger.info("Executing task '{}'.", name)
+
+        val app = koinApplication {
+            modules(listOf(configModule()) + requiredModules())
+        }
+
+        runCatching {
+            TaskRunnerComponent(app.koin).execute()
+        }.onFailure { exception ->
+            logger.error("Exception when executing task '{}':", name, exception)
```
mnonnenmacher, line 85: As commented in the previous commit, swallowing all errors here makes it impossible for the runtime that is managing the tasks to respond to errors.

mnonnenmacher (changes requested):
IMO this PR approaches the implementation of maintenance tasks from the wrong direction by introducing an abstraction layer before there is any task implementation and thereby making some assumptions too early:
* It adds an `spi` Gradle module indicating that tasks should be implemented in separate Gradle modules. I would start with a single Gradle module and think about splitting it when we see that it provides value.
* It adds a quite complex dependency injection management in the `AbstractDiTask` before we even know what the actual task implementations will look like, and if they will require it. Given that we currently only plan to implement two maintenance tasks, I would start by simply hardcoding them and afterward think about abstractions.

oheger-bosch:
> IMO this PR approaches the implementation of maintenance tasks from the wrong direction by introducing an abstraction layer before there is any task implementation and thereby making some assumptions too early:
> 
Well, I typically think before starting to hack.

At least, this started a discussion about return values for tasks.

oheger-bosch:
>     * It adds an `spi` Gradle module indicating that tasks should be implemented in separate Gradle modules. I would start with a single Gradle module and think about splitting it when we see that it provides value.
> 
There are the maintenance tasks about rentention policies. I also have the hope that some functionality used within Kubernetes Job Monitor can be implemented as periodic tasks. In terms of loose coupling and high coherence, it would be good to have those in different modules. Also, a gradle module is not that big overhead. However, I admit that `spi` is probably not a suitable name, since the module is different from other `spi` modules we have so far. So, maybe just name it `task-api` or something like that?

>     * It adds a quite complex dependency injection management in the `AbstractDiTask` before we even know what the actual task implementations will look like, and if they will require it. Given that we currently only plan to implement two maintenance tasks, I would start by simply hardcoding them and afterward think about abstractions.
The tasks will access the database via repositories. To simplify their implementation and make them consistent with other parts of the codebase, they should use the same programming model as is used everywhere - and this is based on dependency injection via coin.

tasks/spi/src/main/kotlin/Task.kt
```diff
@@ -0,0 +1,30 @@
...
+package org.eclipse.apoapsis.ortserver.tasks.spi
+
+/**
+ * An interface describing a task that can be executed by an external scheduler implementation.
+ */
+interface Task {
+    /**
+     * Execute the specific logic of this task.
+     */
+    suspend fun execute()
```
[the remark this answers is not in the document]
oheger-bosch: I have added documentation that tasks can throw arbitrary exceptions. Using a `Result` with a generic type as return value seems a bit strange to me, as it would imply that such results could be processed somehow in a generic way.

tasks/spi/src/main/kotlin/AbstractDITask.kt
oheger-bosch: The original idea was to use the simple class name as fallback, but this did not work out, since the name is `null` for anonymous classes. I have made the task name now mandatory.

tasks/spi/src/main/kotlin/AbstractDITask.kt
oheger-bosch: Exceptions are now propagated to the caller.

mnonnenmacher:
> Well, I typically think before starting to hack.

I do not question that you have thought this through, I just want to avoid that we add unnecessary complexity. And I find it close to impossible to judge if the complexity added by an abstraction is worth it when not seeing what the code would look like without it. Therefore I prefer to start with a simple implementation and do the abstraction later when the benefit becomes obvious.

> There are the maintenance tasks about rentention policies. I also have the hope that some functionality used within Kubernetes Job Monitor can be implemented as periodic tasks. In terms of loose coupling and high coherence, it would be good to have those in different modules. Also, a gradle module is not that big overhead.
> The tasks will access the database via repositories. To simplify their implementation and make them consistent with other parts of the codebase, they should use the same programming model as is used everywhere - and this is based on dependency injection via coin.

The Keycloak sync task would basically look like this:

```kotlin
class KeycloakSyncTask(private val authenticationService: AuthenticationService) : Task {
  fun execute() {
    authenticationService.ensureSuperuserAndSynchronizeRolesAndPermissions()
  }
}
```

For the retention task I would assume that it would look very similar and the logic would be implemented in a service, because that's where the business logic belongs.

So what remains is setting up the required services and their dependencies. This can be done using Koin, just like in the `core` module or in the workers.

If we add more tasks in the future there might be a point where separation of concerns becomes more relevant and if so, adding an abstraction at this point will make this obvious. But right now my feeling is that the proposed implementation adds complexity to solve a problem we do not have yet and might never have.

oheger-bosch:
> So what remains is setting up the required services and their dependencies. This can be done using Koin, just like in the `core` module or in the workers.
> 
> If we add more tasks in the future there might be a point where separation of concerns becomes more relevant and if so, adding an abstraction at this point will make this obvious. But right now my feeling is that the proposed implementation adds complexity to solve a problem we do not have yet and might never have.

I was hoping to skip some iterations and take a short track. With a naive Koin setup, you end up with a module containing _n_ task definitions and _m_ beans for the dependencies of these tasks. It is then rather easy to loose the overview which dependency is required for which task. Every change on this module requires testing all tasks.

With the mechanism of `AbstractDiTask` - if I am allowed to say this -, I find it quite elegant that a concrete task implementation can declare the ORT Server modules it depends on and has an isolated dependency space.

But maybe it is too early. So, how to proceed? Should I drop the commit that introduces this task base class? Are you fine with the other task interfaces?

oheger-bosch:
Closing this in favour of #1782. The discussion points should have been addressed in this new PR.

mnonnenmacher:
> > So what remains is setting up the required services and their dependencies. This can be done using Koin, just like in the `core` module or in the workers.
> > If we add more tasks in the future there might be a point where separation of concerns becomes more relevant and if so, adding an abstraction at this point will make this obvious. But right now my feeling is that the proposed implementation adds complexity to solve a problem we do not have yet and might never have.
> 
> I was hoping to skip some iterations and take a short track. With a naive Koin setup, you end up with a module containing _n_ task definitions and _m_ beans for the dependencies of these tasks. It is then rather easy to loose the overview which dependency is required for which task. Every change on this module requires testing all tasks.
> 
> With the mechanism of `AbstractDiTask` - if I am allowed to say this -, I find it quite elegant that a concrete task implementation can declare the ORT Server modules it depends on and has an isolated dependency space.

I understand that, but like I said, my feeling is that we are not going to need it anytime soon, and I found it a bit difficult to understand how it works. However, if we have a few actual tasks implemented, maybe reintroducing it at this point shows the benefit better.

> But maybe it is too early. So, how to proceed? Should I drop the commit that introduces this task base class? Are you fine with the other task interfaces?

A little late, but yes, I'm fine with the task interfaces.

PR #1643, opened 2024-12-13
ci(docker): Improve the Docker job names

mnonnenmacher:
Use individual names for the Docker build jobs as the auto-generated unique names are based on the matrix parameters and are too long which makes it difficult to configure the required checks.

PR #1644, opened 2024-12-13
Fix Keycloak health check in Compose

mnonnenmacher:
This will have to be rebased after #1643 was merged.

@lamppu @Etsija FYI


sschuberth (approved):

sschuberth:
So, with these fixes, there is no more need to revert https://github.com/eclipse-apoapsis/ort-server/pull/1153, correct?

mnonnenmacher:
> So, with these fixes, there is no more need to revert #1153, correct?

Yes.

PR #1654, opened 2024-12-16
Optimize access to custom license texts in Reporter

oheger-bosch:
This PR adds a special `LicenseTextProvider` implementation that optimizes loading of license texts from the config repository.

### DIFF
--- a/workers/reporter/src/main/kotlin/reporter/CustomLicenseTextProvider.kt
+++ b/workers/reporter/src/main/kotlin/reporter/CustomLicenseTextProvider.kt
@@ -1,0 +1,90 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


sschuberth:
Will this make https://github.com/oss-review-toolkit/ort/pull/9523 from @nnobelis obsolete?

oheger-bosch:
> Will this make [oss-review-toolkit/ort#9523](https://github.com/oss-review-toolkit/ort/pull/9523) from @nnobelis obsolete?

Yes, this is an alternative approach to solve the performance issue we have.

sschuberth:
> This PR adds a special `LicenseTextProvider` implementation that optimizes loading of license texts from the config repository.

FYI, I also just created https://github.com/oss-review-toolkit/ort/issues/9620 which generalizes the idea to a `LicenseDataProvider` API on the ORT core side.

workers/reporter/src/main/kotlin/reporter/CustomLicenseTextProvider.kt
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * The [LicenseTextProvider] implementations available in ORT expect that license texts are available in a local
+ * folder. For ORT Server, this is not necessarily the case, but depends on the provider for config files. This
+ * implementation therefore uses the configuration manager API to check whether a requested license text is available
+ * in the configuration. If it is, it obtains the license text from the configuration. Otherwise, it delegates to
+ * another [LicenseTextProvider] implementation which searches the default paths used by ORT.
+ */
+internal class CustomLicenseTextProvider(
+    /** The current worker context. */
+    val workerContext: WorkerContext,
```
mnonnenmacher, line 45: `workerContext` has lots of properties but this class needs only the `configManager` and the `resolvedConfigurationContext`. Passing those properties here instead of the `workerContext` would reduce dependencies of the class and simplify the test.

mnonnenmacher (changes requested):

PR #1656, opened 2024-12-16
Add functions to count items by severity/rating

lamppu:


services/hierarchy/src/main/kotlin/IssueService.kt
```diff
@@ -89,6 +90,24 @@ class IssueService(private val db: Database) {
             .count()
     }
 
+    /** Count issues by severity for provided ORT runs. */
+    suspend fun countBySeverityForOrtRunIds(vararg ortRunIds: Long): Map<Severity, Long> = db.dbQuery {
```
oheger-bosch, line 94: IIUC, the map returned by the function only contains keys for severities for which issues have been found. Should it be initialized to set all possible keys to 0, so that it is guaranteed that every enum constant is assigned a value? At least, the behavior should be documented for the function.

Maybe a map is not the most convenient structure for this use case, since clients typically have to deal with possible `null` values. So, would it make sense to introduce a thin wrapper class around a `Map<T, Long>`, say `CountsByCategory` or something that has a

```
fun getCount(key: T): Long = map[key} ?: 0
```
The same applies to the other functions introduced in this PR as well.

model/src/commonMain/kotlin/CountByCategory.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package org.eclipse.apoapsis.ortserver.model
+
+/** Wrapper class for a map of counts by category. */
+data class CountByCategory<T>(
+    /** The map of counts by category. */
+    val map: Map<T, Long>
+) {
+    /** Get the count for the given [key]. */
+    fun getCount(key: T): Long = map[key] ?: 0
```
sschuberth, line 28: This is completely fine, just to share an alternative Kotlin syntax which might be more speaking: `getOrDefault(key, 0)`

services/hierarchy/src/main/kotlin/IssueService.kt
```diff
@@ -89,6 +91,26 @@ class IssueService(private val db: Database) {
             .count()
     }
 
+    /** Count issues by severity for provided ORT runs. */
+    suspend fun countBySeverityForOrtRunIds(vararg ortRunIds: Long): CountByCategory<Severity> = db.dbQuery {
+        val countAlias = Count(OrtRunsIssuesTable.id, true)
+
+        val severityToCountMap = mutableMapOf<Severity, Long>().apply {
+            Severity.entries.forEach { this[it] = 0 }
+        }
```
sschuberth, line 100: A bit shorter:

    val severityToCountMap = Severity.entries.associateWithTo(mutableMapOf()) { 0L }

sschuberth: Similar several times below.

services/hierarchy/src/main/kotlin/VulnerabilityService.kt
```diff
@@ -139,10 +141,44 @@ class VulnerabilityService(private val db: Database) {
...
+            .groupBy(VulnerabilitiesTable.id, IdentifiersTable.id)
+            .alias("subQuery")
+
+        val rating = subQuery[ratingAlias]
+
+        subQuery
+            .select(rating, countAlias)
+            .groupBy(rating)
+            .forEach { row ->
+                ratingToCountMap[VulnerabilityRating.entries[(row[rating] ?: 0)]] = row[countAlias]
```
sschuberth, line 173: Just double-checking: The `[]`-operator is an alias for `get()` which seems to throw instead of returning `null` for absent keys. Can someone confirm / dissent?

sschuberth: Note that I did not check whether tests cover that case.

lamppu: Yes, using `get()`, the editor gives me a hint _"Should be replaced with indexing"_

lamppu: I can't really produce a test that would produce an out-of-bounce rating, as the ratingAlias should only give an overall rating within those bounds

sschuberth: Right, but do `[]` / `get()` throw or return `null` for missing keys?

sschuberth: Please see

```
class ResultRow(
    /** Mapping of the expressions stored on this row to their index positions. */
    val fieldIndex: Map<Expression<*>, Int>,
    private val data: Array<Any?> = arrayOfNulls<Any?>(fieldIndex.size)
) {
    private val database: Database? = TransactionManager.currentOrNull()?.db

    private val lookUpCache = ResultRowCache()

    /**
     * Retrieves the value of a given expression on this row.
     *
     * @param expression expression to evaluate
     * @throws IllegalStateException if expression is not in record set or if result value is uninitialized
     *
     * @see [getOrNull] to get null in the cases an exception would be thrown
     */
    @Suppress("UNCHECKED_CAST")
    operator fun <T> get(expression: Expression<T>): T {
```

So I guess you should use `getOrNull()` instead.

sschuberth (approved):

Issue #1662, opened 2024-12-17
Downloading logs fails in ORT Server


Etsija:
Yesterday, I downloaded logs from ORT runs just fine, but now I get an unknown error (500 response) from back-end for all download attempts. Downloading code has not been touched for a while now in UI. Anyone know of any recent PR which might have caused this?

(This fails both in local Compose environment and in our infra.)

![Screenshot from 2024-12-17 13-50-52](https://github.com/user-attachments/assets/fa3a7155-971e-40a1-b781-c2f636ea4737)

The request is failing also when tested with PostMan:

![Screenshot from 2024-12-17 14-10-09](https://github.com/user-attachments/assets/7ae3b9e4-10e1-428f-b76a-0d6b05db7335)

Looks like the Reporter log is failing.

mnonnenmacher:
@Etsija Did you check the if the server logs contain more details about why the download is failing?

Etsija:
This is what I was able to extract from core logs:

[Explore-logs-2024-12-18 07_52_58.txt](https://github.com/user-attachments/files/18175741/Explore-logs-2024-12-18.07_52_58.txt)

Etsija:
I think this might be the interesting part:
```
2024-12-18 07:52:37.744		Caused by: java.util.concurrent.CancellationException: Parent job is Cancelling
2024-12-18 07:52:37.744			at kotlinx.coroutines.InvokeOnCancelling.invoke(JobSupport.kt:1571)
2024-12-18 07:52:37.744			at io.ktor.client.engine.UtilsKt$attachToUserJob$cleanupHandler$1.invoke(Utils.kt:97)
2024-12-18 07:52:37.744			at kotlinx.coroutines.JobSupport.notifyCancelling(JobSupport.kt:1604)
2024-12-18 07:52:37.744			at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
2024-12-18 07:52:37.744			at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:99)
2024-12-18 07:52:37.744			at org.eclipse.apoapsis.ortserver.logaccess.LogFileService.access$downloadLogFile(LogFileService.kt:51)
2024-12-18 07:52:37.744			at org.eclipse.apoapsis.ortserver.logaccess.LogFileService$downloadLogFile$1.invokeSuspend(LogFileService.kt)
[...]

PR #1674, opened 2024-12-18
ci(docker): Always load the base images to the local registry

mnonnenmacher:
Always load the base images to the local registry as this is where Jib is looking for them since 808a3fc.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -142,7 +142,7 @@ jobs:
         context: ${{ matrix.docker.context }}
         file: ${{ matrix.docker.context }}/${{ matrix.docker.dockerfile }}
         push: ${{ !env.IS_PR }}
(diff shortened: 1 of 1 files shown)


sschuberth (approved):

mnonnenmacher:
> So, does this fix the Docker build as mentioned by @mmurto over [here](https://github.com/eclipse-apoapsis/ort-server/issues/1662#issuecomment-2550950107)?

Yes, at least that's the plan.

sschuberth:
> > So, does this fix the Docker build as mentioned by @mmurto over [here](https://github.com/eclipse-apoapsis/ort-server/issues/1662#issuecomment-2550950107)?
> 
> Yes, at least that's the plan.

Looks [good](https://github.com/eclipse-apoapsis/ort-server/actions/runs/12393646781)!

PR #1678, opened 2024-12-19
fix(compose): Remove explicit admin credentials

sschuberth:
Keycloak 26 introduced a feature to bootstrap admin accounts [1]. This feature seem to interfere with specifying admin credentials explicitly via the Docker Compose environment, leading to

Detail: Key (realm_id, username)=(0b009e14-6671-4ae5-8e3f-ddcc98049624, admin) already exists.

when starting the Keycloak image. Removing these variables prevents that while still being able to log into Keycloak with "admin" / "admin".

Note that at the same time these variables were anyway deprecated in favor of `KC_BOOTSTRAP_ADMIN_USERNAME` / `KC_BOOTSTRAP_ADMIN_PASSWORD`, causing a separate deprecation warning in the logs. However, using those variables does not prevent the above error. But removing the original variables also gets rid of that deprecation warning.

[1]: https://www.keycloak.org/server/bootstrap-admin-recovery

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -86,8 +86,6 @@ services:
       retries: 15
       start_period: 90s
     environment:
-      KEYCLOAK_ADMIN: admin
-      KEYCLOAK_ADMIN_PASSWORD: admin
       KC_DB: postgres
       KC_DB_USERNAME: postgres
       KC_DB_PASSWORD: postgres


Etsija:
Looks good to me.

sschuberth:
> this is related in a way to issue #1662, since it was completely preventing from debugging that?

Well, I don't think it's related to #1662 any more than to any other issue that you'd want to debug / bisect. In that sense, I believe the comment you just made is enough of a reference 😸

mnonnenmacher:
> Detail: Key (realm_id, username)=(0b009e14-6671-4ae5-8e3f-ddcc98049624, admin) already exists.

I wonder why I don't see this message when I do `docker compose logs keycloak`.

> Note that at the same time these variables were anyway deprecated in favor of KC_BOOTSTRAP_ADMIN_USERNAME / KC_BOOTSTRAP_ADMIN_PASSWORD, causing a separate deprecation warning in the logs. However, using those variables does not prevent the above error. But removing the original variables also gets rid of that deprecation warning.

The [official docs](https://www.keycloak.org/getting-started/getting-started-docker#_start_keycloak) use those variables, so I believe in general using them should not cause errors.

sschuberth:
> I wonder why I don't see this message when I do `docker compose logs keycloak`.

I had to do a full `docker compose down -v` / `docker compose up keycloak` in order to see these.

> The [official docs](https://www.keycloak.org/getting-started/getting-started-docker#_start_keycloak) use those variables, so I believe in general using them should not cause errors.

I Agree that's how it should be. But unfortunately it does not seem to be the case.

mmurto:
> > Detail: Key (realm_id, username)=(0b009e14-6671-4ae5-8e3f-ddcc98049624, admin) already exists.
> 
> I wonder why I don't see this message when I do `docker compose logs keycloak`.
> 
> > Note that at the same time these variables were anyway deprecated in favor of KC_BOOTSTRAP_ADMIN_USERNAME / KC_BOOTSTRAP_ADMIN_PASSWORD, causing a separate deprecation warning in the logs. However, using those variables does not prevent the above error. But removing the original variables also gets rid of that deprecation warning.
> 
> The [official docs](https://www.keycloak.org/getting-started/getting-started-docker#_start_keycloak) use those variables, so I believe in general using them should not cause errors.

I think it's more correct t osay that instead of those variables causing errors, the error is caused by our realm import [containing a user with the same username](https://github.com/eclipse-apoapsis/ort-server/blob/1c1b23256028e450fecafb9e26926ae6268e196e/scripts/docker/keycloak/master-realm.json#L413) than the one created with the environment variables. So the error is fixed by removing the environment variables so there are no duplicate usernames, but the thing that actually causes the crash with them is our realm JSON.

sschuberth:
> the thing that actually causes the crash with them is our realm JSON.

But is there a way to modify our realm JSON to not cause the crash? At a quick glance I couldn't find at. Because I agree that this (together with keeping but renaming the variables) would be the preferable solution.

mmurto:
> > the thing that actually causes the crash with them is our realm JSON.
> 
> But is there a way to modify our realm JSON to not cause the crash? At a quick glance I couldn't find at. Because I agree that this (together with keeping but renaming the variables) would be the preferable solution.

The `admin` user could be removed, though then `admin` would not be added to the `SUPERUSERS` group and updating would be harder because then you'd have to remove `admin` from the JSON every time the file is updated.

I think the solution in this PR is the best for now, before Keycloak management is migrated to Terraform.

mmurto (approved):

PR #1680, opened 2024-12-19
fix(reporter): Add ScanCode license texts to reporter docker image

wkl3nk:
Dump the ScanCode license texts to directory _/opt/scancode-license-data_ when creating the reporter docker container.
This directory is used by ORT as fallback option if the ScanCode license texts cannot be located by the existing heuristic look-up algorithm.

See oss-review-toolkit/ort#9622.

### DIFF
--- a/workers/reporter/docker/Reporter.Dockerfile
+++ b/workers/reporter/docker/Reporter.Dockerfile
@@ -19,6 +19,24 @@
 # SPDX-License-Identifier: Apache-2.0
 # License-Filename: LICENSE
 
+# Build-Stage for Python executing scancode-license-data to get the license texts in a directory
+FROM python:3.11-slim AS scancode-license-data-build
+
+# Keep in sync with Scanner.Dockerfile
+ARG SCANCODE_VERSION=32.2.1
+
+RUN apt-get update && apt-get install -y curl libgomp1 && rm -rf /var/lib/apt/lists/*
+
+# Use pip to install ScanCode
+RUN curl -Os https://raw.githubusercontent.com/nexB/scancode-toolkit/v$SCANCODE_VERSION/requirements.txt && \
+    pip install -U --constraint requirements.txt scancode-toolkit==$SCANCODE_VERSION && \
+    rm requirements.txt
+
+# Extract ScanCode license data to directory.
+RUN scancode-license-data --path /opt/scancode-license-data \
+    && find /opt/scancode-license-data -type f -not -name "*.LICENSE" -exec rm -f {} + \
+    && rm -rf /opt/scancode-license-data/static
+
 # When updating this version make sure to keep it in sync with the other worker Dockerfiles and libs.version.toml.
 FROM eclipse-temurin:21.0.4_7-jdk-jammy@sha256:0472478e22da0f66043fa6acd8cd30126592349f47937adafc2340794e5bf06a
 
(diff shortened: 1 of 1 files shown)


sschuberth:
> See [oss-review-toolkit/ort#9622](https://github.com/oss-review-toolkit/ort/pull/9622).

Actually, I just added a [comment](https://github.com/oss-review-toolkit/ort/issues/8147#issuecomment-2556965097) as I don't understand why the fix was even necessary, in retrospect. Can you help me understand?

workers/reporter/docker/Reporter.Dockerfile
```diff
@@ -51,4 +51,24 @@ RUN chgrp -R 0 /home/ort && chmod -R g+rwX /home/ort
...
+
+# Use pip to install ScanCode
+RUN curl -Os https://raw.githubusercontent.com/nexB/scancode-toolkit/v$SCANCODE_VERSION/requirements.txt && \
+    pip install -U --constraint requirements.txt scancode-toolkit==$SCANCODE_VERSION && \
+    rm requirements.txt
+
+# Extract ScanCode license data to directory.
+RUN scancode-license-data --path /opt/scancode-license-data \
+    && find /opt/scancode-license-data -type f -not -name "*.LICENSE" -exec rm -f {} + \
+    && rm -rf /opt/scancode-license-data/static
```
mnonnenmacher, line 74: Did you test this? I thought the last `FROM` statement defines what the final image will be.

workers/reporter/docker/Reporter.Dockerfile
```diff
@@ -51,4 +51,24 @@ RUN chgrp -R 0 /home/ort && chmod -R g+rwX /home/ort
...
+# Build-Stage for Python executing scancode-license-data to get the license texts in a directory
+FROM python:3.11-slim AS scancode-license-data-build
+
+# Keep in sync with Scanner.Dockerfile
+ARG SCANCODE_VERSION=32.2.1
+
+RUN apt-get update && apt-get install -y curl libgomp1 && rm -rf /var/lib/apt/lists/*
+
+# Use pip to install ScanCode
+RUN curl -Os https://raw.githubusercontent.com/nexB/scancode-toolkit/v$SCANCODE_VERSION/requirements.txt && \
```
sschuberth, line 67: I wonder whether in this case, where ScanCode itself it not really needed, it wouldn't be more efficient to use e.g. `curl` (and some `jq` magic) on the [LicenseDB API](https://github.com/aboutcode-org/scancode-licensedb?tab=readme-ov-file#api) and download the `*.LICENSE` files directly.

workers/reporter/docker/Reporter.Dockerfile
wkl3nk: Tested ... yes ... in the final image, the directory _/opt/scancode-license-data_ exists and contains the expected files. But: You are right, the last `FROM` statement defines the base of the final image. I will update the PR accordingly and move the new section to the top of the Dockerfile.

workers/reporter/docker/Reporter.Dockerfile
wkl3nk: But then the procedere to fetch the `*.LICENSE` files would be different from how it is done in the ORT docker image. I would prefer to do it in the same way for both docker images, the ORT docker image and the ORT Server Reporter docker image.

mnonnenmacher (approved):

workers/reporter/docker/Reporter.Dockerfile
mnonnenmacher: I agree that we should first improve the way we get license texts in ORT itself. There might be ways to implement this more efficiently, but as this only affects build performance in the multi-stage build and not the final image I don't see it as a pressing issue.

PR #1682, opened 2024-12-19
Add run statistics endpoint for products

lamppu:

### DIFF
--- a/core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
+++ b/core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
@@ -31,6 +31,7 @@ import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.file.aFile
 import io.kotest.matchers.longs.shouldBeGreaterThan
 import io.kotest.matchers.longs.shouldBeLessThan
+import io.kotest.matchers.maps.shouldContainExactly
 import io.kotest.matchers.maps.shouldHaveSize
 import io.kotest.matchers.nulls.beNull
 import io.kotest.matchers.nulls.shouldNotBeNull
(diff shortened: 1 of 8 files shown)


api/v1/model/src/commonMain/kotlin/OrtRunSummary.kt
```diff
@@ -31,7 +31,7 @@ data class OrtRunSummary(
      * The unique identifier.
      */
     val id: Long,
-
+
```
sschuberth, line 34: Nit: This should be undone.

core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
```diff
@@ -1816,11 +1727,16 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
 
                 val statistics = response.body<OrtRunStatistics>()
 
-                statistics.issuesCount shouldBe 1
-                statistics.packagesCount shouldBe 2
-                statistics.ecosystems?.size shouldBe 2
-                statistics.vulnerabilitiesCount shouldBe null
-                statistics.ruleViolationsCount shouldBe null
+                with(statistics) {
+                    issuesCount.shouldBeNull()
```
nnobelis, line 1731: _nit_ isn't it `shouldBe null` ?
@sschuberth  right ?

+ other places

sschuberth: Actually, it's `should beNull()` 😸

lamppu: Just out of curiosity is that a syntax choice, or is there some difference?

sschuberth: Just syntax. In general, we prefer infix functions, i.e. those without the need for a `.` before the function call, to make the code read more fluent.

Issue #876, opened 2024-08-22
Run IDs should probably not be labels in Loki


mmurto:
The current logaccess from Loki requires the `runId` to be a label:

https://github.com/eclipse-apoapsis/ort-server/blob/3f1f5bc44eda6125ef23c8116b5ca22a7d72f59f/logaccess/loki/src/main/kotlin/LokiLogFileProvider.kt#L127

This does not seem to be advised with Loki, [according to Loki's docs](https://grafana.com/docs/loki/latest/get-started/labels/bp-labels/#label-values-must-always-be-bounded): 

> If you are dynamically setting labels, never use a label which can have unbounded or infinite values. This will always result in big problems for Loki.

I believe (according to the same doc) that the correct solution would be to not label them and use a regex filtering to get logs for the specific run.

mnonnenmacher:
@mmurto As part of fixing #168 this will have to be implemented in a more generic way. AFAIK the `run_id` labels are currently only added in our internal configuration based on the labels of the Kubernetes pods. With #875 it will be possible to take them directly from the log statements.

sschuberth:
This was resolved by https://github.com/eclipse-apoapsis/ort-server/pull/1690.

Issue #168, opened 2024-04-29
Retrieving logs for an ORT Run fails


Etsija:
When trying to retrieve logs for a successful ORT Run, we get

```
Uncaught ApiError: Internal Server Error

{"message":"Error when processing the request.","cause":"Could not create instance for '[Singleton:'org.eclipse.apoapsis.ortserver.logaccess.LogFileService']'"}

```

mnonnenmacher:
@Etsija This is for the Docker Compose setup, right?

sschuberth:
> @Etsija This is for the Docker Compose setup, right?

Yes, we have no other setup yet.

lamppu:
Same kind of error coming for (at least) GET and POST requests for organization/product/repository secrets
`{
    "message": "Error when processing the request.",
    "cause": "Could not create instance for '[Singleton:'org.eclipse.apoapsis.ortserver.services.SecretService']'"
}`

mnonnenmacher:
> Same kind of error coming for (at least) GET and POST requests for organization/product/repository secrets `{ "message": "Error when processing the request.", "cause": "Could not create instance for '[Singleton:'org.eclipse.apoapsis.ortserver.services.SecretService']'" }`

@lamppu Can you create a separate issue for this problem? It's probably easier to solve than the log issue.

sschuberth:
@mnonnenmacher this issue is becoming rather pressing for us. Do you think you'll have a chance to address it any time soon?

mnonnenmacher:
I think this cannot be solved without solving #569 first. Currently it is not possible to associate the logs of the workers in Compose with a specific ORT run.

sschuberth:
This was resolved by https://github.com/eclipse-apoapsis/ort-server/pull/1690.

PR #1690, opened 2024-12-20
Loki logging improvements

mnonnenmacher:
This PR applies several improvements to the Loki logging (see the commit messages for details):
* Change the log format to simplify parsing.
* Adapt the Loki query for better performance.
* Fix the Loki configuration in Compose and add a log dashboard.

### DIFF
--- a/core/src/main/resources/logback.xml
+++ b/core/src/main/resources/logback.xml
@@ -21,7 +21,7 @@
     <conversionRule conversionWord="customMdc" converterClass="org.eclipse.apoapsis.ortserver.utils.logging.MdcConverter" />
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>time=%d{YYYY-MM-dd HH:mm:ss.SSS} thread=%thread %customMdc level=%-5level logger=%logger{36} message="%replace(%msg){'\"','\\\"'}"%n</pattern>
+            <pattern>time=%d{YYYY-MM-dd'T'HH:mm:ss.SSS} thread=%thread %customMdc level=%-5level logger=%logger{36} message="%replace(%msg){'\"','\\\"'}"%n</pattern>
         </encoder>
     </appender>
     <root level="trace">
--- a/core/src/main/resources/logback.xml
+++ b/core/src/main/resources/logback.xml
@@ -20,7 +20,7 @@
 <configuration>
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>%d{YYYY-MM-dd HH:mm:ss.SSS} [%thread] [%mdc] level=%-5level %logger{36} - %msg%n</pattern>
+            <pattern>time=%d{YYYY-MM-dd HH:mm:ss.SSS} thread=%thread %mdc level=%-5level logger=%logger{36} message="%replace(%msg){'\"','\\\"'}"%n</pattern>
         </encoder>
     </appender>
     <root level="trace">
--- a/core/src/main/resources/logback.xml
+++ b/core/src/main/resources/logback.xml
@@ -18,9 +18,10 @@
   -->
 
 <configuration>
+    <conversionRule conversionWord="customMdc" converterClass="org.eclipse.apoapsis.ortserver.utils.logging.MdcConverter" />
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>time=%d{YYYY-MM-dd HH:mm:ss.SSS} thread=%thread %mdc level=%-5level logger=%logger{36} message="%replace(%msg){'\"','\\\"'}"%n</pattern>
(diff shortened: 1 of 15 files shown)


docker-compose.yml
```diff
@@ -174,6 +174,9 @@ services:
       GRAPHITE_PROTOCOL: PLAINTEXT
       CORE_SECRET_PROVIDER: secret-file
       CORE_SECRET_FILES: /mnt/secrets.properties
+      LOG_FILE_PROVIDER_NAME: loki
+      LOKI_SERVER_URL: "http://loki:3100"
+      LOKI_NAMESPACE: "compose"
```
sschuberth, line 179: @Etsija you might be interested in this change.

mnonnenmacher:
I still need to fix the `LokiLogFileProviderTest` but please review already.

core/src/main/resources/logback.xml
nnobelis, line 1: I am not too sure about this change:

- Aren't the log lines not going to be too long ?
- More characters = more data to store in the DB. Isn't it inefficient ?

Using regular expression is not wrong/inefficient _per se_: for instance the whole UNIX/Linux/Posix rely on them.

mnonnenmacher: > Aren't the log lines not going to be too long ?

It doesn't make that much of a difference, but having clearly structured logs makes parsing and querying a LOT easier. It's a middle ground between switching to JSON logging which would be even better for parsing but make the logs are to read for humans.

> More characters = more data to store in the DB. Isn't it inefficient ?

No, the inefficiency we are currently seeing is caused by the fact that we are using the unbounded run_id and trace_id as labels, see https://grafana.com/docs/loki/latest/get-started/labels/bp-labels/#use-dynamic-labels-sparingly. Also, we use logfmt to parse the log level which is a lot less efficient than having a label for the log level (which is a bounded value) like also introduced in this PR.
So overall, yes, it uses a little more storage space, but it allows to make the queries we run more efficient in all other ways.

> Using regular expression is not wrong/inefficient per se: for instance the whole UNIX/Linux/Posix rely on them.

The structured logs help with ingestion (with logstash in the Compose environment or promtail in our Kubernetes environment), because it makes it easier to extract key value pairs to use as labels, compared to using regular expressions for the same purpose.

mnonnenmacher: Actually, after re-reading the link above I will rework this PR to not use a label for the log level.

mnonnenmacher: > Actually, after re-reading the link above I will rework this PR to not use a label for the log level.

Done.

docker-compose.yml
sschuberth: So, should this commit close https://github.com/eclipse-apoapsis/ort-server/issues/168?

logaccess/loki/src/main/kotlin/LokiLogFileProvider.kt
```diff
@@ -159,10 +159,9 @@ class LokiLogFileProvider(
         source: LogSource,
         levels: Set<LogLevel>
     ): String {
-        val levelCriterion = levels.joinToString(" or ") { """level="${it.name}"""" }
-        return """{namespace="${config.namespace}",container=~"${source.name.lowercase()}.*",""" +
-                """run_id="$ortRunId"} | logfmt level | $levelCriterion """ +
-                "| drop __error__, __error_details__, level"
+        val levelCriterion = levels.joinToString("|") { it.name }
```
sschuberth, line 162: Should this commit close https://github.com/eclipse-apoapsis/ort-server/issues/876?

logaccess/loki/build.gradle.kts
```diff
@@ -41,6 +41,7 @@ dependencies {
     implementation(libs.ktorKotlinxSerialization)
 
     testImplementation(testFixtures(projects.logaccess.logaccessSpi))
+    testImplementation(projects.utils.logging)
```
sschuberth, line 44: Why is this required here now if no code changes are done in this module?

scripts/compose/grafana/provisioning/dashboards/dashboard.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: 1
```
sschuberth, line 1: Commit message: Something seems to be wrong in "to view the Loki logs to the Grafana configuration". Maybe "to add..."?

scripts/compose/grafana/provisioning/dashboards/log-viewer.json
```diff
@@ -0,0 +1,200 @@
+{
```
sschuberth, line 1: Could this be YAML as well for consistency?

logaccess/loki/build.gradle.kts
mnonnenmacher: It's not, I have removed it.

scripts/compose/grafana/provisioning/dashboards/log-viewer.json
mnonnenmacher: No, dashboards have to be in JSON, I have added a link to the docs to the commit message.

sschuberth (approved):

PR #1701, opened 2024-12-23
refactor(ui): Use API for statistics cards

Etsija:
With new API support, we can finally switch to use back-end queries for all statistics cards (in the product and ORT run overview pages). This will most probably have a positive effect to the UI performace by improving the rendering speed of the statistics cards.


lamppu:
Not sure if this should be fixed in this PR, or a separate one, but for a product with no runs yet I'm seeing this:
kuva

I guess these should rather be dashes or something?

Etsija:
> Not sure if this should be fixed in this PR, or a separate one, but for a product with no runs yet I'm seeing this: kuva
> 
> I guess these should rather be dashes or something?

I'll take a look.

Etsija:
What do you say @lamppu ?  I think with the current code structure it's difficult to say whether there's no data (as no runs done / empty product) or whether data fetching has failed, so could this be "N/A" than "Failed"?
```
          <div className='text-2xl font-bold'>
            {value !== undefined ? value : 'Failed'}
          </div>
```

lamppu:
> What do you say @lamppu ? I think with the current code structure it's difficult to say whether there's no data (as no runs done / empty product) or whether data fetching has failed, so could this be rather "N/A" than "Failed"?

Oh actually the counts will be returned as null if there are no corresponding valid runs to gather the counts from, so maybe some value could be passed in that case?

mmurto:
> What do you say @lamppu ? I think with the current code structure it's difficult to say whether there's no data (as no runs done / empty product) or whether data fetching has failed, so could this be rather "N/A" than "Failed"?
> 
> ```
>           <div className='text-2xl font-bold'>
>             {value !== undefined ? value : 'Failed'}
>           </div>
> ```

The suspense hooks will throw if they don't succeed or there's no old (stale) data: https://tanstack.com/query/latest/docs/framework/react/guides/suspense. So if we're at that point after a suspense query, the data fetching has succeeded.

lamppu (approved):

PR #1702, opened 2024-12-23
refactor(services): Reverse entries order in maps

lamppu:
Reverse order of severities and vulnerability ratings in maps so they are ordered from most severe to least severe, and will be rendered in this order everywhere in the UI as well where the colored bars showcasing the distributions are rendered.

### DIFF
--- a/services/hierarchy/src/main/kotlin/IssueService.kt
+++ b/services/hierarchy/src/main/kotlin/IssueService.kt
@@ -95,7 +95,7 @@ class IssueService(private val db: Database) {
     suspend fun countBySeverityForOrtRunIds(vararg ortRunIds: Long): CountByCategory<Severity> = db.dbQuery {
         val countAlias = Count(OrtRunsIssuesTable.id, true)
 
-        val severityToCountMap = Severity.entries.associateWithTo(mutableMapOf()) { 0L }
+        val severityToCountMap = Severity.entries.reversed().associateWithTo(mutableMapOf()) { 0L }
 
         OrtRunsIssuesTable
             .innerJoin(IssuesTable)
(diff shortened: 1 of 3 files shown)


services/hierarchy/src/main/kotlin/IssueService.kt
```diff
@@ -95,7 +95,7 @@ class IssueService(private val db: Database) {
     suspend fun countBySeverityForOrtRunIds(vararg ortRunIds: Long): CountByCategory<Severity> = db.dbQuery {
         val countAlias = Count(OrtRunsIssuesTable.id, true)
 
-        val severityToCountMap = Severity.entries.associateWithTo(mutableMapOf()) { 0L }
+        val severityToCountMap = Severity.entries.reversed().associateWithTo(mutableMapOf()) { 0L }
```
mnonnenmacher, line 98: `severityToCountMap` is still of type `Map` which makes no promise with regards to ordering, even if the underlying `LinkedHashMap` might preverse the order of the keys and during serialization this might be respected. If sorting becomes a feature of this function, `CountByCategory.map` should become a `SortedMap` to make it explicit.

However, I wonder if sorting on the client side would not be a better option to not make the presentation depend on an implementation detail of the backend.

mnonnenmacher (changes requested):

services/hierarchy/src/main/kotlin/IssueService.kt
lamppu: Ah ok, it seemed to preserve the order in my local dev environment, so I didn't realize that. Perhaps it should be handled in the client side after all.

PR #1712, opened 2024-12-30
fix(ui): Fix one redirect

Etsija:

### DIFF
--- a/ui/src/routes/runs/$runId/index.tsx
+++ b/ui/src/routes/runs/$runId/index.tsx
@@ -28,10 +28,10 @@ export const Route = createFileRoute('/runs/$runId/')({
       const ortRun = await RunsService.getOrtRunById({
         runId: Number.parseInt(params.runId),
       });
-      organizationId = ortRun.organizationId;
-      productId = ortRun.productId;
-      repositoryId = ortRun.repositoryId;
-      index = ortRun.index;
+      organizationId = ortRun.organizationId.toString();
+      productId = ortRun.productId.toString();
+      repositoryId = ortRun.repositoryId.toString();
+      index = ortRun.index.toString();
     } catch (error) {
       if (error instanceof ApiError) {
         throw notFound();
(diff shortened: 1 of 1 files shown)


ui/src/routes/runs/$runId/index.tsx
```diff
@@ -28,19 +28,25 @@ export const Route = createFileRoute('/runs/$runId/')({
...
         runId: Number.parseInt(params.runId),
       });
-      organizationId = ortRun.organizationId;
-      productId = ortRun.productId;
-      repositoryId = ortRun.repositoryId;
-      index = ortRun.index;
+      organizationId = ortRun.organizationId.toString();
+      productId = ortRun.productId.toString();
+      repositoryId = ortRun.repositoryId.toString();
+      index = ortRun.index.toString();
```
sschuberth, line 34: Sorry for a TS beginner question maybe, but how is this "typesafe" when longs are converted to strings here? Isn't this just faking typesafety, as instead of the implicit to-string-conversion in the interpolation of original line 42, you now just do it explicitly in these lines?

Etsija: Better type-safety is brought by feeding the `params` object to `redirect()`, which now recognizes the `to` as a valid route. I believe the earlier format was not type-safe in this manner.

Etsija: The PR was triggered by (supposedly) tightened type-checking of TanStack Router, which lead to [this error](https://github.com/eclipse-apoapsis/ort-server/actions/runs/12538694250/job/34975031666?pr=1711#step:8:49).

sschuberth:
For background, @Etsija just told me this is supposed to make the update in https://github.com/eclipse-apoapsis/ort-server/pull/1711 work.

Etsija:
One code block was accidentally in a wrong place. Now ready for review; verified to work as before, and potentially helping to merge #1711.

sschuberth:
> and potentially helping to merge #1711.

Could you verify that it actually *does* help before we merge by cherry-picking / applying the change from #1711 locally and testing it?

Etsija:
> > and potentially helping to merge #1711.
> 
> Could you verify that it actually _does_ help before we merge by cherry-picking / applying the change from #1711 locally and testing it?

I've now verified that the build fails with #1711 applied locally without this fix - and with the fix, the build succeeds.

sschuberth (approved):

sschuberth:
> I've now verified that the build fails with #1711 applied locally without this fix - and with the fix, the build succeeds.

Great, I'll rebase #1711 on top of this.

PR #1739, opened 2025-01-07
Start untangling orchestrator

mnonnenmacher:
The few classes in the `orchestrator` currently have a very high degree of coupling and no clear separation of concerns. For example:
* The logic to determine the next jobs is partly implemented in `WorkerScheduleInfo` and partly in `WorkerScheduleContext`.
* Messaging is handled by `Orchestrator`, `WorkerScheduleContext`, and `WorkerScheduleInfo`.
* `Orchestrator`, `WorkerJobRepositories`, and `WorkerScheduleInfo` all interact with the repositories to create jobs or update their status.

Besides some minor clean ups, this PR separates the logic to determine the next jobs into a new class that has no dependencies on infrastructure. It is the first in a series of changes to untangle the different responsibilities of the orchestrator to make them testable in isolation and improve overview.

### DIFF
--- a/orchestrator/src/main/kotlin/WorkerScheduleInfo.kt
+++ b/orchestrator/src/main/kotlin/WorkerScheduleInfo.kt
@@ -56,18 +56,18 @@ internal enum class WorkerScheduleInfo(
      * A list defining the worker jobs that this job depends on. This job will only be executed after all the
      * dependencies have been successfully completed.
      */
-    private val dependsOn: List<WorkerScheduleInfo> = emptyList(),
+    val dependsOn: List<WorkerScheduleInfo> = emptyList(),
 
     /**
      * A list defining the worker jobs that must run before this job. The difference to [dependsOn] is that this job
      * can also run if these other jobs will not be executed. It is only guaranteed that it runs after all of them.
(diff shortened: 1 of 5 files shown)


orchestrator/src/main/kotlin/WorkerScheduleInfo.kt
```diff
@@ -125,7 +125,7 @@ internal enum class WorkerScheduleInfo(
             configs.evaluator != null
     },
 
-    REPORTER(ReporterEndpoint, runsAfter = listOf(EVALUATOR), runAfterFailure = true) {
+    REPORTER(ReporterEndpoint, dependsOn = listOf(ANALYZER), runsAfter = listOf(EVALUATOR), runAfterFailure = true) {
```
oheger-bosch, line 122: It was a deliberate decision that the Reporter step should always be executed to enable use cases like an "ORT run report" containing information about the whole run with its successful and failing steps. So, I would be reluctant to say that it generally makes no sense to run the reporter after a failed analyzer step. In every case, the type "fix" is not correct for this commit because this is no bug, but the behavior was by design.

Issue #1745, opened 2025-01-08
CycloneDX XML SBOM doesn't get written


Etsija:
I have configured the Reporter job with this:
```
  "jobConfigs": {
    "reporter": {
      "formats": [
        "CycloneDx",
        "SpdxDocument",
        "WebApp"
      ],
      "config": {
        "SpdxDocument": {
          "options": {
            "outputFileFormats": "YAML,JSON"
          },
          "secrets": {}
        },
        "CycloneDx": {
          "options": {
            "outputFileFormats": "XML,JSON"
          },
          "secrets": {}
        }
      }
  }
```

But the ORT run only produces WebApp, SPDX YAML+JSON, CycloneDX JSON - CycloneDX XML is always missing from the reports.

![Screenshot from 2025-01-08 07-38-48](https://github.com/user-attachments/assets/0f758e0c-a96d-4d7d-b369-1f039290956d)

Verified from the logs that it is indeed not written:
```
time=2025-01-08T05:27:13.605 thread=DefaultDispatcher-worker-1 traceId=b672f9d2-a8fe-40bf-a79a-22f9652fe4cc ortRunId=12 component=reporter-worker reporterJobId=12 level=INFO  logger=o.e.a.o.w.reporter.ReportStorage message="Storing 'bom.cyclonedx.json' under key '12|bom.cyclonedx.json'."
time=2025-01-08T05:27:13.611 thread=DefaultDispatcher-worker-4 traceId=b672f9d2-a8fe-40bf-a79a-22f9652fe4cc ortRunId=12 component=reporter-worker reporterJobId=12 level=INFO  logger=o.e.a.o.w.reporter.ReportStorage message="Storing 'scan-report-web-app.html' under key '12|scan-report-web-app.html'."
time=2025-01-08T05:27:13.631 thread=DefaultDispatcher-worker-2 traceId=b672f9d2-a8fe-40bf-a79a-22f9652fe4cc ortRunId=12 component=reporter-worker reporterJobId=12 level=INFO  logger=o.e.a.o.w.reporter.ReportStorage message="Storing 'bom.spdx.yml' under key '12|bom.spdx.yml'."
[...]

PR #1748, opened 2025-01-08
ui: More report formats

Etsija:
This PR adds more output formats to CycloneDX, SPDX, and NOTICE file in Reporter.

Screenshot from 2025-01-08 10-49-33

Note: please also see issue #1745.  Earlier in ORT core, plugin naming was following the class naming convention, and therefore, `CycloneDx` was used as an id. With the new plugin structure in ORT core, each plugin should have an id which should be used when configuring the plugin for an ORT Run.  For CycloneDX, [the correct id is actually now "CycloneDX"](https://github.com/oss-review-toolkit/ort/blob/8ebd8c0aea0ffb0770edc92a8c477a4e2f4f0586/plugins/reporters/cyclonedx/src/main/kotlin/CycloneDxReporter.kt#L98-L103), not "CycloneDx". Therefore, it needed to be changed in the UI as well, which is done as part of this PR.

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
@@ -146,7 +146,7 @@ export type PackageManager = (typeof packageManagers)[number]['id'];
 
 export const reportFormats = [
   {
-    id: 'CycloneDx',
+    id: 'CycloneDX',
     label: 'CycloneDX SBOM',
   },
   {
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
@@ -147,15 +147,15 @@ export type PackageManager = (typeof packageManagers)[number]['id'];
 export const reportFormats = [
   {
     id: 'CycloneDX',
-    label: 'CycloneDX SBOM',
+    label: 'CycloneDX SBOM (JSON and XML formats)',
   },
   {
     id: 'SpdxDocument',
-    label: 'SPDX Document',
+    label: 'SPDX SBOM (JSON and YAML formats)',
   },
   {
     id: 'PlainTextTemplate',
-    label: 'NOTICE file',
+    label: 'NOTICE file (DEFAULT and SUMMARY formats)',
   },
   {
     id: 'WebApp',
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
@@ -539,6 +539,10 @@ export function formValuesToPayload(
     return Object.keys(options).length > 0 ? options : undefined;
   };
 
+  //
+  // Analyzer configuration
+  //
+
   // In ORT Server, running or not running a job for and ORT Run is decided
   // based on the presence or absence of the corresponding job configuration
   // in the request body. If a job is disabled in the UI, we pass "undefined"
@@ -562,12 +566,20 @@ export function formValuesToPayload(
     ),
   };
 
+  //
+  // Advisor configuration
+  //
+
   const advisorConfig = values.jobConfigs.advisor.enabled
     ? {
         skipExcluded: values.jobConfigs.advisor.skipExcluded,
         advisors: values.jobConfigs.advisor.advisors,
       }
     : undefined;
+
+  //
+  // Scanner configuration
+  //
 
   const scannerConfig = values.jobConfigs.scanner.enabled
     ? {
@@ -576,6 +588,10 @@ export function formValuesToPayload(
         skipExcluded: values.jobConfigs.scanner.skipExcluded,
       }
     : undefined;
+
+  //
+  // Evaluator configuration
+  //
 
   const evaluatorConfig = values.jobConfigs.evaluator.enabled
     ? {
@@ -592,11 +608,19 @@ export function formValuesToPayload(
       }
     : undefined;
 
+  //
+  // Reporter configuration
+  //
+
   const reporterConfig = values.jobConfigs.reporter.enabled
     ? {
         formats: values.jobConfigs.reporter.formats,
       }
     : undefined;
+
+  //
+  // Notifier configuration
+  //
 
   // Convert the recipient addresses back to an array of strings, as expected by the back-end.
   const addresses = values.jobConfigs.notifier.mail.recipientAddresses
@@ -641,6 +665,10 @@ export function formValuesToPayload(
         },
       }
     : undefined;
+
+  //
+  // Create the payload from worker configurations
+  //
 
   // Convert the parameters and labels arrays back to objects, as expected by the back-end.
   const parameters = values.jobConfigs.parameters
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
@@ -20,7 +20,12 @@
 import { FieldErrors } from 'react-hook-form';
 import { z } from 'zod';
 
-import { AnalyzerJobConfiguration, CreateOrtRun, OrtRun } from '@/api/requests';
+import {
+  AnalyzerJobConfiguration,
+  CreateOrtRun,
+  OrtRun,
+  ReporterJobConfiguration,
+} from '@/api/requests';
 import { packageManagers } from '../../-types';
 
 const keyValueSchema = z.object({
@@ -401,8 +406,12 @@ export function defaultValues(
               ortRun.jobConfigs.reporter !== undefined &&
               ortRun.jobConfigs.reporter !== null,
             formats:
-              ortRun.jobConfigs.reporter?.formats ||
-              baseDefaults.jobConfigs.reporter.formats,
+              // To fix retriggering (with "Reuse" button) of old runs with wrong data
+              // ("CycloneDx" instead of "CycloneDX"), convert this to correct plugin
+              // configuration.
+              ortRun.jobConfigs.reporter?.formats?.map((format) =>
+                format === 'CycloneDx' ? 'CycloneDX' : format
+              ) || baseDefaults.jobConfigs.reporter.formats,
           },
           notifier: {
             enabled:
@@ -612,9 +621,49 @@ export function formValuesToPayload(
   // Reporter configuration
   //
 
+  // Check if CycloneDX, SPDX, and/or NOTICE file reports are enabled in the form,
+  // and configure them to use all output formats, accordingly.
+
+  const cycloneDxEnabled =
+    values.jobConfigs.reporter.formats.includes('CycloneDX');
+  const spdxDocumentEnabled =
+    values.jobConfigs.reporter.formats.includes('SpdxDocument');
+  const noticeFileEnabled =
+    values.jobConfigs.reporter.formats.includes('PlainTextTemplate');
+
+  const config: ReporterJobConfiguration['config'] = {};
+
+  if (spdxDocumentEnabled) {
+    config.SpdxDocument = {
+      options: {
+        outputFileFormats: 'YAML,JSON',
+      },
+      secrets: {},
+    };
+  }
+
+  if (cycloneDxEnabled) {
+    config.CycloneDX = {
+      options: {
+        outputFileFormats: 'XML,JSON',
+      },
+      secrets: {},
+    };
+  }
+
+  if (noticeFileEnabled) {
+    config.PlainTextTemplate = {
+      options: {
+        templateIds: 'NOTICE_DEFAULT,NOTICE_SUMMARY',
+      },
+      secrets: {},
+    };
+  }
+
   const reporterConfig = values.jobConfigs.reporter.enabled
     ? {
         formats: values.jobConfigs.reporter.formats,
+        config: Object.keys(config).length > 0 ? config : undefined,
       }
     : undefined;
 
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
@@ -304,7 +304,7 @@ export function defaultValues(
       },
       reporter: {
         enabled: true,
-        formats: ['CycloneDx', 'SpdxDocument', 'WebApp'],
+        formats: ['CycloneDX', 'SpdxDocument', 'WebApp'],
       },
       notifier: {
         enabled: false,
--- a/.github/ort/config.yml
+++ b/.github/ort/config.yml
@@ -9,7 +9,7 @@ ort:
     skipExcluded: true
   reporter:
     config:
-      CycloneDx:
+      CycloneDX:
         output.file.formats: JSON
       SpdxDocument:
         creationInfo.organization: Eclipse Apoapsis


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -33,6 +33,13 @@ const packageManagerOptionsSchema = z.object({
   options: z.array(keyValueSchema).optional(),
 });
 
+const reporterOptionsSchema = z.object({
+  options: z.object({
+    outputFileFormats: z.string().optional(),
+    templateIds: z.string().optional(),
```
sschuberth, line 39: Just a remark: IMO it's a bit weird to combine properties that are supported by different reporters into a single options schema. Ideally, each reporter should have its own schema, as it's more or less just "coincidence" that both CycloneDX and SPDX reporters support the equally named "outputFileFormats" option. But I'm not sure whether it's worth the effort.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -304,7 +318,7 @@ export function defaultValues(
       },
       reporter: {
         enabled: true,
-        formats: ['CycloneDx', 'SpdxDocument', 'WebApp'],
+        formats: ['CycloneDX', 'SpdxDocument', 'WebApp'],
```
sschuberth, line 321: I believe this change should be extracted to a precessing commit as it deserves a dedicated rationale, like:

```
Fix the casing of the "CycloneDX" report format to exactly match the ORT
plugin ID [1]. This is a preparation for configuring reporter-specific
options that are keyed by the plugin ID.

[1]: https://github.com/oss-review-toolkit/ort/blob/69a15f4fbe04b1ed837c311dd510d66c99701089/plugins/reporters/cyclonedx/src/main/kotlin/CycloneDxReporter.kt#L99
```

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -592,12 +622,85 @@ export function formValuesToPayload(
...
+  //
+  // Reporter configuration
+  //
+
+  // Check if CycloneDX, SPDX, and/or NOTICE file reports are enabled in the form,
+  // and configure them to use all output formats, accordingly.
+
+  // Unfortunately, OpenAPI's PluginConfiguration always requires 'secrets' object,
+  // and it cannot be "undefined", so we need to pass in an empty secrets object instead,
+  // which produces an eslint error about empty object types allowing non-nullish values.
```
sschuberth, line 634: @mnonnenmacher is there a way to make secrets optional?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -592,12 +622,85 @@ export function formValuesToPayload(
       }
     : undefined;
 
+  //
+  // Reporter configuration
+  //
+
+  // Check if CycloneDX, SPDX, and/or NOTICE file reports are enabled in the form,
+  // and configure them to use all output formats, accordingly.
```
sschuberth, line 630: Here you say "all formats", which I like better than "both formats" in the commit message. Mind aligning the latter to this?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -592,12 +622,85 @@ export function formValuesToPayload(
...
+  // which produces an eslint error about empty object types allowing non-nullish values.
+
+  /* eslint-disable @typescript-eslint/no-empty-object-type */
+
+  const cycloneDxEnabled =
+    values.jobConfigs.reporter.formats.includes('CycloneDX');
+  const spdxDocumentEnabled =
+    values.jobConfigs.reporter.formats.includes('SpdxDocument');
+  const noticeFileEnabled =
+    values.jobConfigs.reporter.formats.includes('PlainTextTemplate');
```
sschuberth, line 643: Hard-coding this not not really scale well if we had additional formats. Also see my comment below.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -592,12 +622,85 @@ export function formValuesToPayload(
...
+    };
+    PlainTextTemplate?: {
+      options: {
+        templateIds: string;
+      };
+      secrets: {};
+    };
+  } = {};
+
+  if (spdxDocumentEnabled) {
```
sschuberth, line 666: Do these blocks really need to be wrapped in conditionals? As there should be no harm done by adding options for disabled reporters, I'd simply drop the conditionals to simplify code.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -539,6 +553,10 @@ export function formValuesToPayload(
     return Object.keys(options).length > 0 ? options : undefined;
   };
 
+  //
```
sschuberth, line 556: Nit: Maybe extract the addition of these section headers to a separate `docs` commit to keep the diff for the actual changes smaller.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
```diff
@@ -146,16 +146,16 @@ export type PackageManager = (typeof packageManagers)[number]['id'];
 
 export const reportFormats = [
   {
-    id: 'CycloneDx',
-    label: 'CycloneDX SBOM',
+    id: 'CycloneDX',
```
sschuberth, line 149: IMO this line should be squashed into the commit that deals with the CycloneDX renaming to align with the plugin ID.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
```diff
@@ -146,16 +146,16 @@ export type PackageManager = (typeof packageManagers)[number]['id'];
...
+    label: 'CycloneDX SBOM (both JSON and XML)',
```
sschuberth, line 150: I propose to say "(JSON and XML formats)". Similar below.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
```diff
@@ -146,16 +146,16 @@ export type PackageManager = (typeof packageManagers)[number]['id'];
...
   },
   {
     id: 'SpdxDocument',
-    label: 'SPDX Document',
+    label: 'SPDX Document (both JSON and YAML)',
```
sschuberth, line 154: While at it, I'd also change "SPDX Document" to "SPDX SBOM".

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-types.ts
```diff
@@ -146,16 +146,16 @@ export type PackageManager = (typeof packageManagers)[number]['id'];
...
   },
   {
     id: 'PlainTextTemplate',
-    label: 'NOTICE file',
+    label: 'NOTICE file (both NOTICE_DEFAULT and NOTICE_SUMMARY)',
```
sschuberth, line 158: Here, the "NOTICE_" prefixes could be dropped for shortness, and say "(DEFAULT and SUMMARY formats)".

sschuberth (changes requested):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
Etsija: GitButler failed me in this: I tried to do a separate commit, but it tied the changes into another commit.

lamppu:
> Note 2: this PR introduces a partially breaking change. **IMPORTANT: do not use Rerun functionality, if the old run (to be copied) included CycloneDx format in reporter job. The reason is related to Note 1: Rerun copies the old config as-is, and as a consequence, both `CycloneDx` and `CycloneDX` reporter formats are activated, of which only the latter one is correct. This might cause some problems with the Reporter.**

Maybe you could filter out the old 'CycloneDx' [here](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/routes/organizations/%24orgId/products/%24productId/repositories/%24repoId/_repo-layout/create-run/-create-run-utils.ts#L404)?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
Etsija: We have earlier made efforts to not include configurations for disabled jobs/their sub-items, to simplify debugging in problem cases.

Etsija:
> > Note 2: this PR introduces a partially breaking change. **IMPORTANT: do not use Rerun functionality, if the old run (to be copied) included CycloneDx format in reporter job. The reason is related to Note 1: Rerun copies the old config as-is, and as a consequence, both `CycloneDx` and `CycloneDX` reporter formats are activated, of which only the latter one is correct. This might cause some problems with the Reporter.**
> 
> Maybe you could filter out the old 'CycloneDx' [here](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/routes/organizations/%24orgId/products/%24productId/repositories/%24repoId/_repo-layout/create-run/-create-run-utils.ts#L404)? I guess that would be a bit of an ugly hack, but could prevent having to manually reconfigure runs

Good idea, I'll investigate, as it would make this PR more robust.

lamppu:
> > > Note 2: this PR introduces a partially breaking change. **IMPORTANT: do not use Rerun functionality, if the old run (to be copied) included CycloneDx format in reporter job. The reason is related to Note 1: Rerun copies the old config as-is, and as a consequence, both `CycloneDx` and `CycloneDX` reporter formats are activated, of which only the latter one is correct. This might cause some problems with the Reporter.**
> > 
> > 
> > Maybe you could filter out the old 'CycloneDx' [here](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/routes/organizations/%24orgId/products/%24productId/repositories/%24repoId/_repo-layout/create-run/-create-run-utils.ts#L404)? I guess that would be a bit of an ugly hack, but could prevent having to manually reconfigure runs
> 
> Good idea, I'll investigate, as it would make this PR more robust.

Or maybe it could instead replace it with 'CycloneDX', but in any case, something like that

Etsija:
@sschuberth if you could at least mark resolved discussions as resolved, it would make it easier to progress this PR further.

sschuberth:
> @sschuberth if you could at least mark resolved discussions as resolved, it would make it easier to progress this PR further.

Sure, I'm always doing that once I have time for another review iteration. But so far I didn't.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -33,6 +33,13 @@ const packageManagerOptionsSchema = z.object({
   options: z.array(keyValueSchema).optional(),
 });
 
+const reporterOptionsSchema = z.object({
```
sschuberth, line 36: Commit message:

> If none of those report formats is chosen, don't include a config
object in `jobConfigs.reporter`.

That just states something that's obvious from the code diff. But the commit message should be used to actually explain *why* things are done this way. You were mentioning [elsewhere](https://github.com/eclipse-apoapsis/ort-server/pull/1748/commits/54a9132223ffc4ff5d4b0981c55ad5593b0a4d80#r1907095025) that this helps debugging (But again, why?), so that could be the starting point for an argumentation.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -401,8 +415,10 @@ export function defaultValues(
               ortRun.jobConfigs.reporter !== undefined &&
               ortRun.jobConfigs.reporter !== null,
             formats:
-              ortRun.jobConfigs.reporter?.formats ||
-              baseDefaults.jobConfigs.reporter.formats,
+              // Convert this to correct plugin configuration
```
sschuberth, line 418: This should explain a bit more *why* this is needed. Keyword: Retriggering of old runs with "wrong" data. And remember to use a dot at the end of the sentence 😉

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -612,9 +628,74 @@ export function formValuesToPayload(
...
+    values.jobConfigs.reporter.formats.includes('CycloneDX');
+  const spdxDocumentEnabled =
+    values.jobConfigs.reporter.formats.includes('SpdxDocument');
+  const noticeFileEnabled =
+    values.jobConfigs.reporter.formats.includes('PlainTextTemplate');
+
+  const config: {
+    SpdxDocument?: {
+      options: {
+        outputFileFormats: string;
```
sschuberth, line 650: I obviously know too little about TypeSCript, but can't we replace this with something like

    outputFileFormats: spdxDocumentEnabled ? 'YAML,JSON' : string;

etc. to avoid lines 668-694?

sschuberth (changes requested):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
Etsija: Well, my justification to this is as follows: to debug the ORT run creation form and its translation into the config object that is sent to the backend, we have taken lots of measures to _not_ include configuration objects for things which actually aren't part of the created run.

To debug something which is not working when changing the run creation form, I usually comment out the actual part of the code which sends the request to create the run to back-end and console log the JSON object which the code is trying to send.

But the object is _very_ long and complicated, so while I do realize it doesn't harm sending configs to items which are non-existing in the run, it does help debugging when they are not sent at all.

If you really see it necessary to simplify the reporter part of code, I will do so; I do know the code will look a lot simpler.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -592,12 +626,85 @@ export function formValuesToPayload(
...
+  /* eslint-disable @typescript-eslint/no-empty-object-type */
+
+  const cycloneDxEnabled =
+    values.jobConfigs.reporter.formats.includes('CycloneDX');
+  const spdxDocumentEnabled =
+    values.jobConfigs.reporter.formats.includes('SpdxDocument');
+  const noticeFileEnabled =
+    values.jobConfigs.reporter.formats.includes('PlainTextTemplate');
+
+  const config: {
```
mmurto, line 649: Would something like `const config: CreateRunFormValues['jobConfigs']['reporter']['config']` work here?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -592,12 +626,85 @@ export function formValuesToPayload(
...
+  //
+
+  // Check if CycloneDX, SPDX, and/or NOTICE file reports are enabled in the form,
+  // and configure them to use all output formats, accordingly.
+
+  // Unfortunately, OpenAPI's PluginConfiguration always requires 'secrets' object,
+  // and it cannot be "undefined", so we need to pass in an empty secrets object instead,
+  // which produces an eslint error about empty object types allowing non-nullish values.
+
+  /* eslint-disable @typescript-eslint/no-empty-object-type */
```
mmurto, line 640: Is this a left-over?

mmurto (changes requested):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
Etsija: > Is this a left-over?

No.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
sschuberth: Ok, then a short version of that explanation that goes to the commit message could be

> If none of those report formats is chosen, don't include a config
object in `jobConfigs.reporter` to reduce the amount of configuration to consider when debugging things.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
mmurto: Can you then move the `eslint-disable` to where the lint violator is and use `eslint-disable-next-line` instead, or is there a reason to do the override here?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
Etsija: > Would something like `const config: CreateRunFormValues['jobConfigs']['reporter']['config']` work here?

What would you initialize it to?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
sschuberth: Ok, then I obviously really don't understand things. I though type definition was done via Zod [here](https://github.com/eclipse-apoapsis/ort-server/pull/1748/files#diff-9a8b8b3152bca6b691f8c3f876a9924e1e3f599eed2716384a259f74ce161890R37-R41).

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
mmurto: Is the `{}` actually the correct type here? If you read [the lint docs](https://typescript-eslint.io/rules/no-empty-object-type/), it tries to guard against an error:

> The {}, or "empty object" type in TypeScript is a common source of confusion for developers unfamiliar with TypeScript's structural typing. {} represents any non-nullish value, including literals like 0 and "":
>
> let anyNonNullishValue: {} = 'Intentionally allowed by TypeScript.';
>
> Often, developers writing {} actually mean either:
>
> object: representing any object value
> unknown: representing any value at all, including null and undefined
> In other words, the "empty object" type {} really means "any value that is defined". That includes arrays, class > instances, functions, and primitives such as string and symbol.

The comment says that it "requires 'secrets' object", so if it should be an object, the typing is wrong.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
sschuberth: I'm fine either way, and will leave the final approval to @mmurto who's much more knowledgeable in TypeScript than I am.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
mmurto: Here's a nice blog post about it: https://www.totaltypescript.com/the-empty-object-type-in-typescript.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
mmurto: Looking at the Zod definition, this could work, right?

```ts
{
  CycloneDX: {},
  SpdxDocument: {},
  PlainTextTemplate: {}
}
```

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
Etsija: `secrets: object` ??  It doesn't pass the return statement.  Believe me, I've tried (almost) everything.

Etsija: Oh I see: `secrets: Record<PropertyKey, never>` seems to work, as in at least pass eslint without the eslint disable.

mmurto: Let me open the editor.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
Etsija: No it doesn't.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
mmurto: Do you feel it's clear enought with the change I suggested below, so by changing the type and init to `  const config: ReporterJobConfiguration['config'] = {};`?

Etsija: Yes, this is far shorter and easier to understand.

Etsija:
With @mmurto latest suggestion, the code is now clearer and shorter. Tested, all formats generated properly.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -95,6 +107,13 @@ export const createRunFormSchema = z.object({
     reporter: z.object({
       enabled: z.boolean(),
       formats: z.array(z.string()),
+      config: z
+        .object({
+          CycloneDX: reporterOptionsSchema,
+          SpdxDocument: reporterOptionsSchema,
+          PlainTextTemplate: reporterOptionsSchema,
+        })
+        .optional(),
```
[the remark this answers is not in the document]
Etsija: No you're probably right.

mmurto (approved):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
mmurto: To clear this up, the linked Zod definition defines the type for the form visible in the UI for creating new forms. The function converts that type to format accepted by the API, so it outputs different type.

PR #1749, opened 2025-01-08
API client chores

MarcelBochtler:


api/v1/client/src/commonMain/kotlin/OrtServerClient.kt
```diff
@@ -59,7 +59,7 @@ class OrtServerClient(
 
             val client = createOrtHttpClient(JSON) {
                 defaultRequest {
-                    url(config.baseUrl)
+                    url(config.baseUrl.takeIf { it.endsWith("/") } ?: "${config.baseUrl}/")
```
oheger-bosch, line 62: Any chance to have this backed by a test with reasonable effort?

sschuberth, line 62: Commit message: Isn't this more of a conventional thing than a requirement? Like, you could very well use no trailing slash here, but then you'd need to use a leading slash in e.g. `post` calls. Simply put, strings are just concatenated as-is, and you need to decide where to put the slash.

MarcelBochtler: > Commit message: Isn't this more of a conventional thing than a requirement?

No.
Using
```
defaultRequest {
    url(config.baseUrl)
}
```

Without a trailing slash, e.g. with the url `https://example.com/ort`, it will remove the `/ort` part and therefore point to a different URL.

MarcelBochtler: See the documentation of `DefaultRequest`:
```
 *     * Base URL doesn't end with slash ->
 *       remove last path segment of base URL and concat strings.
 *       Example:
 *       base = `https://example.com/dir/deafult_file.html`,
 *       request = `file.html`,
 *       result = `https://example.com/dir/file.html`
```

MarcelBochtler: I tried, but I don't think that testing this is possible without changing the place where this fix is applied.

What I tried:
* Mocking the client in order to verify the call to the correct. Not possible because a client instance is created in lin 60, therefore it cannot be mocked
* Accessing the `defaultRequest` parameters and checking if the slash is correctly applied: I didn't find a way how to do that.

Do you have any other ideas how this could be tested?
Or does this mean that maybe we should apply the fix at a different place?

oheger-bosch: Not sure, but wouldn't a test be possible on a higher level for code that is using the client? So, sort of an integration test?

MarcelBochtler: Moved the change to the CLI to ensure the URL is always the same, which also makes testing simpler.

cli/src/main/kotlin/Extensions.kt
```diff
@@ -0,0 +1,25 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.cli
+
+/**
+ * Add a trailing slash to this [String] if it is missing.
+ */
+fun String.ensureTrailingSlash(): String = takeIf { endsWith("/") } ?: "$this/"
```
sschuberth, line 25: Funnily enough, as ORT already was `fun String?.withoutSuffix()`, I was considering to add `ensureSuffix(suffix: String)`. If that comes, we could replace this again later on. Or already make this also more generic by taking the desired suffix as an argument.

cli/src/test/kotlin/ExtensionsTest.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.eclipse.apoapsis.ortserver.cli.ensureTrailingSlash
+
+class ExtensionsTest : StringSpec({
+    "ensureTrailingSlash" should {
+        "add a trailing slash to a string if missing" {
+            "http://localhost:8080".ensureTrailingSlash() should endWith("/")
+        }
+
+        "not add a trailing slash to a string if already present" {
+            "http://localhost:8080/".ensureTrailingSlash() should endWith("/")
```
sschuberth, line 33: This does not really verify that not two slashes are at the end.

sschuberth (dismissed):

cli/src/test/kotlin/ExtensionsTest.kt
oheger-bosch: Agreed, better check the whole expected string outcome. And you also do not check whether this new function is called when constructing the `OrtServerOptions`. Is there an easy way to test the clikt option processing?

MarcelBochtler: > Is there an easy way to test the clikt option processing?

There isn't.

mnonnenmacher (approved):

cli/src/test/kotlin/ExtensionsTest.kt
sschuberth: Maybe things have changed with the resolution of https://github.com/ajalt/clikt/issues/449?

PR #1760, opened 2025-01-10
chore(reuse): Use the default "precedence" of "closest"

sschuberth:
The migration in 6a9e50a added `precedence = "aggregate"`, probably to resemble the old behavior. However, the most natural and also new default behavior is to use the license declaration that is closest [1] to the file.

[1]: https://reuse.software/spec-3.3/#reusetoml


sschuberth:
FYI @Etsija.

MarcelBochtler (approved):

PR #1761, opened 2025-01-10
feat(cli): Add a `--version` option to show the used CLI version

MarcelBochtler:
At the moment the CLI and the Server use the same version, therefore this is the version which is printed.

### DIFF
--- a/cli/src/main/kotlin/OrtServerMain.kt
+++ b/cli/src/main/kotlin/OrtServerMain.kt
@@ -27,6 +27,7 @@ import com.github.ajalt.clikt.parameters.groups.provideDelegate
 import com.github.ajalt.clikt.parameters.options.convert
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.options.versionOption
 
 import kotlin.system.exitProcess
 
@@ -34,6 +35,7 @@ import kotlinx.serialization.json.Json
 
 import org.eclipse.apoapsis.ortserver.client.OrtServerClient
 import org.eclipse.apoapsis.ortserver.client.OrtServerClientConfig
+import org.eclipse.apoapsis.ortserver.model.ORT_SERVER_VERSION
 
 const val COMMAND_NAME = "ort-server"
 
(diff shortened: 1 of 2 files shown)


cli/src/main/kotlin/OrtServerMain.kt
```diff
@@ -47,6 +48,13 @@ class OrtServerMain : SuspendingNoOpCliktCommand(COMMAND_NAME) {
 
     init {
         subcommands(RunsCommand(ortServerConfig))
+
+        versionOption(
+            version = this::class.java.`package`.implementationVersion ?: "IDE-SNAPSHOT",
```
sschuberth, line 53: Can the existing `ORT_SERVER_VERSION` constant be reused? If necessary, maybe my moving it to a more accessible place the CLI can depend on?

oheger-bosch:
It would probably make sense to extend the output to contain both the versions of the server and the client. The former can be obtained via the versions endpoint. Then users could for instance figure out whether their client is outdated.

MarcelBochtler:
> It would probably make sense to extend the output to contain both the versions of the server and the client. The former can be obtained via the versions endpoint. Then users could for instance figure out whether their client is outdated.

Yes, but as this would require to also implement authentication, and more importantly the handling of missing authentication when using the `--version` command.

I would like to keep this out of this PR.

oheger-bosch:
> Yes, but as this would require handling missing authentication when using the `--version` command, as `--version` is an authenticated command. And I think `--version` should be able to be used without the user configuring authentication.
> 
> I would like to keep this out of this PR.

AFAIK, the version endpoint does not require authentication. But I am fine with starting simple.

MarcelBochtler:
> AFAIK, the version endpoint does not require authentication. But I am fine with starting simple.

See:
https://github.com/eclipse-apoapsis/ort-server/blob/991e2c36e48ca20167bbfcc4d4f3dfb48f370d5d/core/src/main/kotlin/plugins/Routing.kt#L41-L48

mnonnenmacher:
> AFAIK, the version endpoint does not require authentication. But I am fine with starting simple.

For reference, see the discussion here: https://github.com/eclipse-apoapsis/ort-server/issues/95

sschuberth:
> > AFAIK, the version endpoint does not require authentication. But I am fine with starting simple.
> 
> For reference, see the discussion here: #95

Thanks, I was meaning to quote that, too. So, is @oheger-bosch now in favor of an unauthenticated version endpoint? 😸

oheger-bosch:
> > For reference, see the discussion here: #95
> 
> Thanks, I was meaning to quote that, too. So, is @oheger-bosch now in favor of an unauthenticated version endpoint? 😸

Of course, I remember this discussion - but I somehow had in mind that my concerns were ignored.

mnonnenmacher (approved):

PR #1763, opened 2025-01-10
fix(ui): Make the favicon path relative

mnonnenmacher:
This fixes loading the favicon if the UI is not hosted at the root of a domain.

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -21,7 +21,7 @@ License-Filename: LICENSE
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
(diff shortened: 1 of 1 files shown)


sschuberth (approved):

mmurto:
This could be a more general solution: https://vite.dev/guide/build.html#relative-base.

mnonnenmacher:
> This could be a more general solution: https://vite.dev/guide/build.html#relative-base.

Interesting, but if just removing the slash works to me that's good enough for now.

mmurto:
> > This could be a more general solution: https://vite.dev/guide/build.html#relative-base.
> 
> Interesting, but if just removing the slash works to me that's good enough for now.

It works for now, but as referencing images with absolute paths and placing them in the `public` directory is a common way to use images, other images can break later as well.

PR #1765, opened 2025-01-10
feat(core): Add a function to delete runs older than a given timestamp

nnobelis:
See commit.

### DIFF
--- a/dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
+++ b/dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
@@ -20,6 +20,7 @@
 package org.eclipse.apoapsis.ortserver.dao.repositories.ortrun
 
 import kotlinx.datetime.Clock
+import kotlinx.datetime.Instant
 
 import org.eclipse.apoapsis.ortserver.dao.blockingQuery
 import org.eclipse.apoapsis.ortserver.dao.blockingQueryCatching
@@ -119,6 +120,12 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
                 condition
             }
         }
+
+    override fun listRunsBefore(before: Instant): ListQueryResult<OrtRun> = db.blockingQuery {
+        OrtRunDao.listQuery(ListQueryParameters.DEFAULT.copy(limit = Int.MAX_VALUE), OrtRunDao::mapToModel) {
+            OrtRunsTable.finishedAt less before
+        }
+    }
 
     override fun listForRepository(repositoryId: Long, parameters: ListQueryParameters): ListQueryResult<OrtRun> =
         db.blockingQueryCatching {
--- a/services/hierarchy/src/main/kotlin/OrtRunService.kt
+++ b/services/hierarchy/src/main/kotlin/OrtRunService.kt
@@ -19,6 +19,8 @@
 
 package org.eclipse.apoapsis.ortserver.services
 
+import kotlinx.datetime.Instant
+
 import org.eclipse.apoapsis.ortserver.dao.dbQuery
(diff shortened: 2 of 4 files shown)


services/hierarchy/src/main/kotlin/OrtRunService.kt
```diff
@@ -73,4 +77,22 @@ class OrtRunService(
             throw ResourceNotFoundException("ORT run with id '$ortRunId' not found.")
         }
     }
+
+    /**
+     * Delete all ORT runs that are older than the given [before] timestamp. Runs are deleted from the database and
+     * their reports are deleted from storage.
+     */
+    suspend fun deleteAncientRuns(before: Instant) {
```
mnonnenmacher, line 85: You could also use a more specific name like `deleteRunsCreatedBefore()`.

services/hierarchy/src/test/kotlin/OrtRunServiceTest.kt
```diff
@@ -178,6 +181,64 @@ class OrtRunServiceTest : WordSpec() {
                 fixtures.ortRunRepository.get(ortRunId) shouldNotBe null
             }
         }
+
+        "deleteAncientRuns" should {
```
oheger-bosch, line 185: The function has been renamed to "deleteRunsCreatedBefore"; this should be reflected here as well.

services/hierarchy/src/main/kotlin/OrtRunService.kt
```diff
@@ -73,4 +77,22 @@ class OrtRunService(
...
+        val operator = FilterOperatorAndValue(ComparisonOperator.LESS_THAN, before)
+        val listQueryParameters = ListQueryParameters.DEFAULT.copy(limit = Int.MAX_VALUE)
+        val runs = ortRunRepository.list(listQueryParameters, OrtRunFilters(finishedAt = operator))
+
+        logger.info("Deleting ${runs.totalCount} ORT runs older than $before.")
+
+        runs.data.forEach { run ->
+            runCatching {
+                deleteOrtRun(run.id)
+            }
```
oheger-bosch, line 95: Exceptions when deleting a run should be logged.

dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
```diff
@@ -109,11 +110,21 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
             OrtRunDao.listQuery(parameters, OrtRunDao::mapToModel) {
                 var condition: Op<Boolean> = Op.TRUE
 
-                filters?.status?.let { statusFilter ->
-                    condition = condition and OrtRunsTable.status.applyFilter(
-                        statusFilter.operator,
-                        statusFilter.value
-                    )
+                filters?.let {
```
oheger-bosch, line 113: I don't strictly object against this approach with generic filter conditions, but I am worried that we add here significant complexity just to duplicate logic/functionality the ORM framework already contains. Wouldn't it be more straight-forward to add a dedicated `list` function to `OrtRunRepository` which applies the filtering logic for old ORT runs?

services/hierarchy/src/main/kotlin/OrtRunService.kt
nnobelis: They are already logged in `deleteOrtRun`. I would like to avoid double logging.

dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
nnobelis: You right, no need to expose such filtering in the API.

oheger-bosch (approved):

PR #1782, opened 2025-01-13
Maintenance tasks project

oheger-bosch:
This PR adds a new project for the execution of periodic maintenance task. It replaces #1636.


tasks/src/main/kotlin/Task.kt
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.eclipse.apoapsis.ortserver.tasks
+
+/**
+ * An interface describing a task that can be executed.
+ */
+interface Task {
+    /**
+     * Execute the specific logic of this task.
```
wkl3nk, line 27: Nit: Could be reduced to "Execute the task".

settings.gradle.kts
```diff
@@ -58,6 +58,7 @@ include(":storage:azure-blob")
 include(":storage:database")
 include(":storage:s3")
 include(":storage:spi")
+include(":tasks")
```
wkl3nk, line 61: Nit: Just be aware about the english terms. "Task" is more an abstract term, while "Job" would be closer to the Kubernetes terminlogy of "Jobs" or "CronJobs". No discussion required, just pointing out an alternative.

tasks/src/main/kotlin/Task.kt
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * Tasks may fail with arbitrary exceptions. It is in the responsibility of the caller to catch these exceptions
+     * and handle them.
+     */
+    suspend fun execute()
```
wkl3nk, line 32: What about imperative style in comments?

wkl3nk, line 32: Should there be some kind of hint or contract, that, once a task is started, it does **not** return to the caller **until the task is finished**? Just to make sure that no implementation does things in an asynchronous manner internally, possibly returning immediately after `execute()` is called?

oheger-bosch: I could add something to the documentation of the `execute` function. Not sure though whether it is actually needed. There is the `suspend` modifier which indicates that the function is invoked in a (semi-) blocking manner. However, it will probably do no harm to mention this explicitly.

oheger-bosch: I have extended the comment accordingly.

settings.gradle.kts
MarcelBochtler, line 61: Commit message nit: Remove the leading space before `that`.

MarcelBochtler (approved):
LGTM

tasks/src/main/kotlin/TaskRunner.kt
```diff
@@ -0,0 +1,104 @@
...
+ * The tasks to run are defined via the configuration as a comma-separated list of task names. To change the set of
+ * tasks dynamically, the configuration can be overwritten using an environment variable. The task names must match
+ * named bean definitions of type [Task] in the Koin modules. The function fetches all referenced tasks from Koin and
+ * executes them asynchronously.
+ */
+internal suspend fun runTasks(modules: List<Module>) {
+    logger.info("Setting up Koin application.")
+    val app = startKoin {
+        modules(modules)
+        modules(configModule())
```
mnonnenmacher, line 61: The `configModule` is already added in the `main` function.

tasks/src/test/kotlin/TaskRunnerTest.kt
```diff
@@ -0,0 +1,197 @@
...
+                val taskLog = checkTaskExecution("task1", "nonExistingTask", "task2")
+
+                taskLog.getExecutedTasks() shouldContainExactlyInAnyOrder setOf("task1", "task2")
+            }
+        }
+
+        "main" should {
+            "call runTasks with a database module" {
+                checkMain { koin ->
+                    verifyDatabaseModuleIncluded()
```
mnonnenmacher, line 92: Should this also verify that the `configModule` is included?

tasks/src/main/kotlin/TaskRunner.kt
oheger-bosch: Good catch. Fixed.

tasks/src/test/kotlin/TaskRunnerTest.kt
oheger-bosch: This is implicitly tested by 'verifyDatabaseModuleIncluded()`; the function checks whether a `ConfigManager` is available.

mnonnenmacher (approved):

PR #1793, opened 2025-01-14
fix(ui): Correctly apply system theme to decide CycloneDX icon variant

Etsija:

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/sbom/index.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/sbom/index.tsx
@@ -43,7 +43,7 @@ const SBOMComponent = () => {
   // theme automatically. Two different logos are provided instead in
   // https://cyclonedx.org/about/branding/ and the correct one is selected
   // based on the current theme.
-  const { theme } = useTheme();
+  const { activeTheme } = useTheme();
 
   const { data: ortRun } = useRepositoriesServiceGetOrtRunByIndexSuspense({
     repositoryId: Number.parseInt(params.repoId),
@@ -128,7 +128,9 @@ const SBOMComponent = () => {
                 <TooltipTrigger>
                   <img
                     alt='CycloneDX'
-                    src={theme === 'dark' ? CycloneDXLight : CycloneDXDark}
+                    src={
+                      activeTheme === 'dark' ? CycloneDXLight : CycloneDXDark
+                    }
                     width={230}
                   />
                 </TooltipTrigger>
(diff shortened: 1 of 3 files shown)


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/sbom/index.tsx
```diff
@@ -128,7 +132,15 @@ const SBOMComponent = () => {
...
                     alt='CycloneDX'
-                    src={theme === 'dark' ? CycloneDXLight : CycloneDXDark}
+                    src={
+                      theme === 'dark'
+                        ? CycloneDXLight
+                        : theme === 'light'
+                          ? CycloneDXDark
+                          : systemTheme === 'dark'
+                            ? CycloneDXLight
+                            : CycloneDXDark
```
sschuberth, line 142: Isn't there a way to do just *one* query about the theme in use (no matter matter whether set explicitly, or by adhering to the system theme), and then only have *one* conditional in here like `activeTheme == 'dark' ? CycloneDXLight : CycloneDXDark`?

mmurto (changes requested):
IMO this should rather be handled in the theme provider so that the `useTheme` returns the effective theme. I think the selected theme will also need to be returned for the theme switcher.

So maybe enhance the theme provider to return something like

```typescript
{
    selected: 'light' | 'dark' | 'system',
    theme: 'light' | 'dark'
} 
```

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/sbom/index.tsx
sschuberth: Looks to me like it should work when [using React's `useTheme`](https://reactnavigation.org/docs/use-theme/) instead of [our own `useTheme` implementation](https://github.com/eclipse-apoapsis/ort-server/blob/52a7d1491575a00110ab628a7dfb7ca2fa6b0c0d/ui/src/components/theme-provider-context.tsx#L30-L37).

mmurto: > Looks to me like it should work when [using React's `useTheme`](https://reactnavigation.org/docs/use-theme/) instead of [our own `useTheme` implementation](https://github.com/eclipse-apoapsis/ort-server/blob/52a7d1491575a00110ab628a7dfb7ca2fa6b0c0d/ui/src/components/theme-provider-context.tsx#L30-L37).

That's not a React built-in, but a hook from a library called React Navigation.

mmurto (approved):

PR #1800, opened 2025-01-15
Make deleting old ort runs more efficient

oheger-bosch:
This PR introduces a new function to find ORT runs before a given reference time that only returns the IDs of matched runs. For the use case of deleting old ORT runs, this is way more efficient. Previously, fully initialized ORT run objects were retrieved with all their associations. For a larger number of affected runs, this took very long and consumed lot of resources.

### DIFF
--- a/dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
+++ b/dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
@@ -121,12 +121,6 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
             }
         }
 
-    override fun listRunsBefore(before: Instant): ListQueryResult<OrtRun> = db.blockingQuery {
-        OrtRunDao.listQuery(ListQueryParameters.DEFAULT.copy(limit = Int.MAX_VALUE), OrtRunDao::mapToModel) {
-            OrtRunsTable.finishedAt less before
-        }
-    }
-
     override fun listForRepository(repositoryId: Long, parameters: ListQueryParameters): ListQueryResult<OrtRun> =
         db.blockingQueryCatching {
             OrtRunDao.listQuery(parameters, OrtRunDao::mapToModel) { OrtRunsTable.repositoryId eq repositoryId }
@@ -150,6 +144,12 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
         OrtRunDao.find {
             OrtRunsTable.status inList activeRunStatuses
         }.map { ActiveOrtRun(it.id.value, it.createdAt, it.traceId) }
+    }
+
+    override fun findRunsBefore(before: Instant): List<Long> = db.blockingQuery {
+        OrtRunsTable.select(OrtRunsTable.id)
+            .where { OrtRunsTable.finishedAt less before }
+            .map { it[OrtRunsTable.id].value }
     }
 
     override fun update(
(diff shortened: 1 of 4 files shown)


dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
```diff
@@ -152,6 +152,12 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
         }.map { ActiveOrtRun(it.id.value, it.createdAt, it.traceId) }
     }
 
+    override fun findRunsBefore(before: Instant): List<Long> = db.blockingQuery {
```
nnobelis, line 155: Instead of adding a new function, could you maybe replace `listRunsBefore` instead to avoid dead code ?

dao/src/test/kotlin/repositories/ortrun/DaoOrtRunRepositoryTest.kt
```diff
@@ -451,6 +462,58 @@ class DaoOrtRunRepositoryTest : StringSpec({
...
+        val finishedTimes = listOf(
+            refTime - 10.seconds,
+            refTime - 1.seconds,
+            refTime,
+            refTime + 1.seconds
+        )
+
+        // Mock the clock to have defined finishedAt times for the test ORT runs.
+        // For each run, the repository queries the clock twice: for the creation and the completion time.
+        val mockTimes = buildList {
```
nnobelis, line 476: IMO this is unnecessary: I you look at the test I wrote, you can just:

1. create the ORT runs
2. change their states to finished them
3. list the runs with a time before the `finishedAt` of the last run

No need to mock the clock...

oheger-bosch: To be honest, I found your approach with delays to force a later finished time rather problematic.

dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
oheger-bosch: Yes, I was not sure whether there was a use case for the existing function - maybe for the UI, so that users can see older runs that are going to be removed soon? But I am also fine with replacing it. Should the function then be named `listRunsBefore`? Since the `list` functions typically return entities.

nnobelis: Yes, fine with me, or `listRunsIdsBefore`. Or `findRunsIdsBefore`. As you want :)

dao/src/test/kotlin/repositories/ortrun/DaoOrtRunRepositoryTest.kt
nnobelis: Ah, could you elaborate ?
Do you think it can lead to flakiness or you don't like the test losing time ?

I find mocking a static function of the JVM more brutal and hackish, if I may.

oheger-bosch: Losing time and that you cannot specify an exact time. I also find the mocking solution as implemented here not really thrilling. Ideally, the repository would have a `Clock` instance as parameter, so that time-related operations could be controlled from the outside. But this would require a major refactoring.

dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
oheger-bosch: I replaced `listRunsBefore` by `findRunsBefore`.

nnobelis (approved):

PR #1801, opened 2025-01-15
ui: Add mustRunAfter to analyzer configuration

Etsija:

Screenshot from 2025-01-15 12-17-59

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
@@ -35,6 +35,16 @@ const keyValueSchema = z.object({
 
 const packageManagerOptionsSchema = z.object({
   enabled: z.boolean(),
+  mustRunAfter: z
+    .array(
+      z
+        .string()
+        // Only accept valid package manager IDs to the string array in the form.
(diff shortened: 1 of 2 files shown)


sschuberth:
Just as a remark to the public: We're aware that the UI for this feature could be designed to be more slick, but for now we'd like to proceed with this to make the feature available at all in the UI.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -35,6 +35,16 @@ const keyValueSchema = z.object({
 
 const packageManagerOptionsSchema = z.object({
   enabled: z.boolean(),
+  mustRunAfter: z
+    .array(
+      z
+        .string()
+        // Only accept valid package manager IDs to the string array in the form
```
sschuberth, line 42: As you know, I'm a fan of having dots at the end of sentences, even if they're just code comments 😉

lamppu:
Doesn't necessarily have to happen in this PR, but maybe the view in the Job Configurations tab should be updated as well, now it looks like this 
kuva

sschuberth:
> Doesn't necessarily have to happen in this PR

Yeah, let's not add "mustRunAfter" configuration to that view for now. I believe the information displayed there needs some overhaul anyway.

lamppu (approved):

PR #1805, opened 2025-01-15
Add a task to delete old ORT runs

oheger-bosch:
This PR adds a first task to the recently introduced project for maintenance tasks which enforces the data retention policy for ORT runs.

### DIFF
--- a/tasks/src/main/kotlin/impl/DeleteOldOrtRunsTask.kt
+++ b/tasks/src/main/kotlin/impl/DeleteOldOrtRunsTask.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.tasks.impl
+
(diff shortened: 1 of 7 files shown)


tasks/src/main/resources/application.conf
```diff
@@ -18,3 +18,31 @@
...
+  azureBlobStorageAccountName = ${?REPORT_STORAGE_STORAGE_ACCOUNT_NAME}
+  s3AccessKey = ${?REPORT_STORAGE_ACCESS_KEY}
+  s3SecretKey = ${?REPORT_STORAGE_SECRET_KEY}
+  s3Region = ${?REPORT_STORAGE_REGION}
+  s3BucketName = ${?REPORT_STORAGE_BUCKET_NAME}
+  s3EndpointUrl = ${?REPORT_STORAGE_ENDPOINT_URL}
+}
+
+dataRetention {
+  ortRunDays = 90
```
sschuberth, line 46: Is this the default that#s now in place? I'm not sure that number would be fine with us. Of course, we could configure it in our deployment. But I'm wonder whether the feature as a whole should be disabled instead by default to not change behavior compared to before. Only users of the server who want / need it should enabled it. Is there even a way to disable the feature completely now, e.g. by setting the number of days to a negative value?

Same for other maintenance / data retention task introduced recently / to be introduced. What do you think?

oheger-bosch: The way the maintenance tasks are implemented currently (refer to the corresponding [discussion](https://github.com/eclipse-apoapsis/ort-server/discussions/1560)), they are not active automatically, but users have to manually schedule them using an external mechanism. In our environment, we will configure a Kubernetes Cron job to run this task periodically. If you do not do this, you don't have to worry that runs are suddenly deleted.

tasks/src/main/kotlin/impl/DeleteOldOrtRunsTask.kt
```diff
@@ -0,0 +1,67 @@
...
+
+import org.eclipse.apoapsis.ortserver.config.ConfigManager
+import org.eclipse.apoapsis.ortserver.services.OrtRunService
+import org.eclipse.apoapsis.ortserver.tasks.Task
+
+import org.slf4j.LoggerFactory
+
+/**
+ * A task class that handles the deletion of old ORT runs according to the configured data retention policy. The task
+ * delegates to [OrtRunService] to trigger the deletion of old ORT finished before the number of days defined by the
```
bs-ondem, line 35: typo: "of old ORT runs.."

mnonnenmacher (approved):

bs-ondem (approved):

PR #1814, opened 2025-01-16
UI: Streamline filtering into column headers next to sorting

lamppu:
Move the filter buttons to the column headers to create a more slick UI. Also add a text filter component. The filter icon will be colored in blue when a filter is applied.

Select filter:
Screenshot from 2025-01-16 14-20-15

Text filter:
Screenshot from 2025-01-16 14-21-36


Etsija (changes requested):
As the commit 7a7bc5b ("refactor(ui): Streamline...") is the "beef" of your PR, I think it warrants a good explanatory commit body. You do quite a lot of refactoring for the data-table components (deleting, adding and modifying sub-components), but also add 'select' filters to three tables. As it is, this is not a 'refactor' but definitely a 'feat' commit.

Etsija (commented):
Is this needed? Earlier you said this is true by default.

Etsija:
> Is this needed? Earlier you said this is true by default.

Obviously, I don't know how to comment a code line when reviewing, so this can be ignored. :smile:

Etsija:
Sorting icon in tables has a tooltip which is reactive (ie. sort ascending / desc / clear). For the sake of consistency in the UI, could the filtering icons also display tooltips, for example, "Filter this column" / "Remove filtering"?

lamppu:
> Sorting icon in tables has a tooltip which is reactive (ie. sort ascending / desc / clear). For the sake of consistency in the UI, could the filtering icons also display tooltips, for example, "Filter this column" / "Remove filtering"?

@sschuberth we discussed this briefly with Jyrki after the daily, and felt that perhaps we should rather use tooltips more sparsely for things that aren't very self-explanatory. So the conclusion was that for concistency, maybe the tooltip should be removed from the sorting icon instead of adding one for the filtering icon. Just wanted to check if you had any opinions on this though?

sschuberth:
> So the conclusion was that for concistency, maybe the tooltip should be removed from the sorting icon instead of adding one for the filtering icon.

I believe that sorting (in contrast to filtering) should keep the tooltip, as they tell the user what's going to happen next if the icons was clicked again. So the tooltip might prevent the user from accidentally clearing the sorting, for example. Clearing as a "third state" is a nice feature, but maybe a but unusual UX, so it deserves from explanation.

The same does not apply to filtering, so I think it's fine the way it is, to have tooltip for sorting but not for filtering. The latter is easy enough to add later if desired.

lamppu:
@Etsija FYI, I edited the extended types a bit since your last review, cause I noticed I had omitted the generics part `<TData extends RowData, TValue>` for the `ColumnMeta`, it did work, but apparently it should be there to ensure that it matches, and doesn't cause issues in the future.  I also realized I shouldn't use an additional generic `T` type for the select option values etc., but use the `TValue` type of the column instead. And this change is also affecting the `FilterMultiSelect` component.

Etsija (approved):

PR #1819, opened 2025-01-17
ci(docker): Build and push the CLI Docker image

nnobelis:


sschuberth:
Maybe a stupid question, but why does it even make sens e to have a Docker image for the CLI? I mean, AFAIU it's not supposed to run in K8s, and has no dependencies on external tools. (Except maybe a JRE, but I was somewhat assuming that the CLI would be built as a Kotlin native binary.)

nnobelis:
@sschuberth I guess it's for the ease of distribution, to make it easily available to various CI environments.

Having a binary instead would mean that you're to _curl_ it, change its permission, then finally run it.

AFAIK, different CLIs have tasks to directly run a Docker image.

sschuberth:
> I guess it's for the ease of distribution, to make it easily available to various CI environments.

You *guess* why you're doing this? 😸

More seriously, while I kind of see the point of how a Docker image could be slightly easier to deploy in some environments, then I'd at least strong encourage to try minimizing its size.

But...

> Except maybe a JRE, but I was somewhat assuming that the CLI would be built as a Kotlin native binary.

Looks like my assumption was wrong here. The `cli` project is not a multiplatform project yet, unfortunately.

Probably as a follow-up to this PR, I believe what should be done is:
- Convert the `cli` project to multiplatform (which implies not applying the `application` plugin anymore due to [this](https://youtrack.jetbrains.com/issue/KT-37964) and other issues)
- "Misuse" the `jib-native-image-extension-gradle` from [here](https://github.com/GoogleContainerTools/jib-extensions/tree/master/first-party/jib-native-image-extension-gradle) to not copy a GraalVM-created binary, but a Kotlin native binary to a distroless / scratch image similar to like done [here](https://github.com/jamesward/ktor-native-server/blob/main/build.gradle.kts).
- FYI, [this](https://dev.to/mreichelt/christmas-hacking-squeezing-kotlin-native-into-docker-6ao) older blog post also has some nice insights (which, however, might not be fully relevant anymore).

nnobelis:
> You guess why you're doing this?

My guess was to answer your question, not doing the task :sweat_smile: 

Another point I forgot, is that the Docker image is containerized. Also I believe from a security point of view, it's better since you have to explicit enable, through Docker volumes, which folder from the CI agent will be available to the CLI application.

Also with Docker, you get the application update system "for free", plus the optimization of caching the image layers.

Frankly, I understand that Kotlin Native binary is cool and so, but I don't see that added benefit compared to delivering the CLI via Docker.

oheger-bosch:
> 
> Please note that my latest proposal about about shipping a Kotlin native binary _inside_ a Docker image. The main motivation is that this probably reduces the download size by more than 10-fold.

This is certainly a good optimization we should do. But this does not prevent us to already ship what we have, so that our users can put it into use. They will be even happier if we later come with improvements.

oheger-bosch (approved):

PR #1822, opened 2025-01-17
feat(analyzer): Introduce new JobConfiguration parameter for submodules

wkl3nk:
Add a parameter `submoduleFetchStrategy`to the Analyzer JobConfiguration that allows specifying how repository submodules should be fetched. This includes the options to fetch submodules recursively or to fetch only the top level  submodules, providing greater flexibility in submodule handling.

Don't fetch submodules at all: _DISABLED_ (same as _recursiveCheckout_ = false)
Fetch all nested submodules recursively: _FULLY_RECURSIVE_ (same as _recursiveCheckout_ = true)
New: Only fetch the top level submodules: _TOP_LEVEL_ONLY_ 

I decided to keep the parameter _recursiveCheckout_ working for now and mark it as deprecated, because there exist job configurations in the database that still hold it, and these would no longer work after an ORT Server update. Additionally, this allows to relaxed take care for any Kotlin scripts that generate the JobConfigurations. But: Configurations parameter `submoduleFetchStrategy` supersedes `recursiveCheckout`.

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
+++ b/api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
@@ -76,8 +76,16 @@ data class AnalyzerJobConfiguration(
     /**
      * A flag indicating whether the submodules of the repository should be downloaded during the download process.
      * If set to `true`, submodules will be downloaded; if `false`, they will be ignored.
+     *
+     * Note: This attribute is deprecated and will be removed in a future release. Use [submoduleFetchStrategy] instead.
+     *
      */
     val recursiveCheckout: Boolean = true,
+
+    /**
+     * The strategy to use for fetching submodules.
+     */
+    val submoduleFetchStrategy: SubmoduleFetchStrategy? = null,
 
     /**
      * The list of package curation providers to use.
@@ -372,3 +380,24 @@ data class NotifierJobConfiguration(
      */
     val jira: JiraNotificationConfiguration? = null
 )
+
+@Serializable
+/**
+ * The strategy to use for fetching submodules.
+ */
+enum class SubmoduleFetchStrategy {
+    /**
+     * Don't fetch submodules at all.
+     */
+    DISABLED,
+
+    /**
+     * Only fetch the top level of submodules.
+     */
+    TOP_LEVEL_ONLY,
+
+    /**
+     * Fetch all nested submodules recursively.
+     */
+    FULLY_RECURSIVE
+}
--- a/workers/analyzer/src/main/kotlin/analyzer/AnalyzerDownloader.kt
+++ b/workers/analyzer/src/main/kotlin/analyzer/AnalyzerDownloader.kt
@@ -21,8 +21,13 @@ package org.eclipse.apoapsis.ortserver.workers.analyzer
 
 import java.io.File
 
+import org.eclipse.apoapsis.ortserver.model.SubmoduleFetchStrategy
+
+import org.ossreviewtoolkit.downloader.VcsHost
 import org.ossreviewtoolkit.downloader.VersionControlSystem
 import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
+import org.ossreviewtoolkit.plugins.api.PluginConfig
 import org.ossreviewtoolkit.utils.ort.createOrtTempDir
 
 import org.slf4j.LoggerFactory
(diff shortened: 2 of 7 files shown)


api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
```diff
@@ -372,3 +380,24 @@ data class NotifierJobConfiguration(
...
+enum class SubmoduleFetchStrategy {
+    /**
+     * Don't fetch submodules at all.
+     */
+    DISABLED,
+
+    /**
+     * Only fetch the first layer of submodules.
+     */
+    FIRST_LAYER_ONLY,
```
sschuberth, line 397: Maybe better: `TOP_LEVEL_ONLY`?

wkl3nk: Agreed. It is the better term.

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
```diff
@@ -76,9 +76,17 @@ data class AnalyzerJobConfiguration(
     /**
      * A flag indicating whether the submodules of the repository should be downloaded during the download process.
      * If set to `true`, submodules will be downloaded; if `false`, they will be ignored.
+     *
+     * Note: This attribute is deprecated and will be removed in a future release. Use [submoduleFetchStrategy] instead.
```
oheger-bosch, line 80: Since there has not yet been an official 1.0 release of ORT Server, we can still remove this field. Not sure from where it is used.

workers/analyzer/src/main/kotlin/analyzer/AnalyzerDownloader.kt
```diff
@@ -34,22 +39,68 @@ class AnalyzerDownloader {
...
         logger.info("Finished downloading '$repositoryUrl' revision '$revision'.")
 
         return outputDir
     }
+
+    /**
+     * Build custom [PluginConfig] for Git VCS if the [submoduleFetchStrategy] is
+     * [SubmoduleFetchStrategy.TOP_LEVEL_ONLY].
+     */
+    internal fun buildCustomVcsPluginConfiguMap(
```
oheger-bosch, line 75: Should this function be named `buildCustomVcsPluginConfigMap` instead of `...ConfiguMap` ?

workers/analyzer/src/main/kotlin/analyzer/AnalyzerDownloader.kt
```diff
@@ -34,22 +39,68 @@ class AnalyzerDownloader {
...
+        repositoryUrl: String, submoduleFetchStrategy: SubmoduleFetchStrategy?
+    ) =
+        if (submoduleFetchStrategy == SubmoduleFetchStrategy.TOP_LEVEL_ONLY) {
+            val vcsType = VcsHost.parseUrl(repositoryUrl).type
+            require(vcsType == VcsType.GIT) {
```
oheger-bosch, line 80: How is this error handled upstream? Will this fail the Analyzer run? And is this error visible to the user? Would it maybe be a better strategy to continue, but create an issue of level WARN?

workers/analyzer/src/main/kotlin/analyzer/AnalyzerDownloader.kt
```diff
@@ -34,22 +39,68 @@ class AnalyzerDownloader {
...
+        } else {
+            emptyMap()
+        }
+
+    /**
+     * Evaluate the [recursiveCheckout] and the [submoduleFetchStrategy] parameter to determine if the working tree
+     * should be checked out recursively. The [submoduleFetchStrategy] parameter takes precedence over the
+     * deprecated [recursiveCheckout] parameter.
+     */
+    internal fun evaluateRecursiveCheckoutParameter(
```
oheger-bosch, line 98: Again, this function could be obsolete when dropping the `recursiveCheckout` parameter.

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
wkl3nk: (At least as Bosch) we cannot simply drop this configuration field, because it is used in our Mini Server data model and Mini Server CLI. My plan was to keep it at least we have rolled out the changes to the production environment, and then as a second step actually remove it from Mini Server data model and Mini Server CLI.

oheger-bosch: Can we in MiniServer map it to the new property?

workers/analyzer/src/main/kotlin/analyzer/AnalyzerDownloader.kt
wkl3nk: It currently fails the Analyzer run, with the following message in the logs:
````
 logger=o.e.a.o.w.analyzer.AnalyzerComponent message="Analyzer job '83' failed."
java.lang.IllegalArgumentException: Submodule fetch strategy TOP_LEVEL_ONLY is only supported for Git repositories, but got VCS type 'Mercurial'.
	at org.eclipse.apoapsis.ortserver.workers.analyzer.AnalyzerDownloader.buildCustomVcsPluginConfigMap$analyzer(AnalyzerDownloader.kt:80)
	at org.eclipse.apoapsis.ortserver.workers.analyzer.AnalyzerDownloader.downloadRepository(AnalyzerDownloader.kt:49)
	at org.eclipse.apoapsis.ortserver.workers.analyzer.AnalyzerWorker.run(AnalyzerWorker.kt:73)
````

wkl3nk: For the following reasons I would like to keep it as it is and intentionally fail the Analyzer run:

- Yes, intentionally failing the Analyzer is a little bit harsh, but when it happens, then already when setting up / onboarding a repository. In these cases, customers should be able to investigate the log files to find out why the scan aborted in the Analyzer phase.
- If we would continue ... we don't even know if the VCS type that is used supports the concept of submodules like Git does.
- Only one customer (up to now) required the functionality to check out the top level submodules only. It is not worth the effort to change the code so deeply that it allows to create _issues_ already in the Downloader. Currently it is only possible to create _issues_ during the Analyzer run (after the download of the repository)

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
wkl3nk: Sure. In a second step, we would build some mapping logic in the Mini Server that enables us to get rid of the (then outdated) parameter in ORT Server.

workers/analyzer/src/main/kotlin/analyzer/AnalyzerDownloader.kt
wkl3nk: Fixed.

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
oheger-bosch: Ok, IIUC, this second step will happen in a matter of days rather than weeks or months, right?

workers/analyzer/src/main/kotlin/analyzer/AnalyzerDownloader.kt
oheger-bosch: Well, the error message is not ideal, especially with the reference to Mercurial. But you are probably right that the vast majority of users has Git and will not be affected. So, I won't object.

oheger-bosch:
There are failing tests; otherwise I would approve.

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
wkl3nk: Yes, it could be migrated then after **this one** is deployed.

wkl3nk:
@oheger-bosch There is still a little bug in ORT upstream that makes the tests unstable. I will create a bug ticket and provide a workaround for the test, but then referencing the bug ticket in the code so that the workaround can be removed as soon as the bug is fixed.

oheger-bosch (dismissed):

wkl3nk:
Created a ticket for the little bug in ORT upstream that makes the tests unstable: https://github.com/oss-review-toolkit/ort/issues/9795

oheger-bosch (approved):

PR #1841, opened 2025-01-21
feat(cli): Change the name of the ORT Server CLI executable

MarcelBochtler:
The new name is `osc` which was decided in a public vote, see [1].

[1]: https://github.com/eclipse-apoapsis/ort-server/discussions/1794


cli/src/main/kotlin/OrtServerMain.kt
```diff
@@ -58,6 +59,10 @@ class OrtServerMain : SuspendingNoOpCliktCommand(COMMAND_NAME) {
             message = { "$commandName CLI version $it" }
         )
     }
+
+    override fun help(context: Context) = """
+        The ORT Server Client (osc) CLI is a command line interface to interact with an ORT Server instance.
```
sschuberth, line 64: To avoid "CLI is a command line interface", maybe just drop "CLI"? Or say "The ORT Server Client (OSC) is a Command Line Interface (CLI) to interact with an ORT Server instance."

bs-ondem (approved):

PR #1845, opened 2025-01-21
Fix inefficient project and package comparison

mnonnenmacher:
Improve performance when looking for duplicate projects and packages. See the commit messages for details.

### DIFF
--- a/dao/src/main/kotlin/repositories/analyzerrun/PackagesTable.kt
+++ b/dao/src/main/kotlin/repositories/analyzerrun/PackagesTable.kt
@@ -19,21 +19,26 @@
 
 package org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun
 
-import org.eclipse.apoapsis.ortserver.dao.mapAndCompare
 import org.eclipse.apoapsis.ortserver.dao.tables.shared.DeclaredLicenseDao
+import org.eclipse.apoapsis.ortserver.dao.tables.shared.DeclaredLicensesTable
 import org.eclipse.apoapsis.ortserver.dao.tables.shared.IdentifierDao
 import org.eclipse.apoapsis.ortserver.dao.tables.shared.IdentifiersTable
 import org.eclipse.apoapsis.ortserver.dao.tables.shared.RemoteArtifactDao
 import org.eclipse.apoapsis.ortserver.dao.tables.shared.RemoteArtifactsTable
 import org.eclipse.apoapsis.ortserver.dao.tables.shared.VcsInfoDao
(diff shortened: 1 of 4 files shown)


oheger-bosch (approved):
Changes look good to me. Not sure about the test failures though.

mnonnenmacher:
> Changes look good to me. Not sure about the test failures though.

I found two issues related to missing metadata which I could fix. I have added tests for that and for projects the logic was already broken before if some metadata was missing.

nnobelis:
I will believe you that the proposed queries do the same thing as the previous ones.

However I wonder about an alternative: could we not store for each package a hash of its (nested) values ?
Hence, to fast lookup a package, we would just compute its hash and load the matching package from the DB ?

Same idea for the projects.

oheger-bosch:
> However I wonder about an alternative: could we not store for each package an hash of its (nested) values ? Hence, to fast lookup a package, we would just compute its hash and load the matching package from the DB ?
> 
> Same idea for the projects.

This is exactly the approach I am currently trying for storing scan summaries.

One problem with hashes is that they might need to be updated when there are changes on the data model. Other than that, this is probably quite effective.

mnonnenmacher:
> I will believe you that the proposed queries do the same thing as the previous ones.
> 
> However I wonder about an alternative: could we not store for each package an hash of its (nested) values ? Hence, to fast lookup a package, we would just compute its hash and load the matching package from the DB ?
> 
> Same idea for the projects.

Hashes are one way to solve this, but I wanted to implement a fix that does not require any database migrations, because this makes it a lot easier to test it in on different setups and also makes the deployment less risky. Last time when we did a big migration of the project and packages tables some problems only became visible when applying the migrations to the production database.

Also, like in a hash table there is a non zero chance for hash collisions so if a hash matches you would still have to do the equality check as well.

> This is exactly the approach I am currently trying for storing scan summaries.
> 
> One problem with hashes is that they might need to be updated when there are changes on the data model. Other than that, this is probably quite effective.

This the other reason I did not want to use hashes, with any future schema change we would have to consider if they affect the hashes and probably recalculate them, making trivial migrations like adding a single column to a table a lot more difficult to implement and more expensive to execute.

I expect that we will optimize the schema especially in performance critical areas like analyzer and scanner results and I don't want to make those changes harder than they need to be.

mnonnenmacher:
@nnobelis One more thing, I mentioned in the commit message that I still use the DAO API for loading projects and packages. This is inefficient because it uses a lot of queries. I would like to change this to also load the entities with a single query, but did not do this yet due to timing constraints and because I think that the schema should be changed anyway, so I wanted to focus only on the critical path and fix the performance of the comparison.

PR #1850, opened 2025-01-22
feat(cli): Accept CreateOrtRun object in run command

Juli0q:
The command now accepts the CreateOrtRun object either as a JSON string or as a path to a JSON file.

### DIFF
--- a/cli/src/main/kotlin/StartCommand.kt
+++ b/cli/src/main/kotlin/StartCommand.kt
@@ -20,9 +20,13 @@
 package org.eclipse.apoapsis.ortserver.cli
 
 import com.github.ajalt.clikt.command.SuspendingCliktCommand
+import com.github.ajalt.clikt.parameters.groups.mutuallyExclusiveOptions
+import com.github.ajalt.clikt.parameters.groups.required
+import com.github.ajalt.clikt.parameters.options.convert
 import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.types.inputStream
 import com.github.ajalt.clikt.parameters.types.long
 
 import kotlin.time.Duration.Companion.milliseconds
@@ -31,7 +35,6 @@ import kotlin.time.Duration.Companion.seconds
 import kotlinx.coroutines.delay
 
 import org.eclipse.apoapsis.ortserver.api.v1.model.CreateOrtRun
-import org.eclipse.apoapsis.ortserver.api.v1.model.JobConfigurations
 import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRun
 import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunStatus
 import org.eclipse.apoapsis.ortserver.client.OrtServerClient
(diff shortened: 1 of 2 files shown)


cli/src/main/kotlin/StartCommand.kt
```diff
@@ -23,6 +23,7 @@ import com.github.ajalt.clikt.command.SuspendingCliktCommand
 import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.types.file
```
MarcelBochtler, line 26: Commit message: This is a breaking change.
Please use a `!` to indicate this.

```
feat(cli)!: Accept CreateOrtRun object in run command
```

You should also mention that you replace the current options with the JSON option.

cli/src/main/kotlin/StartCommand.kt
```diff
@@ -45,41 +45,34 @@ class StartCommand(private val config: OrtServerOptions) : SuspendingCliktComman
...
-    )
-
     private val wait by option(
         "--wait",
         envvar = "WAIT",
         help = "Wait for the run to finish."
     ).flag()
 
+    private val ortRunConfigFile by option(
+        "--ort-run-config-file",
```
MarcelBochtler, line 55: This command will be called like
```
osc run start --ort-run-config-file ...
```

I'm not sure about the option naming, because you're somehow duplicating the `run` part, and I'm not sure if this is really for setting run __configurations__.

How about `--parameters-file`? Or is this to generic?

cli/src/main/kotlin/StartCommand.kt
```diff
@@ -45,41 +45,34 @@ class StartCommand(private val config: OrtServerOptions) : SuspendingCliktComman
...
+        envvar = "ORT_RUN_CONFIG_FILE",
+        help = "The path to the Create ORT run configuration file."
+    ).file()
+
+    private val ortRunConfigString by option(
+        "--ort-run-config-string",
```
MarcelBochtler, line 61: We should drop the `string` part. CLI parameters are by definition a `string`.
I don't think that you need this to distinguish from the `file` version of the option

cli/src/main/kotlin/StartCommand.kt
```diff
@@ -45,41 +45,34 @@ class StartCommand(private val config: OrtServerOptions) : SuspendingCliktComman
...
+        envvar = "ORT_RUN_CONFIG_STRING",
+        help = "The Create ORT run configuration as a string."
+    )
```
MarcelBochtler, line 64: These two options should be mutually exclusive. Right?
https://ajalt.github.io/clikt/api/clikt/com.github.ajalt.clikt.parameters.groups/mutually-exclusive-options.html

At least one of them is required, which can be implemented using above option.

MarcelBochtler (changes requested):

cli/src/main/kotlin/StartCommand.kt
Juli0q: The problem with the mutually exclusive is that both options have to be of the same data type. And if I understand the clikt documentation correctly, not using .file() means that auto-completion in the terminal does not work. I've opted for this approach because, imo auto-completion is really convenient.

MarcelBochtler: > I've opted for this approach because, imo auto-completion is really convenient.

Yes, I agree. Having this auto-completion is more important than removing some boilerplate code.

cli/src/test/kotlin/StartCommandTest.kt
```diff
@@ -92,7 +92,7 @@ class StartCommandTest : StringSpec({
                 "runs",
                 "start",
                 "--repository-id", "$repositoryId",
-                "--vcs-revision", revision
+                "--parameters", "{\"revision\": \"$revision\", \"jobConfigs\": {}}"
```
MarcelBochtler, line 95: Consider using `"""` to avoid the escaping of `"`s.

oheger-bosch:
I found [this feature](https://ajalt.github.io/clikt/parameters/#file-path-inputstream-and-outputstreams) of clickt. Would this be an option to simplify the implementation an allow reading from stdin?

cli/src/main/kotlin/StartCommand.kt
```diff
@@ -45,41 +48,33 @@ class StartCommand(private val config: OrtServerOptions) : SuspendingCliktComman
...
     private val wait by option(
         "--wait",
         envvar = "WAIT",
         help = "Wait for the run to finish."
     ).flag()
 
+    private val parameters by mutuallyExclusiveOptions(
+        option(
+            "--parameters-file",
+            envvar = "PARAMETERS_FILE",
```
MarcelBochtler, line 60: I think that `PARAMETERS_FILE` is too generic for an environment variable.
At the time the user configures these env vars, they are not in the `run start` context.

For the download commands I solved this issue by using a path with an ORT prefix:
`ORT_DOWNLOAD_LOGS_OUTPUT_DIR`

I know that this isn't completely consistent everywhere (e.g. we still use `REPOSITORY_ID` instead of `ORT_REPOSITORY_ID`).

What do you think about `ORT_RUNS_START_PARAMETERS_FILE`?

MarcelBochtler (approved):

PR #1851, opened 2025-01-22
Scan result storage optimization

oheger-bosch:
This PR optimizes the deduplication logic for scan summaries significantly by using hash values to find duplicates.


workers/scanner/src/main/kotlin/scanner/ScanSummaryHash.kt
```diff
@@ -0,0 +1,103 @@
...
+/** The algorithm to use for hashing. */
+private const val HASH_ALGORITHM = "SHA-256"
+
+/**
+ * Calculate a hash value for the given [ortSummary]. This value is then stored together with the summary to allow
+ * finding duplicates efficiently.
+ */
+@OptIn(ExperimentalStdlibApi::class)
+internal fun calculateScanSummaryHash(ortSummary: ScanSummary): String {
+    val digest = MessageDigest.getInstance(HASH_ALGORITHM)
```
nnobelis, line 41: Can the digest be static and be reused ?

nnobelis (commented):
LGTM but maybe we should had a team discussion regarding the remarks made by @mnonnenmacher  on https://github.com/eclipse-apoapsis/ort-server/pull/1845 ?

While I disagree that the hash collisions are problematic, the need for a migration and an hash regeneration on model changes are indeed pain points.

dao/src/main/resources/db/migration/V90__scanSummaryHash.sql
```diff
@@ -0,0 +1,3 @@
+-- Add a hash column to the scan_summary table to make it easy to find duplicates.
+
+ALTER TABLE scan_summaries ADD COLUMN hash TEXT NOT NULL DEFAULT '';
```
mnonnenmacher, line 3: I don't understand, why does the hash value not have to be calculated for existing summaries?

workers/scanner/src/main/kotlin/scanner/OrtServerScanResultStorage.kt
```diff
@@ -159,24 +158,19 @@ class OrtServerScanResultStorage(
...
-            (ScanSummariesTable.startTime eq summary.startTime.toKotlinInstant().toDatabasePrecision()) and
-                    (ScanSummariesTable.endTime eq summary.endTime.toKotlinInstant().toDatabasePrecision())
-        }.toList()
-
-        return summaries.takeUnless { it.isEmpty() }?.let { existingSummaries ->
-            existingSummaries.firstOrNull { compareScanSummaries(summary, it.mapToModel()) }
-        } ?: createScanSummaryDao(summary)
+        val summaryHash = calculateScanSummaryHash(summary)
+        return ScanSummaryDao.find { (ScanSummariesTable.hash eq summaryHash) }.firstOrNull()
+            ?: createScanSummaryDao(summary, summaryHash)
```
mnonnenmacher, line 163: > Note that this deduplication only affects scan summaries created by a
single ORT run. Therefore, the hash values are actually used only
temporarily, and it will not cause issues if the hash calculation was
changed in the future.

So doesn't this mean that scan summaries are basically not deduplicated anymore? I guess I first need to understand what the purpose of the replaced logic was, but right now I don't get why we even need a hash then.

mnonnenmacher:
> While I disagree that the hash collisions are problematic, 

They are highly unlikely, but if there ever is one it would be very hard to even notice that there is an issue.

> the need for a migration and an hash regeneration on model changes are indeed pain points.

According to the commit message of the second commit this is not an issue here, but I don't yet understand why.

workers/scanner/src/main/kotlin/scanner/OrtServerScanResultStorage.kt
oheger-bosch: The deduplication of scan summaries has been introduced because under some specific circumstances duplicates were stored. I don't remember the full details any more, but it happened when monorepos were scanned with FossID; then each project in the repo was assigned the full summary with all findings. The way the ORT Scanner is integrated with ORT Server (via the scan result storage interface), there is no easy way to detect or prevent such duplicates, but it has to be checked whether an identical summary is already stored.

Since a scan summary has a start and end timestamp, a deduplication over different ORT runs typically cannot happen. So, this applied only to this special case that duplicates were created in a single run. This is the reason why the hash is only relevant during the lifetime of the run, and there is no need to update it or calculate one for existing summaries.

I have noticed that the current deduplication check takes very long for larger scan results and issues an insane number of database queries. (On one project I tested, the check took about 20 minutes.) This can be fully prevented with the hash approach.

mnonnenmacher: Please update the commit message to explain this a bit better. Also, the deduplication was introduced in 95effa1 and 8b41d58 so maybe reference them as well.

dao/src/main/resources/db/migration/V90__scanSummaryHash.sql
mnonnenmacher: Explained below.

workers/scanner/src/main/kotlin/scanner/OrtServerScanResultStorage.kt
oheger-bosch: I tried to explain this better in the commit message.

mnonnenmacher (approved):

PR #1860, opened 2025-01-23
ui: Add issue category to issues data table

Etsija:
This PR uses regular expressions on the issue messages, to categorize the issues to eg. infrastructure or build related, which will help in assigning the issues to different teams. Please note that the categories and regular expressions used will need to be tweaked and expanded in the future, pending on [ORT PR #4393](https://github.com/oss-review-toolkit/ort/issues/4393), which would somewhat move the responsibility of categorizing the issues to ORT itself.

Screenshot from 2025-01-23 12-09-21
Screenshot from 2025-01-23 12-09-50


Etsija:
@sschuberth @mnonnenmacher commit-lint fails, but I do not understand why. I did try to be extra careful when writing the commit messages. Could you help?

ui/src/helpers/get-issue-category.ts
```diff
@@ -0,0 +1,90 @@
...
+  const infrastructureRegEx: RegExp = new RegExp(
+    'The .* worker failed due to an unexpected error.*' +
+      "|ERROR: Timeout after .* seconds while scanning file '.*'\\."
+  );
+  const missingDataRegEx: RegExp = new RegExp(
+    'IOException: Could not resolve provenance for .*'
+  );
+  const buildSystemRegEx: RegExp = new RegExp(
+    '.* failed to resolve dependencies for .*'
+  );
```
mmurto, line 44: Building the regexp is often quite an expensive operation, so these should be moved outside of the function body to ensure they only execute once, and not once for each function call. Also, is the type annotation required? If not, it should be removed.

ui/src/helpers/get-issue-category.ts
```diff
@@ -0,0 +1,90 @@
...
+  switch (true) {
+    case infrastructureRegEx.test(message):
+      return 'Infrastructure';
+    case missingDataRegEx.test(message):
+      return 'Missing Data';
+    case buildSystemRegEx.test(message):
+      return 'Build System';
+    default:
+      return 'Other';
+  }
```
mmurto, line 55: As it's not immediately obvious what `switch (true)` does here and there's no need for fall through, I'd replace this with if-elseif-else or ifs with early returns.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/issues/index.tsx
```diff
@@ -151,6 +154,40 @@ const IssuesComponent = () => {
...
+        return getIssueCategory(issue.message);
+      },
+      {
+        id: 'category',
+        header: 'Category',
+        cell: ({ getValue }) => {
+          return <div>{getValue()}</div>;
+        },
+        filterFn: (row, _columnId, filterValue): boolean => {
+          // Is this correct?
```
mmurto, line 168: What does this relate to?

ui/src/schemas/index.ts
```diff
@@ -35,6 +36,11 @@ export const severitySchema: z.ZodEnum<[Severity, ...Severity[]]> = z.enum([
   'ERROR',
 ]);
 
+// Enum schema for the possible values of the issue categories.
+export const issueCategorySchema: z.ZodEnum<
+  [IssueCategory, ...IssueCategory[]]
+> = z.enum(['Build System', 'Missing Data', 'Infrastructure', 'Other']);
```
mmurto, line 42: To avoid repetition, would it make sense to define the Zod schema as the source of thruth and infer the type from it?

mmurto (changes requested):

ui/src/schemas/index.ts
Etsija: I thought it wasn't possible with Zod, but must've remembered some other related thing about the library.

mmurto: https://zod.dev/?id=type-inference

sschuberth:
> commit-lint fails, but I do not understand why.

There seems to be an issue with issue references. Changing "on ORT PR #4393 [3]" to "an ORT PR [3]" fixes it.

ui/src/schemas/index.ts
```diff
@@ -35,6 +35,12 @@ export const severitySchema: z.ZodEnum<[Severity, ...Severity[]]> = z.enum([
   'ERROR',
 ]);
 
+// Enum schema and type for the possible values of the issue categories.
+export const issueCategorySchema: z.ZodEnum<
+  ['Build System', 'Missing Data', 'Infrastructure', 'Other']
+> = z.enum(['Build System', 'Missing Data', 'Infrastructure', 'Other']);
```
mmurto, line 41: The type annotation can be removed from here now that there's no need to align the schema definition to anything else.

mmurto (changes requested):

ui/src/schemas/index.ts
Etsija: Please explain.

mmurto (approved):

PR #1865, opened 2025-01-24
fix(dao): Add missing index on the `scan_summaries_issues` table

nnobelis:


dao/src/main/resources/db/migration/V90__addMissingScanSummaryIndex.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX IF NOT EXISTS scan_summaries_issues_scan_summary_id_idx ON scan_summaries_issues (scan_summary_id);
```
oheger-bosch, line 1: Fine with me, just one question: Why is 'IF NOT EXISTS` used here, as we do not use such constraints anywhere else?

oheger-bosch (approved):

PR #1867, opened 2025-01-24
Scan storage read optimization

oheger-bosch:
When reading scan results, do not include the findings of the results,
but return stripped down `ScanResult` objects with only basic properties
initialized. The findings are never evaluated because they are already
contained in the database. This is a significant optimization since
loading and mapping large scan results is an expensive operation.

### DIFF
--- a/dao/src/main/kotlin/tables/ScanSummariesTable.kt
+++ b/dao/src/main/kotlin/tables/ScanSummariesTable.kt
@@ -46,12 +46,28 @@ class ScanSummaryDao(id: EntityID<Long>) : LongEntity(id) {
     val snippetFindings by SnippetFindingDao referrersOn SnippetFindingsTable.scanSummaryId
     val issues by ScanSummariesIssuesDao referrersOn ScanSummariesIssuesTable.scanSummaryId
 
-    fun mapToModel() = ScanSummary(
+    /**
+     * Map this DAO to a [ScanSummary] model object. Based on the [withFindings] flag, either include all findings
+     * or return a summary without findings.
+     */
+    fun mapToModel(withFindings: Boolean = true) = ScanSummary(
         startTime = startTime,
         endTime = endTime,
-        licenseFindings = licenseFindings.mapTo(mutableSetOf(), LicenseFindingDao::mapToModel),
-        copyrightFindings = copyrightFindings.mapTo(mutableSetOf(), CopyrightFindingDao::mapToModel),
-        snippetFindings = snippetFindings.mapTo(mutableSetOf(), SnippetFindingDao::mapToModel),
+        licenseFindings = if (withFindings) {
+            licenseFindings.mapTo(mutableSetOf(), LicenseFindingDao::mapToModel)
+        } else {
+            emptySet()
+        },
+        copyrightFindings = if (withFindings) {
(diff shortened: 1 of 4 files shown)


workers/scanner/src/main/kotlin/scanner/OrtServerScanResultStorage.kt
```diff
@@ -65,15 +65,21 @@ import org.slf4j.LoggerFactory
...
+ * The [ScannerWorker] creates an instance of this class and passes it to the ORT scanner. Via this instance, the
+ * scanner can then read already existing scan results from the ORT Server database and write new scan results to it.
+ * Read and written scan results are associated to the scanner run with the provided [scannerRunId]; this establishes
+ * the link to the current ORT run.
  *
- * Read and written scan results are associated to the scanner run with the provided [scannerRunId].
+ * [ScanResult]s are searched by details of a [KnownProvenance] which it is associated with. If the [KnownProvenance]
+ * is an [ArtifactProvenance], the URL and the hash value must match. If the [KnownProvenance] is a
+ * [RepositoryProvenance], the VCS type, URL and the resolved revision must match. Note that for read operations,
+ * only the basic properties of scan results are relevant. Based on this information, the ORT scanner decides for
```
nnobelis, line 78: Which are the basic properties which are relevant ?

Does this include also the issues ?

dao/src/main/kotlin/tables/ScanSummariesTable.kt
```diff
@@ -46,12 +46,28 @@ class ScanSummaryDao(id: EntityID<Long>) : LongEntity(id) {
...
+        } else {
+            emptySet()
+        },
+        copyrightFindings = if (withFindings) {
+            copyrightFindings.mapTo(mutableSetOf(), CopyrightFindingDao::mapToModel)
+        } else {
+            emptySet()
+        },
+        snippetFindings = if (withFindings) {
+            snippetFindings.mapTo(mutableSetOf(), SnippetFindingDao::mapToModel)
```
nnobelis, line 67: I am confused.

This function `SnippetFindingDao::mapToModel()` does not do any DB access, just mapping some data classes.
Is this in-memory mapping expensive ? Which time improvement did your tests show you ?

mnonnenmacher (dismissed):

dao/src/main/kotlin/tables/ScanSummariesTable.kt
mnonnenmacher: `licenseFindings` and so on are delegate properties referencing other tables, accessing them actually loads the linked data from the database which can be very expensive if there are a lot of findings.

nnobelis: Ah yes, property delegation from Exposed. Got it !

workers/scanner/src/main/kotlin/scanner/OrtServerScanResultStorage.kt
oheger-bosch: I left this intentionally a bit vague. Currently the issues are still loaded, but I hope that a further optimization could prevent this.

nnobelis (approved):

PR #1870, opened 2025-01-24
feat(core): Add more duration-related metrics

bs-ondem:


scripts/compose/grafana/provisioning/dashboards/ort-run-and-job-duration.json
```diff
@@ -0,0 +1,254 @@
...
+    "from": "now-6h",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "browser",
+  "title": "ORT Run & Job Duration",
+  "uid": "feax6cgendypse",
+  "version": 11,
+  "weekStart": ""
+}
```
mnonnenmacher, line 612: Please add the missing line break.

core/src/main/kotlin/utils/JobMetrics.kt
```diff
@@ -192,6 +251,21 @@ class JobMetrics(private val application: Application) : MeterBinder {
...
+                }.andWhere {
+                    AnalyzerJobsTable.finishedAt.isNull()
+                }.andWhere {
+                    AnalyzerJobsTable.startedAt greaterEq timestamp
+            }.mapNotNull {
+                val createdAt = it[AnalyzerJobsTable.createdAt]
+                val startedAt = it[AnalyzerJobsTable.startedAt] ?: return@mapNotNull null
+                startedAt - createdAt
+            }
+        }
```
mnonnenmacher, line 267: Should we merge this with the function above to reduce the amount of database queries? I'm a bit undecided because it would also delay the reporting of queue times, what do you think?

mnonnenmacher (changes requested):

core/src/main/kotlin/utils/JobMetrics.kt
bs-ondem: Do you want to merge the functions `getAnalyzerJobDurations` and `getAnalyzerJobQueueDurations`?

mnonnenmacher: That was the idea at least.

bs-ondem: It makes sense to use a single query, but I'm not sure how it would look, since the `WHERE` clauses are completely different.

mnonnenmacher: I would just use `AnalyzerJobsTable.finishedAt greaterEq timestamp` and report the queue time once the jobs have finished. That's what I meant with "it would also delay the reporting of queue times".

mnonnenmacher: But it was just an idea, I would be fine with also keeping it the way it is. We should just verify that the queries use indexes, then the performance should not be an issue.

bs-ondem: I would keep it as it is to avoid the delay of reporting the queue durations.

mnonnenmacher: I just checked and we are not having indexes on the columns used in these queries. I think the easiest would be to add a composite index for the status and the timestamp columns (status should go first because the index can then also be used for queries filtering only by status which should also speed up the status related metrics).

Please do some testing to verify that PostgreSQL would use the index for all queries made in the duration related metrics and add the required migration.

mnonnenmacher (changes requested):

bs-ondem:
@mnonnenmacher  I ran the queries without any indexes to observe the execution plans. I used the `EXPLAIN ANALYZE` command to get detailed insights into how PostgreSQL executed the queries. 

Here are the results from the `EXPLAIN ANALYZE` for the queries of `AnalyzerJobs` **before** adding the indexes:
```
ort_server=# EXPLAIN ANALYZE
SELECT ort_server.analyzer_jobs.started_at, ort_server.analyzer_jobs.finished_at
FROM ort_server.analyzer_jobs
WHERE ((ort_server.analyzer_jobs.status = 'FINISHED') OR (ort_server.analyzer_jobs.status = 'FINISHED_WITH_ISSUES'))
  AND (ort_server.analyzer_jobs.started_at IS NOT NULL)
  AND (ort_server.analyzer_jobs.finished_at IS NOT NULL)
  AND (ort_server.analyzer_jobs.finished_at >= '2025-02-04T09:34:29.58225117');
                                                                                                           QUERY PLAN                                                                                                           
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Seq Scan on analyzer_jobs  (cost=0.00..7.56 rows=1 width=16) (actual time=0.403..0.431 rows=0 loops=1)
   Filter: ((started_at IS NOT NULL) AND (finished_at IS NOT NULL) AND (finished_at >= '2025-02-04 09:34:29.582251'::timestamp without time zone) AND ((status = 'FINISHED'::text) OR (status = 'FINISHED_WITH_ISSUES'::text)))
   Rows Removed by Filter: 85
 Planning Time: 20.010 ms
 Execution Time: 2.013 ms
(5 rows)

Time: 46.527 ms
```
**After** adding indexes:
```
SELECT ort_server.analyzer_jobs.started_at, ort_server.analyzer_jobs.finished_at
FROM ort_server.analyzer_jobs
WHERE ((ort_server.analyzer_jobs.status = 'FINISHED') OR (ort_server.analyzer_jobs.status = 'FINISHED_WITH_ISSUES'))
  AND (ort_server.analyzer_jobs.started_at IS NOT NULL)
  AND (ort_server.analyzer_jobs.finished_at IS NOT NULL)
  AND (ort_server.analyzer_jobs.finished_at >= '2025-02-04T09:34:29.58225117');
SET
Time: 0.880 ms
                                                                      QUERY PLAN                                                                       
-------------------------------------------------------------------------------------------------------------------------------------------------------
 Index Only Scan using analyzer_jobs_durations on analyzer_jobs  (cost=0.14..13.23 rows=1 width=16) (actual time=0.523..0.526 rows=1 loops=1)
   Index Cond: ((started_at IS NOT NULL) AND (finished_at IS NOT NULL) AND (finished_at >= '2025-02-04 09:34:29.582251'::timestamp without time zone))
   Filter: ((status = 'FINISHED'::text) OR (status = 'FINISHED_WITH_ISSUES'::text))
   Heap Fetches: 1
 Planning Time: 1.518 ms
 Execution Time: 0.745 ms
(6 rows)

Time: 3.265 ms
```

I did the same for the `OrtRuns` table:
```
ort_server=# EXPLAIN ANALYZE
SELECT ort_server.ort_runs.created_at, ort_server.ort_runs.finished_at
FROM ort_server.ort_runs
WHERE ((ort_server.ort_runs.status = 'FINISHED') OR (ort_server.ort_runs.status = 'FINISHED_WITH_ISSUES'))
  AND (ort_server.ort_runs.finished_at IS NOT NULL)
  AND (ort_server.ort_runs.finished_at >= '2025-02-04T09:47:00.387446795');
                                                                                            QUERY PLAN                                                                                             
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Seq Scan on ort_runs  (cost=0.00..30.40 rows=1 width=16) (actual time=0.257..0.281 rows=1 loops=1)
   Filter: ((finished_at IS NOT NULL) AND (finished_at >= '2025-02-04 09:47:00.387447'::timestamp without time zone) AND ((status = 'FINISHED'::text) OR (status = 'FINISHED_WITH_ISSUES'::text)))
   Rows Removed by Filter: 91
 Planning Time: 0.695 ms
 Execution Time: 0.313 ms
(5 rows)

Time: 2.509 ms
```
After adding indexes:
```
ort_server=# EXPLAIN ANALYZE
SELECT ort_server.ort_runs.created_at, ort_server.ort_runs.finished_at
FROM ort_server.ort_runs
WHERE ((ort_server.ort_runs.status = 'FINISHED') OR (ort_server.ort_runs.status = 'FINISHED_WITH_ISSUES'))
  AND (ort_server.ort_runs.finished_at IS NOT NULL)
  AND (ort_server.ort_runs.finished_at >= '2025-02-04T09:47:00.387446795');
                                                                                                                                                       QUERY PLAN                                                                                                                                                       
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on ort_runs  (cost=8.31..12.33 rows=1 width=16) (actual time=0.030..0.032 rows=1 loops=1)
   Recheck Cond: (((status = 'FINISHED'::text) AND (finished_at IS NOT NULL) AND (finished_at >= '2025-02-04 09:47:00.387447'::timestamp without time zone)) OR ((status = 'FINISHED_WITH_ISSUES'::text) AND (finished_at IS NOT NULL) AND (finished_at >= '2025-02-04 09:47:00.387447'::timestamp without time zone)))
   Heap Blocks: exact=1
   ->  BitmapOr  (cost=8.31..8.31 rows=1 width=0) (actual time=0.021..0.022 rows=0 loops=1)
         ->  Bitmap Index Scan on ort_runs_durations  (cost=0.00..4.16 rows=1 width=0) (actual time=0.017..0.017 rows=1 loops=1)
               Index Cond: ((status = 'FINISHED'::text) AND (finished_at IS NOT NULL) AND (finished_at >= '2025-02-04 09:47:00.387447'::timestamp without time zone))
         ->  Bitmap Index Scan on ort_runs_durations  (cost=0.00..4.16 rows=1 width=0) (actual time=0.004..0.004 rows=0 loops=1)
               Index Cond: ((status = 'FINISHED_WITH_ISSUES'::text) AND (finished_at IS NOT NULL) AND (finished_at >= '2025-02-04 09:47:00.387447'::timestamp without time zone))
 Planning Time: 0.159 ms
 Execution Time: 0.059 ms
(10 rows)

Time: 0.598 ms
```

mnonnenmacher (approved):

PR #1873, opened 2025-01-24
Minor CLI improvements

mnonnenmacher:

### DIFF
--- a/cli/src/main/kotlin/DownloadCommand.kt
+++ b/cli/src/main/kotlin/DownloadCommand.kt
@@ -20,10 +20,13 @@
 package org.eclipse.apoapsis.ortserver.cli
 
 import com.github.ajalt.clikt.command.SuspendingNoOpCliktCommand
+import com.github.ajalt.clikt.core.Context
 import com.github.ajalt.clikt.core.subcommands
 
 class DownloadCommand(config: OrtServerOptions) : SuspendingNoOpCliktCommand(name = "download") {
     init {
         subcommands(LogsCommand(config), ReportsCommand(config))
     }
+
+    override fun help(context: Context) = "Commands to download files for a run."
 }
--- a/cli/src/main/kotlin/OrtServerMain.kt
+++ b/cli/src/main/kotlin/OrtServerMain.kt
@@ -56,7 +56,7 @@ class OrtServerMain : SuspendingNoOpCliktCommand(COMMAND_NAME) {
             names = setOf("--version", "-v"),
             help = "Show the version and exit.",
             // TODO: Also add the server version given by `baseUrl` to this output.
-            message = { "$commandName CLI version $it" }
+            message = { "$commandName version $it" }
         )
     }
 
--- a/cli/src/main/kotlin/StartCommand.kt
+++ b/cli/src/main/kotlin/StartCommand.kt
@@ -20,6 +20,7 @@
 package org.eclipse.apoapsis.ortserver.cli
 
 import com.github.ajalt.clikt.command.SuspendingCliktCommand
+import com.github.ajalt.clikt.core.Context
 import com.github.ajalt.clikt.parameters.groups.mutuallyExclusiveOptions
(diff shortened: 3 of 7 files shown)


cli/src/main/kotlin/OrtServerMain.kt
```diff
@@ -56,7 +56,7 @@ class OrtServerMain : SuspendingNoOpCliktCommand(COMMAND_NAME) {
             names = setOf("--version", "-v"),
             help = "Show the version and exit.",
             // TODO: Also add the server version given by `baseUrl` to this output.
-            message = { "$commandName CLI version $it" }
+            message = { "$commandName version $it" }
```
sschuberth, line 59: Commit message: But "CLI" does not stand for "Client" 😉 So while there IMO is no "redundancy", I still agree that is reads better without "CLI".

cli/src/main/kotlin/DownloadCommand.kt
```diff
@@ -20,10 +20,13 @@
...
 import com.github.ajalt.clikt.command.SuspendingNoOpCliktCommand
+import com.github.ajalt.clikt.core.Context
 import com.github.ajalt.clikt.core.subcommands
 
 class DownloadCommand(config: OrtServerOptions) : SuspendingNoOpCliktCommand(name = "download") {
     init {
         subcommands(LogsCommand(config), ReportsCommand(config))
     }
+
+    override fun help(context: Context) = "Commands to download files for an ORT run."
```
sschuberth, line 31: That sounds a bit like "a run *on* ORT". Why not just "a run"?

sschuberth: Similar below.

cli/src/main/kotlin/OrtServerMain.kt
mnonnenmacher: It does:
*  https://github.com/eclipse-apoapsis/ort-server/discussions/1794#discussioncomment-11830535
* https://github.com/eclipse-apoapsis/ort-server/blob/efadae752d97db1b23256bb6ca52e5b293b87154/cli/src/main/kotlin/OrtServerMain.kt#L64

sschuberth: No 😉 "OSC" stands for "ORT Server Client" and "CLI" stands for "Command Line Interface", but neither does "CLI" stand for "client", nor does the "OSC" acronym in any way refer to "client" literally.

cli/src/main/kotlin/StartCommand.kt
```diff
@@ -59,12 +59,14 @@ class StartCommand(private val config: OrtServerOptions) : SuspendingCliktComman
         option(
             "--parameters-file",
             envvar = "ORT_RUNS_START_PARAMETERS_FILE",
-            help = "The path to the Create ORT run configuration file!"
+            help = "The path to a JSON file containing the ORT run configuration " +
```
sschuberth, line 62: Now with the previous commit dropping "ORT" from "ORT run", consider doing the same here, too.

cli/src/main/kotlin/StartCommand.kt
```diff
@@ -59,12 +59,14 @@ class StartCommand(private val config: OrtServerOptions) : SuspendingCliktComman
...
+                    "(see https://eclipse-apoapsis.github.io/ort-server/api/post-ort-run)."
         ).inputStream().convert { it.bufferedReader().readText() },
         option(
             "--parameters",
             envvar = "ORT_RUNS_START_PARAMETERS",
-            help = "The Create ORT run configuration as a string."
+            help = "The ORT run configuration as a JSON string " +
```
sschuberth, line 68: See above.

cli/src/main/kotlin/OrtServerMain.kt
mnonnenmacher: I think I misunderstood your first comment, I'm using a different rationale now.

sschuberth (approved):

PR #1882, opened 2025-01-27
Extend the packages endpoint with the shortest dependency path

lamppu:
Add information about shortest dependency paths to the packages endpoint response. This will give a quick overview for how deeply nested a dependency is in the project. For simplicity, only one path for a package for each project is returned (only the shortest path across scopes is picked).

Screenshot from 2025-02-17 15-10-53

### DIFF
--- a/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
+++ b/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
@@ -75,6 +75,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.ScannerJob as ApiScannerJob
 import org.eclipse.apoapsis.ortserver.api.v1.model.ScannerJobConfiguration as ApiScannerJobConfiguration
 import org.eclipse.apoapsis.ortserver.api.v1.model.Secret as ApiSecret
 import org.eclipse.apoapsis.ortserver.api.v1.model.Severity as ApiSeverity
+import org.eclipse.apoapsis.ortserver.api.v1.model.ShortestDependencyPath as ApiShortestDependencyPath
 import org.eclipse.apoapsis.ortserver.api.v1.model.SortDirection as ApiSortDirection
 import org.eclipse.apoapsis.ortserver.api.v1.model.SortProperty as ApiSortProperty
 import org.eclipse.apoapsis.ortserver.api.v1.model.SourceCodeOrigin as ApiSourceCodeOrigin
@@ -131,10 +132,11 @@ import org.eclipse.apoapsis.ortserver.model.VulnerabilityWithIdentifier
 import org.eclipse.apoapsis.ortserver.model.runs.Identifier
 import org.eclipse.apoapsis.ortserver.model.runs.Issue
 import org.eclipse.apoapsis.ortserver.model.runs.OrtRuleViolation
-import org.eclipse.apoapsis.ortserver.model.runs.Package
 import org.eclipse.apoapsis.ortserver.model.runs.PackageManagerConfiguration
+import org.eclipse.apoapsis.ortserver.model.runs.PackageWithShortestDependencyPaths
 import org.eclipse.apoapsis.ortserver.model.runs.ProcessedDeclaredLicense
 import org.eclipse.apoapsis.ortserver.model.runs.RemoteArtifact
+import org.eclipse.apoapsis.ortserver.model.runs.ShortestDependencyPath
 import org.eclipse.apoapsis.ortserver.model.runs.VcsInfo
 import org.eclipse.apoapsis.ortserver.model.runs.advisor.Vulnerability
 import org.eclipse.apoapsis.ortserver.model.runs.advisor.VulnerabilityReference
@@ -533,24 +535,6 @@ fun ApiIdentifier.mapToModel() = Identifier(type, namespace, name, version)
 
 fun VulnerabilityReference.mapToApi() = ApiVulnerabilityReference(url, scoringSystem, severity, score, vector)
 
-fun Package.mapToApi() =
-    ApiPackage(
-        identifier.mapToApi(),
-        purl,
-        cpe,
-        authors,
-        declaredLicenses,
-        processedDeclaredLicense.mapToApi(),
-        description,
-        homepageUrl,
-        binaryArtifact.mapToApi(),
-        sourceArtifact.mapToApi(),
-        vcs.mapToApi(),
-        vcsProcessed.mapToApi(),
-        isMetadataOnly,
-        isModified
-    )
-
 fun ProcessedDeclaredLicense.mapToApi() =
     ApiProcessedDeclaredLicense(
         spdxExpression,
@@ -789,3 +773,27 @@ fun ApiSubmoduleFetchStrategy.mapToModel() = when (this) {
     ApiSubmoduleFetchStrategy.TOP_LEVEL_ONLY -> SubmoduleFetchStrategy.TOP_LEVEL_ONLY
     ApiSubmoduleFetchStrategy.FULLY_RECURSIVE -> SubmoduleFetchStrategy.FULLY_RECURSIVE
 }
+
+fun ShortestDependencyPath.mapToApi() = ApiShortestDependencyPath(
+    scope = scope,
+    projectIdentifier = projectIdentifier.mapToApi(),
+    path = path.map { it.mapToApi() }
+)
+
+fun PackageWithShortestDependencyPaths.mapToApi() = ApiPackage(
+    pkg.identifier.mapToApi(),
+    pkg.purl,
+    pkg.cpe,
+    pkg.authors,
+    pkg.declaredLicenses,
+    pkg.processedDeclaredLicense.mapToApi(),
+    pkg.description,
+    pkg.homepageUrl,
+    pkg.binaryArtifact.mapToApi(),
+    pkg.sourceArtifact.mapToApi(),
+    pkg.vcs.mapToApi(),
+    pkg.vcsProcessed.mapToApi(),
+    pkg.isMetadataOnly,
+    pkg.isModified,
+    shortestDependencyPaths.map { it.mapToApi() }
+)
--- a/api/v1/model/src/commonMain/kotlin/ShortestDependencyPath.kt
+++ b/api/v1/model/src/commonMain/kotlin/ShortestDependencyPath.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,21 +21,15 @@ package org.eclipse.apoapsis.ortserver.api.v1.model
 
 import kotlinx.serialization.Serializable
 
+/** The shortest dependency path for a Package. */
 @Serializable
-data class Package(
-    val identifier: Identifier,
-    val purl: String,
-    val cpe: String? = null,
-    val authors: Set<String>,
-    val declaredLicenses: Set<String>,
-    val processedDeclaredLicense: ProcessedDeclaredLicense,
-    val description: String,
-    val homepageUrl: String,
-    val binaryArtifact: RemoteArtifact,
-    val sourceArtifact: RemoteArtifact,
-    val vcs: VcsInfo,
-    val vcsProcessed: VcsInfo,
-    val isMetadataOnly: Boolean = false,
-    val isModified: Boolean = false,
-    val shortestDependencyPaths: List<ShortestDependencyPath>
+data class ShortestDependencyPath(
+    /** The identifier of the root project of this path. */
+    val projectIdentifier: Identifier,
+
+    /** The scope in which this shortest path of the dependency is found in. */
+    val scope: String,
+
+    /** Path of dependency identifiers to the dependency. */
+    val path: List<Identifier>
 )
--- a/core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
+++ b/core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
@@ -26,6 +26,7 @@ import io.kotest.assertions.ktor.client.haveHeader
 import io.kotest.assertions.ktor.client.shouldHaveStatus
 import io.kotest.engine.spec.tempdir
 import io.kotest.inspectors.forAll
+import io.kotest.matchers.collections.shouldBeSingleton
 import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
 import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.file.aFile
@@ -112,6 +113,7 @@ import org.eclipse.apoapsis.ortserver.model.runs.OrtRuleViolation
 import org.eclipse.apoapsis.ortserver.model.runs.Package
 import org.eclipse.apoapsis.ortserver.model.runs.ProcessedDeclaredLicense
 import org.eclipse.apoapsis.ortserver.model.runs.RemoteArtifact
+import org.eclipse.apoapsis.ortserver.model.runs.ShortestDependencyPath
 import org.eclipse.apoapsis.ortserver.model.runs.VcsInfo
 import org.eclipse.apoapsis.ortserver.model.runs.advisor.AdvisorConfiguration
 import org.eclipse.apoapsis.ortserver.model.runs.advisor.AdvisorResult
@@ -906,6 +908,10 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
                     configuration = AnalyzerJobConfiguration()
                 )
 
+                val project = dbExtension.fixtures.getProject()
+                val identifier1 = Identifier("Maven", "com.example", "example", "1.0")
+                val identifier2 = Identifier("Maven", "com.example", "example2", "1.0")
+
                 dbExtension.fixtures.analyzerRunRepository.create(
                     analyzerJobId = analyzerJob.id,
                     startTime = Clock.System.now().toDatabasePrecision(),
@@ -926,10 +932,10 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
                         packageManagers = emptyMap(),
                         skipExcluded = true
                     ),
-                    projects = emptySet(),
+                    projects = setOf(project),
                     packages = setOf(
                         Package(
-                            Identifier("Maven", "com.example", "example", "1.0"),
+                            identifier1,
                             purl = "pkg:maven/com.example/example@1.0",
                             cpe = null,
                             authors = setOf("Author One", "Author Two"),
@@ -970,12 +976,7 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
                             isModified = false
                         ),
                         Package(
-                            Identifier(
-                                type = "Maven",
-                                namespace = "com.example",
-                                name = "example2",
-                                version = "1.0"
-                            ),
+                            identifier2,
                             purl = "pkg:maven/com.example/example2@1.0",
                             cpe = null,
                             authors = emptySet(),
@@ -1014,7 +1015,23 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
                         )
                     ),
                     issues = emptyList(),
-                    dependencyGraphs = emptyMap()
+                    dependencyGraphs = emptyMap(),
+                    shortestDependencyPaths = mapOf(
+                        identifier1 to listOf(
+                            ShortestDependencyPath(
+                                project.identifier,
+                                "compileClassPath",
+                                emptyList()
+                            )
+                        ),
+                        identifier2 to listOf(
+                            ShortestDependencyPath(
+                                project.identifier,
+                                "compileClassPath",
+                                listOf(identifier1)
+                            )
+                        )
+                    )
                 )
 
                 val response = superuserClient.get("/api/v1/runs/${ortRun.id}/packages")
@@ -1024,11 +1041,26 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
 
                 with(packages.data) {
                     shouldHaveSize(2)
-                    first().identifier.name shouldBe "example"
-                    first().authors shouldHaveSize 2
-                    first().declaredLicenses shouldHaveSize 3
-                    first().processedDeclaredLicense.mappedLicenses shouldHaveSize 2
-                    first().processedDeclaredLicense.unmappedLicenses shouldHaveSize 4
+
+                    with(first()) {
+                        identifier.name shouldBe "example"
+                        authors shouldHaveSize 2
+                        declaredLicenses shouldHaveSize 3
+                        processedDeclaredLicense.mappedLicenses shouldHaveSize 2
+                        processedDeclaredLicense.unmappedLicenses shouldHaveSize 4
+
+                        shortestDependencyPaths.shouldBeSingleton {
+                            it.projectIdentifier shouldBe project.identifier.mapToApi()
+                            it.scope shouldBe "compileClassPath"
+                            it.path shouldBe emptyList()
+                        }
+                    }
+
+                    last().shortestDependencyPaths.shouldBeSingleton {
+                        it.projectIdentifier shouldBe project.identifier.mapToApi()
+                        it.scope shouldBe "compileClassPath"
+                        it.path shouldBe listOf(identifier1.mapToApi())
+                    }
                     last().identifier.name shouldBe "example2"
                 }
             }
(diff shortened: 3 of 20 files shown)


workers/analyzer/src/test/kotlin/UtilsTest.kt
```diff
@@ -0,0 +1,111 @@
...
+
+import org.eclipse.apoapsis.ortserver.model.runs.Identifier
+
+import org.ossreviewtoolkit.model.Identifier as OrtIdentifier
+import org.ossreviewtoolkit.model.Project as OrtProject
+import org.ossreviewtoolkit.model.VcsInfo as OrtVcsInfo
+import org.ossreviewtoolkit.model.VcsType
+
+class UtilsTest : WordSpec({
+    "getIdentifierToShortestPathMap" should {
```
sschuberth, line 34: Nit: We tend to use "()" after the function name to denote it's a function that's being tested.

services/hierarchy/src/main/kotlin/PackageService.kt
```diff
@@ -78,7 +81,15 @@ class PackageService(private val db: Database) {
         }
 }
 
-private fun ResultRow.toPackage(): Package = PackageDao.wrapRow(this).mapToModel()
+private fun ResultRow.toPackageWithShortestDependencyPath(): PackageWithShortestDependencyPath =
+    PackageWithShortestDependencyPath(
+        pkg = PackageDao.wrapRow(this).mapToModel(),
+        shortestDependencyPath = if (this.getOrNull(ShortestDependencyPathsTable.id) !== null) {
```
sschuberth, line 87: Drop the explicit `this.´, and stick to just `!=`.

services/hierarchy/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -196,7 +243,7 @@ class PackageServiceTest : WordSpec() {
 
                 service.countForOrtRunIds(ortRun1Id, ortRun2Id) shouldBe 3
             }
-    }
+        }
```
sschuberth, line 246: Unintended change?

lamppu: Yes, though it seems the indentation was wrong before. Should this be fixed in a separate commit?

sschuberth: If indentation was really wrong before, yes, please fix this is a separate commit to split functional from non-functional changes.

services/hierarchy/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -196,7 +196,7 @@ class PackageServiceTest : WordSpec() {
...
```
sschuberth, line 246: Nit: Not for this time, but maybe the next: For single-file changes you may use the name of the file as the scope of the change in the commit title, like `chore(PackageServiceTest): ...`.

api/v1/model/src/commonMain/kotlin/Package.kt
```diff
@@ -36,5 +36,6 @@ data class Package(
     val vcs: VcsInfo,
     val vcsProcessed: VcsInfo,
     val isMetadataOnly: Boolean = false,
-    val isModified: Boolean = false
+    val isModified: Boolean = false,
+    val shortestDependencyPath: ShortestDependencyPath? = null,
```
sschuberth, line 40: There's still a trailing comma 😉 We really should have a rule to disallow that, but unfortunately [detekt / ktlint only have rules to *enforce* the usage](https://detekt.dev/docs/rules/formatting/#trailingcommaoncallsite) 🙄.

sschuberth (commented):
Basically LGTM, but I'd also like @mnonnenmacher to have a look.

api/v1/model/src/commonMain/kotlin/Package.kt
Etsija: @sschuberth you mean the trailing comma after `shortestDependencyPath`? What's even more confusing for us TypeScript people is that in the UI code, contrary to (seems like) back-end code, trailing commas are _used_ and actually `prettier` puts them there automatically... :smile:

sschuberth: > @sschuberth you mean the trailing comma after `shortestDependencyPath`?

Yes.

> contrary to (seems like) back-end code, trailing commas are used and actually prettier puts them there automatically... 😄

Well, that's probably because we've configured Prettier to do so. We *could* do the same for Detekt, and align formatting in that regard. But I'd say it's not unusual for front-end and back-end code to use different formatting styles, as these are mostly also using different languages.

So maybe that's a good opportunity to ask @eclipse-apoapsis/technology-apoapsis-committers: Should be enforce trailing commas at the declaration and / or caller sites for Kotlin code?

Etsija: Trailing commas is a [well known convention](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas) in JavaScript land, and is actually pretty helpful.

sschuberth: Let's continue that specific discussion [elsewhere](https://github.com/eclipse-apoapsis/ort-server/discussions/1913).

sschuberth:
Looks like there are test failures now:

```
 org.eclipse.apoapsis.ortserver.dao.UniqueConstrainsViolationTest > An exception about a unique constraint violation is handled by repeating the transaction FAILED
    io.kotest.engine.extensions.ExtensionException$BeforeEachException: org.flywaydb.core.api.FlywayException: Found more than one migration with version 91
    Offenders:
    -> /home/runner/work/ort-server/ort-server/dao/file:/home/runner/work/ort-server/ort-server/dao/build/libs/dao-0.0.1-RC.001.sha.5af34b7.jar!/db/migration/V91__scanSummaryHash.sql (SQL)
    -> /home/runner/work/ort-server/ort-server/dao/file:/home/runner/work/ort-server/ort-server/dao/build/libs/dao-0.0.1-RC.001.sha.5af34b7.jar!/db/migration/V91__addShortestDependencyPathsTable.sql (SQL)
        at app//io.kotest.engine.test.TestExtensions$beforeTestBeforeAnyBeforeContainer$errors$2$2.invoke(TestExtensions.kt:81)
        at app//io.kotest.engine.test.TestExtensions$beforeTestBeforeAnyBeforeContainer$errors$2$2.invoke(TestExtensions.kt:81)
```

lamppu:
> Looks like there are test failures now:
> 
> ```
>  org.eclipse.apoapsis.ortserver.dao.UniqueConstrainsViolationTest > An exception about a unique constraint violation is handled by repeating the transaction FAILED
>     io.kotest.engine.extensions.ExtensionException$BeforeEachException: org.flywaydb.core.api.FlywayException: Found more than one migration with version 91
>     Offenders:
>     -> /home/runner/work/ort-server/ort-server/dao/file:/home/runner/work/ort-server/ort-server/dao/build/libs/dao-0.0.1-RC.001.sha.5af34b7.jar!/db/migration/V91__scanSummaryHash.sql (SQL)
>     -> /home/runner/work/ort-server/ort-server/dao/file:/home/runner/work/ort-server/ort-server/dao/build/libs/dao-0.0.1-RC.001.sha.5af34b7.jar!/db/migration/V91__addShortestDependencyPathsTable.sql (SQL)
>         at app//io.kotest.engine.test.TestExtensions$beforeTestBeforeAnyBeforeContainer$errors$2$2.invoke(TestExtensions.kt:81)
>         at app//io.kotest.engine.test.TestExtensions$beforeTestBeforeAnyBeforeContainer$errors$2$2.invoke(TestExtensions.kt:81)
> ```

Yes, I needed to rename the migration file, because a different migration was merged in between.

services/hierarchy/src/test/kotlin/PackageServiceTest.kt
mnonnenmacher: I'd rather keep the scopes limited to module names: https://github.com/eclipse-apoapsis/ort-server/blob/main/CONTRIBUTING.md#scopes

model/src/commonMain/kotlin/runs/PackageWithShortestDependencyPath.kt
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * A data class representing a package, and the shortest dependency path that the package is found in (relative to a
+ * project found in a run).
+ */
+data class PackageWithShortestDependencyPath(
+    /** A package. */
+    val pkg: Package,
+
+    /** The shortest dependency path for the package. */
+    val shortestDependencyPath: ShortestDependencyPath? = null
```
mnonnenmacher, line 31: I don't have a full overview of the changes in this PR, but previously there was code that got the shortest path by scope, and now only one shortest path is here. Why is that?

mnonnenmacher (changes requested):

model/src/commonMain/kotlin/runs/PackageWithShortestDependencyPath.kt
Etsija: IIRC it was a team decision to at this point only show one shortest dependency path, instead of shortest path for every scope.

sschuberth: Yes, that's mostly out of UI considerations: Esp. in Android projects, there could be a lot of scopes, mostly with duplicate dependencies and probably the same shortest paths. So we decided to (for now) only show "the shortest of the shortest paths" from an arbitrary scope (but name the scope).

lamppu: Just to make sure I haven't misunderstood and oversimplified this, the way this works now, it will save just one shortest path for each individual package identifier, across all scopes, but also across projects. Should it be returning one per project instead (if it is found in more than one project, that is)?

sschuberth: > Should it be returning one per project instead (if it is found in more than one project, that is)?

Yes. The root of each dependency path should be the respective project, which means storing the shortest path *across* projects would be wrong. So the path hierarchy is:


`root project (which we do not display in the packages table, but that's the logic)` -> `direct dependency of the project` -> `transitive dependency along the path` -> ... -> `transitive dependency along the path` -> `leaf dependency (that has no further dependencies)`

sschuberth:
@lamppu, for a quick overview of the feature, could you maybe add a screenshot to the top post?

Etsija:
I'm already preparing the UI for this, so will then add a relevant screenshot to my PR as well. As with this feature, the packages table starts to be something useful to users, I'm also adding sorting and filtering to the table.

sschuberth:
> @lamppu, for a quick overview of the feature, could you maybe add a screenshot to the top post?

My bad, I totally forgot that the front-end UI is not implemented yet... that's what I was referring to as a screenshot 😬 No need to add screenshots for code 😅

lamppu:
> I'm also adding sorting and filtering to the table.

I think this is pending backend sorting and filtering, because the API request will be very slow if it's done client side.

Etsija:
> > I'm also adding sorting and filtering to the table.
> 
> I think this is pending backend sorting and filtering, because the API request will be very slow if it's done client side.

Do you mean with the dependency paths? Because  I didn't experience any particular slowness with ~2000 packages. TanStack Table is supposed to do client-side manipulation quite efficiently.

model/src/commonMain/kotlin/runs/ShortestDependencyPath.kt
```diff
@@ -0,0 +1,32 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.model.runs
+
+/** The shortest dependency path for a Package. */
+data class ShortestDependencyPath(
+    /** The identifier of the root project of this path. */
+    val projectIdentifier: Identifier,
```
sschuberth, line 25: Nit: Maybe just `projectId`?

dao/src/main/kotlin/repositories/analyzerrun/DaoAnalyzerRunRepository.kt
```diff
@@ -76,8 +79,23 @@ class DaoAnalyzerRunRepository(private val db: Database) : AnalyzerRunRepository
 
             createAnalyzerConfiguration(analyzerRun, config)
 
-            projects.forEach { createProject(analyzerRun, it) }
-            packages.forEach { createPackage(analyzerRun, it) }
+            val projectsMap = projects.associateBy({ it.identifier }) { createProject(analyzerRun, it) }
```
sschuberth, line 82: Nit: Personally, I find the following syntax a bit more readable

    associate { it.identifier to createProject(analyzerRun, it) }

dao/src/main/kotlin/repositories/analyzerrun/DaoAnalyzerRunRepository.kt
```diff
@@ -76,8 +79,23 @@ class DaoAnalyzerRunRepository(private val db: Database) : AnalyzerRunRepository
...
+
+            packages.forEach {
+                val pkgDao = createPackage(analyzerRun, it)
+
+                shortestDependencyPaths[it.identifier]?.onEach { shortestDependencyPath ->
```
sschuberth, line 87: Nit: Prefer `forEach` over `onEach` if the return value is unused (like here).

dao/src/main/kotlin/repositories/analyzerrun/DaoAnalyzerRunRepository.kt
```diff
@@ -76,8 +79,23 @@ class DaoAnalyzerRunRepository(private val db: Database) : AnalyzerRunRepository
...
+                    projectsMap[shortestDependencyPath.projectIdentifier]?.also { projectDao ->
+                        ShortestDependencyPathDao.new {
+                            this.pkg = pkgDao
+                            this.analyzerRun = analyzerRun
+                            this.project = projectDao
+                            this.scope = shortestDependencyPath.scope
+                            this.path = shortestDependencyPath.path
+                        }
+                    }
```
sschuberth, line 96: Just double-checking: This creates a new `ShortestDependencyPathDao`, but does not seem to store it anywhere. Is this really correct?

workers/analyzer/src/main/kotlin/analyzer/AnalyzerWorker.kt
```diff
@@ -97,9 +99,20 @@ internal class AnalyzerWorker(
...
         )
 
+        val shortestPathsByIdentifier = mutableMapOf<Identifier, MutableList<ShortestDependencyPath>>()
+
+        analyzerRun.result.projects.forEach { project ->
+            getIdentifierToShortestPathsMap(
+                project.id.mapToModel(),
+                ortResult.dependencyNavigator.getShortestPaths(project)
+            ).forEach { (identifier, path) ->
+                shortestPathsByIdentifier.getOrPut(identifier) { mutableListOf() }.add(path)
```
sschuberth, line 109: Nit: Prefer the `+=` operator over calling `.add()` here.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
```diff
@@ -791,3 +775,27 @@ fun ApiSubmoduleFetchStrategy.mapToModel() = when (this) {
...
+    pkg.processedDeclaredLicense.mapToApi(),
+    pkg.description,
+    pkg.homepageUrl,
+    pkg.binaryArtifact.mapToApi(),
+    pkg.sourceArtifact.mapToApi(),
+    pkg.vcs.mapToApi(),
+    pkg.vcsProcessed.mapToApi(),
+    pkg.isMetadataOnly,
+    pkg.isModified,
+    shortestDependencyPaths?.map { it.mapToApi() } ?: emptyList()
```
sschuberth, line 800: Nit: Prefer `.orEmpty()` over a fallback with the `?:` operator.

api/v1/model/src/commonMain/kotlin/Package.kt
```diff
@@ -36,5 +36,6 @@ data class Package(
     val vcs: VcsInfo,
     val vcsProcessed: VcsInfo,
     val isMetadataOnly: Boolean = false,
-    val isModified: Boolean = false
+    val isModified: Boolean = false,
+    val shortestDependencyPaths: List<ShortestDependencyPath>? = null
```
sschuberth, line 40: Just asking: What's the semantic difference between `shortestDependencyPaths` being `null` vs empty? Can both happen? Would it make sense to explain the difference in a code comment for the property?

api/v1/model/src/commonMain/kotlin/ShortestDependencyPath.kt
```diff
@@ -0,0 +1,35 @@
...
+
+package org.eclipse.apoapsis.ortserver.api.v1.model
+
+import kotlinx.serialization.Serializable
+
+/** The shortest dependency path for a Package. */
+@Serializable
+data class ShortestDependencyPath(
+    /** The identifier of the root project of this path. */
+    val projectIdentifier: Identifier,
```
sschuberth, line 28: Also here, maybe just `projectId`.

core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
```diff
@@ -1024,11 +1040,34 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
...
+                        identifier.name shouldBe "example"
+                        authors shouldHaveSize 2
+                        declaredLicenses shouldHaveSize 3
+                        processedDeclaredLicense.mappedLicenses shouldHaveSize 2
+                        processedDeclaredLicense.unmappedLicenses shouldHaveSize 4
+
+                        shortestDependencyPaths shouldNotBeNull {
+                            shouldHaveSize(1)
+
+                            with(first()) {
```
sschuberth, line 1054: `shouldHaveSize(1)` / `first()` calls can be combined into `shouldBeSingleton { ... }`. Similar below.

model/src/commonMain/kotlin/runs/PackageWithShortestDependencyPaths.kt
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * A data class representing a package, and the shortest dependency path that the package is found in (relative to a
+ * project found in a run).
+ */
+data class PackageWithShortestDependencyPaths(
+    /** A package. */
+    val pkg: Package,
+
+    /** Package ID */
+    val pkgId: Long,
```
sschuberth, line 31: This is not the same as a `Package`'s `Identifier`, so would it make sense to use a different property name to avoid confusion (but I do not have a good idea for any)?

model/src/commonMain/kotlin/runs/PackageWithShortestDependencyPaths.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    /** The shortest dependency path for the package. */
```
sschuberth, line 33: Esp. here where you already have a comment, please explain the semantic difference between `null` and empty.

services/hierarchy/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -100,7 +98,7 @@ class PackageServiceTest : WordSpec() {
                                 mappedLicenses = mapOf(
                                     "License 1" to "Mapped License 1",
                                     "License 2" to "Mapped License 2",
-                                    ),
+                                ),
```
sschuberth, line 101: This indentation fix should probably also go to the first commit.

sschuberth (commented):
Mostly nits. @mnonnenmacher please also have another look.

dao/src/main/kotlin/repositories/analyzerrun/DaoAnalyzerRunRepository.kt
lamppu: It does create a new table row [1](https://www.jetbrains.com/help/exposed/dao-crud-operations.html#create) [2](https://www.jetbrains.com/help/exposed/dao-entity-definition.html#defining-entity-class), note also that the test in `DaoAnalyzerRunRepositoryTest.kt` would fail if it didn't, as well as the tests in the following commit.

api/v1/model/src/commonMain/kotlin/Package.kt
lamppu: Yeah I actually realized it really doesn't need to be `null`, I think I didn't quite think this through properly.

model/src/commonMain/kotlin/runs/ShortestDependencyPath.kt
lamppu: Since at least the models [Project](https://github.com/eclipse-apoapsis/ort-server/blob/main/model/src/commonMain/kotlin/runs/Project.kt#L23), [VulnerabilityWithIdentifier](https://github.com/eclipse-apoapsis/ort-server/blob/main/model/src/commonMain/kotlin/VulnerabilityWithIdentifier.kt#L30) and [Package](https://github.com/doubleopen-project/ort-server/blob/main/model/src/commonMain/kotlin/runs/Package.kt#L23) talk about `identifier`s, I would like to just ensure some consensus on whether and when to shorten it to `id` before making this change.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
lamppu: (Fixed by not having the `shortestDependencyPaths` nullable)

model/src/commonMain/kotlin/runs/PackageWithShortestDependencyPaths.kt
lamppu: (Fixed by not having the `shortestDependencyPaths` nullable)

model/src/commonMain/kotlin/runs/PackageWithShortestDependencyPaths.kt
lamppu: I guess this could be something like `pkgDbId` to make it explicitly clear that it refers to the database ID (or at least the comment should be better)?

model/src/commonMain/kotlin/runs/ShortestDependencyPath.kt
sschuberth: @mnonnenmacher, what's the guideline here?

model/src/commonMain/kotlin/runs/PackageWithShortestDependencyPaths.kt
sschuberth: Also pinging @mnonnenmacher for this one.

mnonnenmacher: As above, I would recommend to use `id` for database ids and `identifier` for ORT identifiers.

sschuberth (approved):

workers/analyzer/src/main/kotlin/analyzer/Utils.kt
```diff
@@ -0,0 +1,49 @@
...
+): Map<Identifier, ShortestDependencyPath> {
+    val map = mutableMapOf<Identifier, ShortestDependencyPath>()
+
+    pathsByScopeMap.forEach { (scope, pathsMap) ->
+        pathsMap.forEach { (pkgIdentifier, pathList) ->
+            val existingEntry = map[pkgIdentifier.mapToModel()]
+
+            if (existingEntry == null || existingEntry.path.size > pathList.size) {
+                map[pkgIdentifier.mapToModel()] =
+                    ShortestDependencyPath(projectIdentifier, scope, pathList.map { it.mapToModel() })
```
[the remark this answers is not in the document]
mnonnenmacher: I think I would just not mention the scope at all in this case, does it matter which scope contains the shortest path?
But as a separate feature I would like to store all scope names that a package appears in, because this could also be used to filter packages by scope.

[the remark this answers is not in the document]
lamppu: > If we show this to users, this can be confusing, for example, if the runtime and the test scopes have an equally short path, and test is chosen here, in the UI it will appear as if this was a test dependency.

Yeah and of course that's also the case if a longer path has a different scope, that scope will "disappear" as well.

> But as a separate feature I would like to store all scope names that a package appears in, because this could also be used to filter packages by scope.

So I guess the scopes should be saved separately in their own table, and a separate field for scopes should be added in the response of the packages endpoint and this should be displayed in the details section of the package somehow.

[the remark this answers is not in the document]
sschuberth: > I think I would just not mention the scope at all in this case, does it matter which scope contains the shortest path?

I'd also be ok with not displaying the scope at all in the UI, but as I cannot rule out use-cases for having it, I'd still include it in the backend API response.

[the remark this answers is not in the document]
mnonnenmacher: > I'd still include it in the backend API response

But if we don't want to show it anywhere and it is randomly selected anyway, why even store it in the database?

I don't know what the use case for showing the shortest path is anyway, so I can't say if it makes sense to show the scope here. Getting a list of direct dependencies that introduce a certain transitive dependency, or respectively getting all paths from different roots, would appear more useful to me.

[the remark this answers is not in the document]
mnonnenmacher: Ok, maybe we should think how we could make it better than the WebApp as part of the migration. But anyway, I don't have any issue with the implementation so I'll approve the PR.

> To me, that sounds more like what should eventually be show on the "Dependencies" tab, or be implemented as a "reverse search" feature to show all products that contain a certain dependency.

I thought of this because in the Workbench packages can be filtered by scope.

mnonnenmacher (approved):

PR #1883, opened 2025-01-27
chore(ui): Remove the note about the analyzer job being mandatory

sschuberth:
The note sounded a bit as if there were immediate plans to be able to run without the analyzer, but this is not the case.


mnonnenmacher (approved):

sschuberth:
@mnonnenmacher any idea why [this](https://github.com/eclipse-apoapsis/ort-server/actions/runs/12989497197/job/36222632642#step:7:22) still fails for PRs from forks?

mnonnenmacher:
> @mnonnenmacher any idea why [this](https://github.com/eclipse-apoapsis/ort-server/actions/runs/12989497197/job/36222632642#step:7:22) still fails for PRs from forks?

I guess it's a permission issue with forks, maybe we have to use this trigger instead:
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target

sschuberth:
> This event allows your workflow to do things like label or comment on pull requests from forks.

This sounds good, I'll give it a try.

PR #1888, opened 2025-01-27
fix(github): Check issue links in the context of the base of the PR

sschuberth:
This should fix another permission issue [1] when creating a PR from a fork, see the docs at [2].

[1]: https://github.com/eclipse-apoapsis/ort-server/actions/runs/12989497197/job/36222632642#step:7:22
[2]: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target

### DIFF
--- a/.github/workflows/check-issue-links.yml
+++ b/.github/workflows/check-issue-links.yml
@@ -1,7 +1,7 @@
 name: Check Issue Links
 
 on:
-  pull_request:
+  pull_request_target:
     branches:
     - main
 


sschuberth:
Hmm, this now does not seem to run the `check-issue-links` workflow at all anymore...

mnonnenmacher:
> Hmm, this now does not seem to run the `check-issue-links` workflow at all anymore...

I guess that's because the workflow in the main branch does not have this trigger yet so we would have to merge it first. We could give it a try, but if it doesn't work out of the box we should continue testing on a fork to make it faster.

mnonnenmacher (approved):

PR #1889, opened 2025-01-27
feat(scanner): Generate .gitconfig file with URL insteadOf sections

wkl3nk:
Before a scan starts in a Scanner pod, generate a `.gitconfig` environment file containing URL `insteadOf` sections. These are required if the repository contains submodules specified using SSH URLs.

The `insteadOf` sections can either be generated using default values in `application.conf` or customized via the environment variable `GIT_CONFIG_URL_INSTEAD_OF`. This allows flexibility in defining what URL `insteadOf` sections are created in `.gitconfig`.

### DIFF
--- a/workers/common/src/main/kotlin/common/env/GitConfigGenerator.kt
+++ b/workers/common/src/main/kotlin/common/env/GitConfigGenerator.kt
@@ -1,0 +1,127 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.workers.common.env
+
+import org.eclipse.apoapsis.ortserver.config.ConfigManager
+import org.eclipse.apoapsis.ortserver.model.CredentialsType
+import org.eclipse.apoapsis.ortserver.utils.config.getStringOrNull
+import org.eclipse.apoapsis.ortserver.workers.common.env.definition.EnvironmentServiceDefinition
+
+import org.slf4j.LoggerFactory
+
+/**
+ * A specialized generator class to generate a Git configuration file.
+ *
+ * Source code repositories may have submodules locations specified using SSH URLs. This imposes a problem, because
+ * the required SSH keys to fetch the submodules are not available in the execution environment. To work around this
+ * issue, Git allows to define URL `insteadOf` sections in the _.gitconfig_ file. This generator creates such
+ * sections from a static configuration file, which can be customized via an environment variable.
+ *
+ * The `insteadOf` sections can either be generated using default values in
+ * `application.conf` or customized via the environment variable
+ * `GIT_CONFIG_URL_INSTEAD_OF`. This allows flexibility in defining what
+ * URL `insteadOf` sections are created in `.gitconfig`.
+ *
+ * The expected format is a comma-separated list of base URLs and their corresponding insteadOf URLs,
+ * separated by an equal sign. For example:
+ * `https://github.com=ssh://git@github.com,https://github.com/=git@github.com:`
+ *
+ * See also: [Git documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf)
+ *
+ * There is a coupling between this generator class and the generator class for Git credentials
+ * [GitCredentialsGenerator]. If Git credentials are generated, then this class also generates a `credential`
+ * section in the Git configuration file _.gitconfig_ in order to reference the generated _.git-credentials_ file.
+ */
+class GitConfigGenerator(private val gitConfigUrlInsteadOfPairs: Map<String, String>) :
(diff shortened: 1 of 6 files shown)


workers/common/src/main/kotlin/common/env/GitConfigGenerator.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.eclipse.apoapsis.ortserver.model.CredentialsType
+import org.eclipse.apoapsis.ortserver.utils.config.getStringOrNull
+import org.eclipse.apoapsis.ortserver.workers.common.env.definition.EnvironmentServiceDefinition
+
+import org.slf4j.LoggerFactory
+
+/**
+ * A specialized generator class to generate a Git configuration file.
+ */
+class GitConfigGenerator(val configManager: ConfigManager) : EnvironmentConfigGenerator<EnvironmentServiceDefinition> {
```
oheger-bosch, line 34: There is now an implicit coupling between this class and `GitCredentialsGenerator`, as the latter requires the former to work properly. I would like to avoid this, but have no idea how this could be achieved easily. At least, it should be stated in the class comment that `GitConfigGenerator` will create a `credential` section if necessary. The class comment of `GitCredentialsGenerator` needs to be adapted as well; it says that this class creates a `.gitconfig` file.

oheger-bosch, line 34: Suggestion: Do not pass the `ConfigManager` to the constructor, but the already parsed `insteadOf` declarations. The companion object could offer a `create` function that expects the `ConfigManager`, evaluates the configuration, and returns an initialized instance. This allows for a more flexible usage and would also simplify testing.

workers/common/src/main/kotlin/common/env/GitConfigGenerator.kt
```diff
@@ -0,0 +1,89 @@
...
+    companion object {
+        private const val GIT_CONFIG_FILE_NAME = ".gitconfig"
+
+        private val logger = LoggerFactory.getLogger(GitConfigGenerator::class.java)
+
+        // Create url.<base>.insteadOf sections in the Git configuration file.
+        // See https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf
+        internal const val GIT_CONFIG_URL_INSTEAD_OF = "gitConfigUrlInsteadOf"
+
+        internal fun parseGitConfigUrlInsteadOf(config: String) =
```
oheger-bosch, line 44: Please document the expected format of the configuration string in the class comment.

workers/common/src/test/kotlin/common/env/GitConfigGeneratorTest.kt
```diff
@@ -0,0 +1,167 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
```
oheger-bosch, line 2: Nit: Copyright year should be 2025.

workers/common/src/main/kotlin/common/env/GitConfigGenerator.kt
```diff
@@ -0,0 +1,89 @@
...
+            config.split(",")
+                .filter { it.contains("=") }
```
oheger-bosch, line 46: Probably a warning should be logged if syntactically incorrect declarations are found. Otherwise, it is hard to diagnose problems with the configuration.

workers/common/src/test/kotlin/common/env/GitConfigGeneratorTest.kt
```diff
@@ -0,0 +1,167 @@
...
+        "parse a syntactically correct configuration with 2 entries" {
+            val configLine = " https://github.com = ssh://git@github.com , https://github.com/ = git@github.com: "
+
+            val result = GitConfigGenerator.parseGitConfigUrlInsteadOf(configLine)
+
+            result shouldContainExactly mapOf(
+                "https://github.com" to "ssh://git@github.com",
+                "https://github.com/" to "git@github.com:"
+            )
+        }
```
oheger-bosch, line 165: Maybe add a test case for an incorrect configuration?

workers/scanner/src/main/resources/application.conf
```diff
@@ -48,3 +48,6 @@ scanner {
     password = ${?SCANNER_RECEIVER_TRANSPORT_PASSWORD}
   }
 }
+
+gitConfigUrlInsteadOf = "https://github.com=ssh://git@github.com,https://github.com/=git@github.com:"
```
oheger-bosch, line 52: I would rather use an empty default. The URLs listed here could be perfectly fine in some environment.

workers/common/src/main/kotlin/common/env/GitConfigGenerator.kt
```diff
@@ -0,0 +1,89 @@
...
+                .map {
+                    val (base, url) = it.split("=")
+                    base.trim() to url.trim()
+                }
+                .toMap()
+
+        internal suspend fun generateGitConfig(
+            builder: ConfigFileBuilder, definitions: Collection<EnvironmentServiceDefinition>, config: Config
```
oheger-bosch, line 54: Style nit: Convention is to place each parameter on a separate line.

workers/common/src/main/kotlin/common/env/GitConfigGenerator.kt
wkl3nk: Yes, I agree. There should be a hint in both classes there is there is a small coupling between the two generators. But I like this better an less error prone than both generators fiddle around with the same _.gitconfig_ configuration file.

wkl3nk: But `GitConfigGenerator` is instantiated by Kotlin Koin as DI implementation. So at which level should the parsing take place? Can you give me a hint?

oheger-bosch: If there is a function `fun create(configManager: ConfigManager): GitConfigGenerator` in the companion object, it can be called in the Koin module as `GitConfigGenerator.create(get())`.

wkl3nk: Ah. I understand, some kind of factory method to create the instance, instead to use the constructor.

oheger-bosch (approved):

PR #1892, opened 2025-01-28
chore(github): Do not comment on PRs when issue references are in sync

sschuberth:
Reduce noise. Just complain if something is wrong.

### DIFF
--- a/.github/workflows/check-issue-links.yml
+++ b/.github/workflows/check-issue-links.yml
@@ -111,4 +111,4 @@ jobs:
         done
 
         # Add a comment to the PR.
-        gh pr comment ${{ github.event.pull_request.number }} --body "$(echo -e "${{ env.COMMENT_TEXT }}")"
+        gh pr comment ${{ github.event.pull_request.number }} --body "$(echo -e "$COMMENT_TEXT")"
--- a/.github/workflows/check-issue-links.yml
+++ b/.github/workflows/check-issue-links.yml
@@ -55,14 +55,14 @@ jobs:
             }
           }
         }'
-        
+
         # Send the GraphQL query using gh API.
         response=$(gh api graphql -f query="$query")
-        
+
         # Extract linked issues.
         linked_issues=$(echo $response | jq -r '.data.repository.pullRequest.closingIssuesReferences.edges[].node.number' | sort | uniq)
         echo "Found currently linked issues: $linked_issues"
-        
+
         # Export linked issues as an environment variable.
         echo "LINKED_ISSUES=$linked_issues" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


sschuberth:
@mnonnenmacher I believe this also needs to be merged first in order to see its effect.

.github/workflows/check-issue-links.yml
```diff
@@ -90,13 +88,14 @@ jobs:
...
-        echo "COMMENT_TEXT=$comment_text" >> $GITHUB_ENV
+        if [[ -n "$comment_text" ]]; then
+          echo $comment_text
 
+          # Export the comment text as an environment variable.
+          echo "COMMENT_TEXT=$comment_text" >> $GITHUB_ENV
+        fi
     - name: Add PR comment
-      if: always()
+      if: ${{ env.COMMENT_TEXT != '' }}
```
mnonnenmacher, line 98: Old comments should still be deleted, so either move the condition into the script or split the step.

mnonnenmacher (changes requested):

.github/workflows/check-issue-links.yml
```diff
@@ -90,11 +88,12 @@ jobs:
...
-        echo $comment_text
-
-        # Export the comment text as an environment variable.
-        echo "COMMENT_TEXT=$comment_text" >> $GITHUB_ENV
+        if [[ -n "$comment_text" ]]; then
+          echo $comment_text
 
+          # Export the comment text as an environment variable.
+          echo "COMMENT_TEXT=$comment_text" >> $GITHUB_ENV
+        fi
```
mnonnenmacher, line 96: I just noticed, could you add an else case here which prints "Issues referenced in commit messages and issues linked to this PR are in sync.", just so that the logs confirm that everything is ok?

mnonnenmacher (changes requested):

mnonnenmacher:
Any idea why the action is now failing with `fatal: Invalid revision range 8efd9be29479e34bd983daa7a85472cddeb66cb8..e88a56aff89859166aac5c91d7c1fd45540ee1c8`? Could the reason be that your branch is not on top of main?
This is getting much more complicated than I anticipated.

mnonnenmacher (dismissed):

mnonnenmacher:
> > Could the reason be that your branch is not on top of main?
> 
> I've rebased meanwhile, but [as mentioned](https://github.com/eclipse-apoapsis/ort-server/pull/1892#issuecomment-2619271260) I guess we need to merge to `main` before relying on the action's output.

Sure, but that's not the reason why the action fails. I could reproduce it locally, the problem seems to be that `origin/main` is not fetched at all. Adding `git fetch origin main` before the `git log --format...` line should fix it. Do you want to add this to this PR or should I make a separate one?

sschuberth:
> Do you want to add this to this PR or should I make a separate one?

I'll add it.

mnonnenmacher (approved):

PR #1893, opened 2025-01-28
CLI: Add auth login command

MarcelBochtler:


cli/src/main/kotlin/StartCommand.kt
```diff
@@ -45,13 +45,13 @@ internal val POLL_INTERVAL = System.getProperty("POLL_INTERVAL")?.toLongOrNull()
 class StartCommand : SuspendingCliktCommand(name = "start") {
     private val repositoryId by option(
         "--repository-id",
-        envvar = "REPOSITORY_ID",
+        envvar = "ORT_RUNS_START_REPOSITORY_ID",
```
sschuberth, line 48: I find the `ORT_RUNS_` prefix a bit weird. I'm aware that other commands already use it, but I guess all commands should align on not using `ORT_` at all, but use `OSC_` instead as a prefix.

cli/src/test/kotlin/AuthLoginCommandTest.kt
```diff
@@ -0,0 +1,88 @@
...
+                    "login",
+                    "--base-url", "http://localhost:8080/",
+                    "--token-url", "http://localhost/token",
+                    "--username", "testUser",
+                    "--password", "testPassword",
+                    "--client-id", "test-client-id"
+                )
+            )
+
+            Yaml.default.decodeFromStream<AuthenticationStorage>(testAuthFile.inputStream()).let {
```
sschuberth, line 75: Also here the `InputStream` should be closed.

sschuberth: Prefer something like `with()` or `also {}` over `let {}` if the return value is not used.

cli/src/main/kotlin/AuthStatusCommand.kt
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Command to show information about the current authentication status.
+ */
+class AuthStatusCommand : SuspendingCliktCommand(name = "status") {
+    override fun help(context: Context) = "Show information about the current authentication status."
+
+    override suspend fun run() {
+        if (authFile.exists()) {
+            Yaml.default.decodeFromStream<AuthenticationStorage>(authFile.inputStream()).let {
```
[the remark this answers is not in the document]
sschuberth: Prefer something like `with()` or `also {}` over `let {}` if the return value is not used.

cli/src/main/kotlin/utils/ConfigDirHelper.kt
```diff
@@ -0,0 +1,63 @@
...
+
+        if (!dir.exists()) dir.safeMkdirs()
+
+        return dir
+    }
+
+/**
+ * The file where the authentications are stored.
+ */
+internal val authFile: File
```
oheger-bosch, line 62: Would this property better belong to `AuthenticationUtils`?

In general, I have some difficulties to understand the various components spread over multiple projects that are involved in storing authentication data. Exposing the `authFile` here reminds me of the concept of global variables in pre-OO times.

As a suggestion, could there be an `object AuthenticationStorage` in the `cli` module that encapsulates all the internal data types and details? It could expose an API on a more higher level, to query and set authentication data without clients have to deal with file access.

Also, is it really necessary that the model for storing authentication data is part of the `api` module?

cli/src/main/kotlin/DownloadCommand.kt
```diff
@@ -20,9 +20,12 @@
 package org.eclipse.apoapsis.ortserver.cli
```
oheger-bosch, line 20: Commit message typo: "missing help *texts*".

cli/src/main/kotlin/AuthStatusCommand.kt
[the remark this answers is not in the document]
MarcelBochtler: Obsolete with the refactoring of the `AuthenticationStorage`.

cli/src/test/kotlin/AuthLoginCommandTest.kt
MarcelBochtler: Obsolete with the `AuthenticationStorage` refactoring.

cli/src/main/kotlin/model/AuthenticationStorage.kt
```diff
@@ -0,0 +1,99 @@
...
+            storage.addAll(savedAuth)
+        }
+    }
+
+    fun add(authentication: HostAuthenticationDetails) {
+        if (storage.isEmpty()) {
+            storage.add(authentication)
+        } else {
+            storage[0] = authentication
+        }
```
sschuberth, line 54: Can you explain the semantic here a bit (e.g. in function documentation)? Why is it intended to overwrite a first entry, instead of prepending it, for example?

cli/src/main/kotlin/model/AuthenticationStorage.kt
```diff
@@ -0,0 +1,99 @@
...
+
+        saveToFile()
+    }
+
+    fun get(): HostAuthenticationDetails? = storage.firstOrNull()
+
+    fun clear() {
+        storage.clear()
+
+        saveToFile()
```
sschuberth, line 64: Should this maybe rather delete the file, instead of clearing its contents? Also here some function docs would be appreciated.

cli/src/main/kotlin/model/AuthenticationStorage.kt
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private fun saveToFile() {
+        authFile.writeText(Yaml.default.encodeToString(storage))
+
+        // Ensure file permissions 600.
```
sschuberth, line 70: Please also explain *why*.

cli/src/main/kotlin/model/AuthenticationStorage.kt
```diff
@@ -0,0 +1,99 @@
...
+    val baseUrl: String,
+
+    /** The URL to request a new access token. */
+    val tokenUrl: String,
+
+    /** The client ID to use for authentication. */
+    val clientId: String,
+
+    val username: String,
+    val tokens: Tokens
```
sschuberth, line 92: Consider documenting these as well, so that the whole class is documented.

cli/src/main/kotlin/utils/AuthenticationUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+                    val updatedTokenInfo = runCatching {
+                        val auth = AuthService(
+                            client = createDefaultHttpClient(JSON),
+                            tokenUrl = authDetails.tokenUrl,
+                            clientId = authDetails.clientId
+                        )
+
+                        auth.refreshToken(authDetails.tokens.refresh)
+                    }.getOrElse {
+                        println("Failed to refresh authentication. Please run '$COMMAND_NAME auth login'.")
```
sschuberth, line 73: It feels wrong to me to do console output in this utility function (and ideally this should use `echo()` anyway, but it lacks the terminal context to do so).

How about instead using `getOrThrow()` here and leaving it the callers to do console output / map to the desired type of exception?

cli/src/main/kotlin/utils/ConfigDirHelper.kt
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * The directory where the configuration files are stored, respecting the XDG Base Directory specification [1].
+ *
+ * [1]: https://specifications.freedesktop.org/basedir-spec/latest/
+ */
+internal val configDir: File
+    get() {
+        val dir = when {
+            Os.isLinux -> File(
+                System.getenv("XDG_CONFIG_HOME") ?: "${Os.fixupUserHomeProperty()}/.config/$COMMAND_NAME"
```
sschuberth, line 38: Even if this case is for Linux, you should generally prefer `Os.env[]` to look up environment variables, as it does the right thing on Windows and looks them up case-insensitively. Similar below.

cli/src/main/kotlin/utils/ConfigDirHelper.kt
```diff
@@ -0,0 +1,57 @@
...
+
+            Os.isWindows -> File(
+                System.getenv("XDG_CONFIG_HOME")
+                    ?: "${
+                        System.getenv("LOCALAPPDATA")
+                            ?: "${Os.fixupUserHomeProperty()}/.config"
+                    }/$COMMAND_NAME"
+            )
+
+            else -> File("${Os.fixupUserHomeProperty()}/.config/$COMMAND_NAME")
```
sschuberth, line 51: Maybe refer `Os.userHomeDirectory.resolve(".config/$COMMAND_NAME")`. Similar above.

cli/src/main/kotlin/model/AuthenticationStorage.kt
MarcelBochtler: IMO I already did that with the note in the class documentation:
```
Note: Currently only single host authentication is supported.
```

So the logic basically is:
Add the single authentication or overwrite the existing one.

sschuberth: Ok, I didn't look there. Would it make sense to instead use a single (nullable) `storage` property then until this restriction is lifted? Or are you going to extend the functionality very soon to a real list?

MarcelBochtler: > Or are you going to extend the functionality very soon to a real list?

No plans to do so, yet.
Just a feature that I want at some point in time, therefore I already designed the file format in a way to support it.

cli/src/main/kotlin/utils/AuthenticationUtils.kt
MarcelBochtler: > It feels wrong to me to do console output in this utility function

Yes, I agree.
However, the error message is printed in the `refreshToken` lambda.
That means the caller (the CLI part), would need to wrap every _usage_ of the `OrtServerClient` into a try catch, in order to handle this in the same way, and not only at the time of creation of the client.

Alternatively, the CLI could offer another abstraction layer that handles these errors.

I'm not sure if this is worth the effort, or am I missing a simpler solution?

api/v1/client/src/commonMain/kotlin/auth/AuthService.kt
```diff
@@ -29,7 +29,7 @@ import io.ktor.http.isSuccess
 /**
  * A service to generate/refresh an OAuth token to authenticate with the ORT server.
  */
-internal class AuthService(
+class AuthService(
```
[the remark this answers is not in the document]
mnonnenmacher: Nit: Before 1.0.0 commits should not be marked as breaking: https://github.com/eclipse-apoapsis/ort-server/blob/main/CONTRIBUTING.md#breaking-changes

cli/src/main/kotlin/utils/AuthenticationUtils.kt
mnonnenmacher: > That means the caller (the CLI part), would need to wrap every usage of the OrtServerClient into a try catch, in order to handle this in the same way, and not only at the time of creation of the client.

Could this not be done in the `main` function?

cli/src/test/kotlin/AuthLogoutCommandTest.kt
```diff
@@ -0,0 +1,58 @@
...
+
+import org.eclipse.apoapsis.ortserver.cli.OrtServerMain
+import org.eclipse.apoapsis.ortserver.cli.model.AuthenticationStorage
+
+class AuthLogoutCommandTest : StringSpec({
+    afterEach { unmockkAll() }
+
+    "logout command" should {
+        "delete the authentication file" {
+            mockkObject(AuthenticationStorage)
```
mnonnenmacher, line 41: Why not just use the `AuthenticationStorage` instead of mocking it?

cli/src/main/kotlin/DownloadCommand.kt
```diff
@@ -20,9 +20,12 @@
...
 
 import com.github.ajalt.clikt.command.SuspendingNoOpCliktCommand
+import com.github.ajalt.clikt.core.Context
 import com.github.ajalt.clikt.core.subcommands
 
 class DownloadCommand : SuspendingNoOpCliktCommand(name = "download") {
+    override fun help(context: Context) = "Commands for downloading results and detailed information about ORT runs."
```
mnonnenmacher, line 27: FYI, I have a similar commit in #1873 already.

mnonnenmacher (changes requested):

cli/src/main/kotlin/utils/AuthenticationUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.eclipse.apoapsis.ortserver.client.OrtServerClient.Companion.JSON
+import org.eclipse.apoapsis.ortserver.client.auth.AuthService
+import org.eclipse.apoapsis.ortserver.client.createDefaultHttpClient
+import org.eclipse.apoapsis.ortserver.client.createOrtHttpClient
+
+/**
+ * Create an authenticated ORT Server client using the stored authentication. Returns null if the user is not
+ * authenticated.
+ */
+fun createOrtServerClient() = AuthenticationStorage.get()?.let { createOrtServerClient(it) }
```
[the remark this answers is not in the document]
MarcelBochtler: > So, should this already be handled here?

The reason to not handle it here, is similar to this discussion: https://github.com/eclipse-apoapsis/ort-server/pull/1893#discussion_r1935335734

This error should be handled by the calling CLI function, because the handling is printing an error message to the user, which should not be done in this Utils class.

> Alternatively, `AuthenticationStorage` could already throw if no information is available or offering a `getOrThrow()` function.

Not being authenticated is a valid state which should not throw an Exception. That's why I prefer a nullable `HostAuthenticationDetails` as response of `AuthenticationStorage.get()`

I would like to keep this as-is.

cli/src/test/kotlin/AuthLogoutCommandTest.kt
MarcelBochtler: Because I don't want to overwrite the actual `auth.yml` file on the machine of the developer executing the test.

But you're right I shouldn't mock the object under test.

cli/src/main/kotlin/DownloadCommand.kt
MarcelBochtler: Dropped the commit

cli/src/test/kotlin/AuthLogoutCommandTest.kt
mnonnenmacher: Ok, I guess I just don't like that we need mocking to test a trivial function.

mnonnenmacher (dismissed):
@sschuberth Looks good to me, but you had far more comments, so I think you should have another look.

cli/src/main/kotlin/OrtServerMain.kt
```diff
@@ -23,33 +23,35 @@ import com.github.ajalt.clikt.command.SuspendingNoOpCliktCommand
...
 
 const val COMMAND_NAME = "osc"
 
 suspend fun main(args: Array<String>) {
-    OrtServerMain().main(args)
+    runCatching {
+        OrtServerMain().main(args)
+    }.getOrElse { t ->
+        if (t.cause is AuthenticationException) {
+            println("Authentication failed. Please run '${COMMAND_NAME} auth login' to authenticate.")
```
sschuberth, line 44: Maybe [this](https://ajalt.github.io/clikt/exceptions/#handling-exceptions-manually) would be the right approach to be able to use `echo` instead of `print` here.

cli/src/main/kotlin/OrtServerMain.kt
```diff
@@ -23,33 +23,35 @@ import com.github.ajalt.clikt.command.SuspendingNoOpCliktCommand
...
+import org.eclipse.apoapsis.ortserver.client.auth.AuthenticationException
 import org.eclipse.apoapsis.ortserver.model.ORT_SERVER_VERSION
 
 const val COMMAND_NAME = "osc"
 
 suspend fun main(args: Array<String>) {
-    OrtServerMain().main(args)
+    runCatching {
+        OrtServerMain().main(args)
+    }.getOrElse { t ->
```
sschuberth, line 42: Prefer `onFailure` here as the return value is unused.

cli/src/main/kotlin/utils/ConfigDirHelper.kt
```diff
@@ -0,0 +1,54 @@
...
+            )
+
+            Os.isMac -> File(Os.env["XDG_CONFIG_HOME"] ?: "${Os.userHomeDirectory.resolve(".config/$COMMAND_NAME")}")
+
+            Os.isWindows -> File(
+                Os.env["XDG_CONFIG_HOME"]
+                    ?: "${Os.env["LOCALAPPDATA"] ?: "${Os.fixupUserHomeProperty()}/.config"}/$COMMAND_NAME"
+            )
+
+            else -> File("${Os.userHomeDirectory.resolve(".config/$COMMAND_NAME")}")
```
sschuberth, line 48: Just `Os.userHomeDirectory.resolve(".config/$COMMAND_NAME")` is enough 😉

cli/src/main/kotlin/utils/ConfigDirHelper.kt
```diff
@@ -0,0 +1,54 @@
...
+ * The directory where the configuration files are stored, respecting the XDG Base Directory specification [1].
+ *
+ * [1]: https://specifications.freedesktop.org/basedir-spec/latest/
+ */
+internal val configDir: File
+    get() {
+        val dir = when {
+            Os.isLinux -> File(
+                Os.env["XDG_CONFIG_HOME"] ?: "${Os.userHomeDirectory.resolve(".config/$COMMAND_NAME")}"
+            )
```
sschuberth, line 39: Personally, I'd prefer

            Os.isLinux -> Os.env["XDG_CONFIG_HOME"]?.let { File(it) }
                ?: Os.userHomeDirectory.resolve(".config/$COMMAND_NAME")

to avoid nesting of double-quoted strings.

sschuberth: Similar below.

cli/src/main/kotlin/utils/ConfigDirHelper.kt
```diff
@@ -0,0 +1,54 @@
...
+
+            Os.isMac -> File(Os.env["XDG_CONFIG_HOME"] ?: "${Os.userHomeDirectory.resolve(".config/$COMMAND_NAME")}")
+
+            Os.isWindows -> File(
+                Os.env["XDG_CONFIG_HOME"]
+                    ?: "${Os.env["LOCALAPPDATA"] ?: "${Os.fixupUserHomeProperty()}/.config"}/$COMMAND_NAME"
```
sschuberth, line 45: Please stick to `userHomeDirectory`.

sschuberth (changes requested):

cli/src/main/kotlin/utils/AuthenticationUtils.kt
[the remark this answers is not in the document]
oheger-bosch: I do not fully understand these arguments. Not being authenticated is certainly problematic when creating an HTTP client that requires authentication? So, why can't this function throw an exception that then bubbles up to an outer layer of the CLI?

But I won't insist; keep it as is if you like.

MarcelBochtler:
I dropped the `auth status` command because it didn't actually check if the authentication was valid.
It just checked if there are any credentials stored, which might be confusing.

cli/src/main/kotlin/utils/ConfigDirHelper.kt
```diff
@@ -0,0 +1,53 @@
...
+ * [1]: https://specifications.freedesktop.org/basedir-spec/latest/
+ */
+internal val configDir: File
+    get() {
+        val fallbackDir = Os.userHomeDirectory.resolve(".config/$COMMAND_NAME")
+
+        val dir = when {
+            Os.isLinux -> Os.env["XDG_CONFIG_HOME"]?.let { File(it) } ?: fallbackDir
+
+            Os.isMac -> Os.env["XDG_CONFIG_HOME"]?.let { File(it) } ?: fallbackDir
```
sschuberth, line 41: As these are the same, prefer `Os.isLinux || Os.isMac -> ...`.

sschuberth (approved):
Only one optional nit left from my side.

sschuberth (approved):

PR #1895, opened 2025-01-28
Scan storage read issues optimization

oheger-bosch:
This PR further optimizes the storing of scan results by preventing that issues are unnecessarily loaded when constructing an ORT scanner run. Refer to the single commits for further information.

### DIFF
--- a/dao/src/main/kotlin/tables/shared/IssuesTable.kt
+++ b/dao/src/main/kotlin/tables/shared/IssuesTable.kt
@@ -30,6 +30,8 @@ import org.eclipse.apoapsis.ortserver.model.runs.Issue
 
 import org.jetbrains.exposed.dao.LongEntity
 import org.jetbrains.exposed.dao.id.EntityID
+import org.jetbrains.exposed.sql.Query
+import org.jetbrains.exposed.sql.ResultRow
 import org.jetbrains.exposed.sql.and
 import org.jetbrains.exposed.sql.stringLiteral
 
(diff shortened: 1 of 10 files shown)


dao/src/main/kotlin/tables/shared/IssuesTable.kt
```diff
@@ -83,3 +103,36 @@ class IssueDao(id: EntityID<Long>) : LongEntity(id) {
...
+/**
+ * Create an [Issue] from the fields of this [ResultRow]. Using this function, issues can be created from query
+ * results that may include other data as well. The fields of the result define the issue; they must appear in the
+ * order as described in [IssueDao.createFromQuery].
+ */
+@Suppress("ComplexCondition")
+fun ResultRow.toIssue(): Issue {
+    // The exposed library seems not to fully support fields on alias queries when unions are used.
+    // Use the field indexes in order to extract the values from the ResultRow instead of the field names.
+    // Disadvantage: More fragility, losing type safety at compile time.
```
nnobelis, line 116: Another disadvantage is that, if you refactor the `issues` table, e.g adding or removing a column, you'll silently break this function.

I wonder if we could file a solution with column name lookup instead of column index lookup, referencing the properties' name in `IssueDao`. What do you think ?

workers/scanner/src/main/kotlin/scanner/OrtServerScanResultStorage.kt
```diff
@@ -85,6 +89,12 @@ class OrtServerScanResultStorage(
     private val db: Database,
     private val scannerRunId: Long
 ) : ProvenanceBasedScanStorage {
+    /**
+     * A [Map] to store the issues encountered for the scanned provenances. This is used to track issues that need to
+     * be associated with the current ORT run.
+     */
+    private val issuesMap = ConcurrentHashMap<Provenance, Set<Issue>>()
```
nnobelis, line 96: What is the lifespan for this collection ?

Is the `OrtserverScanResultStorage` instantiated and destroyed per ORT run or is it per storage service i.e. persistent in the application ?

dao/src/main/kotlin/tables/shared/IssuesTable.kt
oheger-bosch: Actually, I did not look deeper into this code. It existed before, and in this PR I just exposed it. It is certainly beneficial to find a better solution here.

Not sure how to proceed here. Can you live with the current solution or should I drop the first commit (it is not required for the other changes)?

workers/scanner/src/main/kotlin/scanner/OrtServerScanResultStorage.kt
oheger-bosch: The lifespan is the current ORT run only. For each run, a dedicated instance of `OrtServerScanResultStorage` is created.

dao/src/main/kotlin/tables/shared/IssuesTable.kt
nnobelis: No problem I misread the diffs and thought you introduced the changes.
Let's leave it like that for the moment then.

nnobelis (approved):

nnobelis (approved):

PR #1896, opened 2025-01-28
fix(mapping): Fix mapping of environment variable declarations

### LINKED ISSUE #1894: Values of environment variables always null
We have been trying to pass some dummy environment variables to the Analyzer, which would help us with technical onboarding of one project. I've implemented the UI for it (no PR yet):

Image

and verified that the proper (I think) configuration object is sent to the create ORT run endpoint:

This is what the API docs say:
Image

This is the relevant part of what we send to back-end from the form, when creating the run:
Image

However, by tweaking the error logging a bit (to enable multiline logging), this is the ERROR message we get:
```
message="Error while processing message with traceId '8168b4ad-4beb-43fe-9263-578d934210da': Found invalid environment variable definitions:
RepositoryEnvironmentVariableDefinition(name=KEYSTORE, secretName=null, value=null)
RepositoryEnvironmentVariableDefinition(name=SIGNING_STORE_PASSWORD, secretName=null, value=null)
RepositoryEnvironmentVariableDefinition(name=SIGNING_KEY_ALIAS, secretName=null, value=null)
RepositoryEnvironmentVariableDefinition(name=SIGNING_KEY_PASSWORD, secretName=null, value=null)"
```

As you can see, although the name is taken in properly, the value is always null.

The error message seems to come [from here](https://github.com/eclipse-apoapsis/ort-server/blob/d29de32154a28b2c82d380f25f0c2d8253e70afd/workers/common/src/main/kotlin/common/env/config/EnvironmentConfigLoader.kt#L347-L367). (This part I tweaked by replacing `appendLine` with `append`, to get the whole error message from back-end, as multiline logging doesn't seem to work.)

It might well be that there's something we don't understand in setting up the environment variables, so this is all based on the current API documentation. That's why I also labelled this ticket with a "question" tag.

oheger-bosch:
The `value` property was not mapped when converting an API object to a model object.

Fixes #1894.

### DIFF
--- a/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
+++ b/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
@@ -598,7 +598,7 @@ fun ApiInfrastructureService.mapToModel() =
 fun InfrastructureServiceDeclaration.mapToApi() =
     ApiInfrastructureService(name, url, description, usernameSecret, passwordSecret, credentialsTypes.mapToApi())
 
-fun ApiEnvironmentVariableDeclaration.mapToModel() = EnvironmentVariableDeclaration(name, secretName)
+fun ApiEnvironmentVariableDeclaration.mapToModel() = EnvironmentVariableDeclaration(name, secretName, value)
 
 fun EnvironmentVariableDeclaration.mapToApi() = ApiEnvironmentVariableDeclaration(name, secretName, value)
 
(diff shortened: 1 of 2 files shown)


oheger-bosch:
> Pushing from fork seems to still cause issues with "Check issue links". I understood this should be fixed already?

I did not follow the recent changes in this area, but obviously, there is still a problem.

sschuberth:
> > Pushing from fork seems to still cause issues with "Check issue links". I understood this should be fixed already?
> 
> I did not follow the recent changes in this area, but obviously, there is still a problem.

I was planning to address this as part of https://github.com/eclipse-apoapsis/ort-server/pull/1892.

sschuberth:
In any case, the failed check is not blocking, so this *can* be merged.

sschuberth (approved):

PR #1898, opened 2025-01-29
ui: Add environment variable support for the Analyzer

Etsija:

Screenshot from 2025-01-29 09-22-19

Screenshot from 2025-01-29 09-23-06

Open question (@sschuberth ): a more helpful text is needed for the environment variables description in the form. Code search didn't reveal any good descriptions to me.

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/analyzer-fields.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/analyzer-fields.tsx
@@ -17,13 +17,15 @@
  * License-Filename: LICENSE
  */
 
-import { UseFormReturn } from 'react-hook-form';
+import { PlusIcon, TrashIcon } from 'lucide-react';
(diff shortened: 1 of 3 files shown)


mmurto:
As the environment variables often contain secrets, and the API allows it, the form should allow giving a secret name in addition to the value directly.

Etsija:
> As the environment variables often contain secrets, and the API allows it, the form should allow giving a secret name in addition to the value directly.

It was a team decision to omit it for now, as we have a strict need of only the name and value as part of technical onboarding for a customer.

mmurto (approved):

sschuberth:
> a more helpful text is needed for the environment variables description in the form.

Done as part of https://github.com/eclipse-apoapsis/ort-server/pull/1902.

lamppu:
@Etsija Did you test this without adding any fake variables? For me, if I don't add any variables in the form, it sends this in the request and then the analyzer fails:
```
"environmentConfig": {
    "infrastructureServices": [],
    "environmentVariables": [{ "name": "" }]
 }
 ```

PR #1903, opened 2025-01-29
feat: Return overall vulnerability rating for vulnerabilities of a run

### LINKED ISSUE #1839: Return advisory overall vulnerability rating for all vulnerabilities in a run
Similar to how `/api/v1/products/:productId/vulnerabilities` works:

Image

the similar endpoint for individual ORT runs, `/api/v1/runs/:runId/vulnerabilities` should also return the advisory overall rating for each vulnerability, which it currently doesn't:

Image

Then we could get rid of this overall rating being calculated in the UI (which is far less optimal).

The overall rating calculation was added to ORT Server in PR #1562, the core of the implementation being [here](https://github.com/eclipse-apoapsis/ort-server/blob/05757af2290c42f94ed55df48b929115c60cd308/services/hierarchy/src/main/kotlin/VulnerabilityService.kt#L178-L188).

lamppu:
Calculate the overall rating for the vulnerability from the vulnerability references in the database query, and return this rating in the response of the vulnerabilities endpoint. This will allow for removing the calculation from the frontend side, and any future improvements in the calculation of the rating can be done in one place on the backend side.

Resolves #1839.

### DIFF
--- a/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
+++ b/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
@@ -515,7 +515,7 @@ fun ApiScannerJobConfiguration.mapToModel() = ScannerJobConfiguration(
 fun Secret.mapToApi() = ApiSecret(name, description)
 
 fun VulnerabilityWithIdentifier.mapToApi() =
-    ApiVulnerabilityWithIdentifier(vulnerability.mapToApi(), identifier.mapToApi())
+    ApiVulnerabilityWithIdentifier(vulnerability.mapToApi(), identifier.mapToApi(), rating.mapToApi())
 
 fun Vulnerability.mapToApi() = ApiVulnerability(externalId, summary, description, references.map { it.mapToApi() })
 
--- a/api/v1/model/src/commonMain/kotlin/VulnerabilityWithIdentifier.kt
+++ b/api/v1/model/src/commonMain/kotlin/VulnerabilityWithIdentifier.kt
@@ -27,5 +27,8 @@ import kotlinx.serialization.Serializable
 @Serializable
 data class VulnerabilityWithIdentifier(
     val vulnerability: Vulnerability,
-    val identifier: Identifier
+    val identifier: Identifier,
+
+    /** An advisory rating for the [Vulnerability], derived from the individual references of the [Vulnerability]. */
+    val rating: VulnerabilityRating,
 )
--- a/model/src/commonMain/kotlin/VulnerabilityWithIdentifier.kt
+++ b/model/src/commonMain/kotlin/VulnerabilityWithIdentifier.kt
@@ -27,5 +27,8 @@ import org.eclipse.apoapsis.ortserver.model.runs.advisor.Vulnerability
  */
 data class VulnerabilityWithIdentifier(
     val vulnerability: Vulnerability,
-    val identifier: Identifier
+    val identifier: Identifier,
+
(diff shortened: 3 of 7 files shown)


api/v1/mapping/src/commonMain/kotlin/Mappings.kt
```diff
@@ -515,7 +515,7 @@ fun ApiScannerJobConfiguration.mapToModel() = ScannerJobConfiguration(
 fun Secret.mapToApi() = ApiSecret(name, description)
 
 fun VulnerabilityWithIdentifier.mapToApi() =
-    ApiVulnerabilityWithIdentifier(vulnerability.mapToApi(), identifier.mapToApi())
+    ApiVulnerabilityWithIdentifier(vulnerability.mapToApi(), identifier.mapToApi(), rating.mapToApi())
```
sschuberth, line 518: Commit message nit: The wrapping seems to be too narrow, i.e. you wrap at column 55 (?) instead of 75.

api/v1/model/src/commonMain/kotlin/VulnerabilityWithIdentifier.kt
```diff
@@ -27,5 +27,8 @@ import kotlinx.serialization.Serializable
 @Serializable
 data class VulnerabilityWithIdentifier(
     val vulnerability: Vulnerability,
-    val identifier: Identifier
+    val identifier: Identifier,
+
+    /** An advisory rating for the [Vulnerability], derived from the individual references of the [Vulnerability]. */
+    val rating: VulnerabilityRating,
```
sschuberth, line 33: Please don't use trailing commas.

model/src/commonMain/kotlin/VulnerabilityWithIdentifier.kt
```diff
@@ -27,5 +27,8 @@ import org.eclipse.apoapsis.ortserver.model.runs.advisor.Vulnerability
  */
 data class VulnerabilityWithIdentifier(
     val vulnerability: Vulnerability,
-    val identifier: Identifier
+    val identifier: Identifier,
+
+    /** An advisory rating for the [Vulnerability], derived from the individual references of the [Vulnerability]. */
+    val rating: VulnerabilityRating,
```
sschuberth, line 33: Please don't use trailing commas.

services/hierarchy/src/main/kotlin/VulnerabilityService.kt
```diff
@@ -210,5 +214,6 @@ private fun ResultRow.toVulnerabilityWithAccumulatedData() = VulnerabilityWithAc
 
 private fun ResultRow.toVulnerabilityWithIdentifierAndReference() = VulnerabilityWithIdentifier(
     vulnerability = VulnerabilityDao.wrapRow(this).mapToModel(),
-    identifier = IdentifierDao.wrapRow(this).mapToModel()
+    identifier = IdentifierDao.wrapRow(this).mapToModel(),
+    rating = VulnerabilityRating.entries[(this[ratingAlias()] ?: 0)]
```
sschuberth, line 218: No need for parentheses inside the square brackets here.

Etsija:
Would be nice if you'd include a commit which does a corresponding clean-up for the UI.

sschuberth:
> I don't think we have any policy for it, so I can also do a follow-up PR once this is merged.

Yes, let's do that as a follow-up, to have the freedom that someone else than @lamppu is doing the UI part.

mnonnenmacher (approved):

PR #1912, opened 2025-01-30
feat(ui): Enhance "Rerun" button with ellipsis for clarity

wkl3nk:
Add an ellipsis ("Rerun...") to the button label to indicate that the user will be prompted to specify parameters before the scan is actually rerun. This improves UX by aligning with common UI conventions where an ellipsis suggests additional input or a dialog before an action executes.

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
@@ -85,7 +85,7 @@ const RunComponent = () => {
                         rerunIndex: ortRun.index,
                       }}
                     >
-                      Rerun
+                      Rerun...
                       <Repeat className='ml-1 h-4 w-4' />
                     </Link>
                   </Button>


nnobelis (approved):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
```diff
@@ -85,7 +85,7 @@ const RunComponent = () => {
                         rerunIndex: ortRun.index,
                       }}
                     >
-                      Rerun
+                      Rerun...
```
Etsija, line 88: As the `...repositories/{repoId}/runs` page, which shows the table of all runs for a repository, also has the rerun button, which is implemented differently to save in horizontal space, we should probably find a way to align the UI in both places. I'm not sure that an icon + "Rerun..." would be the correct solution.  Any suggestions?

Etsija (changes requested):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
wkl3nk: image

wkl3nk: image
Putting three dots to the right of the symbol would not look good. And I also did not see that before. So ... maybe it is better to forget the idea with the 3 dots ... not to have it this way in one part of the UI, and then different in another part.

nnobelis: Would it not be possible for this compact buttons to put the "..." in the tooltip ?

sschuberth: If we put something in the tooltip, then better a proper explanation Like "Rerun with edited parameters" or so.

sschuberth: Maybe in general the way to go is via a tooltip, and not change the button text at all.

mnonnenmacher: >Maybe in general the way to go is via a tooltip, and not change the button text at all.

I agree, I'm not sure if adding an ellipsis is clarifying anything.

Etsija: @wkl3nk you want to try out doing it with `Tooltip` component instead? There's plenty of examples in the UI code.

Etsija: The other "Rerun" button is [here](https://github.com/eclipse-apoapsis/ort-server/blob/c080593a1f3fca8befc7116b6cc13b42ffbf181c/ui/src/routes/organizations/%24orgId/products/%24productId/repositories/%24repoId/-components/repository-runs-table.tsx#L197-L218), and it actually already has a `Tooltip` specified, but maybe the text could be clearer.

PR #1924, opened 2025-01-30
refactor(ui): Make the delete confirmation dialog less verbose

sschuberth:


Etsija (commented):
The `DeleteDialog` component has had a DOM nesting bug from the start, which I'm bringing up here, as you are anyway modifying the component: `AlertDialogDescription` is specified as a HTML paragraph element, which according to the HTML standards [cannot contain any block elements inside it](https://github.com/shadcn-ui/ui/issues/693#issuecomment-1605514264). Therefore, any time a `DeleteDialog` is opened in the UI, browser throws an error, which you can see in the Web Developer Tool's console:
```
In HTML, <div> cannot be a descendant of <p>.
This will cause a hydration error.
```
To fix the errors, one would need to find a way to represent the contents inside `AlertDialogDescription` without using any block elements like `<div>` or `<p>` (`<span>` is an inline element, which is allowed in this case).  The best way I think is to not have one description, but several (as it's quite allowed).  This refactoring gets rid of the DOM nesting errors, as you can see, now the description components do not contain any HTML block elements:
```jsx
        <AlertDialogHeader>
          <div className='flex items-center'>
            <OctagonAlert className='h-8 w-8 pr-2 text-red-500' />
            <AlertDialogTitle>Confirm deletion</AlertDialogTitle>
          </div>
        </AlertDialogHeader>
        <div className='flex flex-col gap-2'>
          <AlertDialogDescription>
            Note that deletion is irreversible and might have unwanted side
            effects.
          </AlertDialogDescription>
          <AlertDialogDescription>
            If you are sure to delete the {thingName}{' '}
            <span className='font-bold'>{thingId}</span>, enter the bold text
            below for confirmation.
          </AlertDialogDescription>
          <AlertDialogDescription>
            <Input
              autoFocus
              value={input}
              onChange={(e) => setInput(e.target.value)}
            />
          </AlertDialogDescription>
        </div>
```
Maybe put that rearrangement of the DOM into a separate fix commit?

ui/src/components/delete-dialog.tsx
```diff
@@ -88,25 +97,20 @@ export const DeleteDialog = ({
...
+            <div>
+              If you are sure to delete the {thingName}{' '}
+              <span className='font-bold'>{thingId}</span>, enter the bold text
+              below for confirmation.
+            </div>
+            <Input
+              autoFocus
+              value={input}
+              onChange={(e) => setInput(e.target.value)}
+            />
```
lamppu: Shouldn't this only be shown if `thingId` is given?

ui/src/components/delete-dialog.tsx
```diff
@@ -95,24 +95,24 @@ export const DeleteDialog = ({
...
+          <AlertDialogDescription>
             <Input
               autoFocus
               value={input}
               onChange={(e) => setInput(e.target.value)}
             />
-          </div>
-        </AlertDialogDescription>
+          </AlertDialogDescription>
+        </div>
```
lamppu, line 115: Not really sure here, would it be better just to get rid of the `AlertDialogDescription` completely and just apply the classes that it brings `text-sm text-muted-foreground`  in the surrounding `div`? Does it really bring any extra value here when it needs to be repeated? Also I guess the one around the `Input` in this approach is quite redundant.

lamppu (changes requested):

ui/src/components/delete-dialog.tsx
Etsija: That would create this:
Screenshot from 2025-02-06 11-35-04

So for accessibility reasons, we need `AlertDialogDescription`.

ui/src/components/delete-dialog.tsx
lamppu: @Etsija did you accidentally do this change in the second commit, I think it belongs in the first commit

Etsija: Well dang.

lamppu (approved):

PR #1925, opened 2025-01-31
refactor(ui): Delete vulnerability rating calculation

Etsija:
This is a follow-up PR to #1903 .


lamppu:
I wonder if there's some way we could make sure that [this](https://github.com/doubleopen-project/ort-server/blob/main/ui/src/helpers/get-status-class.ts#L31-L38) is aligned with the API ratings, so that there would be some build error if that model was changed in the backend, so that it would need to be handled in the frontend too.

lamppu:
> Or hold on, sorry, actually, I _think_ that the `compareVulnerabilityRating` function would produce a build error in such a case actually.

Yes it does, just tested by modifying the `vulnerabilityRatings` a bit. So ignore these :smile:

lamppu (approved):

PR #1927, opened 2025-01-31
ui: Prevent sending empty environment variables to Analyzer

Etsija:

Screenshot from 2025-01-31 11-26-07


lamppu (approved):

sschuberth:
So do I get this correctly, that currently the back-end / API definition does not allow us to just define environment variables without explicitly setting a value for them (i.e. implicitly have empty values)?

PR #1937, opened 2025-02-03
feat(database): Remove entities orphaned with removing ort run

kamil-bielecki-bosch:
When archive scans are removed from database, some of orphaned entities stays in database, just occupying space, and deteriorating performance. To prevent this, obsolete (orphaned) records are deleted.

### DIFF
--- a/dao/src/main/resources/db/migration/V93__projectsPackagesCascadeDelete.sql
+++ b/dao/src/main/resources/db/migration/V93__projectsPackagesCascadeDelete.sql
@@ -1,0 +1,59 @@
+-- This migration script modifies the foreign key constraints of the tables that reference projects and packages
+-- to add an ON DELETE CASCADE clause. Cascade deletion is performed only on connecting (associating) tables to prevent
+-- deletion of actual data. Orphaned data removal is performed by another process.
+-- This script also creates necessary indexes to speed up queries.
+
+ALTER TABLE packages_declared_licenses
+    DROP CONSTRAINT packages_declared_licenses_package_id_fkey,
+  ADD CONSTRAINT packages_declared_licenses_package_id_fkey
+    FOREIGN KEY (package_id) REFERENCES packages (id) ON DELETE CASCADE;
+
+ALTER TABLE packages_authors
+    DROP CONSTRAINT packages_authors_package_id_fkey,
+  ADD CONSTRAINT packages_authors_package_id_fkey
+    FOREIGN KEY (package_id) REFERENCES packages (id) ON DELETE CASCADE;
+
+ALTER TABLE projects_declared_licenses
+    DROP CONSTRAINT projects_declared_licenses_project_id_fkey,
+  ADD CONSTRAINT projects_declared_licenses_project_id_fkey
+    FOREIGN KEY (project_id) REFERENCES projects (id) ON DELETE CASCADE;
+
+ALTER TABLE projects_authors
+    DROP CONSTRAINT projects_authors_project_id_fkey,
+  ADD CONSTRAINT projects_authors_project_id_fkey
+    FOREIGN KEY (project_id) REFERENCES projects (id) ON DELETE CASCADE;
+
+ALTER TABLE processed_declared_licenses
+    DROP CONSTRAINT processed_declared_licenses_package_id_fkey,
+  ADD CONSTRAINT processed_declared_licenses_package_id_fkey
+    FOREIGN KEY (package_id) REFERENCES packages (id) ON DELETE CASCADE;
+
+ALTER TABLE processed_declared_licenses
+    DROP CONSTRAINT processed_declared_licenses_project_id_fkey,
+  ADD CONSTRAINT processed_declared_licenses_project_id_fkey
+    FOREIGN KEY (project_id) REFERENCES projects (id) ON DELETE CASCADE;
+
+
+ALTER TABLE processed_declared_licenses_mapped_declared_licenses
+    DROP CONSTRAINT processed_declared_licenses_m_processed_declared_license_i_fkey,
+  ADD CONSTRAINT processed_declared_licenses_m_processed_declared_license_i_fkey
+    FOREIGN KEY (processed_declared_license_id) REFERENCES processed_declared_licenses (id) ON DELETE CASCADE;
+
+
+ALTER TABLE processed_declared_licenses_unmapped_declared_licenses
+    DROP CONSTRAINT processed_declared_licenses_u_processed_declared_license_i_fkey,
+  ADD CONSTRAINT processed_declared_licenses_u_processed_declared_license_i_fkey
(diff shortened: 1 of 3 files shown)


services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
```diff
@@ -0,0 +1,378 @@
+/*
```
oheger-bosch, line 1: Commit message nits:
* ORT is spelled in capital letters.
* Stick to the term "run" and avoid "scan" as synonym.
* Is the first sentence of the body grammatically correct?

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
```diff
@@ -0,0 +1,378 @@
...
+
+/**
+ * Service to remove orphaned entities.
+ */
+class OrphanRemovalService(
+    private val db: Database
+) {
+    private val logger = LoggerFactory.getLogger(OrphanRemovalService::class.java)
+
+    fun deleteRunsOrphanedEntities() {
```
oheger-bosch, line 77: Make this a `suspend` function; then you can skip the `runBlocking`. As a public function, it also should have a documentation.

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
```diff
@@ -0,0 +1,378 @@
...
+                        .union(
+                            ProjectsDeclaredLicensesTable
+                                .select(ProjectsDeclaredLicensesTable.projectId.alias("id"))
+                                .where(ProjectsDeclaredLicensesTable.projectId.isNotNull())
+                        )
+                    )
+            }
+        }
+
+    private suspend fun deleteOrphanedProducts() =
```
oheger-bosch, line 147: Do not delete products automatically. They are under the control of users. The same is true for repositories.

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
```diff
@@ -0,0 +1,378 @@
...
+                    )
+            }
+        }
+
+    private suspend fun deleteOrphanedDeclaredLicenses() =
+        db.dbQuery {
+            DeclaredLicensesTable.deleteWhere {
+                id notInSubQuery (
+                    PackagesDeclaredLicensesTable
+                        .select(PackagesDeclaredLicensesTable.declaredLicenseId.alias("id"))
```
oheger-bosch, line 210: I am not sure whether this will work. Isn't here a cyclic dependency between projects and packages on one side and declared licenses on the other side? So, a package/project will not be deleted because there is still a license, and a license will not be deleted because there is still a package/project?

Is this situation covered by the tests?

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
kamil-bielecki-bosch: Ok. That part has been removed

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
kamil-bielecki-bosch: First two points: fixed
Third one: I'm not an English Native, so have no idea. But changed description anyway to more descriptive one.

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
kamil-bielecki-bosch: Done.

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
kamil-bielecki-bosch: Added cascade delete constraints

oheger-bosch: Maybe I am missing something, but I don't think that the cascade option alone solves this problem.

In `deleteOrphanedPackages()` / `deleteOrphanedProjects()`, you add the condition that the ID must not be referenced by a declared license (via the `projects_declared_licenses` or `packages_declared_licenses` tables). But from checking on our database, every project / package is associated with a declared license. So, no projects / packages will ever be deleted.

The correct approach should delete the projects and packages that are no longer referenced by an ORT run. This may create orphans in declared licenses which then need to be deleted.

dao/src/main/resources/db/migration/V93__projectsPackagesCascadeDelete.sql
```diff
@@ -0,0 +1,23 @@
+-- This migration script modifies the foreign key constraints of the tables that reference projects and packages
+-- to add an ON DELETE CASCADE clause.
+
+ALTER TABLE packages_declared_licenses
```
oheger-bosch, line 4: Just for my understanding: As these tables define m:n relations between packages/projects and licenses, the effect of this cascade constraint is that only rows in these tables are removed automatically, but no license entities. Is this correct? Maybe add some more explanation to the comment, so that this is directly obvious.

dao/src/main/resources/db/migration/V93__projectsPackagesCascadeDelete.sql
```diff
@@ -0,0 +1,63 @@
+-- This migration script modifies the foreign key constraints of the tables that reference projects and packages
+-- to add an ON DELETE CASCADE clause. Cascade deletion is performed only on connecting (associating) tables to prevent
+-- deletion of actual data. Orphaned data removal is performed by another process
+-- This script also creates necessary indexes to speed up queries
```
oheger-bosch, line 4: Nit: Missing dots at end of sentences.

dao/src/main/resources/db/migration/V93__projectsPackagesCascadeDelete.sql
```diff
@@ -0,0 +1,63 @@
...
+
+
+ALTER TABLE processed_declared_licenses_unmapped_declared_licenses
+    DROP CONSTRAINT processed_declared_licenses_u_processed_declared_license_i_fkey,
+  ADD CONSTRAINT processed_declared_licenses_u_processed_declared_license_i_fkey
+    FOREIGN KEY (processed_declared_license_id) REFERENCES processed_declared_licenses (id) ON DELETE CASCADE;
+
+-- Create necessary indexes to speed up queries
+CREATE INDEX IF NOT EXISTS packages_analyzer_runs_package_id
+    ON packages_analyzer_runs USING btree
```
oheger-bosch, line 50: Nit: According to the [documentation](https://www.postgresql.org/docs/current/indexes-types.html), type `btree` is the default. So, the `USING` clause should be dropped to be in-line with other `CREATE INDEX` statements.

oheger-bosch (commented):
Two minor nits; otherwise looks good to me.

Do you have any estimate how long the single `DELETE` statements will take for their execution?

oheger-bosch (approved):

PR #1942, opened 2025-02-03
fix(cli): Update authentication details on token refresh

MarcelBochtler:


nnobelis (approved):

mnonnenmacher (approved):

mnonnenmacher (approved):

cli/src/main/kotlin/utils/AuthenticationUtils.kt
```diff
@@ -75,9 +75,10 @@ private fun createOrtServerClient(authDetails: HostAuthenticationDetails): OrtSe
 
                             AuthenticationStorage.store(updatedAuthDetails)
                         }
-                    }.getOrElse {
+                    }.getOrElse { t ->
```
mnonnenmacher, line 78: Nit: We usually use either the default `it` or `e` as common in catch blocks.

cli/src/main/kotlin/utils/AuthenticationUtils.kt
```diff
@@ -75,9 +75,10 @@ private fun createOrtServerClient(authDetails: HostAuthenticationDetails): OrtSe
...
                         throw AuthenticationException(
-                            "Authentication refresh failed.$it Please run '$COMMAND_NAME auth login'."
+                            "Authentication refresh failed. Please run '$COMMAND_NAME auth login'.",
```
mnonnenmacher, line 80: Nit: Maybe say "Authentication token refresh..."?

PR #1943, opened 2025-02-03
feat(auth): Add "offline_access" scope for long-lived refresh tokens

wkl3nk:
Add the `offline_access` scope when requesting authentication and refresh tokens in the CLI. This ensures that the authentication provider issues **long-lived refresh tokens**, allowing the CLI to obtain new access tokens without requiring the user to log in frequently.

This allows for having long-lived refresh tokens for the CLI, while still having the refresh tokens for the web application expire after a shorter period of time in case there is no user activity.

Such refresh token will no langer have a `exp` (expiration) claim. However, in authentication providers like Keycloak there **are** policies that enable to invalidate them nevertheless after a period of time of inactivity. This mechanism allows to have a regular automatic **clean-up of long-lived CLI sessions**, avoiding the number of long-lived sessions to grow in Keycloak without any bounds. 

While for web applications it is good practice to invalidate the refresh token after 15 - 30 minutes of inactivity, for CLIs it is common to have them last for several hours or even days.

### DIFF
--- a/api/v1/client/src/commonMain/kotlin/auth/AuthService.kt
+++ b/api/v1/client/src/commonMain/kotlin/auth/AuthService.kt
@@ -35,9 +35,9 @@ class AuthService(
     private val clientId: String
 ) {
     /**
-     * Generate a token for the given [username] and [password].
+     * Generate a token for the given [username] and [password] using password grant type with optional [scopes].
      */
-    suspend fun generateToken(username: String, password: String): TokenInfo =
+    suspend fun generateToken(username: String, password: String, scopes: Set<String> = emptySet()): TokenInfo =
         client.submitForm(
             url = tokenUrl,
             formParameters = Parameters.build {
@@ -45,6 +45,9 @@ class AuthService(
                 append("username", username)
                 append("password", password)
                 append("grant_type", "password")
+                if (scopes.isNotEmpty()) {
+                    append("scope", scopes.joinToString(" "))
+                }
             }
         ).let { response ->
             if (!response.status.isSuccess()) {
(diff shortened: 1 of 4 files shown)


api/v1/client/src/commonMain/kotlin/auth/AuthService.kt
```diff
@@ -45,6 +45,7 @@ class AuthService(
                 append("username", username)
                 append("password", password)
                 append("grant_type", "password")
+                append("scope", "offline_access")
```
MarcelBochtler, line 48: > This allows for having long-lived refresh tokens for the
CLI, while still having the refresh tokens for the web application
expire after a shorter period of time in case there is no user
activity.

You cannot be sure that this `AuthService` is only used from the CLI, as it is intentionally put into the `api` module.
Therefore, you cannot hard code the `offline_access` here.

The caller / user of `AuthService` should be able to set the scope when creating this service, or when actually generating the tokens.

MarcelBochtler (changes requested):

api/v1/client/src/commonMain/kotlin/auth/AuthService.kt
```diff
@@ -37,14 +37,17 @@ class AuthService(
     /**
      * Generate a token for the given [username] and [password].
      */
-    suspend fun generateToken(username: String, password: String): TokenInfo =
+    suspend fun generateToken(username: String, password: String, scopes: Set<String> = emptySet()): TokenInfo =
```
MarcelBochtler, line 40: Consider adding `[scopes]` to the documentation.
Same below.

MarcelBochtler (approved):

MarcelBochtler (approved):

PR #1947, opened 2025-02-03
feat(ui): Hide secret values by default

### LINKED ISSUE #1940: Hide secret values by default
In the form to add a new secret, the secret value is currently shown in plain text. Instead, the value field should behave like typical password fields in forms and obfuscate the value by default, and have a small button next to it to optionally show the value.

Image

mnonnenmacher:
Add a new `PasswordInput` component that hides the value by default. The input contains a button to toggle between showing and hiding the input.

Use this component to hide the values of secrets by default.

Fixes #1940.

### DIFF
--- a/ui/src/components/form/password-input.tsx
+++ b/ui/src/components/form/password-input.tsx
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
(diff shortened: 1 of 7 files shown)


Etsija:
LGTM. If you want to get rid of some, if not all, React errors regarding uncontrolled->controlled state in the secrets forms, please see my answer in #1946.

This could also be a separate follow-up PR, as I believe there are many forms for which we haven't yet set defaults, and they will give more errors to dev console.

Etsija (approved):

PR #1951, opened 2025-02-04
feat(ui): Change default page size for products and repositories to 10

mnonnenmacher:


mnonnenmacher:
@lamppu @Etsija I don't know what was the reason to set this to 5 for the product and repository lists, but on my QHD screen there is a lot of empty space below the table with the default setting, and I found myself constantly changing this to 10.

Etsija:
It was Sebastian's request, so maybe ask him before changing it back to 10 😀

Anyway, this should really be a personal choice, which would be stored to browser's local storage. IIRC we have an internal Double open ticket for that.

sschuberth:
I don't have any objection to change (though not everybody has an UHD screen), but would appreciate a rationale in the commit message (like alignment with other tables, if that's the case).

mnonnenmacher:
> Our ClearanceUI uses Zustand (yes, German) lightweight library for that kind of persistent state handling, and with it, it's pretty easy to implement support for persisting UI choices transparently.

Having that would be great, if we had that I'd be totally fine with keeping the current defaults. Do you think there will be time to implement that anytime soon?

Etsija:
> > Our ClearanceUI uses Zustand (yes, German) lightweight library for that kind of persistent state handling, and with it, it's pretty easy to implement support for persisting UI choices transparently.
> 
> Having that would be great, if we had that I'd be totally fine with keeping the current defaults. Do you think there will be time to implement that anytime soon?

I could save some spare time to have a look at it soon, as we have some good examples of the feature in our ClearanceUI. I'm currently in the middle of doing technical onboarding for a customer, so probably can't do it yet.

mnonnenmacher:
> I could save some spare time to have a look at it soon, as we have some good examples of the feature in our ClearanceUI. I'm currently in the middle of doing technical onboarding for a customer, so probably can't do it yet.
> 
> I created a ticket for this: #1954

Thanks, then I'll close this PR, storing the settings is a much better solution and I think for smaller screens the current defaults are better.

PR #1953, opened 2025-02-05
ui: Add dependency tree deduplication option for WebApp

Etsija:
This PR adds an option (off by default) to deduplicate the dependency tree for the different reports. This is a global option currently, but only used for the WebApp due to none of the other supported reports having the option implemented in ORT. 

Screenshot from 2025-02-05 08-50-32
Screenshot from 2025-02-05 08-51-23

When creating a new ORT run from scratch, the option needs to specifically be set to on, and even then it will only affect the WebApp, so there shouldn't be any side effects for any other reports as such.

### DIFF
--- a/ui/src/helpers/get-status-class.ts
+++ b/ui/src/helpers/get-status-class.ts
@@ -23,7 +23,7 @@ import {
   Severity,
   VulnerabilityRating,
 } from '@/api/requests';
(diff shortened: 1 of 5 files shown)


mnonnenmacher (approved):
LGTM, should @lamppu have a look as well?

PR #1958, opened 2025-02-05
Degrade SQL queries logging level to trace

kamil-bielecki-bosch:


dao/src/main/kotlin/SQLQueryTraceLogger.kt
```diff
@@ -0,0 +1,40 @@
...
+package org.eclipse.apoapsis.ortserver.dao
+
+import org.jetbrains.exposed.sql.SqlLogger
+import org.jetbrains.exposed.sql.Transaction
+import org.jetbrains.exposed.sql.exposedLogger
+import org.jetbrains.exposed.sql.statements.StatementContext
+import org.jetbrains.exposed.sql.statements.expandArgs
+import org.jetbrains.exposed.sql.transactions.TransactionManager
+
+object SQLQueryTraceLogger : SqlLogger {
```
oheger-bosch, line 29: Could you explain a bit why this object is needed? What is the effect if just the log level of the Exposed logger is set to TRACE?

core/src/main/resources/logback.xml
```diff
@@ -29,7 +29,7 @@
     </root>
     <logger name="org.eclipse.jetty" level="INFO"/>
     <logger name="io.netty" level="INFO"/>
-    <logger name="Exposed" level="DEBUG"/> <!-- Log SQL statements executed by Exposed -->
+    <logger name="Exposed" level="TRACE"/> <!-- Log SQL statements executed by Exposed -->
```
oheger-bosch, line 32: I think there are some more `logback.xml` files in other modules that would need to be adapted as well.

dao/src/main/kotlin/SQLQueryTraceLogger.kt
kamil-bielecki-bosch: If log level for Exposed will be set to TRACE, it will just lower minimum level of logged messages. 
In default configuration this -> https://github.com/JetBrains/Exposed/blob/c3029fe4b8a493451b9608b04a185a48b8c95624/exposed-core/src/main/kotlin/org/jetbrains/exposed/sql/SQLLog.kt#L29 logger is used to log SQL Queries. As it is in the code, it always logs messages of DEBUG level. To change this behavior, other logger have to be introduced, then level of logging decreased to TRACE.
And as `Slf4jSqlDebugLogger` is used as default SQL Query logger (https://github.com/JetBrains/Exposed/blob/c3029fe4b8a493451b9608b04a185a48b8c95624/exposed-core/src/main/kotlin/org/jetbrains/exposed/sql/DatabaseConfig.kt#L175) some of configuration has to be done for that.

oheger-bosch: One more nit regarding the class name: When it comes to acronyms, we typically use a spelling like `SqlQueryTraceLogger` instead of `SQLQueryTraceLogger` - but I have no strong feelings about this.

oheger-bosch (approved):

PR #1963, opened 2025-02-06
ui: Improve form handling

Etsija:
This PR gets rid of most React warnings about form components transferring from `uncontrolled` to `controlled` state by:
1. Providing default values for mandatory form fields.
2. Handling optional fields in a special way, also allowing proper `Zod` validation for optional fields, whenever user inputs something to them.

As an example of handling optional fields, the user creation form is modified accordingly. If this approach is acceptable, other forms would be fixed in a follow-up PR.

This is a more proper alternative to #1946.


mnonnenmacher (approved):
Looks good to me, but I would appreciate if @lamppu could also have a look as my React knowledge is limited.

lamppu (approved):

PR #1966, opened 2025-02-06
feat(reporter): Make source code origin priority configurable

bs-ondem:

### DIFF
--- a/workers/reporter/docker/Reporter.Dockerfile
+++ b/workers/reporter/docker/Reporter.Dockerfile
@@ -45,6 +45,8 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     apt-get update \
     && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
     git \
+    mercurial \
+    repo \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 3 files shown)


workers/reporter/docker/Reporter.Dockerfile
```diff
@@ -45,6 +45,9 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     apt-get update \
```
nnobelis, line 45: Commit message nit: instead of _commands_, maybe rather _programs_ or _executables_ or even _clients_ ?

workers/reporter/src/test/kotlin/reporter/SourceCodeBundleReporterTest.kt
nnobelis, line 1: Can you maybe add a test with `sourceCodeOrigins` set ?

workers/reporter/docker/Reporter.Dockerfile
bs-ondem: I would prefer to use `packages` to stay consistent with the apt-get [documentation](https://linux.die.net/man/8/apt-get).

nnobelis (approved):

PR #1982, opened 2025-02-10
fix(cli): Use the correct file permissions for the auth.yml file

MarcelBochtler:
Ensure that only the owner can read and write this file.


sschuberth:
> Ensure that only the owner can read and write this file.

Could you please add more detail to the commit message about what was wrong before, and what's correct now?

mnonnenmacher (approved):

PR #1987, opened 2025-02-10
chore(dao): Add indexes to speed up statistics queries

mnonnenmacher:
Add several indexes to speed up the queries executed by the statistics endpoints. Testing with several organizations and products with many repositories shows that these indexes improve performance by more than 50 percent.


mnonnenmacher:
Note that for organizations and products with a lot of repositories the endpoint is still slow and to make the endpoints fast also in those cases a different solution needs to be found. But adding the indexes at least mitigates the issue and reduces load on the database.

MarcelBochtler (approved):

oheger-bosch (approved):

PR #1993, opened 2025-02-11
fix(loki): Fix log level filter

mnonnenmacher:
Surround the regular expression to match log levels with parenthesis to not include the prefix `level=` in the first option. For example, instead of `level=INFO|WARN` the expression is now `level=(INFO|WARN)`. The old expression accidentally matched all log lines containing the string `WARN`.

### DIFF
--- a/logaccess/loki/src/main/kotlin/LokiLogFileProvider.kt
+++ b/logaccess/loki/src/main/kotlin/LokiLogFileProvider.kt
@@ -159,7 +159,7 @@ class LokiLogFileProvider(
         source: LogSource,
         levels: Set<LogLevel>
     ): String {
-        val levelCriterion = levels.joinToString("|") { it.name }
+        val levelCriterion = levels.joinToString(separator = "|", prefix = "(", postfix = ")") { it.name }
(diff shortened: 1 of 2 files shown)


MarcelBochtler (approved):

sschuberth:
@mnonnenmacher, does this change also require adjustments in our infrastructure setup? Because we somehow started to not see logs anymore, and this seems related.

mnonnenmacher:
> @mnonnenmacher, does this change also require adjustments in our infrastructure setup? Because we somehow started to not see logs anymore, and this seems related.

No, it only changes the way logs are queried for download but nothing else.

Etsija:
Well we are seeing that no logs whatsoever have arrived at our deployment environment after 10.2.2025
Screenshot from 2025-02-14 10-33-35

mnonnenmacher:
> Well we are seeing that no logs whatsoever have arrived at our deployment environment after 10.2.2025 Screenshot from 2025-02-14 10-33-35

That was one day before this PR was merged so the root cause must be something else.

PR #2000, opened 2025-02-12
feat(database): Create maintenance task for orphan deletion

kamil-bielecki-bosch:


tasks/src/main/kotlin/TaskRunner.kt
```diff
@@ -125,6 +126,7 @@ private fun tasksModule(): Module =
         singleOf(::ReportStorageService)
 
         single<Task>(named("delete-old-ort-runs")) { DeleteOldOrtRunsTask.create(get(), get()) }
+        single<Task>(named("delete-orphaned-ort-run-entities")) { DeleteOrphanedOrtRunEntitiesTask.create(get()) }
```
oheger-bosch, line 129: Not sure about the task name. Should it just be "delete-orphaned-entities"? These entities are not exclusively associated with a single ORT run.

tasks/src/main/kotlin/impl/DeleteOrphanedOrtRunEntitiesTask.kt
```diff
@@ -0,0 +1,51 @@
...
+         * Create a new instance of [DeleteOldOrtRunsTask].
+         */
+        fun create(orphanRemovalService: OrphanRemovalService): DeleteOrphanedOrtRunEntitiesTask {
+            return DeleteOrphanedOrtRunEntitiesTask(orphanRemovalService)
+        }
+    }
+
+    override suspend fun execute() {
+        logger.info("Deleting orphaned ORT run entities.")
+        orphanRemovalService.deleteRunsOrphanedEntities()
```
oheger-bosch, line 49: Would it be possible to log how many entities from which tables are deleted? Or is this already done in the service?

tasks/src/main/kotlin/impl/DeleteOrphanedOrtRunEntitiesTask.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.eclipse.apoapsis.ortserver.tasks.Task
+
+import org.slf4j.LoggerFactory
+
+/**
+ * A task class that handles the deletion of database entities orphaned after deletion of ORT run. The task delegates
+ * to [OrphanRemovalService] to trigger the deletion of entities that are not longer referencing ORT runs deleted
+ * during [DeleteOldOrtRunsTask] run.
+ */
+class DeleteOrphanedOrtRunEntitiesTask(
```
oheger-bosch, line 32: Please add a test.

tasks/src/main/kotlin/impl/DeleteOrphanedOrtRunEntitiesTask.kt
kamil-bielecki-bosch: It's done in the service itself. Every kind of entity deleted reports number of deleted entities:
https://github.com/eclipse-apoapsis/ort-server/blob/e77ca30c8da3949a92a5638633fd8ec808ce9045/services/hierarchy/src/main/kotlin/OrphanRemovalService.kt#L78

tasks/src/main/kotlin/TaskRunner.kt
kamil-bielecki-bosch: Changed.

tasks/src/main/kotlin/impl/DeleteOrphanedOrtRunEntitiesTask.kt
kamil-bielecki-bosch: Test added

kamil-bielecki-bosch: Test added

oheger-bosch (approved):

PR #2011, opened 2025-02-13
feat(database): Improvement on orphan deletion performance

kamil-bielecki-bosch:
To prevent long running single SQL queries, removing of orphaned entities is refactored to be performed in batches of 1000 records.

### DIFF
--- a/services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
+++ b/services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
@@ -52,8 +52,7 @@ import org.eclipse.apoapsis.ortserver.dao.tables.shared.RemoteArtifactsTable
 import org.eclipse.apoapsis.ortserver.dao.tables.shared.VcsInfoTable
 
 import org.jetbrains.exposed.sql.Database
-import org.jetbrains.exposed.sql.SqlExpressionBuilder.isNotNull
-import org.jetbrains.exposed.sql.SqlExpressionBuilder.notInSubQuery
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
 import org.jetbrains.exposed.sql.alias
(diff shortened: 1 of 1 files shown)


services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
```diff
@@ -66,6 +65,10 @@ import org.slf4j.LoggerFactory
 class OrphanRemovalService(
     private val db: Database
 ) {
+    companion object {
+        private const val DELETION_BATCH_SIZE = 100
```
mnonnenmacher, line 69: Commit message says 1000 but here the batch size is 100.

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
```diff
@@ -86,220 +89,324 @@ class OrphanRemovalService(
...
+            db.dbQuery {
+                PackagesTable.select(PackagesTable.id).where {
+                    PackagesTable.id notInSubQuery (
+                        PackagesAnalyzerRunsTable
+                            .select(PackagesAnalyzerRunsTable.packageId.alias("id"))
+                            .where(PackagesAnalyzerRunsTable.packageId.isNotNull())
+                        )
+                }.limit(DELETION_BATCH_SIZE).forEach {
+                    val packageId = it[PackagesTable.id]
+                    PackagesTable.deleteWhere { PackagesTable.id eq packageId }
```
mnonnenmacher, line 105: So actually you are not deleting batches of packages but individual packages now. Did you do any performance comparisons between before and after this change?

mnonnenmacher (changes requested):

kamil-bielecki-bosch:
After some tests performed this PR seems to be obsolete.

PR #2016, opened 2025-02-14
Add server version information to `--version`

MarcelBochtler:

### DIFF
--- a/cli/src/main/kotlin/OrtServerMain.kt
+++ b/cli/src/main/kotlin/OrtServerMain.kt
@@ -29,8 +29,10 @@ import com.github.ajalt.clikt.parameters.options.versionOption
 
 import kotlin.system.exitProcess
 
+import kotlinx.coroutines.runBlocking
 import kotlinx.serialization.json.Json
 
+import org.eclipse.apoapsis.ortserver.cli.utils.createOrtServerClient
 import org.eclipse.apoapsis.ortserver.client.OrtServerClient
 import org.eclipse.apoapsis.ortserver.client.OrtServerClientException
 import org.eclipse.apoapsis.ortserver.client.auth.AuthenticationException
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/api/VersionsRoute.kt
mnonnenmacher, line 1: Nit: "chore" would be the better commit type

core/src/main/kotlin/api/VersionsRoute.kt
```diff
@@ -21,7 +21,6 @@ package org.eclipse.apoapsis.ortserver.core.api
 
 import io.github.smiley4.ktorswaggerui.dsl.routing.get
 
-import io.ktor.server.application.call
```
sschuberth, line 24: Any idea why this was not catched by detekt?

cli/src/main/kotlin/OrtServerMain.kt
```diff
@@ -65,18 +67,36 @@ class OrtServerMain : SuspendingNoOpCliktCommand(COMMAND_NAME) {
...
+        val serverVersions = runCatching { createOrtServerClient()?.versions?.getVersions() }.getOrNull()
+
+        return buildString {
+            appendLine("$commandName version $cliVersion")
+            if (serverVersions != null) {
+                val serverVersion = serverVersions["ORT Server"]
+                val ortVersion = serverVersions["ORT Core"]
+
+                appendLine("Server version ${serverVersion ?: "<unknown>"}")
+                appendLine("ORT Core version ${ortVersion ?: "<unknown>"}")
```
mnonnenmacher, line 96: Maybe `<requires authentication>` instead?

cli/src/main/kotlin/OrtServerMain.kt
```diff
@@ -65,18 +67,36 @@ class OrtServerMain : SuspendingNoOpCliktCommand(COMMAND_NAME) {
 
         subcommands(AuthCommand(), RunsCommand())
 
+        @Suppress("ForbiddenMethodCall")
```
sschuberth, line 70: Maybe add a comment why it's ok to not use `org.eclipse.apoapsis.ortserver.utils.logging.runBlocking` here?

core/src/main/kotlin/api/VersionsRoute.kt
MarcelBochtler: No, unfortunately not.

cli/src/main/kotlin/OrtServerMain.kt
MarcelBochtler: <unknown> is not printed when the authentication is not available.
It is printed when the authentication was available, but the server returned a map that does not contain `ORT Server`.

cli/src/main/kotlin/OrtServerMain.kt
```diff
@@ -65,18 +67,36 @@ class OrtServerMain : SuspendingNoOpCliktCommand(COMMAND_NAME) {
...
+
+    /**
+     * Build the version information for the CLI and (if authenticated) for the ORT Server.
+     */
+    private suspend fun buildVersionInformation(cliVersion: String): String {
+        val serverVersions = runCatching { createOrtServerClient()?.versions?.getVersions() }.getOrNull()
+
+        return buildString {
+            appendLine("$commandName version $cliVersion")
+            if (serverVersions != null) {
```
oheger-bosch, line 91: Should an error message be printed if the server versions could not be obtained?

cli/src/main/kotlin/OrtServerMain.kt
MarcelBochtler: The CLI does not have a dependency to the utils package, and at this point I don't want to introduce it (due to #2031).
Maybe we should introduce an exception for the `ForbiddenMethodCall` for the CLI?

cli/src/main/kotlin/OrtServerMain.kt
MarcelBochtler: I'm not sure about this.
This change was not intended to create some sort of `osc status` command.
Printing the version should IMO work without printing an error if the user is not authenticated or not connected to a network.

cli/src/main/kotlin/OrtServerMain.kt
sschuberth: > Maybe we should introduce an exception for the `ForbiddenMethodCall` for the CLI?

Sounds reasonable to me, if it can be done with detekt.

sschuberth: Looks like the [generic `excludes` mechanism](https://detekt.dev/docs/introduction/configurations#path-filters--excludes--includes) could be used for this.

mnonnenmacher (approved):

PR #2017, opened 2025-02-14
feat(workers): Parallelize file list downloads

mnonnenmacher:
Parallelize the download of file lists when generating an ORT result to improve performance. In tests this made the download of file lists about 80% faster.

### DIFF
--- a/workers/common/src/main/kotlin/common/Utils.kt
+++ b/workers/common/src/main/kotlin/common/Utils.kt
@@ -19,6 +19,12 @@
 
 package org.eclipse.apoapsis.ortserver.workers.common
 
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+
+import org.eclipse.apoapsis.ortserver.utils.logging.runBlocking
+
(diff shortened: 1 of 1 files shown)


sschuberth (approved):

workers/common/src/main/kotlin/common/Utils.kt
```diff
@@ -36,11 +42,15 @@ fun enableOrtStackTraces() {
...
-        .mapTo(mutableSetOf()) { (provenance, fileList) ->
-            FileList(
-                provenance,
-                fileList.files.mapTo(mutableSetOf()) { FileList.Entry(it.path, it.sha1) }
-            )
+    runBlocking(Dispatchers.IO.limitedParallelism(20)) {
+        provenances.map { provenance ->
+            async { provenance to fileListResolver.get(provenance) }
+        }.awaitAll().mapNotNull { (provenance, fileList) ->
+            fileList?.let {
```
oheger-bosch, line 49: This transformation could have been done asynchronously as well.

mnonnenmacher: @oheger-bosch See #2030.

PR #2035, opened 2025-02-18
fix(database): Add missing constraint for project cascade deletion

kamil-bielecki-bosch:
Add project_scopes_project_id_fkey constraint for project_scopes table to enable cascade deletion of orphaned projects

### DIFF
--- a/dao/src/main/resources/db/migration/V95__projectsOrphanEntitiesFix.sql
+++ b/dao/src/main/resources/db/migration/V95__projectsOrphanEntitiesFix.sql
@@ -1,0 +1,6 @@
+-- This migration script modifies the foreign key constraint on project_scopes table that reference projects table.
+
+ALTER TABLE project_scopes
+    DROP CONSTRAINT project_scopes_project_id_fkey,
+  ADD CONSTRAINT project_scopes_project_id_fkey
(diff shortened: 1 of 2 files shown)


dao/src/main/resources/db/migration/V95__projectsOrphanEntitiesFix.sql
```diff
@@ -0,0 +1,6 @@
+-- This migration script modifies the foreign key constraint on project_scopes table that reference projects table.
```
oheger-bosch, line 1: Commit message: Can you mention that this is a fixup of the commit that originally introduced the orphan deletion?

kamil-bielecki-bosch: Done.

oheger-bosch: The commit lint check is now failing. It should be sufficient to use the short Git commit SHA hash; GitHub will render this as a link.

kamil-bielecki-bosch: Looks like it's enough to use [note] notation with long link and it's fine.

PR #2036, opened 2025-02-18
Use the default remote branch in case of an empty revision

sschuberth:


sschuberth:
> This should not be a `docs` commit as it has user-facing effects.

Huh, isn't documentation pretty much always user-facing? Like on a web page? The other things are code comments 😉

sschuberth:
Putting to draft state as it seems the "empty revision means default" logic only applies to configuration provider, but not to the analyzer job cloning the project repository.

mmurto:
> > This should not be a `docs` commit as it has user-facing effects.
> 
> Huh, isn't documentation pretty much always user-facing? Like on a web page? The other things are code comments 😉

I'll rephrase user-facing as something that affects the versioned build output. Code comments or web page changes don't change what the user sees if they deploy a release after them, but changes to UI text do. At least the release automation tools that I've used (like [Release Please](https://github.com/googleapis/release-please)) don't generate new releases for documentation changes, but a release should be generated for UI improvements like this.

Etsija:
> Putting to draft state as it seems the "empty revision means default" logic only applies to configuration provider, but not to the analyzer job cloning the project repository.

We should maybe create an issue for this.

sschuberth:
> We should maybe create an issue for this.

I was actually planning to solve this as part of this PR.

mmurto (changes requested):
I believe the commit should be `feat`. Of course we haven't really defined the types in detail, but IMO chore should have no visible effects. Some points [here](https://stackoverflow.com/a/26944812).

sschuberth:
> I believe the commit should be `feat`.

Thanks for finally coming up with what it *should* be in your opinion, after telling me that it should *not* be `docs` or `chore` 😉 

That said, I really believe it to be exaggerated to call a wording improvement (user-facing or not) a feature... but hey, if it helps to get an approval, I'll reword.

mmurto (approved):

sschuberth:
> I really believe it to be exaggerated to call a wording improvement (user-facing or not) a feature...

One last additional thought: According to semantic versioning, a new feature implies a minor version bump. And I really don't think the minor version should be bumped due to a wording change like this. Here, it does not matter, however, as there are other "real" `feat` commits in the PR.

mmurto:
> > I really believe it to be exaggerated to call a wording improvement (user-facing or not) a feature...
> 
> 
> 
> One last additional thought: According to semantic versioning, a new feature implies a minor version bump. And I really don't think the minor version should be bumped due to a wording change like this. Here, it does not matter, however, as there are other "real" `feat` commits in the PR.

I agree that it feels a little big to bump minor for it, but if following semantic versioning, chores and docs don't usually bump anything, which means that this sort of improvements would not be released and deployed on their own, but they would be kept waiting for bigger feats and fixes. I think that's a worse option.

PR #2044, opened 2025-02-19
test(ui): Fix one failing unit test

Etsija:

### DIFF
--- a/ui/src/lib/utils.ts
+++ b/ui/src/lib/utils.ts
@@ -52,8 +52,10 @@ if (import.meta.vitest) {
       '11.06.2024, 13:07:45'
     );
 
-    expect(formatTimestamp('2024-06-11T13:07:45Z', 'UTC', 'fi-FI')).toBe(
-      '11.06.2024 klo 13.07.45'
+    // Test the Finnish locale with a custom format. Matching "klo" is made optional
+    // due to vitest isolated test environment failing to support the full ICU format.
+    expect(formatTimestamp('2024-06-11T13:07:45Z', 'UTC', 'fi-FI')).toMatch(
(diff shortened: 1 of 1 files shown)


ui/src/lib/utils.ts
```diff
@@ -52,8 +52,10 @@ if (import.meta.vitest) {
       '11.06.2024, 13:07:45'
     );
 
-    expect(formatTimestamp('2024-06-11T13:07:45Z', 'UTC', 'fi-FI')).toBe(
-      '11.06.2024 klo 13.07.45'
+    // Test the Finnish locale with a custom format. Matching "klo" is made optional
+    // due to vitest isolated test environment failing to support the full ICU format.
+    expect(formatTimestamp('2024-06-11T13:07:45Z', 'UTC', 'fi-FI')).toMatch(
```
mmurto, line 57: I'd use [`toEqual`](https://vitest.dev/api/expect#toequal) instead to avoid the regexp.

mmurto: Oh it was actually needed. :D

mmurto (approved):

sschuberth:
Any idea why this was not catch by us running the test in CI?

https://github.com/eclipse-apoapsis/ort-server/blob/89cb8b164f99b93f03676dfb9015978d18f75951/.github/workflows/build-and-test.yml#L59-L60

sschuberth:
PS: The original test also passes fine for me locally.

Etsija:
This seems to actually be passing for everyone else but me, so I will investigate my local environment.

Etsija:
Tried so many things, verified everything, but locally, this one test still fails for me. I'll probably close this PR and just bear with it failing, as I cannot find the root cause to fix it for me locally.

PR #2058, opened 2025-02-19
feat(workers): Add sleep loop to keep pod alive for problem analysis

wkl3nk:
This PR is split up in two commits: 

- One commit deals with sending the worker pods to sleep as long a `keep-alive.lock` file exists in the user's home directory. This lock file can be created manually by the user by **quickly** creating a terminal session in the respective worker pod and create the file. After that, manual problem analysis can take place. Having finished the analysis, the user can remove the lock file, and the worker pod will finish and hand over to the next worker pod in the sequence.
- The other commit deals with creating the lock file `keep-alive.lock` by means of a job configuration of the respective worker pod. There is no longer the need to manually create this lock file, but the user still has the responsibility to remove the lock file after the problem analysis is finished.

For details, see the respective commit messages.


transport/spi/src/main/kotlin/EndpointComponent.kt
```diff
@@ -52,6 +55,27 @@ abstract class EndpointComponent<T : Any>(
...
+         */
+        fun sleepWhileKeepAliveFileExists() {
+            val checkIntervalSeconds = 60L
+            val file = File(System.getProperty("user.home"), "keep-alive.lock")
+            while (file.exists()) {
+                logger.info(
+                    "Keep alive lock file ${file.absolutePath} exists. " +
+                        "Sleeping for $checkIntervalSeconds seconds until next check ..."
+                )
+                TimeUnit.SECONDS.sleep(checkIntervalSeconds)
```
mnonnenmacher, line 74: The endpoint handlers that are calling this function are suspend functions which should never call `Thread.sleep` inside. Instead, make this also a suspend function and use `delay(60.seconds)`.

mnonnenmacher (changes requested):

workers/advisor/src/main/kotlin/advisor/AdvisorComponent.kt
```diff
@@ -77,6 +77,9 @@ class AdvisorComponent : EndpointComponent<AdvisorRequest>(AdvisorEndpoint) {
                 is RunResult.Ignored -> null
             }
 
+            // Check if there is a demand to keep the pod alive for manual problem analysis.
+            sleepWhileKeepAliveFileExists()
```
oheger-bosch, line 81: I think it should be possible to have this line at the end of the `start()` function in `EndpointComponent` instead of duplicating it in all workers.

transport/spi/src/main/kotlin/EndpointComponent.kt
```diff
@@ -52,6 +55,27 @@ abstract class EndpointComponent<T : Any>(
...
 ) : KoinComponent {
+    companion object {
+        private val logger: Logger = LoggerFactory.getLogger(EndpointComponent::class.java)
+
+        /**
+         * Sleep while a keep-alive file exists. This is helpful in case a user terminal session is opened in the
+         * Kubernetes pod, and the pod should not terminate immediately after its work is done, giving the user
+         * arbitrary extra time for detailed problem analysis directly in the pod's execution environment.
+         */
+        fun sleepWhileKeepAliveFileExists() {
```
oheger-bosch, line 66: This new feature should be documented more prominently; maybe the class comment for `EndpointComponent` could be a good place?

oheger-bosch (commented):
This is already quite helpful. I see the following problems or missing features (which not necessarily have to be addressed in this PR):
* Pods may terminate fast, so that there is not sufficient time to log in and create a lock file. For instance, when an Analyzer pod fails to check out the repository.
* Especially in the Analyzer, I would love to have more control when the sleep happens. For instance, if the Analyzer would wait after the checkout, one could do some modifications on the checked out sources or configuration files before letting the analysis run. This is then probably a feature specific to the Analyzer.

transport/spi/src/main/kotlin/EndpointComponent.kt
wkl3nk: Done.

workers/advisor/src/main/kotlin/advisor/AdvisorComponent.kt
wkl3nk: About position of this "sleep" in the code. Tried it out, does not work in the `start()` function, probably as this only seems to create a message receiver.

transport/spi/src/main/kotlin/EndpointComponent.kt
wkl3nk: Added some sentences to the class documentation.

workers/advisor/src/main/kotlin/advisor/AdvisorComponent.kt
wkl3nk: @oheger-bosch : You comment about "pods may terminate so fast ... to manually create a lock file". I have added job configuration parameters for the respective worker pods that allow to create the lock file by configuration.

wkl3nk: @oheger-bosch : "More control when the sleep happens". Yes ... probably nice to have ... but for the beginning, can we keep it simple? I guess this already is good enough for 95% of problem cases to analyze.

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
```diff
@@ -98,7 +98,13 @@ data class AnalyzerJobConfiguration(
...
      * A flag to control whether excluded scopes and paths should be skipped during the analysis.
      */
-    val skipExcluded: Boolean? = null
+    val skipExcluded: Boolean? = null,
+
+    /**
+     * Keep the worker alive after it has finished. This is useful for manual problem analysis directly
+     * within the pod's execution environment.
+     */
+    val keepAliveWorker: Boolean? = null
```
oheger-bosch, line 107: Can't these be non-nullable booleans? I assume, they have been made nullable to make them optional? But if a default value of *false* is assigned, this should achieve the same effect, and code does not have to deal with ternary logic (true/false/null).

transport/spi/src/main/kotlin/EndpointComponent.kt
```diff
@@ -80,6 +82,14 @@ abstract class EndpointComponent<T : Any>(
                 delay(CHECK_INTERVAL_SECONDS * 1000)
             }
         }
+
+        suspend fun generateKeepAliveFile() =
```
oheger-bosch, line 86: Please add a comment for this function.

transport/spi/src/main/kotlin/EndpointComponent.kt
```diff
@@ -80,6 +82,14 @@ abstract class EndpointComponent<T : Any>(
...
+            withContext(Dispatchers.IO) {
+                val file = File(System.getProperty("user.home"), KEEP_ALIVE_FILE_NAME)
```
oheger-bosch, line 88: Suggestion: This code to locate the lock file is also used when checking for its presence. Maybe it can be extracted to a helper function.

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
wkl3nk: The reason behind was that these options appear in the (resolved) JobConfiguration(s) every time and for each of the workers, if I make them non-nullable booleans. I did not want to "pollute" the JobConfiguration with things that are only required by developers in some special problem solving places. @oheger-bosch Can you give me a short hint please if you still think that non-nullable is the way to go after having read this?

oheger-bosch: Well, I am bit undecided. Technically, there is no distinction between options used by end users and options for diagnostic purposes; so I find it a bit strange to treat them differently and try to hide them. But I have no strong feelings about this.

wkl3nk: Thinking about it, I think you are right: Let's make it transparent and not try to hide it.

transport/spi/src/main/kotlin/EndpointComponent.kt
```diff
@@ -65,21 +67,38 @@ abstract class EndpointComponent<T : Any>(
...
         /**
          * Sleep while a keep-alive file exists. This is helpful in case a user terminal session is opened in the
          * Kubernetes pod, and the pod should not terminate immediately after its work is done, giving the user
          * arbitrary extra time for detailed problem analysis directly in the pod's execution environment.
          */
         suspend fun sleepWhileKeepAliveFileExists() {
-            val file = File(System.getProperty("user.home"), KEEP_ALIVE_FILE_NAME)
-            while (file.exists()) {
+            val file = getKeepAliveFile()
+            while (getKeepAliveFile().exists()) {
```
oheger-bosch, line 82: Nit: This should probably be `while (file.exists())` after extracting the variable.

oheger-bosch (dismissed):
One nit, but feel free to ignore.

oheger-bosch (approved):

PR #2061, opened 2025-02-19
build: Add Docker-Compose-based IntelliJ run configurations for debugging

sschuberth:
This adds a bunch of run configurations for local development where everything, incl. the UI, is built from the latest source code. This allows to conveniently set breakpoints e.g. in the backend code, which is triggered by user actions in the frontend running on localhost.

To make this work, a developer first needs to launch the "Compose Latest" run configuration to build all Docker images locally, which will take some time. Once all containers are up, one or multiple of the remote debug configurations can be attached, depending on which Gradle modules the developer wants to be able to hit breakpoints. For now, only three modules are pre-configured: The orchestrator, the analyzer worker, and the advisor worker. Each come with their own remote debugging port configurations which needs to match between `docker-compose.yml` (where the port needs to be exposed) and the debug configuration in IntelliJ.

### DIFF
--- a/orchestrator/build.gradle.kts
+++ b/orchestrator/build.gradle.kts
@@ -66,5 +66,9 @@ jib {
     container {
         mainClass = "org.eclipse.apoapsis.ortserver.orchestrator.EntrypointKt"
         creationTime.set("USE_CURRENT_TIMESTAMP")
+
+        if (System.getProperty("idea.active").toBoolean()) {
+            jvmFlags = listOf("-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5010")
+        }
     }
 }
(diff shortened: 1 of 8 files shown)


mnonnenmacher (approved):

orchestrator/build.gradle.kts
```diff
@@ -66,5 +66,9 @@ jib {
     container {
         mainClass = "org.eclipse.apoapsis.ortserver.orchestrator.EntrypointKt"
         creationTime.set("USE_CURRENT_TIMESTAMP")
+
+        if (System.getProperty("idea.active").toBoolean()) {
```
bahoop, line 70: What happens if this is executed outside Intellij? Is there a "nullpointer" risk in this gradle scripts? I really like to have this permanent enabled but may think about:

- it will all the time block / open ports on the local machine and also is from a security point of view an not expected default behavior when a developer use the "run" instead of "run debug" configuration.
- the entire project build script contains vendor-specific logic

PR #2062, opened 2025-02-19
feat: Forward worker error messages to the orchestrator

### LINKED ISSUE #2034: Do not fail with "The analyzer worker failed due to an unexpected error" if a non-exitsing branch name is specified
Currently, the analyzer fails with an issue just saying "The analyzer worker failed due to an unexpected error" if a non-existing branch name was specified got a run. We should do a better job and display a more telling issue message to the user here.

sschuberth:
Previously, the orchestrator just created generic issues like "The analyzer worker failed due to an unexpected error", swallowing the message of the underyling exception that occurred in the worker.

This change adds code to forward exception messages to be able to create meaningful issue messages.

Resolves #2034.

### DIFF
--- a/model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
+++ b/model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
@@ -24,6 +24,14 @@ import kotlinx.serialization.Serializable
 import org.eclipse.apoapsis.ortserver.model.OrtRun
 
 /**
+ * A common interface for worker errors that may provide an error message. This is used to forward original exception
+ * messages to the orchestrator to create issues with meaningful messages.
+ */
+interface WorkerErrorMessage {
+    val errorMessage: String?
+}
+
+/**
  * Base class for the hierarchy of messages that can be processed by the Orchestrator component.
  */
 @Serializable
@@ -44,8 +52,10 @@ data class ConfigWorkerResult(
 @Serializable
 data class ConfigWorkerError(
     /** The ID of the ORT run on which the worker failed. */
-    val ortRunId: Long
-) : OrchestratorMessage()
+    val ortRunId: Long,
+    /** The message of the error, if any. */
+    override val errorMessage: String? = null
+) : OrchestratorMessage(), WorkerErrorMessage
 
 /**
  * A common interface for messages that are sent by workers to the Orchestrator. The interface allows access to the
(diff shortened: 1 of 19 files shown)


model/src/commonMain/kotlin/orchestrator/OrchestratorMessage.kt
```diff
@@ -23,6 +23,14 @@ import kotlinx.serialization.Serializable
 
 import org.eclipse.apoapsis.ortserver.model.OrtRun
 
+/**
+ * A common interface for worker errors that may provide an error message. This is used to forward original exception
+ * messages to the orechstrator to create issues with meaningful messages.
```
mnonnenmacher, line 28: Typo in "orchestrator".

mnonnenmacher (changes requested):

orchestrator/src/main/kotlin/Orchestrator.kt
```diff
@@ -472,12 +474,12 @@ class Orchestrator(
 typealias CreatedJobs = List<JobScheduleFunc>
 
 /**
- * Create an [Issue] object representing an error that occurred in any [Endpoint].
+ * Create an [Issue] representing the error with the given [message], if any, that occurred in an [Endpoint].
  */
-fun <T : Any> Endpoint<T>.createErrorIssue(): Issue = Issue(
+fun <T : Any> Endpoint<T>.createErrorIssue(message: String?): Issue = Issue(
```
oheger-bosch, line 479: Nit: Should this function expect the whole `WorkerMessage` object instead of the message string and handle the extraction of the message itself? Then it would be fully responsible for the issue creation, and `handleWorkerError` would not need to partially deal with this.

sschuberth: I basically like the idea. However, as `handleWorkerError` takes a `WorkerMessage`, and that interface has nothing in common with `WorkerErrorMessage`, this requires `createErrorIssue` to take `error: Any`. But that's probably acceptable as the code still looks cleaner, IMO.

oheger-bosch:
Looks like there is a related test failure.

oheger-bosch (dismissed):

sschuberth:
> > Looks like there is a related test failure.
> 
> Just fixed it.

Sorry, there were yet more to fix.

oheger-bosch (approved):

PR #2065, opened 2025-02-20
ui: Show the shortest dependency paths

Etsija:
This is the UI counterpart for #1882. Please note that this whole package details section will be revamped in later PRs, so this is only the initial/minimal implementation for the UI.

Screenshot from 2025-02-20 08-38-38

Question: Do we always want to show the "Shortest dependency paths" section in the package details, and notify the user in case there are none found? The most prominent reason for this being that the run is so old that the data isn't available, so the user should rerun to get the paths.


sschuberth:
> The most prominent reason for this being that the run is so old that the data isn't available, so the user should rerun to get the paths.

Isn't this even the *only* reason? Anyway, I'd say if the data is not available, skip displaying the whole section.

lamppu (approved):

PR #2103, opened 2025-02-24
CLI: Build native binaries on their respective platforms

MarcelBochtler:

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -64,6 +64,9 @@ jobs:
         mkdir -p ${{ matrix.target.artifact }}
         mv ${{ matrix.target.buildPath }} ${{ matrix.target.artifact }}/${{ matrix.target.binName }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -33,12 +33,21 @@ jobs:
           - name: 'Linux x64'
             os: ubuntu-24.04
             task: 'linkReleaseExecutableLinuxX64'
+            artifact: 'osc-cli-linux-x64'
+            buildPath: 'cli/build/bin/linuxX64/releaseExecutable/osc.kexe'
+            binName: 'osc'
```
bs-ondem, line 38: Do you need the binary name as property, since it's always `osc`?

MarcelBochtler: For now, it is always `osc`. For Windows, it will be `osc.exe`, that's why I'm already making it configurable.

bs-ondem (approved):

PR #2101, opened 2025-02-24
feat(cli): Add Windows native builds

MarcelBochtler:

PR #2111, opened 2025-02-25
Release: Add CLI as release artifact

MarcelBochtler:
I tested this change on my personal fork.
You can see there how a new release will look like: https://github.com/MarcelBochtler/ort-server/releases/tag/0.1.0-RC30

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -70,8 +70,63 @@ jobs:
         SONATYPE_CONNECT_TIMEOUT_SECONDS: 300
       run: ./gradlew --stacktrace --no-configuration-cache publishAndReleaseToMavenCentral
 
+  build-cli:
+    strategy:
+      matrix:
+        target:
+        - name: "Linux x64"
+          os: ubuntu-24.04
+          task: :cli:linkReleaseExecutableLinuxX64
+          artifact: osc-cli-linux-x64
+          buildPath: cli/build/bin/linuxX64/releaseExecutable/osc.kexe
+          binName: osc
+        - name: "macOS arm64"
+          os: macos-15
+          task: :cli:linkReleaseExecutableMacosArm64
+          artifact: osc-cli-macos-arm64
+          buildPath: cli/build/bin/macosArm64/releaseExecutable/osc.kexe
+          binName: osc
+        - name: "macOS x64"
+          os: macos-13
+          task: :cli:linkReleaseExecutableMacosX64
+          artifact: osc-cli-macos-x64
+          buildPath: cli/build/bin/macosX64/releaseExecutable/osc.kexe
+          binName: osc
+    runs-on: ${{ matrix.target.os }}
+    steps:
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      with:
+        ref: ${{ env.ORT_SERVER_VERSION }}
+        fetch-depth: 0
+
+    - name: Setup Gradle
+      uses: gradle/actions/setup-gradle@94baf225fe0a508e581a564467443d0e2379123b # v4
+      with:
+        dependency-graph: generate-and-submit
+
+    - name: Build CLI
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -70,8 +70,63 @@ jobs:
...
+    - name: Rename binaries
+      run: |
+        mkdir -p ${{ matrix.target.artifact }}
+        mv ${{ matrix.target.buildPath }} ${{ matrix.target.artifact }}/${{ matrix.target.binName }}
+
+    # Compress the binaries to reduce the size of the artifacts, to keep `osc` as the binary name, and to ensure that
+    # the executable permissions are preserved.
+    - name: Compress binaries
+      run: |
+        tar -czf ${{ matrix.target.artifact }}.tar.gz -C ${{ matrix.target.artifact }} ${{ matrix.target.binName }}
```
mnonnenmacher, line 119: Should we also publish checksums of the archives?

.github/workflows/release.yml
```diff
@@ -92,4 +149,8 @@ jobs:
         if [[ "$ORT_SERVER_VERSION" =~ -RC[0-9]+ ]]; then
           PRERELEASE_ARG="--prerelease"
         fi
-        gh release create $ORT_SERVER_VERSION --notes-file RELEASE_NOTES.md $PRERELEASE_ARG
+        
+        gh release create $ORT_SERVER_VERSION --notes-file ./artifacts/release-notes/RELEASE_NOTES.md $PRERELEASE_ARG \
```
mnonnenmacher, line 153: Should the `release-notes` directory already be added in the previous commit?

.github/workflows/release.yml
MarcelBochtler: Yes we should.
Added checksum creation and upload:
https://github.com/MarcelBochtler/ort-server/releases/tag/0.1.0-RC38

mnonnenmacher: Just curious, what was the motivation to put the checksums into a single file?

MarcelBochtler: 1. `gh` is doing it like this: https://github.com/cli/cli/releases/tag/v2.67.0
2. I may avoid a potential issue when adding Windows builds, that are built on Windows, and I didn't want to deal with checksum creation on Windows (easily avoidable by creating the checksums in a loop).
3. Less clutter in the release assets.

But if you see a reason to do it in separate files, I would be fine with doing it.

.github/workflows/release.yml
```diff
@@ -92,4 +149,8 @@ jobs:
         if [[ "$ORT_SERVER_VERSION" =~ -RC[0-9]+ ]]; then
           PRERELEASE_ARG="--prerelease"
         fi
-        gh release create $ORT_SERVER_VERSION --notes-file release-notes/RELEASE_NOTES.md $PRERELEASE_ARG
+
+        gh release create $ORT_SERVER_VERSION --notes-file ./artifacts/release-notes/RELEASE_NOTES.md $PRERELEASE_ARG \
+          './artifacts/osc-cli-linux-x64/osc-cli-linux-x64.tar.gz#osc-cli-linux-x64' \
+          './artifacts/osc-cli-macos-arm64/osc-cli-macos-arm64.tar.gz#osc-cli-macos-arm64' \
+          './artifacts/osc-cli-macos-x64/osc-cli-macos-x64.tar.gz#osc-cli-macos-x64'
```
nnobelis: So in the end, there is no Windows release ?

MarcelBochtler: See https://github.com/eclipse-apoapsis/ort-server/pull/2101. 
My plan is, to also offer Windows binaries.

.github/workflows/release.yml
mnonnenmacher: It's a bit easier to verify the download if you can directly download the checksum, but that's not so important.

mnonnenmacher (approved):

PR #2114, opened 2025-02-25
chore(cli): Ignore disable targets in native builds

MarcelBochtler:
This removes a warning during project configuration when macOS builds cannot be built.


gradle.properties
```diff
@@ -24,6 +24,7 @@ org.gradle.parallel = true
 
 kotlin.code.style = official
 kotlin.mpp.stability.nowarn = true
+kotlin.native.ignoreDisabledTargets=true
```
sschuberth, line 27: Nit: This should have spaces around `=` to align with existing code.

sschuberth (approved):

Issue #1916, opened 2025-01-30
Favicon not shown for nested URL paths


sschuberth:
It seems that the favicon is now only being displayed for the top-level URL, but not anymore when navigating into an organization / product / repository etc.

Maybe https://github.com/eclipse-apoapsis/ort-server/pull/1763 broke it.

Etsija:
It's not missing from Docker Compose environment, only from our deployment env.

mnonnenmacher:
For me it works as long as I don't reload the page from a subpage. I guess [this](https://github.com/eclipse-apoapsis/ort-server/pull/1763#issuecomment-2582541129) is the way to go.

mnonnenmacher:
I have tried setting the `base` to `./` as proposed in the link, but it did not work because now it's trying to load the assets relative to the page, so when I load `/organizations/1` the paths for the JS and CSS assets would be `/organization/1/assets/...`. This is not what I expected after reading the docs and I'm lost here.

@mmurto @Etsija @lamppu Any idea how to properly fix this?

Etsija:
@mnonnenmacher I'm still trying to reproduce the problem. What's the situation you encounter where the favicon disappears? On Docker Compose, it never disappears for me, it's just always there. Have gone to subpages, then reloaded - it's always visible.

sschuberth:
> [@mnonnenmacher](https://github.com/mnonnenmacher) I'm still trying to reproduce the problem.

@Etsija can you please retry to reproduce with @lamppu's hint from the daily?

Etsija:
I did, and could not reproduce this even when using Docker Compose UI image at `http://localhost:8082`.

lamppu:
> I did, and could not reproduce this even when using Docker Compose UI image at `http://localhost:8082`.

Yeah, it seems it was partially fixed by the revert commit, the revert PR has a comment for why this wasn't closed yet https://github.com/eclipse-apoapsis/ort-server/pull/2119#issuecomment-2685494768.

PR #2119, opened 2025-02-26
Revert "fix(ui): Make the favicon path relative"

mnonnenmacher:
This reverts commit 40fe0e16f7438c6926f257ad07199563b95a428e.

The fix only worked when loading the root URL of the UI, not when loading it from any subpath like `organizations/1`. Revert the change to make the favicon at least work consistently when the UI is not deployed to a subpath.

Relates to #1916.

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -21,7 +21,7 @@ License-Filename: LICENSE
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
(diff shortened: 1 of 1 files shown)


MarcelBochtler (approved):

sschuberth:
> Relates to #1916.

Why only "relates" and not "fixes"?

PR #2123, opened 2025-02-26
fix(api): Additionally return Scanner issues for ORT runs

### LINKED ISSUE #1534: Include scanner issues in the list of returned issues
Based on the work of @MarcelBochtler the PR now supports sorting and pagination. 
Integration tests for the REST API endpoints have been added.
**Scanner issues are intentionally not included** in the returned issues, as there are still some clarifications to do.

_Originally posted by @wkl3nk in https://github.com/eclipse-apoapsis/ort-server/issues/1014#issuecomment-2368837895_

wkl3nk:
Return Scanner issues that are stored as part of Scan Summaries. Previously, these issues were not included in the API endpoints for retrieving issues, leading to discrepancies between issues in reports and those displayed in the ORT Server Web Application.

Fixes #1534.

### DIFF
--- a/services/hierarchy/src/main/kotlin/IssueService.kt
+++ b/services/hierarchy/src/main/kotlin/IssueService.kt
@@ -21,12 +21,18 @@ package org.eclipse.apoapsis.ortserver.services
 
 import org.eclipse.apoapsis.ortserver.dao.QueryParametersException
 import org.eclipse.apoapsis.ortserver.dao.dbQuery
+import org.eclipse.apoapsis.ortserver.dao.repositories.scannerjob.ScannerJobsTable
+import org.eclipse.apoapsis.ortserver.dao.repositories.scannerrun.ScannerRunsScanResultsTable
+import org.eclipse.apoapsis.ortserver.dao.repositories.scannerrun.ScannerRunsTable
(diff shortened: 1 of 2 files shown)


oheger-bosch:
I haven't had a deeper look into this PR, but it seems that a bunch of special logic is required to correctly handle the issues produced by the scanner.

So, I was wondering whether an alternative approach would be beneficial: Would it be possible to assign issues directly to an ORT run (meaning adding them to the `ort_runs_issues` table) when a scan result is associated with an ORT run in `OrtServerScanResultStorage`? This would be slightly redundant, but would allow to treat all issues of the run in a uniform way.

wkl3nk:
I will convert this PR to **draft** again and check Oliver's proposal. This would indeed make the code to access the issues much less-complicated or even not required.

wkl3nk:
What looked like a bug actually wasn't one. Although having spent a lot of time trying to understand the code and tried out several things, everything there seems to work as expected.
Using code of this `IssuesService` in the module were the workers are located does not work and also is no good idea. So I will keep it as it is, but anyone who likes to pick up this branch and do things in a slightly different way is welcomed to do so.

oheger-bosch (approved):
Since this fixes an actual issue, I am fine with the changes.

PR #2126, opened 2025-02-27
fix(ui): Correct UI behaviour when searching for a non-existing run

Etsija:
Screenshot from 2025-02-27 11-41-36

### DIFF
--- a/ui/src/routes/runs/$runId/index.tsx
+++ b/ui/src/routes/runs/$runId/index.tsx
@@ -17,9 +17,10 @@
  * License-Filename: LICENSE
  */
 
-import { createFileRoute, notFound, redirect } from '@tanstack/react-router';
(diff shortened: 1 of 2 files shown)


ui/src/routes/runs/$runId/index.tsx
```diff
@@ -49,7 +50,11 @@ export const Route = createFileRoute('/runs/$runId/')({
       });
     }
   },
-  notFoundComponent: () => {
-    return <div>ORT run not found!</div>;
+  errorComponent: ({ error }) => {
+    if (error instanceof NotFoundError) {
+      return <>Sorry, no run exists with that ID.</>;
```
sschuberth, line 55: Ideally, we'd mention the run ID as part of the error message: Like "There is no run with ID 666."

sschuberth (approved):

PR #2127, opened 2025-02-27
refactor: Align to use camel case for sorting field names

lamppu:
Align with the convention of commit https://github.com/eclipse-apoapsis/ort-server/commit/8c1d23ff0067850069600ef5a122bf57372b924d#diff-07702ede8522e232eb239906b7cf8ef6f3b5451730aeb6b5732cca33fcef54c6R42 that predates the addition of these sorting fields.


mnonnenmacher (commented):
Does this not also require changes in the UI? Or does it not use these sort options?

lamppu:
> Does this not also require changes in the UI? Or does it not use these sort options?

It doesn't yet, as there's still some filtering missing from the backend implementation, so it's all done client side for now.

mnonnenmacher (approved):

PR #2129, opened 2025-02-27
refactor(services): Use custom sorting for packages query

lamppu:
Allow sorting packages by identifier, purl and the processed declared license SPDX expression. This will provide needed sorting options to allow sorting the packages table in the UI.

### DIFF
--- a/services/hierarchy/src/main/kotlin/PackageService.kt
+++ b/services/hierarchy/src/main/kotlin/PackageService.kt
@@ -19,16 +19,18 @@
 
 package org.eclipse.apoapsis.ortserver.services
 
+import org.eclipse.apoapsis.ortserver.dao.QueryParametersException
 import org.eclipse.apoapsis.ortserver.dao.dbQuery
 import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerjob.AnalyzerJobsTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.AnalyzerRunsTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.PackageDao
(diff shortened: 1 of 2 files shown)


services/hierarchy/src/main/kotlin/PackageService.kt
```diff
@@ -47,12 +51,32 @@ class PackageService(private val db: Database) {
...
+
+        parameters.sortFields.forEach {
+            val sortOrder = it.direction.toSortOrder()
+            when (it.name) {
+                "identifier" -> {
+                    orders += IdentifiersTable.type to sortOrder
+                    orders += IdentifiersTable.namespace to sortOrder
+                    orders += IdentifiersTable.name to sortOrder
+                    orders += IdentifiersTable.version to sortOrder
+                }
```
lamppu, line 64: I wasn't sure if this kind of combination would be ok to use, but this is technically what we need right now in the frontend, but of course this could be separated like [here](https://github.com/eclipse-apoapsis/ort-server/blob/main/services/hierarchy/src/main/kotlin/VulnerabilityService.kt#L103-L106), and just use a combination of those in the request from the frontend. I think I'll align these based on what would be the preferred way.

mnonnenmacher: To me the question is what's needed by the frontend, but this should be the same for the packages and the vulnerabilities table. So does the vulnerabilities table actually use the flexibility to sort by the identifier properties separately?

lamppu: > So does the vulnerabilities table actually use the flexibility to sort by the identifier properties separately?

No it doesn't, so I was thinking I could align the vulnerabilities service to have it the same way as it's done here, maybe in a follow-up PR, or here if that's ok.

mnonnenmacher (approved):

PR #2146, opened 2025-02-27
Snippets orphan removal

kamil-bielecki-bosch:


services/hierarchy/src/test/kotlin/TableEntryTestFixtures.kt
```diff
@@ -0,0 +1,579 @@
...
+import org.eclipse.apoapsis.ortserver.model.Severity
+import org.eclipse.apoapsis.ortserver.model.runs.DependencyGraphsWrapper
+
+import org.jetbrains.exposed.sql.insert
+
+/**
+ * Set of Database fixtures for testing purposes.
+ * Every function creates a new entry in the corresponding table and returns the id of the created entry.
+ * Default, unique values are generated for every field, to provide entry uniqueness.
+ * REMARK: The functions are not creating it's own transactions, so should be used in db.dbQuery { ... } blocks.
```
oheger-bosch, line 98: "... *their* own transactions..."

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
```diff
@@ -302,4 +310,33 @@ class OrphanRemovalService(
...
+            }
+        }
+
+    private suspend fun deleteOrphanedSnippets() =
+        db.dbQuery {
+            SnippetsTable.deleteWhere {
+                id notInSubQuery (
+                    SnippetFindingsSnippetsTable
+                        .select(SnippetFindingsSnippetsTable.snippetId.alias("id"))
+                        .where(SnippetFindingsSnippetsTable.snippetId.isNotNull())
```
oheger-bosch, line 338: This condition looks strange. Are you sure it is correct? Are snippets not handled by cascade rules?

oheger-bosch (commented):
One general remark: Scan results produced by ScanCode are shared between multiple ORT runs, so it makes sense to handle them by the orphan service.
For snippet results, this is not the case; they are exclusively assigned to a single ORT run. (Or in other words: For every ORT run that has snippet scanning enabled, another scan is started, and the results are stored.) So, would it be better to directly delete snippet results (if they exist) when the owning ORT run is deleted?

nnobelis:
> For snippet results, this is not the case; they are exclusively assigned to a single ORT run. 

Is it true ? I remember in the past, snippets were assigned to the scan results, and a ORT run could have several scan results because of duplication of them.
Are you really saying that now, an ORT run can only have one se of snippets ? If yes, is this because you removed the aforementioned duplication ?

services/hierarchy/src/test/kotlin/OrphanRemovalServiceTest.kt
```diff
@@ -23,75 +23,58 @@ import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.shouldBe
```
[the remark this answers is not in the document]
kamil-bielecki-bosch: Added.

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
kamil-bielecki-bosch: SnippetFindingsSnippetsTable is an connection table between `snippet_findings` and `snippets` table. So, as far as Snippet can be referenced by more than one scan (at least from DB perspective) it's not safe to delete snippets by cascade delete.

services/hierarchy/src/test/kotlin/TableEntryTestFixtures.kt
```diff
@@ -0,0 +1,579 @@
...
+ */
+internal object TableEntryTestFixtures {
```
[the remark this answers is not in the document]
kamil-bielecki-bosch: Name changed.

kamil-bielecki-bosch:
@oheger-bosch So, in that case, there should be just extension on cascade deletes from ort-runs up to at least `snippet_findings_snippets` table? Including all related entities like findings?
If yes, what should be done with all orphaned entities that has no relation to any ort-run (ones already deleted with maintenance task)? Some migration script with delete? 
Also - what is the point with `snippet_findings_snippets` table if there are no situations that one snippet can be found from more than one scan run?

kamil-bielecki-bosch:
Obsolete, due to design changes

PR #2147, opened 2025-02-28
Add endpoint to get vulnerabilities for an organization

Etsija:

For the most parts, this is similar to #1632.

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/OrganizationVulnerability.kt
+++ b/api/v1/model/src/commonMain/kotlin/OrganizationVulnerability.kt
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.api.v1.model
+
+import kotlinx.serialization.Serializable
+
+/**
+ * A response object that gathers relevant information about a [Vulnerability].
+ */
+@Serializable
+data class OrganizationVulnerability(
+    /** The [Vulnerability] details. */
+    val vulnerability: Vulnerability,
+
+    /** The package [Identifier]. */
+    val identifier: Identifier,
+
+    /** An advisory rating for the [Vulnerability], derived from the individual references of the [Vulnerability]. */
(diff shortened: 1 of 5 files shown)


api/v1/mapping/src/commonMain/kotlin/Mappings.kt
mnonnenmacher, line 1: Commit message nit: Typo in "distinguish"

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
```diff
@@ -794,6 +795,14 @@ fun VulnerabilityWithAccumulatedData.mapToApiProduct() = ApiProductVulnerability
     repositoriesCount = repositoriesCount
 )
 
+fun VulnerabilityWithAccumulatedData.mapToApiOrganization() = ApiOrganizationVulnerability(
+    vulnerability = vulnerability.mapToApi(),
+    identifier = identifier.mapToApi(),
+    rating = rating.mapToApi(),
+    ortRunIds = ortRunIds,
+    repositoriesCount = repositoriesCount
+)
```
mnonnenmacher, line 804: Why are two methods required if the mappings are exactly the same? Do you think `ApiProductVulnerability` and `ApiOrganizationVulnerability` will ever be different?

mnonnenmacher (changes requested):

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
Etsija: I thought in future, organization vulnerabilities might also include a count for how many products include the vulnerability, as is now done for the number of repositories. But I can remove the distinctive mappings.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
Etsija: Fixed by deleting the commit.

wkl3nk:
What problem (for the customer) does this feature (to get vulnerabilities per organization) does this feature solve?

wkl3nk (changes requested):
What problem (for the customer) does this feature (to get vulnerabilities per organization) does this feature solve? I personally can see no benefit, it just adds effort and complexity to the project.

Etsija:
> What problem (for the customer) does this feature (to get vulnerabilities per organization) does this feature solve? I personally can see no benefit, it just adds effort and complexity to the project.

@sschuberth can maybe elaborate this further, but we (Double Open) believe that seeing in which repositories vulnerabilities "pile up" in an organization gives an insight to the org. leader as to where to concentrate the safety efforts.

sschuberth:
> > What problem (for the customer) does this feature (to get vulnerabilities per organization) does this feature solve? I personally can see no benefit, it just adds effort and complexity to the project.
> 
> @sschuberth can maybe elaborate this further

Not much to add, except saying that this cam you of concrete user requests, where product manager are not interesting in "low level details" like vulnerabilities found in a single repository's scan, but prefer to get a high level overview of vulnerability metrics throughout their organization.

That said, the metrics themselves certainly need to be improved; it does not make much sense just just sum up vulnerabilities over products like we currently do. But improving the metrics is a follow-up second step.

sschuberth (approved):

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
mnonnenmacher: I just wanted to understand if it is purely hypothetical or if there are concrete ideas why those classes should be different.

sschuberth: There are concrete plans, but the time frame is unclear.

PR #2149, opened 2025-02-28
fix: Return proper HTTP error code when calling non existing repo

kamil-bielecki-bosch:
Return 404 HTTP error code instead of 500 when triggering ORT run on non existing repository.

### DIFF
--- a/core/src/main/kotlin/api/RepositoriesRoute.kt
+++ b/core/src/main/kotlin/api/RepositoriesRoute.kt
@@ -125,20 +125,22 @@ fun Route.repositories() = route("repositories/{repositoryId}") {
             requirePermission(RepositoryPermission.TRIGGER_ORT_RUN)
 
             val repositoryId = call.requireIdParameter("repositoryId")
-            val createOrtRun = call.receive<CreateOrtRun>()
 
-            call.respond(
-                HttpStatusCode.Created,
-                orchestratorService.createOrtRun(
-                    repositoryId,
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/api/RepositoriesRoute.kt
nnobelis, line 1: Could you also add a test ?

nnobelis (changes requested):

core/src/main/kotlin/api/RepositoriesRoute.kt
kamil-bielecki-bosch: Done

core/src/test/kotlin/api/RepositoriesRouteIntegrationTest.kt
```diff
@@ -715,6 +715,24 @@ class RepositoriesRouteIntegrationTest : AbstractIntegrationTest({
                 responses.forAll { it shouldHaveStatus HttpStatusCode.Created }
             }
         }
+
+        "respond with \"Not Found\" if repository doesn't exists" {
```
nnobelis, line 719: _exist_

nnobelis (changes requested):

nnobelis (approved):

sschuberth (changes requested):
Please remove the merge commit from the PR, @kamil-bielecki-bosch.

PR #2154, opened 2025-02-28
Disable Kotest's auto-scan also for multi-platform projects

sschuberth:


api/v1/mapping/build.gradle.kts
mnonnenmacher, line 1: Commit message: Typo in "properties".

This should also fix the (minor) issue that the custom MDC logger is missing on the test classpath for some projects, if you want to mention that in the commit message.

PR #2160, opened 2025-02-28
CLI: Improve error handling

MarcelBochtler:

### DIFF
--- a/api/v1/client/src/commonMain/kotlin/Exceptions.kt
+++ b/api/v1/client/src/commonMain/kotlin/Exceptions.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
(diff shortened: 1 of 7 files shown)


api/v1/client/src/commonMain/kotlin/HttpClientUtils.kt
```diff
@@ -62,13 +62,6 @@ fun createOrtHttpClient(
 ): HttpClient =
     createDefaultHttpClient(json, engine).config {
         HttpResponseValidator {
-            handleResponseExceptionWithRequest { cause, request ->
```
sschuberth, line 65: Again, the commit message just repeats in prose what's obvious from the diff. Please instead explain *why* this is a good thing to do.

MarcelBochtler: This PR is a draft. The content of it and the commit messages are considered a work in progress and not ready for review yet.

api/v1/client/src/commonMain/kotlin/Exceptions.kt
```diff
@@ -0,0 +1,46 @@
+/*
```
bs-ondem, line 1: Commit msg: lowercase "server"

api/v1/client/src/commonMain/kotlin/HttpClientUtils.kt
```diff
@@ -89,8 +89,3 @@ fun createOrtHttpClient(
 
         config()
     }
```
bs-ondem, line 91: Doesn't this change belong in the previous commit?

cli/src/commonMain/kotlin/InfoCommand.kt
```diff
@@ -31,6 +31,7 @@ import com.github.ajalt.clikt.parameters.options.required
 import com.github.ajalt.clikt.parameters.types.long
```
bs-ondem, line 31: Commit message: lowercase "server"

api/v1/client/src/commonMain/kotlin/HttpClientUtils.kt
MarcelBochtler: No. Below you see that this class was still being used.

bs-ondem (approved):

Issue #2082, opened 2025-02-21
ORT run details query should return more revision information


Etsija:
ORT runs have a "revision" string, which I believe currently just resolves to what the user has written to the create run form; in case the user leaves revision empty, it's also ean mpty string in the [ORT run details query](https://eclipse-apoapsis.github.io/ort-server/api/get-ort-run-by-id).

With #2036 we are correctly also handling the case of leaving the "revision" empty, and in that case, the default branch is used.

However we are missing important information from the UI about the non-resolved and resolved revisions.

I think we need two things from the ORT run details endpoint, to show them properly in the UI:
- the non-resolved revision that was found (ie. EITHER what was input to the run creation form, OR the found default branch name in case the field was left empty in the form)
- the resolved revision (ie. commit SHA)

I believe we could then even ditch the resolved configuration information from the ORT run details page if requested, and add this info instead, as I believe it is far more useful to the users to see which revision exactly was analyzed, rather than seeing which revision of the configuration provider was used.

mnonnenmacher:
> the non-resolved revision that was found (ie. EITHER what was input to the run creation form, OR the found default branch name in case the field was left empty in the form)

For this maybe the recently introduced code to use the default branch could be adapted to store the default branch name in the `revision` property of the run. If we want to keep the information that the run was triggered without providing a revision we would need a separate property, but I'm not sure if that is really required.

>  believe we could then even ditch the resolved configuration information from the ORT run details page if requested

I would prefer to keep it. It is important to know which version of the rules or other configuration files like license classifications were used.

Etsija:
I believe in addition to the default branch name we really would need to see the resolved revision (ie. the exact commit SHA) of the targeted repository as well.

mnonnenmacher:
This is mostly fixed by #2163 but for storing the default branch name I will make a separate PR.

PR #2163, opened 2025-02-28
Store resolved revision of ORT run

mnonnenmacher:
Store and show the resolved revision of an ORT run. See the commit messages for details.

For now I just appended the resolved revision like this:
image


Etsija:
IIUC, when the user leaves the revision empty when creating a run, back-end finds the default branch name (main/master/etc), but the revision is still empty in the UI.  Would it be complicated to in this case return the revision, so the UI could show something like
Revision: main
Resolved revision: 0123456789abcde

mnonnenmacher:
> IIUC, when the user leaves the revision empty when creating a run, back-end finds the default branch name (main/master/etc), but the revision is still empty in the UI. Would it be complicated to in this case return the revision, so the UI could show something like Revision: main Resolved revision: 0123456789abcde

I will do this in a separate PR, see https://github.com/eclipse-apoapsis/ort-server/issues/2082#issuecomment-2691230775.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
```diff
@@ -354,6 +354,7 @@ fun OrtRun.mapToApi(jobs: ApiJobs) =
         productId,
         repositoryId,
         revision,
+        resolvedRevision,
```
sschuberth, line 357: Commit message nits:
- I guess the title intentionally is "compact" and omits "a" before "property", "the" before "resolved revison", and "an" before "ORT run"?
- "store the resolved revision *of the project to analyze*" (or similar) to distinguish from the resolved revision used by the configuration context (which has caused confusion in the past).

mnonnenmacher: >     * I guess the title intentionally is "compact" and omits "a" before "property", "the" before "resolved revison", and "an" before "ORT run"?

Yes.

>     * "store the resolved revision _of the project to analyze_" (or similar) to distinguish from the resolved revision used by the configuration context (which has caused confusion in the past).

Done.

sschuberth (approved):

PR #2191, opened 2025-03-05
ui: Show organization vulnerabilities

Etsija:
This PR concludes the initial "security-concentrated" dashboards for organizations and products, aligning them on the view.  Later on, more tools will be developed to assist especially organization and product owners in assessing the current status of the org. or product.

Screenshot from 2025-03-04 16-50-40


ui/src/routes/organizations/$orgId/vulnerabilities/-components/organization-vulnerability-table.tsx
```diff
@@ -0,0 +1,311 @@
...
+        onClick: () => {},
+      },
+    });
+    return;
+  }
+
+  return (
+    <OrganizationVulnerabilityTableInner
+      vulnerabilities={vulnerabilities.data}
+      totalVulnerabilities={totalVulnerabilities}
```
lamppu, line 158: Shouldn't this rather use the count from the response of the vulnerabilities endpoint? For now it's correct, but it would need to be changed anyway when the filtering is implemented, so also passing it from the parent component will become redundant then.

lamppu (changes requested):

lamppu (approved):

PR #2199, opened 2025-03-06
Add filtering to packages endpoint

lamppu:

### DIFF
--- a/core/src/main/kotlin/apiDocs/RunsDocs.kt
+++ b/core/src/main/kotlin/apiDocs/RunsDocs.kt
@@ -39,6 +39,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunStatistics
 import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunStatus
 import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunSummary
 import org.eclipse.apoapsis.ortserver.api.v1.model.Package
+import org.eclipse.apoapsis.ortserver.api.v1.model.PackageFilters
 import org.eclipse.apoapsis.ortserver.api.v1.model.PagedResponse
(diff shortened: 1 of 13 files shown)


core/src/main/kotlin/apiDocs/RunsDocs.kt
```diff
@@ -370,15 +371,31 @@ val getPackagesByRunId: OpenApiRoute.() -> Unit = {
             description = "The ID of the ORT run."
         }
 
+        queryParameter<String>("identifier") {
+            description = "Defines an ORT package identifier for which the packages are to be retrieved for. This " +
```
Etsija, line 375: Double "for" in description.

core/src/main/kotlin/apiDocs/RunsDocs.kt
```diff
@@ -370,15 +371,31 @@ val getPackagesByRunId: OpenApiRoute.() -> Unit = {
...
+                    "uses a case-insensitive substring match."
+        }
+
+        queryParameter<String>("purl") {
+            description = "Defines a purl for which the packages are to be retrieved for. This uses a " +
```
Etsija, line 380: Double "for" in description.

Etsija (changes requested):

Etsija:
I have tested this by taking the PR to my local, and implementing a freeform text filter (server-side filtering) for the `Package ID` column in the UI. As can be seen from the picture, the text filter is working as specified. As seen in the picture, I also included some special characters seen in package IDs in my tests, and they all seem to be handled properly.

Screenshot from 2025-03-11 07-22-48

As filtering the `Declared Licenses` column with a select filter needs an endpoint to gather deduplicated declared licenses from packages (promised as a follow-up to this PR by @lamppu), I did not test yet filtering for that column.

All in all, functionality-wise, so far looks good to me.

PR #2203, opened 2025-03-07
fix(ui): Improve page navigation in `DataTablePagination` component

### LINKED ISSUE #2135: Input field for the page responds too fast to changes
When manually changing the page number in any paginated table, the page is immediately reloaded which makes it impossible to enter a two-digit number as the page. This is problematic when there are many pages, like here:

Image

I tested this only in Firefox, in case the behavior is browser specific.

Instead, there should either be a delay or the reload should only happen when the focus is changed or the enter button is pressed.

bs-ondem:
Handle page changes only when the user presses the "Enter" key or loses focus on the input field.

Resolves #2135.

### DIFF
--- a/ui/src/components/data-table/data-table-pagination.tsx
+++ b/ui/src/components/data-table/data-table-pagination.tsx
@@ -24,6 +24,7 @@ import {
   DoubleArrowRightIcon,
 } from '@radix-ui/react-icons';
 import { Link, LinkOptions, useNavigate } from '@tanstack/react-router';
+import { useState } from 'react';
 
 import { Button } from '@/components/ui/button';
 import { Input } from '@/components/ui/input';
@@ -59,6 +60,7 @@ export function DataTablePagination({
   setCurrentPageOptions,
 }: DataTablePaginationProps) {
   const navigate = useNavigate();
+  const [page, setPage] = useState(currentPage);
 
   return (
     <div className='flex flex-col items-center justify-between gap-4 sm:flex-row sm:gap-6 lg:gap-8'>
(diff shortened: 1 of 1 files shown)


Etsija:
LGTM, and I wasn't able to break it when testing, but I'd like @lamppu to have a look as well.

ui/src/components/data-table/data-table-pagination.tsx
```diff
@@ -88,13 +90,18 @@ export function DataTablePagination({
           Page{' '}
           <Input
             type='number'
-            value={currentPage}
+            value={page}
+            onKeyUp={(event) => {
+              if (event.key === 'Enter') {
+                navigate(setCurrentPageOptions(page));
+              }
+            }}
```
mmurto, line 98: This could be an accessibility issue. The `Input` should probably be wrapped in a `form` and this should be in the form's `onSubmit` and the input's `onBlur`. The form's `onSubmit` likely requires `event.preventDefault()` to prevent page refresh.

mmurto (changes requested):

ui/src/components/data-table/data-table-pagination.tsx
bs-ondem: I'm not sure I fully understand. Do you mean something like this?

    <form ... onSubmit={}>
       <input ... onBlur={}/>
    </form>

So, I would replace the `onKeyUp` event on the input with the `onSubmit` event on the form instead?

mmurto: Yes, exactly like that. That ensures that the action triggers when sending the submit event also on other input methods than keyboard, like assistive tools and touch devices.

Etsija: For the sake of unity, should we actually use a shadcn-ui form here?

Etsija: I guess [this](https://github.com/eclipse-apoapsis/ort-server/blob/68bae311776bf07e005e80fc188a5767bd3ea30c/ui/src/components/header.tsx#L208-L229) could serve as an example of a simplest form which has just one input field.

mmurto: > For the sake of unity, should we actually use a shadcn-ui form here?

I don't think it brings any value over just using the native form. Do you think otherwise?

Etsija: Validation comes to my mind.

mmurto (approved):

PR #2204, opened 2025-03-07
No console logs in production code

Etsija:

Regarding the second commit, I'd like to get an opinion from @mmurto as the `console.error()` has come from you originally. Personally I've never encountered a situation where it would've been triggered, and would be willing to drop the whole line from code.

### DIFF
--- a/ui/.eslintrc.cjs
+++ b/ui/.eslintrc.cjs
@@ -33,5 +33,6 @@ module.exports = {
       'warn',
       { allowConstantExport: true },
     ],
+    'no-console': ['warn'],
   },
 };
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/package-manager-field.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/package-manager-field.tsx
@@ -183,7 +183,8 @@ const FieldWithOptions = ({ form, pmIndex, pmName }: FieldWithOptionsProps) => {
   const pm = packageManagers[pmIndex];
 
   if (!pm) {
-    console.error(`No package manager found for index ${pmIndex}.`);
+    // This will never actually trigger, because in the form,
+    // we always feed a valid package manager.
     return null;
   }
 


mmurto:
> Regarding the second commit, I'd like to get an opinion from @mmurto as the `console.error()` has come from you originally. Personally I've never encountered a situation where it would've been triggered, and would be willing to drop the whole line from code.

Feel free to remove. On a quick glance, a small refactoring of taking in a `PackageManager` instead of `pmIndex` (and name) would get rid of the check completely, and would be a little clearer while at it.

Etsija:
I removed the console.error() first, then turned the rule on.

mmurto:
> > Feel free to remove. On a quick glance, a small refactoring of taking in a `PackageManager` instead of `pmIndex` (and name) would get rid of the check completely, and would be a little clearer while at it.
> 
> We do not have a `PackageManager` type in OpenAPI, so it would have to be implemented. I'll leave this to another PR.

This component doesn't interact with the API, it interacts with https://github.com/eclipse-apoapsis/ort-server/blob/68bae311776bf07e005e80fc188a5767bd3ea30c/ui/src/routes/organizations/%24orgId/products/%24productId/repositories/%24repoId/-types.ts#L40-L141. A brief look at the code hints that the whole pmIndex could be removed, as it's used only to retrieve `pm.id`, which is the exact same as `pmName` as seen [here](https://github.com/eclipse-apoapsis/ort-server/blob/8244a8509d7236e76879260d9b66d70f73c49408/ui/src/routes/organizations/%24orgId/products/%24productId/repositories/%24repoId/-components/package-manager-field.tsx#L152).

Etsija:
Anyway, it's clearly out of scope or this PR. And seems like there are big changes upcoming for this whole code section in near future, so...

mmurto:
> Anyway, it's clearly out of scope or this PR. And seems like there are big changes upcoming for this whole code section in near future, so...

Clear. As the removal swallows a branch that looks like an error, add a short code comment at the `return null` on why the branch will never trigger.

mmurto (approved):

PR #2223, opened 2025-03-10
ui: Improve infrastructure handling

Etsija:
This PR addresses two issues:

1. Infrastructure service table's overflow off-screen to the right: 
Screenshot from 2025-03-10 12-44-22

2. Simplification of creation and editing of the infrastructure services with usage of dropdown lists for the secrets:
Screenshot from 2025-03-10 12-45-20


ui/src/routes/organizations/$orgId/infrastructure-services/$serviceName/edit/index.tsx
```diff
@@ -67,6 +75,11 @@ const EditInfrastructureServicePage = () => {
   const navigate = useNavigate();
   const params = Route.useParams();
 
+  const { data: secrets } =
+    useSecretsServiceGetApiV1OrganizationsByOrganizationIdSecretsSuspense({
+      organizationId: Number.parseInt(params.orgId),
+    });
```
lamppu, line 81: The secrets endpoint is a paginated endpoint, so I wonder if this should have a higher limit parameter set. Similar below for the create page.

lamppu (approved):

PR #2225, opened 2025-03-10
fix(api): Apply mapping to API for OrtRunSummaries

wkl3nk:
Apply the missing mapping from a Model object to an API object for to the `OrtRunSummary` entities returned from the REST endpoint `/api/v1/repositories/{repositoryId}/runs`.


core/src/main/kotlin/api/RepositoriesRoute.kt
```diff
@@ -26,7 +26,6 @@ import io.github.smiley4.ktorswaggerui.dsl.routing.post
 import io.github.smiley4.ktorswaggerui.dsl.routing.put
 
 import io.ktor.http.HttpStatusCode
-import io.ktor.server.application.call
```
MarcelBochtler, line 29: Nit: This is not related to the change and therefore should be in a separate commit.

MarcelBochtler (approved):

PR #2227, opened 2025-03-11
feat(api): Add name of the user that triggered a scan

### LINKED ISSUE #2132: The runs endpoint should also provide information about the user who triggered a run
Runs can usually be triggered by different users of the organization. Esp. during technical on-boarding of projects, user might need to experiment with different analyzer options, and trigger different runs for them In order to trace the last run of a specific user, the name of the user who triggered a runs should be part of the [details returned for a run](https://eclipse-apoapsis.github.io/ort-server/api/get-ort-run-by-id) in order to display the user in the front-end as part of the runs table.

wkl3nk:
Add username and full name (if present) of the user that triggered a scan to the scan representation in the database, and also provide these names when querying scans using the REST API in order to allow to display them in the UI in a user-friendly way.

Add username and full name (as tooltip) of the user that triggered a scan to the list of scans per repository and in the detail view of the scan.

Fixes #2132.

See the respective commits for details.

### DIFF
--- a/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
+++ b/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
@@ -83,6 +83,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.SortProperty as ApiSortProper
 import org.eclipse.apoapsis.ortserver.api.v1.model.SourceCodeOrigin as ApiSourceCodeOrigin
 import org.eclipse.apoapsis.ortserver.api.v1.model.SubmoduleFetchStrategy as ApiSubmoduleFetchStrategy
 import org.eclipse.apoapsis.ortserver.api.v1.model.User as ApiUser
+import org.eclipse.apoapsis.ortserver.api.v1.model.UserDisplayName as ApiUserDisplayName
 import org.eclipse.apoapsis.ortserver.api.v1.model.VcsInfo as ApiVcsInfo
 import org.eclipse.apoapsis.ortserver.api.v1.model.Vulnerability as ApiVulnerability
 import org.eclipse.apoapsis.ortserver.api.v1.model.VulnerabilityRating as ApiVulnerabilityRating
@@ -133,6 +134,7 @@ import org.eclipse.apoapsis.ortserver.model.Severity
 import org.eclipse.apoapsis.ortserver.model.SourceCodeOrigin
 import org.eclipse.apoapsis.ortserver.model.SubmoduleFetchStrategy
 import org.eclipse.apoapsis.ortserver.model.User
+import org.eclipse.apoapsis.ortserver.model.UserDisplayName
 import org.eclipse.apoapsis.ortserver.model.VulnerabilityRating
 import org.eclipse.apoapsis.ortserver.model.VulnerabilityWithAccumulatedData
 import org.eclipse.apoapsis.ortserver.model.VulnerabilityWithIdentifier
@@ -372,7 +374,8 @@ fun OrtRun.mapToApi(jobs: ApiJobs) =
         jobConfigContext,
         resolvedJobConfigContext,
         environmentConfigPath,
-        traceId
+        traceId,
+        userDisplayName?.mapToApi()
     )
 
 fun OrtRun.mapToApiSummary(jobs: ApiJobSummaries) =
@@ -392,7 +395,8 @@ fun OrtRun.mapToApiSummary(jobs: ApiJobSummaries) =
         labels = labels,
         jobConfigContext = jobConfigContext,
         resolvedJobConfigContext = resolvedJobConfigContext,
-        environmentConfigPath = environmentConfigPath
+        environmentConfigPath = environmentConfigPath,
+        userDisplayName = userDisplayName?.mapToApi()
     )
 
 fun OrtRunSummary.mapToApi() =
@@ -412,7 +416,8 @@ fun OrtRunSummary.mapToApi() =
         labels = labels,
         jobConfigContext = jobConfigContext,
         resolvedJobConfigContext = resolvedJobConfigContext,
-        environmentConfigPath = environmentConfigPath
+        environmentConfigPath = environmentConfigPath,
+        userDisplayName = userDisplayName?.mapToApi()
     )
 
 fun JobSummaries.mapToApi() =
@@ -886,3 +891,5 @@ fun Project.mapToApi() = ApiProject(
     homepageUrl,
     scopeNames
 )
+
+fun UserDisplayName.mapToApi() = ApiUserDisplayName(username, fullName)
--- a/api/v1/model/src/commonMain/kotlin/UserDisplayName.kt
+++ b/api/v1/model/src/commonMain/kotlin/UserDisplayName.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.api.v1.model
+
+import kotlinx.serialization.Serializable
+
+/**
+ * User information to be displayed in the UI.
+ */
+@Serializable
+data class UserDisplayName(
+    /**
+     * Unique, human-readable identifier chosen by the user or administrator, used for authentication and login.
+     * This can change over time.
+     */
+    val username: String,
+
+    /**
+     * Full name of the user: A derived attribute that typically combines a user's first name and last name,
+     * providing a readable display name but not serving as a unique identifier. This can change over time.
+     */
+    val fullName: String? = null
+)
--- a/core/src/main/kotlin/authorization/OrtPrincipal.kt
+++ b/core/src/main/kotlin/authorization/OrtPrincipal.kt
@@ -49,3 +49,9 @@ fun OrtPrincipal?.hasRole(role: String) = this != null && role in roles
  * Return true if this [OrtPrincipal] is not `null` and has the [superuser role][Superuser.ROLE_NAME].
  */
 fun OrtPrincipal?.isSuperuser() = hasRole(Superuser.ROLE_NAME)
+
+fun OrtPrincipal.getUserId(): String = payload.subject
+
+fun OrtPrincipal.getUsername(): String = payload.getClaim("preferred_username").asString()
+
+fun OrtPrincipal.getFullName(): String? = payload.getClaim("name").asString()
--- a/core/src/main/kotlin/services/OrchestratorService.kt
+++ b/core/src/main/kotlin/services/OrchestratorService.kt
@@ -25,6 +25,7 @@ import org.eclipse.apoapsis.ortserver.config.ConfigManager
 import org.eclipse.apoapsis.ortserver.dao.dbQuery
 import org.eclipse.apoapsis.ortserver.model.JobConfigurations
 import org.eclipse.apoapsis.ortserver.model.OrtRun
+import org.eclipse.apoapsis.ortserver.model.UserDisplayName
 import org.eclipse.apoapsis.ortserver.model.orchestrator.CreateOrtRun
 import org.eclipse.apoapsis.ortserver.model.repositories.OrtRunRepository
 import org.eclipse.apoapsis.ortserver.transport.Message
@@ -56,7 +57,8 @@ class OrchestratorService(
         jobConfig: JobConfigurations,
         jobConfigContext: String?,
         labels: Map<String, String>?,
-        environmentConfigPath: String?
+        environmentConfigPath: String?,
+        userDisplayName: UserDisplayName?
     ): OrtRun {
         val traceId = MDC.get("traceId")
 
@@ -70,7 +72,8 @@ class OrchestratorService(
                 jobConfigContext,
                 labels.orEmpty(),
                 traceId = traceId,
-                environmentConfigPath = environmentConfigPath
+                environmentConfigPath = environmentConfigPath,
+                userDisplayName = userDisplayName
             )
         }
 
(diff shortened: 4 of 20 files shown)


api/v1/mapping/src/commonMain/kotlin/Mappings.kt
```diff
@@ -82,6 +82,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.SortProperty as ApiSortProper
 import org.eclipse.apoapsis.ortserver.api.v1.model.SourceCodeOrigin as ApiSourceCodeOrigin
 import org.eclipse.apoapsis.ortserver.api.v1.model.SubmoduleFetchStrategy as ApiSubmoduleFetchStrategy
 import org.eclipse.apoapsis.ortserver.api.v1.model.User as ApiUser
+import org.eclipse.apoapsis.ortserver.api.v1.model.UserDisplayName as ApiUserDisplayName
```
sschuberth, line 85: Nit: In the commit message title, "names" should be singular "name". And "triggers" probably should be "triggered".

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
```diff
@@ -114,6 +119,21 @@ const RunComponent = () => {
                 <Label className='font-semibold'>Created at:</Label>{' '}
                 <TimestampWithUTC timestamp={ortRun.createdAt} />
               </div>
+              <div className='text-sm'>
+                <Label className='font-semibold'>Created by:</Label>{' '}
```
Etsija, line 123: This renders an empty "Created by:" for old scans with non-existing user information.  You can make the whole rendering of the "Created by:" info conditional with something like
```
{ortRun.userDisplayName && (
    <div className='text-sm'>
    ...
    /div>
)}
```

Etsija:
I believe it would be especially useful to also see who triggered the last run in
- organization overview page (products table)
- product overview page (repositories table)

You could achieve this by making corresponding changes/additions to this component: https://github.com/eclipse-apoapsis/ort-server/blob/0708e7820ba9b25854e4a764bac60d6885dd96c8/ui/src/routes/organizations/%24orgId/products/%24productId/-components/last-run-date.tsx#L66-L69

Etsija (changes requested):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/repository-runs-table.tsx
```diff
@@ -91,7 +91,24 @@ const columns = [
   }),
   columnHelper.accessor('createdAt', {
     header: 'Created At',
-    cell: ({ row }) => <TimestampWithUTC timestamp={row.original.createdAt} />,
+    cell: ({ row }) => (
+      <div>
+        <TimestampWithUTC timestamp={row.original.createdAt} />
+        <br />
+        {row.original.userDisplayName?.fullName ? (
```
Etsija, line 98: I believe this will render unnecessarily an empty row for old scans which do not have the user information. You can use the same trick for conditional rendering to prevent empty rows: `{row.original.userDisplayName ?? (...)}`

Etsija (changes requested):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/repository-runs-table.tsx
Etsija: To clarify and correct myself:
```
{row.original.userDisplayName && (
    <br />
    {row.original.userDisplayName.fullName ? (
    ...
    )}
)}
```

api/v1/model/src/commonMain/kotlin/UserDisplayName.kt
```diff
@@ -0,0 +1,31 @@
...
+package org.eclipse.apoapsis.ortserver.api.v1.model
+
+import kotlinx.serialization.Serializable
+
+/**
+ * User information to be displayed in the UI.
+ */
+@Serializable
+data class UserDisplayName(
+    var username: String,
```
oheger-bosch, line 29: Please add documentation. The meaning of these properties is not obvious.

api/v1/model/src/commonMain/kotlin/UserDisplayName.kt
```diff
@@ -0,0 +1,31 @@
...
+    var fullName: String? = null
```
oheger-bosch, line 30: These should be `val`s not `var`s.

core/src/main/kotlin/services/OrchestratorService.kt
```diff
@@ -56,10 +64,16 @@ class OrchestratorService(
         jobConfig: JobConfigurations,
         jobConfigContext: String?,
         labels: Map<String, String>?,
-        environmentConfigPath: String?
+        environmentConfigPath: String?,
+        call: RoutingCall
```
oheger-bosch, line 68: Should the service be responsible for extracting the user information from the call? This couples the service to technology used by the endpoints. I think it would therefore be better to let the route handle the extraction and pass the user model object to the service.

model/src/commonMain/kotlin/UserDisplayName.kt
```diff
@@ -0,0 +1,32 @@
...
+package org.eclipse.apoapsis.ortserver.model
+
+import kotlinx.serialization.Serializable
+
+/**
+ * User information to be displayed in the UI.
+ */
+@Serializable
+data class UserDisplayName(
+    var userId: String, // Unique, stable over time
```
oheger-bosch, line 29: These should be `val`s. Data classes should really be immutable.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/index.tsx
wkl3nk: Good idea. I will do it.

core/src/main/kotlin/services/OrchestratorService.kt
wkl3nk: I agree, another presentation layer (next to the current one using Ktor) might use different ways for authentication, and therefore the extraction code should stay in the presentation layer.

Etsija:
A general comment: I believe it is the reviewer's responsibility to mark a discussion "resolved" after comments are addressed and reviewed by the reviewer, not the author's - or have I misunderstood the process, @sschuberth?

Etsija:
And what about this?

Etsija:
> I believe it would be especially useful to also see who triggered the last run in
> 
>     * organization overview page (products table)
> 
>     * product overview page (repositories table)
> 
> 
> You could achieve this by making corresponding changes/additions to this component:
> 
> https://github.com/eclipse-apoapsis/ort-server/blob/0708e7820ba9b25854e4a764bac60d6885dd96c8/ui/src/routes/organizations/%24orgId/products/%24productId/-components/last-run-date.tsx#L66-L69

What about this?

sschuberth:
> A general comment: I believe it is the reviewer's responsibility to mark a discussion "resolved" after comments are addressed and reviewed by the reviewer, not the author's - or have I misunderstood the process, @sschuberth?

That's how I live and preach it: The author puts 👍🏻 under a review comment to acknowledge that (s)he believes to have addressed the comment, but whether the comment was *really* addressed in the reviewer's sense should be confirmed by the reviewer by resolving the conversation. Because as a reviewer that does many reviews in parallel, I personally find it very annoying that I have to remind myself about my own comments by expanding resolved conversations again when doing another review iteration.

However, I have seen, too, that this is being handled differently in this project.

wkl3nk:
> > I believe it would be especially useful to also see who triggered the last run in
> > ```
> > * organization overview page (products table)
> > 
> > * product overview page (repositories table)
> What about this?

It may be useful, but not in the scope of this ticket. This ticket is just about displaying the user that triggered a scan.

dao/src/main/kotlin/repositories/userDisplayName/UserDisplayNamesTable.kt
```diff
@@ -0,0 +1,86 @@
...
+import org.jetbrains.exposed.sql.Column
+import org.jetbrains.exposed.sql.insertIgnore
+import org.jetbrains.exposed.sql.kotlin.datetime.timestamp
+
+/**
+ * A table to represent names to identify a user. This is **not** used for any authentication or authorization purposes,
+ * it is only used to store users' names in a central table, where it is easier to keep them up to date or remove them
+ * after a time of inactivity.
+ */
+object UserDisplayNamesTable : IdTable<String>("user_display_names") {
```
mnonnenmacher, line 40: Why `user_display_names` and not just `user_names` or even `users` (same goes for the class name `UserDisplayName`)?

dao/src/main/kotlin/repositories/userDisplayName/UserDisplayNamesTable.kt
```diff
@@ -0,0 +1,86 @@
...
+
+class UserDisplayNameDAO(id: EntityID<String>) : Entity<String>(id) {
+    companion object : EntityClass<String, UserDisplayNameDAO>(UserDisplayNamesTable) {
+        /**
+         * Insert a new entry, ignoring if an entry with this [id] already exist.
+         * This is much faster and safer (in perspective of concurrent requests) than looking up if the entry
+         * not already exists, and only then inserting a new entry.
+         */
+        fun insertIgnore(userDisplayName: UserDisplayName?): UserDisplayNameDAO? {
+            userDisplayName ?: return null
```
mnonnenmacher, line 68: This is not JS, I would prefer the more expresssive `if (userDisplayName == null) return null` or actually, not allow null values as input.

dao/src/main/kotlin/repositories/userDisplayName/UserDisplayNamesTable.kt
```diff
@@ -0,0 +1,86 @@
...
+
+            UserDisplayNamesTable.insertIgnore {
```
mnonnenmacher, line 70: Doesn't this mean that if the `username` of `fullName` change they will never get updated because there will be a conflict for the `userId`? Should `upsert` be used instead?

mnonnenmacher (changes requested):

dao/src/main/kotlin/repositories/userDisplayName/UserDisplayNamesTable.kt
wkl3nk: By choosing this name I wanted to emphasize that this feature is just about adding information which user triggered a scan, so it's only about names. I explicitly did not want store `User` entities in the backend, because I think user management should not be added to the domain of the ORT Server, instead the users domain should stay in Keycloak.

Another reason was that there is already a `User` model, for transferring user related information from ORT Server (internal Keycloak Client) to the UI. So this would have been totally confusing.

Etsija:
> > > I believe it would be especially useful to also see who triggered the last run in
> > > ```
> > > * organization overview page (products table)
> > > 
> > > * product overview page (repositories table)
> > > ```
> > 
> > 
> > What about this?
> 
> It may be useful, but not in the scope of this ticket. This ticket is just about displaying the user that triggered a scan.

...and that's exactly what I am proposing here, so I believe it is very much in the scope of this ticket. As I hinted,  exactly the same user information can be attached to those two tables just by making similar changes that you have done already to `LastRunDate` component.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/repository-runs-table.tsx
```diff
@@ -91,7 +91,27 @@ const columns = [
   }),
   columnHelper.accessor('createdAt', {
     header: 'Created At',
```
sschuberth, line 93: Should "Create At" also be renamed to "Create At / By" now, to account for the user being displayed, too? Alternatively, just name the header "Created" and optionally prefix the time in the respective row with "at" and the user with "by".

dao/src/main/kotlin/repositories/userDisplayName/UserDisplayNamesTable.kt
wkl3nk: Yep. Upsert works.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/repository-runs-table.tsx
wkl3nk: Changed to "Created"

Etsija:
Just a heads up @wkl3nk : the "Created" info has been moved to [a "top bar" component](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/routes/organizations/%24orgId/products/%24productId/repositories/%24repoId/runs/%24runIndex/-components/run-details-bar.tsx), and soon it will be [altogether removed](https://github.com/eclipse-apoapsis/ort-server/issues/2238?issue=eclipse-apoapsis%7Cort-server%7C2266) from the run overview page.

It would be best if you do the needed changes in this PR, because if #2266 is merged before yours, it will cause a merge conflict with your PR.

wkl3nk:
@Etsija I am currently still waiting for a positive positive review in order to merge it.

Etsija:
> @Etsija I am currently still waiting for a positive positive review in order to merge it.

I could give it - from my part - but I believe there are some open debate with @mnonnenmacher so I'd rather not intervene.

core/src/main/kotlin/authorization/OrtPrincipal.kt
```diff
@@ -49,3 +49,9 @@ fun OrtPrincipal?.hasRole(role: String) = this != null && role in roles
  * Return true if this [OrtPrincipal] is not `null` and has the [superuser role][Superuser.ROLE_NAME].
  */
 fun OrtPrincipal?.isSuperuser() = hasRole(Superuser.ROLE_NAME)
+
+fun OrtPrincipal.getUserId(): String = payload.subject
```
oheger-bosch, line 53: Nit: Not sure why these are extension functions. For the other ones, it makes sense, since they operate on a nullable target. But the new ones can as well be regular methods of the class, or even better properties.

mnonnenmacher:
> > @Etsija I am currently still waiting for a positive positive review in order to merge it.
> 
> I could give it - from my part - but I believe there are some open debate with @mnonnenmacher so I'd rather not intervene.

Just approve PRs in such cases, if other users still have requested changes the merge will still be blocked.

oheger-bosch (commented):
I have some more minor points. Sorry for not noticing before.

Etsija (approved):

Etsija:
> > > @Etsija I am currently still waiting for a positive positive review in order to merge it.
> > 
> > 
> > I could give it - from my part - but I believe there are some open debate with @mnonnenmacher so I'd rather not intervene.
> 
> @Etsija Just approve PRs in such cases, if other users still have requested changes the merge will still be blocked.

OK I wasn't aware of that. Now approved, but Oliver seems to have some minor comments still. No requested changes though.

sschuberth:
> I have some more minor points. Sorry for not noticing before.

@wkl3nk, are you planning to address these points, or can we merge?

PR #2230, opened 2025-03-11
docs(cli): Add documentation how to remove Apple's quarantine attribute

MarcelBochtler:
On macOS, downloaded applications that are not signed and notarized have the `quarantine` flag assigned by macOS.
This has to be removed before running the binary.


website/docs/getting-started/cli.md
```diff
@@ -6,6 +6,16 @@ The ORT Server Client CLI (`osc`) is a command-line interface designed to intera
 
 The latest version of `osc` can be obtained from the [GitHub releases page](https://github.com/eclipse-apoapsis/ort-server/releases/latest).
 
+### macOS
+
+When using the CLI on macOS, Apple's Gatekeeper restrictions may prevent the application from running.
+To bypass this, you need to remove the quarantine attribute from the `osc` binary.
```
sschuberth, line 12: Should we use passive voice instead of "you" here? I.e. "needs to be removed".

wkl3nk (approved):

Issue #2238, opened 2025-03-12
Implement initial UI rearrangements for the run details view


sschuberth:
This is the parent issue for the UI rearrangements as discussed between @hanna-modica, @mnonnenmacher and @sschuberth at FOSS Backstage. The goal was to agree to a general structure of the run details view and esp. the left-hand side menu / grouping. The following sub-issues are not necessarily final, but go into the right direction of a commonly agreed UI layout.

PR #2239, opened 2025-03-12
Run index and id stuff in the UI

sschuberth:

### DIFF
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -217,7 +217,7 @@ export const Header = () => {
                       {...field}
                       type='text'
                       name='runId'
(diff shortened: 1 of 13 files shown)


mmurto:
If the commit types are intentional based on your judgment and you don't think they should be feat, feel free to merge.

mmurto (approved):

sschuberth:
> If the commit types are intentional based on your judgment and you don't think they should be feat, feel free to merge.

Yes, I regard these to be not-so-noteworthy preparations / clean-ups for the "real features" at https://github.com/eclipse-apoapsis/ort-server/issues/2238.

PR #2295, opened 2025-03-19
ui: Use server-side filtering for declared licenses of the "packages" table

Etsija:
Screenshot from 2025-03-19 12-27-44

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/packages/index.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/packages/index.tsx
@@ -32,6 +32,7 @@ import { prefetchUseRepositoriesServiceGetApiV1RepositoriesByRepositoryIdRunsByO
 import {
   usePackagesServiceGetApiV1RunsByRunIdPackagesSuspense,
   useRepositoriesServiceGetApiV1RepositoriesByRepositoryIdRunsByOrtRunIndexSuspense,
+  useRunsServiceGetApiV1RunsByRunIdPackagesLicensesSuspense,
 } from '@/api/queries/suspense';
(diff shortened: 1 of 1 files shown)


Etsija:
@sschuberth commit-lint fails again, but I don't know why; to me there's nothing obvious failing it.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/packages/index.tsx
```diff
@@ -205,7 +240,32 @@ const PackagesComponent = () => {
             {row.getValue('processedDeclaredLicense')}
           </div>
         ),
-        enableColumnFilter: false,
+        filterFn: (row, _columnId, filterValue): boolean => {
+          return filterValue.includes(
+            row.original.processedDeclaredLicense.spdxExpression
+          );
+        },
```
lamppu, line 247: Correct me if I'm missing something, but I don't think this should need a `filterFn` as this is all done with the API query.

lamppu (changes requested):

lamppu:
The title of this PR maybe sounds a bit misleading to me, I would read it as "the packages table used to use client-side manipulation, but this PR changes it so that it uses server-side manipulation".

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/packages/index.tsx
Etsija: Just some left-overs :smirk:

sschuberth:
> I put the link to Johanna's PR instead as a reference, and now commit-lint works.

Yes, that's the proper work-around.

lamppu (approved):

PR #2297, opened 2025-03-19
feat(ui): Improve layout in About dialog

wkl3nk:
Add a column based layout to the version information where key and value are properly vertically aligned.

image

### DIFF
--- a/ui/src/routes/about/index.tsx
+++ b/ui/src/routes/about/index.tsx
@@ -18,6 +18,7 @@
  */
 
 import { createFileRoute } from '@tanstack/react-router';
+import { Fragment } from 'react';
 
 import { useVersionsServiceGetApiV1VersionsSuspense } from '@/api/queries/suspense';
 import { Card, CardContent, CardHeader } from '@/components/ui/card';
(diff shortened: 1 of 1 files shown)


sschuberth:
I wonder whether we should go one step further, increase the spacing between key / value pairs, and in exchange remove the colons for "lighter" visuals. What do you think, @mmurto?

ui/src/routes/about/index.tsx
```diff
@@ -18,6 +18,7 @@
  */
 
 import { createFileRoute } from '@tanstack/react-router';
+import React from 'react';
```
Etsija, line 21: We prefer to use named imports when possible:
```
import { Fragment } from 'react';
...
<Fragment key='...'>
```

ui/src/routes/about/index.tsx
```diff
@@ -34,14 +35,15 @@ export const About = () => {
       <Separator />
       <CardContent className='pt-6'>
         <div>
-          <h3 className='font-semibold'>Version Information</h3>
-          <div className='text-sm'>
+          <h3 className='mb-4 font-semibold'>Version Information</h3>
+          <div className='grid grid-cols-[auto_1fr] gap-x-1 gap-y-1'>
```
Etsija, line 39: Using `gap-x-8` for example spaces the columns better, then you can lose the ":" from keys and get this:

Screenshot from 2025-03-20 06-40-47

Etsija (changes requested):

Etsija:
BTW nice catch using `grid grid-cols-[auto_1fr]`, I think that could be used in many places.

Etsija (approved):

ui/src/routes/about/index.tsx
sschuberth: @Etsija, should we apply the same to the key-values-pairs in the new overview bar to "lighten" the layout, WDYT?

PR #2298, opened 2025-03-19
Make authentication more flexible

oheger-bosch:
This PR introduces a new, ORT Server-specific `Authenticator` implementation that is based on infrastructure services. In contrast to the existing solution, this supports multiple credentials for the same host.

This PR depends on a change in ORT's `Authenticator` implemented in https://github.com/oss-review-toolkit/ort/pull/10060.

### DIFF
--- a/workers/common/src/main/kotlin/common/context/WorkerContextImpl.kt
+++ b/workers/common/src/main/kotlin/common/context/WorkerContextImpl.kt
@@ -34,6 +34,7 @@ import org.eclipse.apoapsis.ortserver.config.ConfigManager
 import org.eclipse.apoapsis.ortserver.config.Context
 import org.eclipse.apoapsis.ortserver.config.Path as ConfigPath
 import org.eclipse.apoapsis.ortserver.model.Hierarchy
+import org.eclipse.apoapsis.ortserver.model.InfrastructureService
 import org.eclipse.apoapsis.ortserver.model.OrtRun
 import org.eclipse.apoapsis.ortserver.model.PluginConfig
 import org.eclipse.apoapsis.ortserver.model.ProviderPluginConfiguration
@@ -44,6 +45,7 @@ import org.eclipse.apoapsis.ortserver.secrets.Path
 import org.eclipse.apoapsis.ortserver.secrets.SecretStorage
 
 import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
+import org.ossreviewtoolkit.utils.ort.OrtAuthenticator
 import org.ossreviewtoolkit.utils.ort.createOrtTempDir
 
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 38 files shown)


workers/common/src/test/kotlin/common/context/OrtServerAuthenticatorTest.kt
```diff
@@ -0,0 +1,215 @@
...
+
+        afterEach {
+            Authenticator.setDefault(originalAuthenticator)
+        }
+
+        "install()" should {
+            "check whether the authenticator is already installed" {
+                val authenticator = OrtServerAuthenticator.install()
+
+                OrtServerAuthenticator.install() shouldBe authenticator
```
mnonnenmacher, line 49: How does this test work, does it check for referential equality so that calling `install` again does not create a new instance?

workers/common/src/main/kotlin/common/context/WorkerContextImpl.kt
```diff
@@ -163,8 +170,28 @@ internal class WorkerContextImpl(
...
+        }
+
+        authenticator.updateAuthenticatedServices(authServices)
+    }
+
     override fun close() {
         tempDirectories.forEach { it.safeDeleteRecursively() }
+
+        // During the worker execution, ORT's authenticator was typically installed. Uninstall it first, which
+        // will OrtServerAuthenticator as default authenticator.
```
mnonnenmacher, line 192: Nit: Word missing after will, maybe "set" or "restore".

mnonnenmacher (changes requested):

workers/reporter/src/main/kotlin/reporter/ReporterRunner.kt
```diff
@@ -93,157 +89,156 @@ class ReporterRunner(
...
-        return contextFactory.createContext(runId).use { context ->
-            val copyrightGarbageFile =
-                if (evaluatorConfig != null) evaluatorConfig.copyrightGarbageFile else config.copyrightGarbageFile
-            val copyrightGarbage = configManager.readConfigFileValueWithDefault(
-                path = copyrightGarbageFile,
-                defaultPath = ORT_COPYRIGHT_GARBAGE_FILENAME,
-                fallbackValue = CopyrightGarbage(),
-                context = context.resolvedConfigurationContext
-            )
+        val copyrightGarbageFile =
```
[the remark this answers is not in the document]
oheger-bosch: Originally, `ReporterRunner` had a context factory as member property and used it to create a context. The same was true for `ReporterWorker`, so actually two context objects were created. Now, only `ReporterWorker` creates a context and passes it as parameter to the runner's `run()` function.

workers/common/src/test/kotlin/common/context/OrtServerAuthenticatorTest.kt
sschuberth: Maybe make this clearer via the `shouldBeSameInstanceAs` matcher?

oheger-bosch: Done.

mnonnenmacher (approved):

Issue #2176, opened 2025-03-03
Make the primary package ID syntax configurable


sschuberth:
Esp. when interacting with other compliance tools, e.g. in the context of the [OCCTET project](https://occtet.eu/), it can be confusing for users that ORT Server is the only tool that uses ORT's package IDs to refer to packages. Instead, using [package URLs (purl)](https://github.com/package-url/purl-spec/) probably is a better choice, esp. with [purl emerging as an ECMA standard](https://github.com/package-url/purl-spec/issues/295).

However, as right now quite a bit of project-specific configuration still needs to be done in one of ORT's configuration files as no UI to do the same exists, and these [configuration files](https://oss-review-toolkit.org/ort/docs/category/configuration) often need to refer to packages via ORT's IDs, also the latter should still be available for easy copying & pasting.

So the idea is to make the primary package ID that ORT Server uses throughout the UI should become configurable per user. That way, users can switch depending on their background (existing ORT users or new ORT Server users), and no screen space is "wasted" by displaying both ORT's IDs and purls in all places.

sschuberth:
I guess whether that approach works might depend on the column header to use, or? Like, say the column should be named "Package ID" independently of the configured primary package identifier, would that be possible?

Etsija:
[This PR](https://github.com/eclipse-apoapsis/ort-server/pull/2317) implements the requested changes for Packages table.

Other tables need `purl` to be returned from their corresponding endpoints.

PR #2309, opened 2025-03-20
ui: User settings page

Etsija:
This PR adds a User Settings page to the UI, accessible from the top-right menu. The page has so far only one user preference, the package ID type shown in the UI, and the setting does not yet have an effect in the UI. The PR however lays the groundwork for #2176 by persisting the user's preference for the package ID type to the browser's local storage, where it can be accessed when implementing #2176 which will be done in a follow-up PR.

Screenshot from 2025-03-20 13-15-26

### DIFF
--- a/ui/src/routes/settings/index.tsx
+++ b/ui/src/routes/settings/index.tsx
@@ -1,0 +1,129 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


ui/src/routes/settings/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  return (
+    <Card className='mx-auto w-full max-w-4xl'>
+      <CardHeader>
+        <h2 className='text-3xl font-bold tracking-tight'>User Settings</h2>
+        <CardDescription>
+          Here you can configure your own user preferences for ORT Server. The
+          preferences will be saved to your browser, so please note that
+          clearing your browser cache will remove them and return the default
+          preferences.
```
sschuberth, line 78: Wording-wise I believe we should not mix "settings" and "preferences", but stick to either. As the page is called "User Settings", I'd stick to "settings" as the term.

Also passive voice should be used, not directly talking to the user. I propose:

"These user-specific settings are saved as cookies in the browser. Clearing browser cookies will result in settings to be reset to their defaults."

lamppu (approved):

PR #2317, opened 2025-03-21
ui: Display configured Package ID in Packages table

Etsija:
PR #2309 laid the groundwork for setting and saving user preferences, so in this PR, either the ORT ID or PURL will be shown as the package ID in the Packages table. 

The default is ORT ID so there are no changes to the UI by default - the user needs to specifically go to the Settings page (accessible from the top-right menu) and choose the "PURL" option to see PURLs being used in the table.

There are several other tables where we want to show PURL instead of ORT ID in the "Package ID" column, but the corresponding endpoints do not return the purl, so they will need to be extended. Such endpoints are:
- all endpoints returning vulnerabilities
- issues endpoint
- rule violations endpoint

Screenshot from 2025-03-21 09-20-13

Screenshot from 2025-03-21 12-25-16

### DIFF
--- a/ui/src/components/dependency-paths.tsx
+++ b/ui/src/components/dependency-paths.tsx
@@ -19,23 +19,30 @@
 
 import { MoveRight } from 'lucide-react';
 
-import { Identifier, Package, ShortestDependencyPath } from '@/api/requests';
+import { Package, ShortestDependencyPath } from '@/api/requests';
 import { identifierToString } from '@/helpers/identifier-to-string';
 import { cn } from '@/lib/utils';
+import { PackageIdType } from '@/schemas';
 
 type DependencyPathsProps = {
   pkg: Package;
+  pkgIdType?: PackageIdType;
   className?: string;
 };
 
-export const DependencyPaths = ({ pkg, className }: DependencyPathsProps) => {
+export const DependencyPaths = ({
+  pkg,
+  pkgIdType,
+  className,
+}: DependencyPathsProps) => {
   return (
     <div className={className}>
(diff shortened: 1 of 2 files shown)


ui/src/helpers/identifier-to-string.ts
```diff
@@ -18,11 +18,20 @@
...
+  idType?: PackageIdType
+): string {
   if (!pkg) {
     return '';
   }
+
   const { type, namespace, name, version } = pkg;
-  return `${type ? type.concat(':') : ''}${namespace ? namespace.concat('/') : ''}${name ? name : ''}${version ? '@'.concat(version) : ''}`;
+  if (idType === 'PURL') {
+    return `pkg:${type ? type.concat('/') : ''}${namespace ? namespace.concat('/') : ''}${name ? name : ''}${version ? '@'.concat(version) : ''}`;
```
lamppu, line 33: The identifier type isn't one on one with the purl type, so I think this needs similar handling for the type as in ORT core [here](https://github.com/oss-review-toolkit/ort/blob/1f22ea48fb9b73f8213e07ec1e10577fb3801dc2/model/src/main/kotlin/utils/PurlExtensions.kt#L39-L77). And [here's](https://github.com/oss-review-toolkit/ort/blob/1f22ea48fb9b73f8213e07ec1e10577fb3801dc2/model/src/main/kotlin/utils/PurlUtils.kt#L37-L65) the `PurlType` enum class. So for example in your screenshot, Pod would be Cocoapods in the PURL.

lamppu (changes requested):

lamppu (approved):

PR #2318, opened 2025-03-21
build(cli): Remove JIB integration

sschuberth:

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -84,8 +84,6 @@ jobs:
           dockerfile: Scanner.Dockerfile
           jibImage: scanner-worker
           task: :workers:scanner:jibDockerBuild
-        - jibImage: cli
-          task: :cli:jibDockerBuild
         - preparationTask: :core:generateOpenApiSpec
           image: ui
           context: ui
--- a/buildSrc/src/main/kotlin/ort-server-kotlin-multiplatform-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/ort-server-kotlin-multiplatform-conventions.gradle.kts
@@ -28,6 +28,7 @@ plugins {
 }
 
 kotlin {
+    // Always target at least the JVM in multiplatform projects.
     jvm()
 
     sourceSets {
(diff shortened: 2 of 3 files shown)


sschuberth:
FYI @MarcelBochtler, as discussed.

.github/workflows/docker-build.yml
```diff
@@ -84,8 +84,6 @@ jobs:
           dockerfile: Scanner.Dockerfile
           jibImage: scanner-worker
           task: :workers:scanner:jibDockerBuild
-        - jibImage: cli
```
MarcelBochtler, line 87: Commit message: Maybe argue keeping the JVM target to be able to run the CLI on other targets than the native bin targets?
I guess the JVM has more target platforms as our native binary.

cli/build.gradle.kts
```diff
@@ -17,33 +17,20 @@
...
-tasks.withType<JibTask> {
-    notCompatibleWithConfigurationCache("https://github.com/GoogleContainerTools/jib/issues/3132")
-}
-
 kotlin {
-    jvm {
-        // Include Java sources into the JVM target's compilations, to make them available for the Jib plugin.
-        withJava()
-    }
+    jvm()
```
MarcelBochtler, line 33: This can be removed. It is part of the `ort-server-kotlin-multiplatform-conventions` plugin.

MarcelBochtler (changes requested):

.github/workflows/docker-build.yml
sschuberth: That would make sense to me only if we could name such a target that we do not target natively yet. Is there one you have in mind?

MarcelBochtler: Linux ARM 64

buildSrc/src/main/kotlin/ort-server-kotlin-multiplatform-conventions.gradle.kts
```diff
@@ -28,6 +28,7 @@ plugins {
 }
 
 kotlin {
+    // Always target at least the JVM in multiplatform projects.
```
MarcelBochtler, line 31: Commit message: I think it should be `targeted`

MarcelBochtler (approved):

PR #2329, opened 2025-03-21
feat(core): Cache Keycloak roles for one minute

mnonnenmacher:
Requesting the user roles from Keycloak can take some time, slowing down endpoint which would otherwise respond very fast. For example, it was observed in a setup with Keycloak 24 that the Keycloak role requests took between 500-1000ms while the remaining request takes <50ms.

This slows down the UI, especially on pages like the repository list, where a single page makes a lot of API requests.

To mitigate the issue, cache Keycloak roles for one minute. This speeds up such UI pages while at the same time ensuring that changes in the Keycloak roles are quickly reflected by the UI.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -36,6 +36,7 @@ rabbitMq = "5.25.0"
 versionsPlugin = "0.52.0"
 
 # Gradle dependencies
+aedile = "2.0.3"
 awsSdk = "1.4.46"
 azureIdentity = "1.15.4"
 azureMessagingServicebus = "7.17.10"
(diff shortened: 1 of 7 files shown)


mnonnenmacher:
I did some tests and the duration for requesting the last run for each repository on the repositories page goes down from 500-1200ms to 45-250ms in a setup with a lot of Keycloak roles and Keycloak 24.

gradle/libs.versions.toml
```diff
@@ -36,6 +36,7 @@ rabbitMq = "5.25.0"
 versionsPlugin = "0.52.0"
 
 # Gradle dependencies
+aedile = "2.0.3"
```
oheger-bosch, line 39: I am sceptic whether this functionality warrants adding a full-blown caching library. I would prefer a simple manual implementation, so that we know what is going on behind the scenes.

mnonnenmacher: The "simple" manual implementation would still have to deal with time-based eviction and thread/coroutine-safety,  and have to prevent that the cache uses too much memory. What's the benefit over using a library that's proven to work?

MarcelBochtler: On the other hand, using a well known library (which [caffeine](https://github.com/ben-manes/caffeine) is) might solve some issues in edge cases which are currently unknown to us.
I think in this case it's still ok to use a library.

wkl3nk (approved):

PR #2330, opened 2025-03-22
Add a mode to let workers handle only a single message

mnonnenmacher:
Add a way to configure workers to handle only a single message. This is useful when using Keda instead of the Kubernetes transport to create Kubernetes jobs which is used in an Azure deployment because the Kubernetes transport does not support the required configuration, and because it is more straightforward to configure the Kubernetes jobs with YAML files instead of environment variables passed to the orchestrator.

### DIFF
--- a/transport/activemqartemis/src/main/kotlin/ArtemisMessageReceiverFactory.kt
+++ b/transport/activemqartemis/src/main/kotlin/ArtemisMessageReceiverFactory.kt
@@ -27,6 +27,7 @@ import org.apache.qpid.jms.JmsConnectionFactory
 import org.eclipse.apoapsis.ortserver.config.ConfigManager
 import org.eclipse.apoapsis.ortserver.transport.Endpoint
 import org.eclipse.apoapsis.ortserver.transport.EndpointHandler
+import org.eclipse.apoapsis.ortserver.transport.EndpointHandlerResult
 import org.eclipse.apoapsis.ortserver.transport.MessageReceiverFactory
 import org.eclipse.apoapsis.ortserver.transport.json.JsonSerializer
 import org.eclipse.apoapsis.ortserver.utils.logging.withMdcContext
@@ -81,6 +82,8 @@ class ArtemisMessageReceiverFactory : MessageReceiverFactory {
             do {
                 runCatching {
                     val jmsMessage = consumer.receiveSave()
+                    var result = EndpointHandlerResult.CONTINUE
+
                     if (jmsMessage != null) {
                         val message = ArtemisMessageConverter.toTransportMessage(jmsMessage, serializer)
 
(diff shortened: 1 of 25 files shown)


transport/activemqartemis/src/main/kotlin/ArtemisMessageReceiverFactory.kt
```diff
@@ -81,6 +82,8 @@ class ArtemisMessageReceiverFactory : MessageReceiverFactory {
             do {
                 runCatching {
                     val jmsMessage = consumer.receiveSave()
+                    var result = EndpointHandlerResult.CONTINUE
```
sschuberth, line 85: IMO it would be nicer to (be able to) do

                    val result = if (jmsMessage == null) {
                        EndpointHandlerResult.CONTINUE
                    } else {
                        val message = ArtemisMessageConverter.toTransportMessage(jmsMessage, serializer)

                        withMdcContext(
                            "traceId" to message.header.traceId,
                            "ortRunId" to message.header.ortRunId.toString()
                        ) {
                            logger.debug(
                                "Received message '{}' with payload of type {}.",
                                message.header.traceId,
                                message.payload.javaClass.name
                            )

                            handler(message)
                        }
                    }

This requires only a trivial change for `withMdcContext` to return `T`. I'll leave it to you if you want to do that; otherwise this can also be done as a follow-up.

workers/advisor/src/main/kotlin/advisor/AdvisorComponent.kt
```diff
@@ -84,7 +84,8 @@ class AdvisorComponent : EndpointComponent<AdvisorRequest>(AdvisorEndpoint) {
             if (response != null) publisher.publish(OrchestratorEndpoint, response)
         }
 
-        EndpointHandlerResult.CONTINUE
+        val handleSingleJob = configManager.config.getBoolean("advisor.handleSingleJob")
```
sschuberth, line 87: Maybe better call this `singleJobOnly`? That would emphasize that they cannot `also* process single jobs, but *only* those.

mnonnenmacher: I think I'd like to keep "handle", what about `handleSingleJobOnly`?

sschuberth: Yes, a bit longer, but telling.

mnonnenmacher: Done.

sschuberth (approved):

transport/activemqartemis/src/main/kotlin/ArtemisMessageReceiverFactory.kt
sschuberth: See https://github.com/eclipse-apoapsis/ort-server/pull/2331.

oheger-bosch:
Instead of adding more complexity to the transport implementations and worker configurations, would it be a better option to implement a dedicated Keda transport?

mnonnenmacher:
> Instead of adding more complexity to the transport implementations and worker configurations, would it be a better option to implement a dedicated Keda transport?

No, Keda is just watching metrics and scaling jobs/deployments based on them, it cannot transfer any data to the created jobs.

PR #2332, opened 2025-03-23
feat(transport): Add support for Azure Service Bus

mnonnenmacher:
Add a transport implementation for the Azure Service Bus [1]. It was tested manually, functional tests will be added separately because using the emulator [2] turned out to be very complicated even though there is a testcontainer for it [3].

[1]: https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-messaging-overview
[2]: https://learn.microsoft.com/en-us/azure/service-bus-messaging/overview-emulator
[3]: https://testcontainers.com/modules/azure-servicebus/?language=java


transport/azure-servicebus/src/main/kotlin/AzureServicebusConfig.kt
```diff
@@ -0,0 +1,66 @@
...
+
+import org.eclipse.apoapsis.ortserver.config.ConfigManager
+
+import org.slf4j.Logger
+
+/**
+ * A class defining the configuration settings used by the RabbitMQ Transport implementation.
+ */
+class AzureServicebusConfig(
+    val namespace: String,
```
sschuberth, line 30: Nit: Missing docs for `namespace`.

transport/azure-servicebus/src/main/kotlin/AzureServicebusConfig.kt
```diff
@@ -0,0 +1,66 @@
...
+    val queueName: String
+) {
+    companion object {
+        /**
+         * Constant for the name of this transport implementation. This name is used for both the message sender and
+         * receiver factories.
+         */
+        const val TRANSPORT_NAME = "azure-servicebus"
+
+        private const val NAMESPACE_PROPERTY = "namespace"
```
sschuberth, line 42: Nit: Missing docs for `NAMESPACE_PROPERTY`.

transport/azure-servicebus/src/main/kotlin/AzureServicebusMessageReceiverFactory.kt
```diff
@@ -0,0 +1,101 @@
...
+        val stopSignal = CompletableDeferred<Unit>()
+
+        fun processMessage(context: ServiceBusReceivedMessageContext) {
+            context.complete() // Acknowledge the message to remove it from the queue.
+            val message = AzureServicebusMessageConverter.toTransportMessage(context.message, serializer)
+
+            MDC.put("traceId", message.header.traceId)
+            MDC.put("ortRunId", message.header.ortRunId.toString())
+
+            if (logger.isDebugEnabled) {
```
sschuberth, line 63: I'm a little bit confused about these explicit log level checks. When do you actually need them? There are none e.g. for lines 72 and 80.

mnonnenmacher: I guess it was there to avoid unnecessary string concatenation if the debug log level is not enabled. I have removed it because the performance impact is irrelevant here and we should finally switch to the same API we use in ORT.

transport/azure-servicebus/build.gradle.kts
```diff
@@ -0,0 +1,49 @@
...
+    testImplementation(projects.model)
+    testImplementation(testFixtures(projects.config.configSpi))
+    testImplementation(testFixtures(projects.transport.transportSpi))
+
+    testImplementation(libs.kotestAssertionsCore)
+    testImplementation(libs.kotestExtensionsTestContainer)
+    testImplementation(libs.kotestRunnerJunit5)
+    testImplementation(libs.mockk)
+    testImplementation(libs.testContainers)
+    testImplementation(libs.testContainersRabbitMq)
```
sschuberth, line 48: Is the Rabbit MQ stuff a left-over? Hint: You might want to run the `projectHealth` task on this new project.

mnonnenmacher: I have removed a few unused dependencies.

PR #2339, opened 2025-03-24
Don't create empty file on failing download

MarcelBochtler:


sschuberth:
How about extracting that common code to a helper function first, and then only fixing it in one place?

MarcelBochtler:
> How about extracting that common code to a helper function first, and then only fixing it in one place?

There was no common code before this change. The code duplication was introduced by this fixed.
I changed the PR to remove the duplication in the commit, which would have introduced it.

sschuberth:
> There was no common code before this change.

I see, sorry for that. Then I guess it would have been nicer to just add a refactoring commit on top of your original PR that combines the then obviously common code. In the current state, one first needs to confirm to oneself that the code would indeed be common.

mnonnenmacher (approved):

PR #2345, opened 2025-03-24
Git config file provider improvements

mnonnenmacher:
Several improvements to the `GitConfigFileProvider` to prepare for adding support for private repositories.

### DIFF
--- a/config/git/src/main/kotlin/GitConfigFileProvider.kt
+++ b/config/git/src/main/kotlin/GitConfigFileProvider.kt
@@ -33,6 +33,7 @@ import org.ossreviewtoolkit.downloader.WorkingTree
 import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.VcsType
 import org.ossreviewtoolkit.plugins.versioncontrolsystems.git.GitFactory
+import org.ossreviewtoolkit.utils.ort.createOrtTempDir
 
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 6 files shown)


config/git/src/main/kotlin/GitConfigFileProvider.kt
```diff
@@ -70,14 +70,17 @@ class GitConfigFileProvider internal constructor(
     constructor(gitUrl: String) : this(gitUrl, org.ossreviewtoolkit.utils.ort.createOrtTempDir())
 
     private val git = GitFactory.create()
+    private val lock = Any()
```
sschuberth, line 68: Commit message:
- "only performed for test first time a config file is requested" not sure what this should say, maybe "for checking first time"?

sschuberth: Why introduce `lock` instead of just doing `synchronized(this)` below?

config/git/src/main/kotlin/GitConfigFileProvider.kt
```diff
@@ -129,6 +129,7 @@ class GitConfigFileProvider internal constructor(
                 initWorkingTree(Context(requestedRevision))
             }
 
+            resolvedRevision = workingTree.getRevision()
```
sschuberth: Commit message:
- Typo in "con*n*fig worker"

config/git/src/main/kotlin/GitConfigFileProvider.kt
```diff
@@ -72,12 +72,7 @@ class GitConfigFileProvider internal constructor(
     private lateinit var resolvedRevision: String
 
     override fun resolveContext(context: Context): Context {
-        synchronized(lock) {
-            val revision = initWorkingTree(context.name)
-            git.updateWorkingTree(workingTree, revision, recursive = true)
-        }
-
-        resolvedRevision = workingTree.getRevision()
+        updateWorkingTree(context.name)
```
sschuberth: Commit message:
- Typo in "use it it"

config/git/src/main/kotlin/GitConfigFileProvider.kt
sschuberth: > It just felt a bit cleaner to have a dedicated lock object, but it doesn't make a big difference.

To me it's the opposite: If I read `synchronized(lock)`, I first need to convince myself about the lifetime of `lock` to understand the locking mechanism. But if like here the `lock` lifetime equals `this` lifetime, I strongly prefer the latter for clarity.

mnonnenmacher: I have added a commit on top.

sschuberth (approved):

PR #2351, opened 2025-03-25
ui: Visualize run and job duration data for a repository

Etsija:
This PR adds visualization for the run and job durations to the repository runs page. It aims at helping to better spot for example outliers in run durations, and the jobs that caused the outlier. The PR also serves as a start in bringing more data visualization to the UI by importing a modern charting library to the project.

Screenshot from 2025-03-25 09-44-03

Screenshot from 2025-03-25 09-44-39

NOTE: The usability of the "all runs" option of the chart needs to be assessed with real-world data, as testing it extensively was not possible in the development environment. Most likely the option needs to be accompanied in a follow-up PR with a date limiter. It must be noted however that even choosing the table page size as 50 and using the "current view" probably gives a good insight about the repository's "run healthiness", by showing the timing information for the last 50 runs.

After seeing how the charts behave with real-world examples, for a large amount of runs, we can fine-tune the component, for example by providing it a "maximum amount of runs shown in the chart" limiter.


ui/src/components/charts/job-durations.tsx
```diff
@@ -0,0 +1,251 @@
...
+              setFetchMode(value as 'VISIBLE_RUNS' | 'ALL_RUNS')
+            }
+          >
+            <div className='flex items-center space-x-2'>
+              <RadioGroupItem value='VISIBLE_RUNS' id='visible' />
+              <Label htmlFor='visible'>Show runs in current table view</Label>
+            </div>
+            <div className='flex items-center space-x-2'>
+              <RadioGroupItem value='ALL_RUNS' id='all' />
+              <Label htmlFor='all'>Show all runs</Label>
```
sschuberth, line 190: I'd just name this "All runs".

ui/src/components/charts/job-durations.tsx
```diff
@@ -0,0 +1,251 @@
...
+          Durations
+          <RadioGroup
+            value={fetchMode}
+            onValueChange={(value) =>
+              setFetchMode(value as 'VISIBLE_RUNS' | 'ALL_RUNS')
+            }
+          >
+            <div className='flex items-center space-x-2'>
+              <RadioGroupItem value='VISIBLE_RUNS' id='visible' />
+              <Label htmlFor='visible'>Show runs in current table view</Label>
```
sschuberth, line 186: I'd just name this "Current view".

ui/src/components/charts/job-durations.tsx
sschuberth: In combination with the above two, how about making the radio group horizontal instead of vertical, as we have more horizontal than vertical space to spare here?

If my above proposals for shorter names are too short, we could also consider to start a sentence and use the radio button choice to complete the sentence, like

    Show durations for (*) current view ( ) all runs

Etsija: > In combination with the above two, how about making the radio group horizontal instead of vertical, as we have more horizontal than vertical space to spare here?
> 
> If my above proposals for shorter names are too short, we could also consider to start a sentence and use the radio button choice to complete the sentence, like
> 
> ```
> Show durations for (*) current view ( ) all runs
> ```

I think that looks nicest, so implemented the choices in a sentence and updated the screenshots. See what you think.

sschuberth: I like it as well!

lamppu (approved):

PR #2353, opened 2025-03-25
fix(tasks): Move singleton injection to proper module

kamil-bielecki-bosch:
Fixes: https://github.com/eclipse-apoapsis/ort-server/pull/2000


sschuberth:
> Fixes: #2000

That's a PR, not an issue. So I guess the reference is wrong?

core/src/main/kotlin/di/Module.kt
```diff
@@ -67,7 +67,6 @@ import org.eclipse.apoapsis.ortserver.services.DefaultAuthorizationService
 import org.eclipse.apoapsis.ortserver.services.InfrastructureServiceService
```
oheger-bosch, line 67: Commit message: Please also mention that this is a fixup for the commit you created in your previous PR.

PR #2359, opened 2025-03-26
ui: Indicate a sorted column

Etsija:
Screenshot from 2025-03-26 07-55-27

Screenshot from 2025-03-26 07-55-57

For discussion: the sorting and filtering "hints" might be too subtle, especially with the light theme, where the bluish color of the sorting/filtering icons is almost non-distinguishable. Should we improve this by for example providing a greyish background to the icons, when a column is sorted or filtered?

### DIFF
--- a/ui/src/components/data-table/data-table-header.tsx
+++ b/ui/src/components/data-table/data-table-header.tsx
@@ -105,9 +105,9 @@ export function DataTableHeader<TData>({
                       onClick={column.getToggleSortingHandler()}
                     >
                       {column.getIsSorted() === 'asc' ? (
-                        <ChevronUp className='h-4 w-4' />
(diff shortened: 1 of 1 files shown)


sschuberth:
> Should we improve this by for example providing a greyish background to the icons, when a column is sorted or filtered, while making sure the indicator does not look not too intrusive?

At least for me, this has been readable enough so far.

Etsija:
> > Should we improve this by for example providing a greyish background to the icons, when a column is sorted or filtered, while making sure the indicator does not look not too intrusive?
> 
> At least for me, this has been readable enough so far.

OK then we can improve later, if needed.

lamppu (approved):

PR #2366, opened 2025-03-26
feat(tasks): Add logging configuration

oheger-bosch:
Copy `logback.xml` from the Kubernetes Job Monitor deployment to align the logging output and configuration with other ORT Server components.

### DIFF
--- a/tasks/src/main/resources/logback.xml
+++ b/tasks/src/main/resources/logback.xml
@@ -1,0 +1,34 @@
+<!--
+  ~ Copyright (C) 2023 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+  ~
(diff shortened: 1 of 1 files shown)


tasks/src/main/resources/logback.xml
```diff
@@ -0,0 +1,34 @@
+<!--
+  ~ Copyright (C) 2023 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
```
nnobelis, line 2: Nit: copyright year ?

oheger-bosch: As I copied the file, I did not want to change the copyright year.

MarcelBochtler: If the file was copied, the copyright year should not be updated.

MarcelBochtler (approved):

PR #2374, opened 2025-03-27
Improve displaying a user's name

sschuberth:


Etsija:
You misspelled username.

ui/src/helpers/extract-initials.ts
```diff
@@ -0,0 +1,54 @@
...
+ * License-Filename: LICENSE
+ */
+
+export function extractInitials(fullName: string | undefined): string {
+  // Remove non-alphanumeric characters.
+  const strippedName = (fullName ?? '').replace(/[^a-zA-Z0-9 ]/g, '');
+
+  // Split words, omitting empty strings.
+  const words = strippedName.split(' ').filter(Boolean);
+  if (words.length == 0) return '??';
```
mmurto, line 26: I wonder if we should rather return undefined here and handle what's displayed for them where this is used? I'd say we never want to show `??` as the avatar.

ui/src/helpers/extract-initials.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  const firstName = words.at(0)!;
+  const firstInitial = firstName.at(0)!.toUpperCase();
+
+  if (words.length == 1) return firstInitial;
+
+  const lastName = words.at(-1)!;
+  const lastInitial = lastName.at(0)!.toUpperCase();
```
mmurto, line 34: These should check for nullable values and act accordingly instead of using non-null assertion. For index access, `Array.at` seems to be not used as much, so I'd use `array[]` accessor instead as it's much more common, but you can do as you see best.

mmurto (changes requested):

ui/src/helpers/extract-initials.ts
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+export function extractInitials(fullName: string | undefined): string {
+  // Remove non-alphanumeric characters.
+  const strippedName = (fullName ?? '').replace(/[^a-zA-Z0-9 ]/g, '');
```
mmurto, line 22: Many nordic surnames begin with Ö/Ä/Å, so this probably shouldn't strip those. Likely for other languages as well, but not sure of the characters there.

mmurto (changes requested):

ui/src/helpers/extract-initials.ts
sschuberth: I was anyway annoyed by having to write a regex... doesn't TS have a built-in character function like `isAlphanum` like other languages do?

ui/src/helpers/extract-initials.ts
sschuberth: > These should check for nullable values and act accordingly instead of using non-null assertion.

But `lastName` is guaranteed to be non-empty (due to the word splitting) and thus always has a first character. Isn't a non-null assertion the right thing in such cases?

ui/src/helpers/extract-initials.ts
sschuberth: > I wonder if we should rather return undefined here and handle what's displayed for them where this is used?

Yes.

> I'd say we never want to show `??` as the avatar.

Here I disagree. I believe having that circle with `??` is better than an empty circle or no circle if initials could not be determined.

ui/src/helpers/extract-initials.ts
sschuberth: > For index access, `Array.at` seems to be not used as much

I was aligning on `at()` as I need to use an index of `-1` in one place, and `[]` does not seem to handle that.

ui/src/helpers/extract-initials.ts
mmurto: Not that I know of. But is there a reason for removing non-alphanumerics here? It exposes a lot of edge-cases as we want to enable Åke 李 to use the app. **If** we want to be strict with the initials, we probably should remove numerics too, or am I missing something here?

sschuberth: I'm using Unicode character categories now.

mmurto: Unicode regex Letter like [here](https://dev.to/tillsanders/let-s-stop-using-a-za-z-4a0m) maybe?

sschuberth: > Unicode regex Letter like [here](https://dev.to/tillsanders/let-s-stop-using-a-za-z-4a0m) maybe?

Yes, that exactly the same article that I found 😸

ui/src/helpers/extract-initials.ts
mmurto: > > These should check for nullable values and act accordingly instead of using non-null assertion.
> 
> But `lastName` is guaranteed to be non-empty (due to the word splitting) and thus always has a first character. Isn't a non-null assertion the right thing in such cases?

IMO it's better to tell that to TypeScript with proper non-null handling, as it's easier to see when looking at the code rather than requiring thinking about whether this can be null or not, and if the code above changes later on so that it could be null, non-null assertion still overrides it.

lamppu:
Just to let you know, you could also return first name and last name in the [useUser](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/hooks/use-user.ts) hook, extracting them from `auth` like this:

```
const firstName = auth?.user?.profile?.given_name;
const lastName = auth?.user?.profile?.family_name;
```

ui/src/helpers/extract-initials.ts
sschuberth: But this is a generic `extractInitials()` helper function now... who knows on what kind of strings we're going to use it 🤷🏻‍♂️

mmurto: Good point, can see this used for organizations and products as well.

ui/src/helpers/extract-initials.ts
sschuberth: I played around with a few syntaxes, but I liked none of them. Unfortunately, something concise like

    const firstName = words[0] && return undefined;

does not work in TypeScript. So I'd actually prefer to keep the non-null assertion in this case.

mmurto: Would this work?

```suggestion
  const firstName = words[0];

  if (!firstName || !firstName[0]) return undefined;

  const firstInitial = firstName[0].toUpperCase();

  const lastName = words[words.length - 1];

  if (!lastName || !lastName[0]) return firstInitial;

  const lastInitial = lastName[0].toUpperCase();
```

mmurto: > Would this work?

Oh it still needs to check for the words.length, or it will give double firstInitial for one word names.

sschuberth: Done, please have another look.

PR #2378, opened 2025-03-27
feat(git-config): Add support for private repositories

mnonnenmacher:
Add support for password authentication to the `GitConfigFileProvider`.
This is implemented using a custom `Authenticator` which is temporarily
installed when cloning the config repository.

### DIFF
--- a/config/git/src/main/kotlin/GitConfigFileProvider.kt
+++ b/config/git/src/main/kotlin/GitConfigFileProvider.kt
@@ -24,6 +24,8 @@ import com.typesafe.config.Config
 import java.io.File
 import java.io.InputStream
 
+import kotlin.time.measureTime
+
 import org.eclipse.apoapsis.ortserver.config.ConfigException
 import org.eclipse.apoapsis.ortserver.config.ConfigFileProvider
 import org.eclipse.apoapsis.ortserver.config.ConfigSecretProvider
@@ -125,7 +127,10 @@ class GitConfigFileProvider internal constructor(
                     val initRevision = requestedRevision.takeUnless { it.isEmpty() } ?: git.getDefaultBranchName(gitUrl)
                     val vcsInfo = VcsInfo(VcsType.GIT, gitUrl, initRevision)
 
-                    git.initWorkingTree(configDir, vcsInfo)
+                    measureTime { git.initWorkingTree(configDir, vcsInfo) }.also {
+                        logger.debug("Initialized Git working tree in $it.")
+                    }
+
                     initRevision
                 }
             } else {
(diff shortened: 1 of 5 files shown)


config/git/src/main/kotlin/GitConfigFileProvider.kt
```diff
@@ -42,29 +45,45 @@ import org.slf4j.LoggerFactory
...
+            val token = runCatching { secretProvider.getSecret(TOKEN_SECRET) }.onFailure {
+                logger.info("Could not get $TOKEN_SECRET from secret provider, continuing without it.")
+            }.getOrNull()
+
+            return GitConfigFileProvider(gitUrl, createOrtTempDir(), username, token)
         }
     }
 
-    private val git = GitFactory.create()
+    private val git = GitFactory.create(historyDepth = 1)
```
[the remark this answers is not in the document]
mnonnenmacher: I just forgot that I added this and also added some logging which should have been in separate commits.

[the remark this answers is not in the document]
mnonnenmacher: I extracted separate commits now.

config/git/src/main/kotlin/GitConfigFileProvider.kt
```diff
@@ -104,11 +121,13 @@ class GitConfigFileProvider internal constructor(
         synchronized(this) {
             // TODO: There might be a better way to do check if the configDir already contains a Git repository.
             val revision = if (!configDir.resolve(".git").isDirectory) {
-                val initRevision = requestedRevision.takeUnless { it.isEmpty() } ?: git.getDefaultBranchName(gitUrl)
-                val vcsInfo = VcsInfo(VcsType.GIT, gitUrl, initRevision)
+                withAuthenticator(username, token) {
```
oheger-bosch, line 126: Would it be possible to simply add the credentials to the Git URL? Then the whole authenticator implementation would not be needed.

sschuberth: Just some random thoughts:
- Would that work for all URL schemes? E.g. for SSH, always the generic `git` username needs to be used instead of the real one.
- Do we need to be careful to not log / store the URL with credentials then?

mnonnenmacher: * Would it be possible to simply add the credentials to the Git URL? Then the whole authenticator implementation would not be needed.

I don't know, I have not tried it because it did not appear as a good option to me. Isn't the plain password saved in `.git/config` then?

oheger-bosch (approved):

PR #2382, opened 2025-03-28
chore: Remove vulnerability resolutions

MarcelBochtler:
The vulnerable packages are no longer part of the dependency tree. Therefore, the vulnerability resolutions can be removed.


mnonnenmacher (approved):

sschuberth:
> The vulnerable packages are no longer part of the dependency tree. Therefore, the vulnerability resolutions can be removed.

Which reminds me that it would be really nice to have a feature to hint at obsolete resolutions in general...

PR #2384, opened 2025-03-28
Rework deletion of orphaned entities

oheger-bosch:
This PR reworks the service for deleting orphaned entities to speed up the task execution. Refer to the single commits for further details.

### DIFF
--- a/services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
+++ b/services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
@@ -32,7 +32,6 @@ import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.ProjectsDecla
 import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.ProjectsTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.evaluatorrun.RuleViolationsTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.ortrun.OrtRunsTable
-import org.eclipse.apoapsis.ortserver.dao.repositories.product.ProductsTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.repositoryconfiguration.PackageConfigurationsTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.repositoryconfiguration.PackageCurationDataAuthors
 import org.eclipse.apoapsis.ortserver.dao.repositories.repositoryconfiguration.PackageCurationDataTable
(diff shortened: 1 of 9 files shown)


kamil-bielecki-bosch (approved):
Looks fine to me,

wkl3nk (approved):

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
```diff
@@ -86,220 +94,209 @@ class OrphanRemovalService(
...
+                )
+
+            unionCondition(subQuery, id)
         }
 }
+
+/**
+ * Add a condition to match the given [matchColumn] to the given [unionQuery]. This results in a statement of the
+ * form: SELECT * FROM (SELECT * FROM sub_query) WHERE sub_query.id = matchColumn.
+ * TODO: Check whether there is a better way to do this with Exposed.
```
mnonnenmacher: Maybe it's possible to pass the return value of the `alias` call, e.g., `PackagesTable.binaryArtifactId.alias("id")`, to this function and use it directly in the `where` condition.

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
mnonnenmacher, line 1: Commit message nit: whitespace at start of line "of the shorter..."

tasks/src/main/resources/application.conf
```diff
@@ -46,3 +46,14 @@ dataRetention {
...
+orphanHandlers {
+  vcsInfo.limit = 1024
+  vcsInfo.limit = ${?ORPHANED_VCS_INFO_LIMIT}
+  vcsInfo.chunkSize = 64
+  vcsInfo.chunkSize = ${?ORPHANED_VCS_INFO_CHUNK_SIZE}
+  remoteArtifacts.limit = 1024
+  remoteArtifacts.limit = ${?ORPHANED_REMOTE_ARTIFACTS_LIMIT}
+  remoteArtifacts.chunkSize = 64
+  remoteArtifacts.chunkSize = ${?ORPHANED_REMOTE_ARTIFACTS_CHUNK_SIZE}
+}
```
mnonnenmacher, line 59: Do we really need to make this configurable?

oheger-bosch: Yes, this has a high impact on the execution time of the task. I also expect that we might experiment with these values. For instance, start with a higher limit, so that the orphans that have piled up are reduced more quickly; and then adapt the value to the number of orphans that are likely produced between two executions.

services/hierarchy/src/main/kotlin/OrphanRemovalService.kt
oheger-bosch: I have tested this, but it does not work. The alias obviously contains the table name; so the condition that is generated would have the form (in your example): `packages.id = vcs_info.id`.

PR #2392, opened 2025-03-30
feat(analyzer): Set the default branch as ORT run revision

### LINKED ISSUE #2082: ORT run details query should return more revision information

mnonnenmacher:
If an ORT run is started with an empty `revision`, the analyzer worker uses the default branch of the repository as revision. Store that as the `revision` of the ORT run to make clear what revision was used to initialize the repository.

Fixes #2082.

### DIFF
--- a/workers/analyzer/src/main/kotlin/analyzer/AnalyzerDownloader.kt
+++ b/workers/analyzer/src/main/kotlin/analyzer/AnalyzerDownloader.kt
@@ -49,10 +49,12 @@ class AnalyzerDownloader {
         val vcs = VersionControlSystem.forUrl(repositoryUrl, config)
         requireNotNull(vcs) { "Could not determine the VCS for URL '$repositoryUrl'." }
 
+        val initRevision = revision.takeUnless { it.isEmpty() } ?: vcs.getDefaultBranchName(repositoryUrl)
+
         val vcsInfo = VcsInfo(
             type = vcs.type,
             url = repositoryUrl,
-            revision = revision.takeUnless { it.isEmpty() } ?: vcs.getDefaultBranchName(repositoryUrl),
+            revision = initRevision,
             path = path
         )
 
@@ -67,7 +69,7 @@ class AnalyzerDownloader {
                     "'$resolvedRevision'."
         )
 
-        return DownloadResult(outputDir, resolvedRevision)
+        return DownloadResult(outputDir, initRevision, resolvedRevision)
     }
 
     /**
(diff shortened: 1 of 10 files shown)


sschuberth:
Will that "automatically" also fix the default branch name to be shown in the UI (the new overview bar of a run), or is more work needed for that?

mnonnenmacher:
> Will that "automatically" also fix the default branch name to be shown in the UI (the new overview bar of a run), or is more work needed for that?

The happens automatically because the UI uses the `revision` property.

workers/analyzer/src/main/kotlin/analyzer/AnalyzerDownloader.kt
```diff
@@ -49,10 +49,12 @@ class AnalyzerDownloader {
         val vcs = VersionControlSystem.forUrl(repositoryUrl, config)
         requireNotNull(vcs) { "Could not determine the VCS for URL '$repositoryUrl'." }
 
+        val initRevision = revision.takeUnless { it.isEmpty() } ?: vcs.getDefaultBranchName(repositoryUrl)
```
sschuberth, line 52: I'm unsure about the "init" here and in `DownloadResult`, as to me it sounds more like the "definite" / resolved revision used to initialize the repository on a low level. Maybe "originalRevision" instead?

mnonnenmacher: I called it "init" because it's used to initialize the repository. "originalRevision" sound like the raw input, not like a value that might come from `getDefaultBranchName`.

sschuberth (approved):

Issue #984, opened 2024-09-05
Consider Keycloak resources for authorization instead of dynamically created roles


mmurto:
As a warning, I'm not super deep on Keycloak. Nevertheless, the way ORT Server uses Keycloak's client roles dynamically (`permission_organization_$ORGID_write` etc.) seems unintuitive. What ORT Server wants to do is to protect resources (organizations, products etc.), and for that Keycloak has [native functionality to manage resources, policies and permissions](https://www.keycloak.org/docs/latest/authorization_services/index.html#_resource_overview). To me this seems like a better fit than the current method of handling authorization. There have also been some reports where [large number of roles has led to performance issues](https://github.com/keycloak/keycloak/issues/12885), and with the current authorization model the role count grows very large.

What is the reason for going with these dynamic roles for authorization instead of Keycloak's resource authorization service?

mnonnenmacher:
I totally agree with that. I wanted to look into resources since the initial implementation but never found the time.

@mmurto In the context of #504, do you know if the other auth providers you suggest there have similar features?

sschuberth:
> I totally agree with that. I wanted to look into resources since the initial implementation but never found the time.

Is this by now on your agenda @mnonnenmacher?

mnonnenmacher:
> > I totally agree with that. I wanted to look into resources since the initial implementation but never found the time.
> 
> Is this by now on your agenda @mnonnenmacher?

It's low priority to me because the current implementation works and there are other more important tasks.

mnonnenmacher:
I am considering if it would be a better solution to move the role management into the database. This would make it permission handling a lot easier, for example, when doing a paginated query for organizations, they could already be filtered by the database query. This would also simplify the interface to the auth provider and make #504 easier.

@mmurto Do you have an opinion on that?

mmurto:
I think it would likely be a good solution. To me it seems that if more fine-graned authorization is needed than group roles, the ecosystem seems to gravitate towards using a separate service for that, like [OpenFGA](https://github.com/openfga), [Ory Keto](https://github.com/ory/keto) and others. While those could provide some benefits, I'm not sure if our use-case and scale warrant a separate service for just authorization.

What is a part of most of the authentication providers, like Keycloak v25 onwards, is organization support. The providers often provide some general functionality, like member management, invitation emails etc. for organizations, so having that managed by the authentication service could be worthwhile, though I'm not sure how it should be architected.

mmurto:
[...]

PR #2415, opened 2025-04-01
feat(core): Clear role cache after write operation

### LINKED ISSUE #2413: Request fails for product/repository for a few seconds after adding the entity
When a non-superuser user (for example organization writer or admin) creates a new product or repository, and is redirected to said product/repository, the request to fetch information about that product or repository fails for a few seconds. This leads to an unhandled error and the following screen:

Image

This probably should have some error catching in the UI, but we should also check if there is also something going wrong in the backend (for example with the caching of the Keycloak roles).

mnonnenmacher:
Clear the role cache for a user after a write operation as it could have changed the roles. For example, if a new product is created this also creates new roles for that product, therefore the role cache needs to be updated on the call.

Fixes #2413.

### DIFF
--- a/core/src/main/kotlin/plugins/Authentication.kt
+++ b/core/src/main/kotlin/plugins/Authentication.kt
@@ -28,11 +28,13 @@ import com.sksamuel.aedile.core.Cache
 import com.sksamuel.aedile.core.asCache
 import com.sksamuel.aedile.core.expireAfterWrite
 
+import io.ktor.http.HttpMethod
 import io.ktor.server.application.Application
 import io.ktor.server.application.install
 import io.ktor.server.auth.Authentication
 import io.ktor.server.auth.jwt.jwt
 import io.ktor.server.config.ApplicationConfig
+import io.ktor.server.request.httpMethod
 
 import java.net.URI
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/plugins/Authentication.kt
```diff
@@ -84,6 +86,16 @@ fun Application.configureAuthentication() {
...
+
+                    // Clear the cache for the user after a write operation as it could have changed the roles.
+                    @Suppress("ComplexCondition")
+                    if (this.request.httpMethod == HttpMethod.Delete ||
+                        this.request.httpMethod == HttpMethod.Patch ||
+                        this.request.httpMethod == HttpMethod.Post ||
+                        this.request.httpMethod == HttpMethod.Put
+                    ) {
+                        roleCache.invalidate(payload.subject)
+                    }
```
lamppu, line 98: I tested locally moving the `OrtPrincipal(payload, roles)` after the if condition, then it does fix the issue.

mnonnenmacher: Yes, this broke the return value of the function, I have updated the PR.

lamppu (approved):

core/src/main/kotlin/plugins/Authentication.kt
```diff
@@ -83,6 +85,17 @@ fun Application.configureAuthentication() {
             validate { credential ->
                 credential.payload.takeIf(this@configureAuthentication::validateJwtPayload)?.let { payload ->
                     val roles = getRoles(payload.subject, keycloakClient, roleCache)
+
+                    // Clear the cache for the user after a write operation as it could have changed the roles.
+                    @Suppress("ComplexCondition")
+                    if (this.request.httpMethod == HttpMethod.Delete ||
```
oheger-bosch, line 91: As a quick fix, this is probably okay. But on the long run, I would prefer to have a more focused invalidation when it is actually needed. Also, a test would be nice.

mnonnenmacher: Ideally we could find a solution that allows us to remove the cache altogether, like https://github.com/eclipse-apoapsis/ort-server/issues/984#issuecomment-2550639579.

core/src/main/kotlin/plugins/Authentication.kt
sschuberth: Did this resolve the [above comment](https://github.com/eclipse-apoapsis/ort-server/pull/2415#issuecomment-2768877643) as well?

PR #2416, opened 2025-04-01
CLI: Add `--json` option to make the CLI output machine readable

MarcelBochtler:
Example for the human-readable `osc runs info` command:
image

### DIFF
--- a/cli/src/commonMain/kotlin/OrtServerMain.kt
+++ b/cli/src/commonMain/kotlin/OrtServerMain.kt
@@ -47,7 +47,7 @@ const val COMMAND_NAME = "osc"
 fun main(args: Array<String>) {
     val cli = OrtServerMain()
 
-    try {
+    runCatching {
         // Kotlin native does not support suspending entrypoints, see
         // https://youtrack.jetbrains.com/issue/KT-52753/Native-Support-suspending-entrypoints.
         runBlocking {
@@ -55,18 +55,18 @@ fun main(args: Array<String>) {
         }
 
         exitProcess(0)
-    } catch (@Suppress("SwallowedException") e: AuthenticationException) {
-        cli.echoError("Authentication failed. Please check your credentials.")
-    } catch (e: OrtServerCliException) {
-        cli.echoError(e.message)
-    } catch (e: OrtServerException) {
-        cli.echoError(e.message)
-    } catch (e: CliktError) {
-        // The jsonFormat flag is not supported for the help message.
-        cli.echoFormattedHelp(e)
-        cli.currentContext.exitProcess(e.statusCode)
-    } catch (@Suppress("SwallowedException", "TooGenericExceptionCaught") e: Exception) {
-        cli.echoError("An unexpected error occurred.")
+    }.onFailure { e ->
+        when (e) {
+            is AuthenticationException -> cli.echoError("Authentication failed. Please check your credentials.")
+            is OrtServerCliException, is OrtServerException -> cli.echoError(e.message)
+            is CliktError -> {
+                // The jsonFormat flag is not supported for the help message.
+                cli.echoFormattedHelp(e)
+                cli.currentContext.exitProcess(e.statusCode)
+            }
+
+            else -> cli.echoError("An unexpected error occurred.")
+        }
     }
 
     exitProcess(1)
--- a/cli/src/commonMain/kotlin/OrtServerMain.kt
+++ b/cli/src/commonMain/kotlin/OrtServerMain.kt
@@ -19,12 +19,14 @@
 
 package org.eclipse.apoapsis.ortserver.cli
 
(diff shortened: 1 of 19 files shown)


cli/src/commonMain/kotlin/OrtServerMain.kt
```diff
@@ -69,7 +69,7 @@ fun main(args: Array<String>) {
 
 class OrtServerMain : SuspendingNoOpCliktCommand(COMMAND_NAME) {
     init {
-        completionOption()
+        completionOption(hidden = true)
```
mnonnenmacher, line 77: Would you mind adding a [check like in ORT](https://github.com/oss-review-toolkit/ort/blob/f26c647a313a312659768a46220e10d05d289e35/.github/workflows/static-analysis.yml#L35-L59) to ensure that the completions are kept up to date?

cli/src/commonMain/kotlin/utils/PrintHelper.kt
```diff
@@ -41,5 +43,19 @@ internal fun SuspendingCliktCommand.echoError(message: String?) {
...
 
+/**
+ * Print the [message] to stdout. If required as a json object.
+ */
+internal fun SuspendingCliktCommand.echoMessage(message: CliPrintable) {
+    if (useJsonFormat) {
+        echo(message.json())
+    } else {
+        // TODO: Replace with human readable formatting.
+        echo(message.toString())
```
mnonnenmacher: The `MessagePrintable` does not override `toString()`, does this actually work?

cli/src/commonMain/kotlin/OrtServerMain.kt
mnonnenmacher: Because I think you should have updated the completion scripts as part of this PR as well?

cli/src/commonMain/kotlin/utils/PrintHelper.kt
MarcelBochtler: Yes, it prints the object with its values.
Nothing I want to keep, but as the message is changed in the following commit, I think that it is good enough.

scripts/cli/generate_completion_scripts.sh
```diff
@@ -0,0 +1,27 @@
...
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+#
+
+GIT_ROOT=$(git rev-parse --show-toplevel)
+
+(cd $GIT_ROOT &&
+  ./gradlew cli:jvmRun --args='--generate-completion=bash' -DmainClass=org.eclipse.apoapsis.ortserver.cli.OrtServerMainKt --quiet >| integrations/completions/osc-completion.bash &&
```
sschuberth, line 24: Nit: Is it encouraged to use "absolute paths" to Gradle tasks like `:cli:jvmRun`, i.e. with a leading colon.

sschuberth, line 24: A plain `>` instead of `>|` should do, as you don't have `set -o noclobber`.

cli/src/commonMain/kotlin/model/printables/OrtRunPrintable.kt
```diff
@@ -19,14 +19,91 @@
...
+    override fun humanReadable(): Widget {
+        return grid {
+            row(yellow("ORT Run Information"))
+            formattedRow("ID", ortRun.id)
+            formattedRow("Repository ID", ortRun.repositoryId)
+            formattedRow("Index", ortRun.index)
+            formattedStatusRow("Status", ortRun.status)
+            formattedRow("Created", ortRun.createdAt)
+            formattedRow("Revision", ortRun.revision)
+            ortRun.path?.takeIf { it.isNotBlank() }?.let { formattedRow("Path", it) }
```
sschuberth, line 49: Prefer `also` over `let` if the return value is not needed / unused.

cli/src/commonMain/kotlin/OrtServerMain.kt
```diff
@@ -55,8 +55,8 @@ fun main(args: Array<String>) {
         }
 
         exitProcess(0)
-    } catch (e: AuthenticationException) {
-        cli.echoError(e.message)
+    } catch (@Suppress("SwallowedException") e: AuthenticationException) {
```
sschuberth, line 58: BTW, I believe you could avoid those `Suppress` annotations by ruing `runCatching` / `onFailure` with a `when` clause inside.

sschuberth (dismissed):

cli/src/commonMain/kotlin/OrtServerMain.kt
MarcelBochtler: I added a commit with this refactoring.

Issue #377, opened 2024-06-06
Come up with a way to "inject" ORT core plugins into the deployment


sschuberth:
Users of the server might want to operate it with additional ORT plugins installed, like third-party [package configuration providers](https://github.com/oss-review-toolkit/ort/tree/main/plugins/package-configuration-providers) or [scanners](https://github.com/oss-review-toolkit/ort/tree/main/plugins/scanners).

We should investigate what's the best way to take such plugins into use / deploy them with the server. To somewhat simplify the boundary conditions, let's assume the plugin is present in the binary form of a fat JARs that [shades](https://github.com/johnrengelman/shadow) its transitive dependencies (except for ORT itself). How would we "inject" such a plugin JAR into the deployment?

PR #2428, opened 2025-04-02
Show installed ORT plugins

mnonnenmacher:
Add an endpoint to get the installed ORT plugins and show them in the UI.

This PR tests organizing code by feature instead of technology, this requires moving some code to new Gradle modules to make it reusable.

To add context, I propose to the change the Gradle module in future to the following, this is the basis for the new modules I create in this PR:

```
ort-server
+--- shared (stateless utility code, models)
|    +--- keycloak-client
|    +--- ktor-utils
|    +--- database-utils
+--- infrastructure (connections to external systems)
|    +--- transport
|    |    +--- rabbitmq
|    +--- secrets
|    |    +--- vault
+--- components (stateful vertical slices)
|    +--- plugin-manager
|    |    +--- api
|    |    +--- backend
|    +--- organizations
|    |    +--- api
|    |    +--- backend
+--- services (functionality using multiple components)
|    +--- statistics
+--- applications (deployable applications)
|    +--- core
|    +--- orchestrator
|    +--- analyzer-worker
```


components/plugin-manager/implementation/src/main/kotlin/endpoints/GetInstalledPlugins.kt
```diff
@@ -0,0 +1,120 @@
...
+    val scanners = ScannerWrapperFactory.ALL.values.map {
+        it.descriptor.mapToApi(PluginType.SCANNER)
+    }
+
+    val allPlugins = advisors +
+            packageConfigurationProviders +
+            packageCurationProviders +
+            packageManagers +
+            reporters +
+            scanners
```
mnonnenmacher, line 94: The code to get ORT plugin information will be moved to a dedicated class in a follow-up PR as part of adding more functionality, I just left it here to keep the first version simple.

components/authorization/implementation/build.gradle.kts
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+plugins {
+    id("ort-server-kotlin-jvm-conventions")
+    id("ort-server-publication-conventions")
+}
+
+group = "org.eclipse.apoapsis.ortserver.components.authorization"
```
mnonnenmacher, line 25: In this iteration the module contains only code that could also be in a shared module at `shared/authorization`, however, this will change in a later iteration so I made it a component from the beginning.

components/plugin-manager/api/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+
+plugins {
+    id("ort-server-kotlin-multiplatform-conventions")
+    id("ort-server-publication-conventions")
+
+    // Apply third-party plugins.
+    alias(libs.plugins.kotlinSerialization)
+}
+
+group = "org.eclipse.apoapsis.ortserver.components.pluginmanager"
```
mnonnenmacher, line 28: I am considering to rename the `api` and `implementation` modules to `frontend` and `backend` in a follow-up to clarify their purpose, please share your opinion.

sschuberth: Isn't "frontend" usually the UI? Also I'm used to the "api" and "implementation" terms from Gradle dependencies. So if this is about a similar concept on the module level, I think I'd prefer to stick with "api" and "implementation".

components/plugin-manager/api/src/commonMain/kotlin/PluginDescriptor.kt
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.components.pluginmanager
+
+import kotlinx.serialization.Serializable
+
+@Serializable
```
sschuberth, line 24: This should go below the docs.

components/plugin-manager/api/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+
+kotlin {
+    linuxX64()
+    macosArm64()
+    macosX64()
+    mingwX64()
```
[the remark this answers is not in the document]
mnonnenmacher: Yes, the module contains classes for the REST API which should be available to the CLI.

components/plugin-manager/api/build.gradle.kts
mnonnenmacher: This is the misunderstanding I was thinking of, "api" actually contains models for the REST API and therefore needs to be a multiplatform module, "implementation" contains the whole backend implementation including the JVM API to be consumed by other modules.
But yes, "frontend" sounds like it would be the frontend, and "api" is already used for the top-level module, so maybe "api" and "backend"?

oheger-bosch: I am also fine with the terms "api" and "implementation". I would prefer "implementation" over "backend" because of the risk of confusing this with the distinction between frontend and backend.

components/plugin-manager/api/src/commonMain/kotlin/PluginDescriptor.kt
```diff
@@ -0,0 +1,71 @@
...
+
+package org.eclipse.apoapsis.ortserver.components.pluginmanager
+
+import kotlinx.serialization.Serializable
+
+/**
+ * A descriptor holding the metadata of a plugin.
+ */
+@Serializable
+data class PluginDescriptor(
```
oheger-bosch, line 28: Question: Will there be a distinction between classes in the public API and internal data model classes and a mapping layer between those?

components/plugin-manager/implementation/build.gradle.kts
```diff
@@ -39,6 +39,8 @@ repositories {
 dependencies {
     api(projects.components.pluginManager.api)
 
+    implementation(projects.components.authorization.implementation)
```
oheger-bosch, line 42: Hm, so it is allowed for components to depend on the implementation module of another component?

oheger-bosch (approved):
Some questions and comments. But since this is all experimental to a certain degree, a picky review probably does not make sense.

MarcelBochtler (approved):

sschuberth:
BTW, I just want to mention https://github.com/eclipse-apoapsis/ort-server/issues/377 in this context. I would be awesome if in the long run we could have real plugin management (similar to tools like Jenkins) that also allows us to add / remove plugins (from remote sources).

components/plugin-manager/api/src/commonMain/kotlin/PluginDescriptor.kt
mnonnenmacher: I want to try to avoid having pure copies that to reduce the amount of boilerplate, but let's see how that works out.

mnonnenmacher:
> BTW, I just want to mention #377 in this context. I would be awesome if in the long run we could have real plugin management (similar to tools like Jenkins) that also allows us to add / remove plugins (from remote sources).

Yes, that would be nice, but it will be a lot of work I guess.

Issue #1926, opened 2025-01-31
Show inherited secrets


mnonnenmacher:
The pages to edit secrets for products and repositories should also show inherited secrets. So on the product secrets page the inherited organization secrets should be shown as well, and on the repository secrets page the inherited organization and product secrets should be shown.

This makes it easier to see which secrets are actually available. Also, if a secret overrides an inherited secret with the same name, this could be indicated by some icon.

PR #2424, opened 2025-04-02
refactor(ui): Align Action column widths

wkl3nk:
Change the alignment of the Actions column to left, and change the color of the trash icon to red, in order to align it with the
User Management UI for Organizations, Products and Repositories.

image

PR #2436, opened 2025-04-03
feat(ui): Add inherited secrets to secrets page 

bs-ondem:

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/secrets/index.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/secrets/index.tsx
@@ -74,7 +74,7 @@ import {
   productPaginationSearchParameterSchema,
 } from '@/schemas';
 
-const defaultPageSize = 10;
+const defaultPageSize = 5;
 
 const ActionCell = ({ row }: CellContext<Secret, unknown>) => {
   const params = Route.useParams();
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/secrets/index.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/secrets/index.tsx
@@ -28,12 +28,20 @@ import {
 import { EditIcon, PlusIcon } from 'lucide-react';
 
 import {
+  useOrganizationsServiceGetApiV1OrganizationsByOrganizationId,
+  useOrganizationsServiceGetApiV1OrganizationsByOrganizationIdSecrets,
+  useProductsServiceGetApiV1ProductsByProductId,
+  useProductsServiceGetApiV1ProductsByProductIdSecrets,
   useRepositoriesServiceDeleteApiV1RepositoriesByRepositoryIdSecretsBySecretName,
   useRepositoriesServiceGetApiV1RepositoriesByRepositoryId,
   useRepositoriesServiceGetApiV1RepositoriesByRepositoryIdSecrets,
   useRepositoriesServiceGetApiV1RepositoriesByRepositoryIdSecretsKey,
 } from '@/api/queries';
 import {
+  prefetchUseOrganizationsServiceGetApiV1OrganizationsByOrganizationId,
+  prefetchUseOrganizationsServiceGetApiV1OrganizationsByOrganizationIdSecrets,
+  prefetchUseProductsServiceGetApiV1ProductsByProductId,
+  prefetchUseProductsServiceGetApiV1ProductsByProductIdSecrets,
   prefetchUseRepositoriesServiceGetApiV1RepositoriesByRepositoryId,
   prefetchUseRepositoriesServiceGetApiV1RepositoriesByRepositoryIdSecrets,
 } from '@/api/queries/prefetch';
@@ -60,7 +68,11 @@ import {
 } from '@/components/ui/tooltip';
 import { toast } from '@/lib/toast';
 import { cn } from '@/lib/utils';
-import { paginationSearchParameterSchema } from '@/schemas';
+import {
+  orgPaginationSearchParameterSchema,
+  paginationSearchParameterSchema,
+  productPaginationSearchParameterSchema,
+} from '@/schemas';
 
 const defaultPageSize = 10;
 
@@ -134,7 +146,7 @@ const ActionCell = ({ row }: CellContext<Secret, unknown>) => {
   );
 };
 
-const columns: ColumnDef<Secret>[] = [
+const baseColumns: ColumnDef<Secret>[] = [
   {
     accessorKey: 'name',
     header: 'Name',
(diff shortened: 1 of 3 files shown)


bs-ondem:
Product Secrets page: 

Bildschirmfoto 2025-04-03 um 18 29 39

Repository Secrets page:
Bildschirmfoto 2025-04-03 um 18 30 26

bs-ondem:
> Hmm, wouldn't it be nicer to have all of these in _one_ table, plus a column that indicates on which level the secret is defined? IMO that would also be clearer in case of overridden secrets from an "upper" level, as the column would show which "level" in in effect.

I followed the approach used by GitHub Actions secrets, where secrets are displayed in separate tables based on their level (Repository Secrets and Organization Secrets). 
Merging all secrets into a single table would require fetching all secrets from all levels and handling pagination on the client side.

ui/src/routes/organizations/$orgId/products/$productId/secrets/index.tsx
```diff
@@ -203,66 +261,122 @@ const ProductSecrets = () => {
...
+              };
+            }}
+          />
+        </CardContent>
+      </Card>
+      <Card className='mt-4'>
+        <CardHeader>
+          <CardTitle>Organization Secrets</CardTitle>
+          <CardDescription>
+            Inherited secrets from {organization.name} for {product.name}.
```
mnonnenmacher, line 313: I would remove "for {product.name}", it's obvious.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/secrets/index.tsx
```diff
@@ -207,67 +326,185 @@ const RepositorySecrets = () => {
...
+              };
+            }}
+          />
+        </CardContent>
+      </Card>
+      <Card className='mt-4'>
+        <CardHeader>
+          <CardTitle>Product Secrets</CardTitle>
+          <CardDescription>
+            Inherited secrets from {product.name} for {repo.url}.
```
mnonnenmacher, line 379: I would remove "for {repo.url}" (same below).
Also, this commit should close #1926.

mnonnenmacher (changes requested):
You need to get rid of the merge commit.

sschuberth:
> I followed the approach used by GitHub Actions secrets, where secrets are displayed in separate tables based on their level (Repository Secrets and Organization Secrets).

I'm still now convinced about this UX. Can we then at least visually indicate whether a secret has been overridden on some higher level?

> Merging all secrets into a single table would require fetching all secrets from all levels and handling pagination on the client side.

Which would probably not be a problem in this case, given the rather small total number of secrets to expect. Or what do @Etsija and @mmurto think?

mnonnenmacher:
> > I followed the approach used by GitHub Actions secrets, where secrets are displayed in separate tables based on their level (Repository Secrets and Organization Secrets).
> 
> I'm still now convinced about this UX. Can we then at least visually indicate whether a secret has been overridden on some higher level?

That would be good but I would see it as an improvement that can be done later in a separate change, because IMO the information if a secret is overridden should be provided by the endpoints and not be done in the UI.

> > Merging all secrets into a single table would require fetching all secrets from all levels and handling pagination on the client side.
> 
> Which would probably not be a problem in this case, given the rather small total number of secrets to expect. Or what do @Etsija and @mmurto think?

We currently have an org with more than 150 secrets.
We discussed the option to use a single table before the implementation, I prefer the separate tables because to me it is a better UX to not merge the data from different sources.

sschuberth:
> We discussed the option to use a single table before the implementation,

Knowing that beforehand, e.g. as part of the PR description, would have been nice.

> I prefer the separate tables because to me it is a better UX to not merge the data from different sources.

While I don't generally agree to that, I won't block this PR because of that.

However, IMO we need to be a bit careful to not rush UI / UX decisions too much... certainly sometimes having some UI is better than having no UI, but cleaning up / changing a user-facing UI is something that's not as lightheartedly done as refactoring code, which is not visible to the user.

mnonnenmacher:
> However, IMO we need to be a bit careful to not rush UI / UX decisions too much...

I agree, we need some lightweight process to agree on UI decisions that provides a place for feedback without slowing us down too much. In most cases it should work to make sure that issues with proper descriptions are created before the implementation starts, and to take them to the community meeting if there is no agreement on the design, because asynchronous communication can be very slow in such cases.

sschuberth:
BTW, are the screenshots outdated, or are the trash bin icons really not red here yet?

bs-ondem:
> BTW, are the screenshots outdated, or are the trash bin icons really not red here yet?

It seems that they are not red in the `Secrets` page. I just checked the Users/Runs page and there, it is red:
Bildschirmfoto 2025-04-09 um 10 44 05

Bildschirmfoto 2025-04-09 um 10 44 47

lamppu:
> > BTW, are the screenshots outdated, or are the trash bin icons really not red here yet?
> 
> It seems that they are not red in the `Secrets` page. I just checked the Users/Runs page and there, it is red

There's an alignment coming for the trash bin icon color in https://github.com/eclipse-apoapsis/ort-server/pull/2424.

bs-ondem:
I just checked the code and the InfrastructureServices, Admin Users and each Secrets page does not use the red trash bin icon.

lamppu:
Code LGTM.

Etsija:
Chiming in to the discussion a bit late, as I have been busy reviewing other UI PRs: 

Let's take the use case @mnonnenmacher provided, an organization with 150 secrets defined. When creating a new repository secret, a repository admin wants to know whether a particular secret has already been defined (ie. overridden) on organization level. There is _no indicator of this_ in the repository secrets table atm, so as it is, the admin "has to know" that his new secret might be overridden on org. or prod. level, so he goes on looking for it from those tables.

All tables are server-side paginated, and there's no sorting applied to any of the columns. So the admin has to flip through the whole organization secrets table, page by page, to find out whether there indeed is a secret with the same name, which will override his new secret.

I think this will lead to pretty bad UX, and there's numerous ways it could be improved:
1. If we want to keep to server-side data manipulation for the tables (not necessary), at least add sorting to the secret's name column, and make it also clearly visible in the headers, like it is for any sortable tables we have.
2. 150, 1500, or even 15000 items in a table is no fuss for TanStack Table, so we could instead very well switch to client-side data manipulation for all secrets tables and implement sorting _and filtering_.
3. Can we please, please add at least an indicator that your new secret _actually already exists_ on a higher entity level so the admin knows he has to go on to searching for those? Ideally, with client-side data manipulation of the tables, this indicator could actually be a straight link to a view of this secrets page, where the corresponding table is already filtered with the corresponding secret seen as the only item in the table.

I'm not in favor of blocking this PR due to my comments and suggestions, but I'd definitely do a follow-up PR (or, several)  to this, to improve the UX.

mnonnenmacher:
> Let's take the use case @mnonnenmacher provided, an organization with 150 secrets defined. When creating a new repository secret, a repository admin wants to know whether a particular secret has already been defined (ie. overridden) on organization level. There is _no indicator of this_ in the repository secrets table atm, so as it is, the admin just "has to know" that his new secret might be overridden on org. or prod. level, and look for it from those tables.

It works the other way round, repository secrets override secrets from their parent product and organization, otherwise I would agree. However, I'd still like to extend the endpoints to include the information if a secret overwrites a parent secret as a follow-up so that it can be shown in the tables.

PR #2441, opened 2025-04-04
Changes to API documentation

Etsija:
The API documentation (Swagger, webpage) has become unnecessarily fragmented, introducing somewhat arbitrary top-level groups that make it difficult to find a particular endpoint.  Also the new users endpoints were incorrectly documented and left untagged. 

This PR implements some major changes to the API documents, by reducing the number of top-level groups, and grouping the corresponding endpoints to under the groups where they semantically belong. Endpoint summaries and descriptions are also reworded a bit, to keep the summaries as brief as possible and aligned with each other.

Screenshot from 2025-04-04 08-38-42

Open question: @mnonnenmacher I also tried some methods to order the endpoints manually inside groups to a better semantic order, but without success. Swagger seems to order the endpoints in a group somewhat arbitrarily. Would you have suggestions for that?

### DIFF
--- a/core/src/main/kotlin/apiDocs/AdminDocs.kt
+++ b/core/src/main/kotlin/apiDocs/AdminDocs.kt
@@ -28,7 +28,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.User
 
 val runPermissionsSync: OpenApiRoute.() -> Unit = {
     operationId = "runPermissionsSync"
-    summary = "Trigger the synchronization of Keycloak roles."
+    summary = "Trigger the synchronization of Keycloak roles"
     tags = listOf("Admin")
 
     request {
@@ -47,7 +47,7 @@ val runPermissionsSync: OpenApiRoute.() -> Unit = {
 
 val getUsers: OpenApiRoute.() -> Unit = {
     operationId = "getUsers"
-    summary = "Get all users of the server."
+    summary = "Get all users of the server"
     tags = listOf("Admin")
 
     request {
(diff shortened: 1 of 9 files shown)


Etsija:
OK so what I didn't realize that by just changing a tag in `apiDocs` from "Infrastructure services" to "Organizations", for example, this changes all infra related OpenAPI query names, which will lead to huge changes needed in the UI. I'll probably skip all grouping changes from the PR and leave it until later.

mnonnenmacher:
> Open question: @mnonnenmacher I also tried some methods to order the endpoints manually inside groups to a better semantic order, but without success. Swagger seems to order the endpoints in a group somewhat arbitrarily. Would you have suggestions for that?

@Etsija For Swagger I found two options to configure sorting, `operationsSorter` and `tagsSorter`, see: https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/#parameters

These can be configured in the `swaggerUI` call here: https://github.com/eclipse-apoapsis/ort-server/blob/f9e1f6b2b382d052f3856d73226652641bcaac13/core/src/main/kotlin/plugins/OpenApi.kt#L143-L151

For the Kotlin API for this configuration see:
https://smiley4.github.io/ktor-openapi-tools/latest/swaggerui/getting_started/#usage
https://smiley4.github.io/ktor-openapi-tools/latest/dokka/ktor-swagger-ui/ktor-swagger-ui/io.github.smiley4.ktorswaggerui.config/-swagger-u-i-config/index.html#666812852%2FProperties%2F709432301

We aren't using the latest version of the library (#2108 and #2166) because I didn't have time to look at the breaking changes yet, so the API for this config might have changed.

For the Docusaurus plugin we use to add the API docs to the website I haven't found a configuration option for the sorting, but maybe it has a better default sorting: https://github.com/PaloAltoNetworks/docusaurus-openapi-docs?tab=readme-ov-file#plugin-configuration-options

PR #2443, opened 2025-04-04
Reword and refactor some endpoints in the API documentation

Etsija:
The API documentation (Swagger, webpage) has become unnecessarily fragmented, introducing somewhat arbitrary top-level groups that make it difficult to find a particular endpoint. Also the new users endpoints were incorrectly documented and left untagged.

This PR implements some major changes to the API documents, by reducing the number of top-level groups, and grouping the corresponding endpoints to under the groups where they semantically belong. Endpoint summaries and descriptions are also reworded a bit, to keep the summaries as brief as possible and aligned with each other.

Top-level grouping:

Screenshot from 2025-04-04 08-38-42

As an example, the "Runs" group:

Screenshot from 2025-04-04 10-49-56

With these changes, the OpenAPI React query client generation library that the UI uses is also producing simpler query hooks to use and remember, due to reduced amount of "top-level" groups to choose from, when choosing a query hook for an endpoint.

The PR is tested locally by building all images, running Docker Compose and testing the UI image for any errors in queries. Also a test ORT run was done, no API errors seen.

### DIFF
--- a/core/src/main/kotlin/apiDocs/DownloadsDocs.kt
+++ b/core/src/main/kotlin/apiDocs/DownloadsDocs.kt
@@ -27,7 +27,7 @@ val getReportByRunIdAndToken: OpenApiRoute.() -> Unit = {
     operationId = "GetReportByRunIdAndToken"
     summary = "Download a report of an ORT run using a token"
     description = "This endpoint does not require authentication."
-    tags = listOf("Reports")
+    tags = listOf("Runs")
 
     request {
         pathParameter<Long>("runId") {
--- a/core/src/main/kotlin/apiDocs/DownloadsDocs.kt
+++ b/core/src/main/kotlin/apiDocs/DownloadsDocs.kt
@@ -25,7 +25,8 @@ import io.ktor.http.HttpStatusCode
 
 val getReportByRunIdAndToken: OpenApiRoute.() -> Unit = {
     operationId = "GetReportByRunIdAndToken"
-    summary = "Download a report of an ORT run using a token. This endpoint does not require authentication."
+    summary = "Download a report of an ORT run using a token"
+    description = "This endpoint does not require authentication."
     tags = listOf("Reports")
 
     request {
--- a/core/src/main/kotlin/apiDocs/ProductsDocs.kt
+++ b/core/src/main/kotlin/apiDocs/ProductsDocs.kt
@@ -532,10 +532,12 @@ val getOrtRunStatisticsByProductId: OpenApiRoute.() -> Unit = {
 
 val getUsersForProduct: OpenApiRoute.() -> Unit = {
     operationId = "GetUsersForProduct"
-    summary = "Get all users that have rights for a product, including user privileges (groups) that user have " +
-        "within product."
-    description = "Fields available for sorting: 'username', 'firstName', 'lastName', 'email', 'group'. " +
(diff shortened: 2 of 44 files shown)


core/src/main/kotlin/apiDocs/DownloadsDocs.kt
```diff
@@ -25,7 +25,8 @@ import io.ktor.http.HttpStatusCode
 
 val getReportByRunIdAndToken: OpenApiRoute.() -> Unit = {
     operationId = "GetReportByRunIdAndToken"
-    summary = "Download a report of an ORT run using a token. This endpoint does not require authentication."
+    summary = "Download a report of an ORT run using a token"
+    description = "This endpoint does not require authentication."
```
mnonnenmacher, line 29: Maybe:
```
summary = "Download a report"
description = "Download a report of an ORT run using a token. This endpoint does not require authentication."

File: core/src/main/kotlin/apiDocs/ProductsDocs.kt [hunk 2]
```diff
@@ -430,7 +430,8 @@ val deleteUserFromProductGroup: OpenApiRoute.() -> Unit = {
 
 val getVulnerabilitiesAcrossRepositoriesByProductId: OpenApiRoute.() -> Unit = {
     operationId = "GetVulnerabilitiesAcrossRepositoriesByProductId"
-    summary = "Get the vulnerabilities from latest successful advisor runs across the repositories in a product."
+    summary = "Get vulnerabilities from a product"
+    description = "Get the vulnerabilities from latest successful advisor runs across the repositories in a product."
```
mnonnenmacher, line 434: Maybe: "Get the vulnerabilities from last successful runs for each repository in a product."

core/src/main/kotlin/apiDocs/ProductsDocs.kt
```diff
@@ -483,7 +484,7 @@ val getVulnerabilitiesAcrossRepositoriesByProductId: OpenApiRoute.() -> Unit = {
 
 val getOrtRunStatisticsByProductId: OpenApiRoute.() -> Unit = {
     operationId = "GetOrtRunStatisticsByProductId"
-    summary = "Get statistics about ORT runs across the repositories of a product."
+    summary = "Get statistics about ORT runs across the repositories of a product"
```
mnonnenmacher, line 487: Sometimes we say "ORT runs" and sometimes just "runs", I think we should align on the latter.

core/src/main/kotlin/apiDocs/RunsDocs.kt
```diff
@@ -158,7 +158,7 @@ val getReportByRunIdAndFileName: OpenApiRoute.() -> Unit = {
 
 val getLogsByRunId: OpenApiRoute.() -> Unit = {
     operationId = "GetLogsByRunId"
-    summary = "Download an archive with selected logs of an ORT run."
+    summary = "Download an archive with selected logs of an ORT run"
```
mnonnenmacher, line 161: "Download a ZIP archive with the logs of a run"

mnonnenmacher (approved):
@Etsija Thanks for doing the cleanup! I added some comments with ideas for further improvements but I'll leave it up to you if you want to apply them, otherwise I might make a follow-up sometime.

Issue #2439, opened 2025-04-03
Add Endpoint to Run Selected Repositories Under a Product


Ndacyayisenga-droid:
Currently, there's no way to trigger runs for specific repositories within a product - we can only run all repositories or retry single failed ones. We need an endpoint that allows running a custom selection of repositories under a product.

mnonnenmacher:
I would suggest to extend the endpoint from #2407 to take an optional argument with the repository IDs.

Ndacyayisenga-droid:
> I would suggest to extend the endpoint from [#2407](https://github.com/eclipse-apoapsis/ort-server/pull/2407) to take an optional argument with the repository IDs.

Nice suggestion @mnonnenmacher. I would as well look through the possiblity of extending the same for  https://github.com/eclipse-apoapsis/ort-server/issues/2438.  I will look through this asap

PR #2449, opened 2025-04-06
feat(api): Modify endpoint to run selected repositories under a product

Ndacyayisenga-droid:
Fixes https://github.com/eclipse-apoapsis/ort-server/issues/2439

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/OrtRun.kt
+++ b/api/v1/model/src/commonMain/kotlin/OrtRun.kt
@@ -175,6 +175,11 @@ data class CreateOrtRun(
      * read from the default location `.ort.env.yml`.
      */
     val environmentConfigPath: String? = null,
+
+    /**
+     * The IDs of the repositories for this run.
+     */
+    val repositoryIds: List<Long> = emptyList()
 )
 
 @Serializable
--- a/core/src/main/kotlin/api/ProductsRoute.kt
+++ b/core/src/main/kotlin/api/ProductsRoute.kt
@@ -397,12 +397,22 @@ fun Route.products() = route("products/{productId}") {
 
             val productId = call.requireIdParameter("productId")
             val createOrtRun = call.receive<CreateOrtRun>()
-
             val userDisplayName = call.principal<OrtPrincipal>()?.let { principal ->
                 UserDisplayName(principal.getUserId(), principal.getUsername(), principal.getFullName())
             }
 
-            val repositoryIds = productService.getRepositoryIdsForProduct(productId)
+            val repositoryIds = if (createOrtRun.repositoryIds.isEmpty()) {
+                productService.getRepositoryIdsForProduct(productId)
+            } else {
+                val productRepoIds = productService.getRepositoryIdsForProduct(productId)
+                require(createOrtRun.repositoryIds.all { it in productRepoIds }) {
+                    """
+                    The following repository IDs do not belong to product $productId: 
+                    ${createOrtRun.repositoryIds.filter { it !in productRepoIds }}
+                    """.trimIndent()
+                }
+                createOrtRun.repositoryIds
+            }
 
             val createdRuns = repositoryIds.mapNotNull { repositoryId ->
                 orchestratorService.createOrtRun(
--- a/core/src/main/kotlin/apiDocs/ProductsDocs.kt
+++ b/core/src/main/kotlin/apiDocs/ProductsDocs.kt
@@ -592,7 +592,8 @@ private val minimalJobConfigurations = JobConfigurations(
 
 val postOrtRunsForProduct: RouteConfig.() -> Unit = {
     operationId = "postOrtRunsForProduct"
-    summary = "Create ORT runs for all repositories under a product"
+    summary = "Create ORT runs for repositories under a product"
+    description = "Create ORT runs for all repositories under a product or specific repositories"
     tags = listOf("Products")
 
     request {
@@ -601,20 +602,18 @@ val postOrtRunsForProduct: RouteConfig.() -> Unit = {
         }
 
         jsonBody<CreateOrtRun> {
-            example("Create ORT runs using minimal job configurations (defaults)") {
+            example("Create ORT runs for all repositories using minimal job configurations") {
                 value = CreateOrtRun(
                     revision = "main",
                     jobConfigs = minimalJobConfigurations
                 )
             }
 
-            example("Create ORT runs using full job configurations") {
+            example("Create ORT runs for specific repositories using minimal job configurations") {
                 value = CreateOrtRun(
                     revision = "main",
-                    jobConfigs = fullJobConfigurations,
-                    labels = mapOf("label key" to "label value"),
-                    path = "optional VCS sub-path",
-                    jobConfigContext = "optional context",
+                    jobConfigs = minimalJobConfigurations,
+                    repositoryIds = listOf(1, 2)
                 )
             }
         }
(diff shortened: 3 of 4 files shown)


api/v1/model/src/commonMain/kotlin/OrtRun.kt
```diff
@@ -175,6 +175,13 @@ data class CreateOrtRun(
      * read from the default location `.ort.env.yml`.
      */
     val environmentConfigPath: String? = null,
+
+    /**
+     * The optional trace ID that is assigned to this run. This is generated when the run is created. It can be used to
+     * correlate the logs from different components that are taking part in processing of the run.
+     */
+    val repositoryIds: List<Long>? = null,
```
sschuberth, line 183: As a list can always be empty, making it nullable additionally should have a clear semantic purpose. I guess we could just make this non-nullable and empty by default, and by convention treat empty `repositoryIds` as *all* `repositoryIds`.

api/v1/model/src/commonMain/kotlin/OrtRun.kt
```diff
@@ -175,6 +175,13 @@ data class CreateOrtRun(
...
+
```
sschuberth, line 184: Nit: Please remove this superfluous empty line.

core/src/main/kotlin/api/ProductsRoute.kt
```diff
@@ -414,7 +423,9 @@ fun Route.products() = route("products/{productId}") {
                     createOrtRun.labels,
                     createOrtRun.environmentConfigPath,
                     userDisplayName
-                ).mapToApi(Jobs())
+                )
+                println("Run created for repo $repositoryId: $run")
```
sschuberth, line 427: See above.

core/src/main/kotlin/api/ProductsRoute.kt
```diff
@@ -398,14 +398,23 @@ fun Route.products() = route("products/{productId}") {
             val productId = call.requireIdParameter("productId")
             val createOrtRun = call.receive<CreateOrtRun>()
 
+            requireNotNull(createOrtRun.jobConfigs) { "jobConfigs must be provided" }
```
sschuberth, line 401: Why this check now? It seems you're not even accessing `jobConfigs` here.

core/src/main/kotlin/api/ProductsRoute.kt
```diff
@@ -398,14 +398,23 @@ fun Route.products() = route("products/{productId}") {
...
 
-            val repositoryIds = productService.getRepositoryIdsForProduct(productId)
+            val repositoryIds = createOrtRun.repositoryIds?.let { providedIds ->
+                val productRepoIds = productService.getRepositoryIdsForProduct(productId)
+                providedIds.filter { it in productRepoIds }.takeIf { it.isNotEmpty() }
+                    ?:
+                    throw IllegalArgumentException("None of the provided repository IDs belong to product $productId")
+            } ?: productService.getRepositoryIdsForProduct(productId)
+
+            println("Triggering runs for repositories: $repositoryIds with request: $createOrtRun")
```
sschuberth, line 414: Please don't use any `println` in library code (as it cannot be disabled by the application that uses the library). Use logging instead if needed.

sschuberth (changes requested):

sschuberth:
@Ndacyayisenga-droid as for the past PRs, please don't address review comments with new commits, but `git commit --amend` to your existing commit, and then force-push to the existing PR branch.

Now that you already created a new commit, it means to simply squash these into one.

sschuberth:
In your commit message

```
feat(api): Modify endpoint to run selected repositories under a product

Extended POST /products/{productId}/runs to accept an optional
repositoryIds parameter, allowing runs for specific repositories.
Updated logic, docs, and tests to support this feature.

feat(api): Modify endpoint to run selected repos under a product
```

you now have the title a second time at the bottom. Please remove that line (again via `git commit --amend`).

Ndacyayisenga-droid:
> In your commit message
> 
> ```
> feat(api): Modify endpoint to run selected repositories under a product
> 
> Extended POST /products/{productId}/runs to accept an optional
> repositoryIds parameter, allowing runs for specific repositories.
> Updated logic, docs, and tests to support this feature.
> 
> feat(api): Modify endpoint to run selected repos under a product
> ```
> 
> you now have the title a second time at the bottom. Please remove that line (again via `git commit --amend`).

done

sschuberth:
> @mnonnenmacher please also have a look.

Friendly reminder @mnonnenmacher 😉

api/v1/model/src/commonMain/kotlin/OrtRun.kt
```diff
@@ -175,6 +175,12 @@ data class CreateOrtRun(
      * read from the default location `.ort.env.yml`.
      */
     val environmentConfigPath: String? = null,
+
+    /**
+     * The optional trace ID that is assigned to this run. This is generated when the run is created. It can be used to
+     * correlate the logs from different components that are taking part in processing of the run.
+     */
+    val repositoryIds: List<Long> = emptyList()
```
mnonnenmacher, line 183: The description of the property is wrong.

core/src/main/kotlin/api/ProductsRoute.kt
```diff
@@ -397,15 +397,21 @@ fun Route.products() = route("products/{productId}") {
...
             }
 
-            val repositoryIds = productService.getRepositoryIdsForProduct(productId)
+            val repositoryIds = if (createOrtRun.repositoryIds.isEmpty()) {
+                productService.getRepositoryIdsForProduct(productId)
+            } else {
+                val productRepoIds = productService.getRepositoryIdsForProduct(productId)
+                createOrtRun.repositoryIds.filter { it in productRepoIds }.takeIf { it.isNotEmpty() }
+                    ?:
+                    throw IllegalArgumentException("None of the provided repository IDs belong to product $productId")
```
mnonnenmacher, line 410: I wonder if it would be better to always fail if at least one of the provided repository IDs does not belong to the product, otherwise users might not notice their mistake if it is silently ignored.

core/src/main/kotlin/api/ProductsRoute.kt
```diff
@@ -414,7 +420,8 @@ fun Route.products() = route("products/{productId}") {
                     createOrtRun.labels,
                     createOrtRun.environmentConfigPath,
                     userDisplayName
-                ).mapToApi(Jobs())
+                )
+                run.mapToApi(Jobs())
```
mnonnenmacher, line 424: Unnecessary refactoring, the previous code was fine.

core/src/main/kotlin/apiDocs/ProductsDocs.kt
```diff
@@ -668,5 +667,8 @@ val postOrtRunsForProduct: RouteConfig.() -> Unit = {
                 }
             }
         }
+        HttpStatusCode.BadRequest to {
```
mnonnenmacher, line 670: Nit: Add an empty line above.

mnonnenmacher (changes requested):
Sorry for the late review @Ndacyayisenga-droid, I was on a conference last week and it took me a bit to catch up.

core/src/main/kotlin/api/ProductsRoute.kt
Ndacyayisenga-droid: > I wonder if it would be better to always fail if at least one of the provided repository IDs does not belong to the product, otherwise users might not notice their mistake if it is silently ignored.

Screenshot 2025-04-18 at 05 05 10

mnonnenmacher (approved):

PR #2464, opened 2025-04-08
Add custom footer content

### LINKED ISSUE #2344: Add a customizable footer
Add a footer to the UI that can be customized by administrators. The purpose is to add custom information specific to the ORT Server instance, for example, links to internal documentation.

The custom content of the footer should be stored in the database and there should be endpoints to get and set the content.

wkl3nk:
Add a REST API endpoint to consume and update dynamic content for the footer section of the UI, add a footer section to the UI that uses this endpoint, and add a configuration panel in the admin section of the UI that allows to view and update the dynamic footer content.

See the respective commits for details.

Fixes #2344.

### DIFF
--- a/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
+++ b/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
@@ -26,6 +26,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.AdvisorJobConfiguration as Ap
 import org.eclipse.apoapsis.ortserver.api.v1.model.AnalyzerJob as ApiAnalyzerJob
 import org.eclipse.apoapsis.ortserver.api.v1.model.AnalyzerJobConfiguration as ApiAnalyzerJobConfiguration
 import org.eclipse.apoapsis.ortserver.api.v1.model.ComparisonOperator as ApiComparisonOperator
+import org.eclipse.apoapsis.ortserver.api.v1.model.ContentManagementSection as ApiContentManagementSection
 import org.eclipse.apoapsis.ortserver.api.v1.model.CredentialsType as ApiCredentialsType
 import org.eclipse.apoapsis.ortserver.api.v1.model.EcosystemStats as ApiEcosystemStats
 import org.eclipse.apoapsis.ortserver.api.v1.model.EnvironmentConfig as ApiEnvironmentConfig
@@ -95,6 +96,7 @@ import org.eclipse.apoapsis.ortserver.model.AdvisorJob
 import org.eclipse.apoapsis.ortserver.model.AdvisorJobConfiguration
 import org.eclipse.apoapsis.ortserver.model.AnalyzerJob
 import org.eclipse.apoapsis.ortserver.model.AnalyzerJobConfiguration
+import org.eclipse.apoapsis.ortserver.model.ContentManagementSection
 import org.eclipse.apoapsis.ortserver.model.CredentialsType
 import org.eclipse.apoapsis.ortserver.model.EcosystemStats
 import org.eclipse.apoapsis.ortserver.model.EnvironmentConfig
@@ -899,5 +901,12 @@ fun Project.mapToApi() = ApiProject(
 )
 
 fun UserDisplayName.mapToApi() = ApiUserDisplayName(username, fullName)
+
+fun ContentManagementSection.mapToApi() = ApiContentManagementSection(
+    id = id,
+    isEnabled = isEnabled,
+    markdown = markdown,
+    updatedAt = updatedAt
+)
 
 fun OidcConfig.mapToApi() = ApiOidcConfig(accessTokenUrl, clientId)
--- a/ui/src/components/footer.tsx
+++ b/ui/src/components/footer.tsx
@@ -1,0 +1,72 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import { useAdminServiceGetApiV1AdminContentManagementSectionsBySectionId } from '@/api/queries';
+import { MarkdownRenderer } from '@/components/markdown-renderer.tsx';
+
+function extractColumns(
+  markdown: string,
+  alignment: 'left' | 'right'
+): string[] {
+  const regex = new RegExp(
+    `:::\\s?align-${alignment}\\s+([\\s\\S]*?)\\s+:::`,
+    'g'
+  );
+  const matches = [];
+  let match;
+
+  while ((match = regex.exec(markdown)) !== null) {
+    if (match[1] !== undefined) {
+      matches.push(match[1].trim());
+    }
+  }
+
+  return matches;
+}
+
+export function Footer() {
+  const { data } =
+    useAdminServiceGetApiV1AdminContentManagementSectionsBySectionId({
+      sectionId: 'footer',
+    });
+
+  const markdown = data?.markdown || '';
+  const enabled = data?.isEnabled || false;
+
+  const columnsAlignedLeft = extractColumns(markdown, 'left');
+  const columnsAlignedRight = extractColumns(markdown, 'right');
+  return (
+    enabled && (
+      <footer className='bg-muted text-muted-foreground w-full border-t text-sm'>
+        <div className='mx-auto flex max-w-screen-xl flex-col items-center justify-between gap-4 p-4 sm:flex-row md:gap-8 md:p-8'>
+          <nav className='flex flex-col gap-4 sm:flex-row sm:items-start'>
+            {columnsAlignedLeft.map((content, index) => (
+              <MarkdownRenderer key={index} markdown={content} />
+            ))}
+          </nav>
+          <nav className='flex flex-col gap-4 sm:flex-row sm:items-start'>
+            {columnsAlignedRight.map((content, index) => (
+              <MarkdownRenderer key={index} markdown={content} />
+            ))}
+          </nav>
+        </div>
+      </footer>
+    )
+  );
+}
--- a/ui/src/routes/admin/content-management/footer/index.tsx
+++ b/ui/src/routes/admin/content-management/footer/index.tsx
@@ -1,0 +1,216 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import { zodResolver } from '@hookform/resolvers/zod';
+import { useQueryClient } from '@tanstack/react-query';
+import { createFileRoute } from '@tanstack/react-router';
+import { Loader2 } from 'lucide-react';
+import { useForm } from 'react-hook-form';
+import { z } from 'zod';
+
+import {
+  useAdminServiceGetApiV1AdminContentManagementSectionsBySectionId,
+  useAdminServiceGetApiV1AdminContentManagementSectionsBySectionIdKey,
+  useAdminServicePatchApiV1AdminContentManagementSectionsBySectionId,
+} from '@/api/queries';
+import { ApiError } from '@/api/requests';
+import { LoadingIndicator } from '@/components/loading-indicator.tsx';
+import { ToastError } from '@/components/toast-error.tsx';
+import { Button } from '@/components/ui/button.tsx';
+import {
+  Card,
+  CardContent,
+  CardDescription,
+  CardFooter,
+  CardHeader,
+  CardTitle,
+} from '@/components/ui/card.tsx';
+import {
+  Form,
+  FormControl,
+  FormField,
+  FormItem,
+  FormLabel,
+  FormMessage,
+} from '@/components/ui/form';
+import { Switch } from '@/components/ui/switch';
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from '@/components/ui/tooltip.tsx';
+import { toast } from '@/lib/toast.ts';
+
+const formSchema = z.object({
+  markdown: z.string().min(1, 'Markdown content is required'),
+  isEnabled: z.boolean(),
+});
+
+function FooterEditorComponent() {
+  const queryClient = useQueryClient();
+
+  const {
+    data: contentManagementSection,
+    isFetching,
+    error,
+    isError,
+  } = useAdminServiceGetApiV1AdminContentManagementSectionsBySectionId({
+    sectionId: 'footer',
+  });
+
+  const form = useForm<z.infer<typeof formSchema>>({
+    resolver: zodResolver(formSchema),
+    defaultValues: {
+      markdown: contentManagementSection?.markdown || '',
+      isEnabled: contentManagementSection?.isEnabled || false,
+    },
+  });
+
+  const { mutateAsync, isPending } =
+    useAdminServicePatchApiV1AdminContentManagementSectionsBySectionId({
+      onSuccess() {
+        queryClient.invalidateQueries({
+          queryKey: [
+            useAdminServiceGetApiV1AdminContentManagementSectionsBySectionIdKey,
+          ],
+        });
+        toast.info('Footer saved', {
+          description: `Footer saved successfully.`,
+        });
+      },
+      onError(error: ApiError) {
+        toast.error(error.message, {
+          description: <ToastError error={error} />,
+          duration: Infinity,
+          cancel: {
+            label: 'Dismiss',
+            onClick: () => {},
+          },
+        });
+      },
+    });
+
+  async function onSubmit(values: z.infer<typeof formSchema>) {
+    await mutateAsync({
+      requestBody: {
+        isEnabled: values.isEnabled,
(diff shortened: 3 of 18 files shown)


wkl3nk:
A database migration will insert a default footer, which is well documented and helps the admins to do customizations. This footer template is disabled by default and first needs to get enabled to get effective.

image

wkl3nk:
A small configuration panel to do the footer customizations in the UI was also added to the admin section.

image

core/src/main/kotlin/api/ContentManagementRoute.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.eclipse.apoapsis.ortserver.core.apiDocs.getSectionById
+import org.eclipse.apoapsis.ortserver.core.apiDocs.patchSectionById
+import org.eclipse.apoapsis.ortserver.core.authorization.requireAuthenticated
+import org.eclipse.apoapsis.ortserver.core.authorization.requireSuperuser
+import org.eclipse.apoapsis.ortserver.core.utils.requireParameter
+import org.eclipse.apoapsis.ortserver.services.ContentManagementService
+
+import org.koin.ktor.ext.inject
+
+fun Route.contentManagement() = route("content-management") {
```
Etsija, line 42: In #2443 I cleaned up the API documentation by getting rid of top-level groups which didn't make sense semantically and made the documentation quite unfocused. I'm questioning the need of yet another top-level apiDoc for this endpoint; why can't it be simply listed under Admin group? It's certainly meant to be used by server admins (superusers) only.

core/src/main/kotlin/plugins/OpenApi.kt
```diff
@@ -106,6 +106,7 @@ fun Application.configureOpenApi() {
             tag("Runs") { }
             tag("Admin") { }
             tag("Versions") { }
+            tag("Content Management") { }
```
Etsija, line 109: Related to my other comment: I don't think we need a top-level group "Content Management".

ui/src/routes/admin/content-management/footer/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+  CardHeader,
+  CardTitle,
+} from '@/components/ui/card.tsx';
+import { Label } from '@/components/ui/label';
+import { Switch } from '@/components/ui/switch';
+import { toast } from '@/lib/toast.ts';
+
+export const Route = createFileRoute('/admin/content-management/footer/')({
+  component: RouteComponent,
+});
```
Etsija, line 48: If you look at any index file in the UI code, we put the `Route` instantiation to the bottom of the index file, not at the top. Can you please align this.

ui/src/routes/admin/content-management/footer/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+    }
+  }, [data?.markdown, isFetching]);
+
+  const [enabled, setEnabled] = useState(false);
+
+  useEffect(() => {
+    if (!isFetching && data?.isEnabled !== undefined) {
+      setEnabled(data.isEnabled);
+    }
+  }, [data?.isEnabled, isFetching]);
```
Etsija, line 96: TanStack Query is a very powerful server state management system in itself, so I don't think you need any of the useState and useEffect hooks here. Basically, the query hook you use will return these variables for you:

```typescript
const { data, isPending, isError, error } =
    useContentManagementServiceGetApiV1ContentManagementSectionsBySectionId({
      sectionId: 'footer',
    });
```

Then you test for `isPending`, `isError` and `error` just before your return statement:

```typescript
  if (isPending) {
    return <LoadingIndicator />;
  }

  if (isError) {
    toast.error('Unable to load data', {
      description: <ToastError error={error} />,
      duration: Infinity,
      cancel: {
        label: 'Dismiss',
        onClick: () => {},
      },
    });
    return;
  }
```
after which you can be sure your `data` will be available to render in your component.

ui/src/routes/admin/content-management/footer/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+      <CardHeader>
+        <CardTitle>Footer</CardTitle>
+        <CardDescription>
+          Use Markdown text to customize the content in the footer section.
+        </CardDescription>
+      </CardHeader>
+      <CardContent className='space-y-4'>
+        {isLoading ? (
+          <LoadingIndicator />
+        ) : (
```
Etsija, line 119: With the above comment, this will not be needed, as the `isPending` handling before the main return statement already handles the loading state.

Etsija (changes requested):

ui/src/routes/admin/content-management/footer/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+            rows={25}
+            onChange={(e) => setMarkdown(e.target.value)}
+            disabled={isSaving || isLoading || isFetching}
+            className='border-input bg-background bg-muted/50 min-h-[200px] w-full rounded-md border p-2 font-mono text-sm'
+          />
+        )}
+        <div className='flex gap-2'>
+          <Switch
+            id='footer-enabled'
+            checked={enabled}
```
Etsija, line 131: I believe you can just use `data.isEnabled` here.

ui/src/routes/admin/content-management/footer/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+      </CardHeader>
+      <CardContent className='space-y-4'>
+        {isLoading ? (
+          <LoadingIndicator />
+        ) : (
+          <textarea
+            value={markdown}
+            rows={25}
+            onChange={(e) => setMarkdown(e.target.value)}
+            disabled={isSaving || isLoading || isFetching}
```
Etsija, line 124: I don't understand the meaning of this.

ui/src/routes/admin/content-management/footer/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+} from '@/components/ui/card.tsx';
+import { Label } from '@/components/ui/label';
+import { Switch } from '@/components/ui/switch';
+import { toast } from '@/lib/toast.ts';
+
+export const Route = createFileRoute('/admin/content-management/footer/')({
+  component: RouteComponent,
+});
+
+function RouteComponent() {
```
Etsija, line 50: Please rename the component into something less general, for example `FooterManagementComponent`. You can look at examples throughout the UI code.

ui/src/routes/admin/content-management/footer/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+              <Loader2 size={16} className='mx-3 animate-spin' />
+            </>
+          ) : (
+            'Save'
+          )}
+        </Button>
+      </CardFooter>
+    </Card>
+  );
+}
```
Etsija, line 165: This whole component is essentially an edit form, so it should be changed as such; then, setting up the initial values (markdown, isEnabled) will be as easy as setting up the form defaults from the query.

core/src/test/kotlin/api/ContentManagementIntegrationTest.kt
```diff
@@ -0,0 +1,117 @@
...
+
+import org.eclipse.apoapsis.ortserver.api.v1.model.ContentManagementSection
+import org.eclipse.apoapsis.ortserver.api.v1.model.UpdateContentManagementSection
+import org.eclipse.apoapsis.ortserver.utils.test.Integration
+
+class ContentManagementIntegrationTest : AbstractIntegrationTest({
+    tags(Integration)
+
+    "GET /content-management/sections/footer" should {
+        "return the footer section" {
```
bs-ondem, line 40: Should it include maybe  an additional test to cover the `requireAuthentication` case?

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
```diff
@@ -897,3 +899,10 @@ fun Project.mapToApi() = ApiProject(
 )
 
 fun UserDisplayName.mapToApi() = ApiUserDisplayName(username, fullName)
+
+fun ContentManagementSection.mapToApi() = ApiContentManagementSection(
```
bs-ondem, line 903: This change does not belong to this commit. This can be moved to c729c40a42050f56539bcdbc6a68cd3c8dd59f02.

core/src/main/kotlin/api/ContentManagementRoute.kt
wkl3nk: I agree. As the functionality is located in the Admin menu in the UI, it also makes sense to show the REST endpoints in the "Admin" section of the OpenAPI Spec and documentation.

ui/src/routes/admin/content-management/footer/index.tsx
wkl3nk: Regarding the React state hooks, things are not so easy. I tried to directly use that "data" hook that is provided by TanStack query, but it does not work out, because these TextArea and the Switch want to modify the data, but it is not possible with the data directly from the Query, as this is **immutable**. I invested some time in analyzing and made the code a little bit more precise, but still want to stick to the React State hooks.

ui/src/routes/admin/content-management/footer/index.tsx
wkl3nk: Agree, this was a little bit too much. In fact, the most important thing is to avoid duplicate submissions of data.

ui/src/routes/admin/content-management/footer/index.tsx
wkl3nk: I would like to keep it as it is. Introducing all this form logic would make the code more complex.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
wkl3nk: I want to keep it here, because this commit has both the _model_ and _api/v1/model_ class. So why not also provide the mapper in this commit, that provides the mapping between these two classes.

ui/src/routes/admin/content-management/footer/index.tsx
Etsija: Related to this comment and my other comment for using a simple edit form instead of `useState` and `useEffect` hooks, can you describe me how this footer editor differs from any edit form we have throughout the UI?

See for example this form, which handles editing of the details of a repository and has a text input for the URL and a switch for the repository type (essentially, that's what you have in your view as well):

https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/routes/organizations/%24orgId/products/%24productId/repositories/%24repoId/_repo-layout/settings/index.tsx

- it sets up a simple form schema for validation in lines 61-64 (you can use very similar one)
- it specifies two queries for handling: one for retrieving data (72-75), one for mutating  (77-102)
- it sets the current URL and repository type (104-110) to the form as `defaultValues` that are based on data retrieval query
- it handles submitting the data from the form (you can omit all deleting part of the code, as your component doesn't include deleting functionality)
- it doesn't need any state hooks (`useState`, `useEffect`) at all, as TanStack Query handles the state for you via the parameters returned from the data retrieval and data mutation hooks

The whole meaning of TanStack Query is to handle the server state for us, and we are doing our best to use its capabilities to avoid using React state hooks if at all possible. I still think your case for the footer edit component doesn't differ at all from any of our other edit forms throughout the UI. So this is an edit form, and it should really be changed to be just that, instead of a component with elaborate manual state handling.

Screenshot from 2025-04-10 07-46-12

Etsija: Hmm, now that I think of it a bit more, is it the Reload functionality that needs the manual state handling?

Etsija: Hmm, now to think of it a bit more: is it the Reload functionality that you need manual state handling for?

core/src/test/kotlin/api/ContentManagementIntegrationTest.kt
wkl3nk: Yes. I have added a test now.

ui/src/routes/admin/content-management/footer/index.tsx
wkl3nk: "Reload" is not a requirement, I could drop it. I will have a look in the example you provided.

wkl3nk: Instead of "Reload", now using "Reset" to reset the form to the defaults. I was able to make it a Form and get rid of the React `useState` and `useEffect`.

ui/src/routes/admin/content-management/footer/index.tsx
wkl3nk: Changed the code into a `Form`

bs-ondem (dismissed):
My comments are addressed. It needs UI approval.

Etsija (approved):

ui/src/components/footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import { useAdminServiceGetApiV1AdminContentManagementSectionsBySectionId } from '@/api/queries';
+import { MarkdownRenderer } from '@/components/markdown-renderer.tsx';
+
+function extractCompanyTag(markdown: string): string | undefined {
+  const match = markdown.match(/:::\s?company\s+(.+?)\s+:::/s);
```
sschuberth, line 24: I'm a bit late to the party, but why is this *layout* marker named after what it's possible *contents* are? I mean, you could but any information in here, not necessarily related to a company. So why not name it `:::left-align(ed)` or so?

ui/src/routes/admin/content-management/footer/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+            />
+          </CardContent>
+          <CardFooter className='flex gap-2'>
+            <Button
+              type='button'
+              variant='outline'
+              onClick={() => form.reset()}
+            >
+              Reset
+            </Button>
```
Etsija, line 184: IIUC the functionality of the "Reset" button would be to reset the form back to what was stored in the database, right?  [The documentation](https://react-hook-form.com/docs/useform/reset) is a bit unclear on whether a plain `reset()` will reset to the default values, so have you tested that it doesn't instead reset the form to empty?

Anyway, as "Reset" can be interpreted in a couple ways by the user (reset to completely empty form / reset to what was stored in the database), UX-wise it would be best to add a tooltip to the button to explain its meaning.

Etsija (changes requested):

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
Etsija: Please see one more comment from @sschuberth and one change request from myself.

ui/src/routes/admin/content-management/footer/index.tsx
wkl3nk: I added a tooltip on the "Reset" button.

ui/src/components/footer.tsx
sschuberth: Similarly, I would probably rename `:::column` to `:::right-align`. (Or mention "align" first for both -> `:::align-left` and `:::align-right`.)

sschuberth: And just like for `:::column` (aka `:::aligh-right`), multiple `:::align-left` (aka `:::company`) should be allowed, if that's not already the case (I did not check the code).

ui/src/components/footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return match ? match[1] : undefined;
+}
+
+function extractColumns(markdown: string): string[] {
+  const regex = /:::\s?column\s+([\s\S]*?)\s+:::/g;
```
sschuberth, line 29: Why is this regex actually different from the "company" one? Don't both "company" and "column" markers follow exactly the same syntax?

ui/src/components/footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const markdown = data?.markdown || '';
+  const enabled = data?.isEnabled || false;
+
+  const companyTag = extractCompanyTag(markdown);
+  const columns = extractColumns(markdown);
+  return (
+    enabled && (
+      <footer className='bg-muted text-muted-foreground w-full border-t text-sm'>
+        <div className='mx-auto flex max-w-screen-xl flex-col items-center justify-between gap-4 p-4 sm:flex-row md:w-full md:gap-8 md:p-8'>
+          {/* Left: copyright or app info */}
```
sschuberth, line 57: Do comments here and below really need to have both `{` and `/*`?

lamppu: Yes, when having comments inside JSX, that's how it needs to be wrapped.

Etsija (approved):

sschuberth (changes requested):
I'd like to discuss the naming of markers a bit more before we all start to write footers with the current marker names.

mnonnenmacher:
> I'd like to discuss the naming of markers a bit more before we all start to write footers with the current marker names.

@sschuberth Can you say what you would like to change? For example, something more generic than "company"?

sschuberth:
> Can you say what you would like to change? For example, something more generic than "company"?

See my comments [above](https://github.com/eclipse-apoapsis/ort-server/pull/2464#discussion_r2037250749). To sum up, I propose to:
- rename `:::column` to `::align-right`
- rename `:::company` to `:::align-left`
- allow multiple `:::align-left` just like for `::align-right`
- probably also add `:::align-center` then (this just came up in my head)

wkl3nk:
@sschuberth: So be it. Multiple column-align-left layout markers are possible now.

````
Please use layout markers like this:

Content inside a "column-align-left" layout marker is aligned to the left of the footer.
Content inside a "column-align-right" layout marker is aligned to the right of the footer.

::: column-align-left
**ORT Server**
A scalable application to automate software compliance checks.
:::

::: column-align-right
[API](https://eclipse-apoapsis.github.io/ort-server/api/ort-server-api)
:::

::: column-align-right
[Issues](https://github.com/eclipse-apoapsis/ort-server/issues)
:::

::: column-align-right
[Visit on Github.com](https://github.com/eclipse-apoapsis/ort-server)
:::
    
````

**Comment by sschuberth** [comment]
> Multiple column-align-left layout markers are possible now.

Thanks. Can we also strip the "column-" prefixes as proposed?

**Comment by wkl3nk** [comment]
@sschuberth I hope the user still recognize then that they can actually build columns if they have many things to display on the footer.

**Comment by wkl3nk** [comment]
@sschuberth Okay, it now has this format:

````

Please use layout markers like this:

Content inside a "align-left" layout marker is aligned to the left of the footer.
Content inside a "align-right" layout marker is aligned to the right of the footer.

::: align-left
**ORT Server**
A scalable application to automate software compliance checks.
:::

::: align-right
[API](https://eclipse-apoapsis.github.io/ort-server/api/ort-server-api)
:::

::: align-right
[Issues](https://github.com/eclipse-apoapsis/ort-server/issues)
:::

::: align-right
[Visit on Github.com](https://github.com/eclipse-apoapsis/ort-server)
:::
    
````

**Comment by sschuberth** [comment]
> @sschuberth I hope the user still recognize then that they can actually build columns if they have many things to display on the footer.

By "build columns" you mean the feature to use multiple e.g. `:::align-right` markers? If so, that's clear from the example, IMO.

Or do you mean something else, like multiple lines within a *single* marker, like

```
:::align-right
line 1
line 2
line 3
:::
```

wkl3nk:
@sschuberth Yes, the second one. Forget it. They will find out.

````
:::align-right
row 1

row 2

row 3
:::
````

sschuberth:
> Forget it. They will find out.

Well, you do have a valid point there. But I think this can be easily resolved by extending the example.

wkl3nk:
@sschuberth I personally find it nice to have only a 1-row footer as default example, that can be used more or less out of the box. So, I would like to keep the example as it is now. Can we finish this PR now?

PR #2468, opened 2025-04-08
test(core): Add test for organizations endpoint

mnonnenmacher:
Add a test to verify that the organizations endpoint returns an empty result for a user without read permission for any organization.


sschuberth:
@Etsija please verify that this test matches your use-case.

Etsija:
> @Etsija please verify that this test matches your use-case.

In a sense it obviously does, but what I currently do not understand why it doesn't work on the UI side; the endpoint always returns a 401 Unauthorized instead of 200 Success in the use case of no organization permissions given to the user. I suspect this has something to do with Keycloak, rather than the endpoint itself.

Issue #2462, opened 2025-04-08
Always show all properties of projects / packages


sschuberth:
Currently, non-present properties of projects / packages are not shown at all in the UI. While this creates a slick experience, it hides from the user what to theoretically expect in terms of metadata. Also, it makes it hard for developers to double-check that all properties ORT knows about are shown in the UI (and exposed by the server API).

Therefore, I propose to always show the titles for all theoretically supported properties, and display a grayed-out "n/a" if a value for a property is missing.

Issue #2340, opened 2025-03-24
More endpoints should return the purl in addition to the package's id


sschuberth:
With #2176 in mind, endpoints like ~~packages~~ rule violations, vulnerabilities and issues should be extended to not only return the `identifier`, but also the `purl`.

Note that it's important to return the `purl` as generated by ORT instead of just calling `identifier.toPurl()` to properly take any curations for purls into account.

Etsija:
I corrected Sebastian's description of the issue, since packages endpoint already returns `purl`.

Moreover, these endpoints:
- vulnerabilities in ORT run
- vulnerabilities for a product
- vulnerabilities for an organization
- rule violations in ORT run
- issues in ORT run
should return the `purl` _when feasible_. 

This means that for example, for issues and rule violations that are triggered by a project, not a package, the purl should not be returned. Then the UI could use this logic as follows:
- if the user prefers to show the `ORT ID`, always show the package identifier in the UI
- if the user prefers to show the `purl`, then show it _when it is available_, ie. returned by the corresponding endpoint.

Etsija:
From the UI perspective, this would mean for example that these rule violations would always show the `ORT ID` in the UI, as they are triggered by the project:

![Image](https://github.com/user-attachments/assets/57cd0c82-8055-4115-8ef7-b2b996d5909c)

while these would either show `ORT ID` or `purl` in the respective column, depending on the user preference:

![Image](https://github.com/user-attachments/assets/e5902daa-6525-4fe3-b119-5295693a1a6c)

sschuberth:
> Moreover, these endpoints:
> 
>     * vulnerabilities in ORT run
> 
>     * vulnerabilities for a product
> 
>     * vulnerabilities for an organization
> 
>     * rule violations in ORT run
> 
>     * issues in ORT run
> 
> 
> should return the `purl` _when feasible_.

@Ndacyayisenga-droid if you're looking for work, these might also be good things to implement.

Etsija:
I totally agree, implementing these would conclude one biggish UI feature (to show either purls or ORT IDs throughout the UI, which is an important feature to Double Open team).

lamppu:
>     * vulnerabilities for a product
> 
>     * vulnerabilities for an organization

I wonder how to handle the purl for these, as I now realized that if the same vulnerability is found in runs for several products/organizations, couldn't it be possible that the curations for the package could differ, if different runs are configured differently?

sschuberth:
[...]

PR #2483, opened 2025-04-10
Extend packages with curations info

kamil-bielecki-bosch:
As UI part is not specified yet, please let me know if any additional fields are necessary to add into curation object.

### DIFF
--- a/services/hierarchy/src/main/kotlin/PackageService.kt
+++ b/services/hierarchy/src/main/kotlin/PackageService.kt
@@ -29,27 +29,27 @@ import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.PackagesTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.ProcessedDeclaredLicensesTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.ShortestDependencyPathDao
 import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.ShortestDependencyPathsTable
+import org.eclipse.apoapsis.ortserver.dao.repositories.repositoryconfiguration.PackageCurationDataDao
+import org.eclipse.apoapsis.ortserver.dao.repositories.repositoryconfiguration.PackageCurationDataTable
+import org.eclipse.apoapsis.ortserver.dao.repositories.repositoryconfiguration.PackageCurationsTable
+import org.eclipse.apoapsis.ortserver.dao.repositories.repositoryconfiguration.RepositoryConfigurationsPackageCurationsTable
+import org.eclipse.apoapsis.ortserver.dao.repositories.repositoryconfiguration.RepositoryConfigurationsTable
 import org.eclipse.apoapsis.ortserver.dao.tables.shared.IdentifiersTable
-import org.eclipse.apoapsis.ortserver.dao.utils.applyFilterNullable
-import org.eclipse.apoapsis.ortserver.dao.utils.applyILike
-import org.eclipse.apoapsis.ortserver.dao.utils.listCustomQueryCustomOrders
 import org.eclipse.apoapsis.ortserver.model.EcosystemStats
+import org.eclipse.apoapsis.ortserver.model.runs.Identifier
+import org.eclipse.apoapsis.ortserver.model.runs.Package
 import org.eclipse.apoapsis.ortserver.model.runs.PackageFilters
-import org.eclipse.apoapsis.ortserver.model.runs.PackageWithShortestDependencyPaths
+import org.eclipse.apoapsis.ortserver.model.runs.PackageRunData
+import org.eclipse.apoapsis.ortserver.model.runs.repository.PackageCurationData
 import org.eclipse.apoapsis.ortserver.model.util.ListQueryParameters
 import org.eclipse.apoapsis.ortserver.model.util.ListQueryResult
+import org.eclipse.apoapsis.ortserver.model.util.OrderDirection
+import org.eclipse.apoapsis.ortserver.model.util.OrderField
 
-import org.jetbrains.exposed.sql.Case
 import org.jetbrains.exposed.sql.Count
 import org.jetbrains.exposed.sql.Database
-import org.jetbrains.exposed.sql.Expression
-import org.jetbrains.exposed.sql.Op
-import org.jetbrains.exposed.sql.ResultRow
-import org.jetbrains.exposed.sql.SortOrder
-import org.jetbrains.exposed.sql.SqlExpressionBuilder.concat
-import org.jetbrains.exposed.sql.SqlExpressionBuilder.neq
-import org.jetbrains.exposed.sql.and
-import org.jetbrains.exposed.sql.stringLiteral
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
+import org.jetbrains.exposed.sql.andWhere
 
 /**
  * A service to interact with packages.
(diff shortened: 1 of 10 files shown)


kamil-bielecki-bosch:
As UI part is not specified yet, please let me know if any additional fields are necessary to add into curation object.

model/src/commonMain/kotlin/runs/PackageWithShortestDependencyPathsAndCurations.kt
```diff
@@ -19,17 +19,22 @@
...
 
     /** Package ID */
     val pkgId: Long,
 
     /** The shortest dependency path for the package. */
-    val shortestDependencyPaths: List<ShortestDependencyPath>
+    val shortestDependencyPaths: List<ShortestDependencyPath>,
+
+    /** The curations for the package. */
+    var curations: Set<PackageCurationData> = mutableSetOf()
```
oheger-bosch, line 39: Data classes should be immutable. Use a `val` here. Replace `mutableSetOf` by `emptySet`.

services/hierarchy/src/main/kotlin/PackageService.kt
```diff
@@ -133,9 +140,35 @@ class PackageService(private val db: Database) {
...
+            .innerJoin(RepositoryConfigurationsPackageCurationsTable)
+            .innerJoin(PackageCurationsTable)
+            .innerJoin(IdentifiersTable)
+            .innerJoin(PackagesTable)
+            .select(listOf(PackagesTable.id, PackageCurationsTable.id))
+            .where(RepositoryConfigurationsTable.ortRunId eq ortRunId)
+            .distinct()
+            .map { curation ->
+                packages.find { it.pkgId == curation[PackagesTable.id].value }?.let { pkg ->
+                    val packageCurations = PackageCurationDataTable
```
oheger-bosch, line 162: IIUC, for each element in the result set, you do another SELECT? So, this is the famous n + 1 SELECT problem?

I am not that familiar with the data structures in this area, but isn't it possible to add the curation data table to the join and do the required processing when iterating over the result set?

sschuberth:
> As UI part is not specified yet, please let me know if any additional fields are necessary to add into curation object.

As a general remark from my side (without having looked at your code): What I would expect happening with this PR merged is that the UI in its current state would "automatically" display the curated data. That's because IMO as a user you're rarely interested in uncurated data, so showing curated data should be the default.

*If* at some point we decide to also show uncurated data, or the chain of curations that lead to the curated data similar to like the [ORT Workbench](https://github.com/oss-review-toolkit/ort-workbench) does, *then* we need to specify the UI part for showing that additional data.

services/hierarchy/src/main/kotlin/PackageService.kt
kamil-bielecki-bosch: Indeed! It's a famous n+1 problem. 
All that is a matter of getting more-or-less full objects of curation data. Including stuff like authors (external table), artifacts etc. So, if there is no need to retrieve it - I can down it to single query. All that depends on what data scope UI really needs to display with package.

kamil-bielecki-bosch:
@sschuberth You're right. That's how this endpoint works. It's getting all the packages found during ORT run, and all curations applied to packages found. So UI can display them automatically. Question is: what data UI want to display? Because curation data is quite wide and I'm not sure if every user needs that wide scope

sschuberth:
> Question is: what data UI want to display? Because curation data is quite wide and I'm not sure if every user needs that wide scope

Your question seems to relate a bit to https://github.com/eclipse-apoapsis/ort-server/issues/2462. In general, I believe we should display *all* properties of a package / project (no matter whether such properties have curations or not).

model/src/commonMain/kotlin/runs/PackageWithShortestDependencyPathsAndCurations.kt
kamil-bielecki-bosch: Good point - changed.

model/src/commonMain/kotlin/runs/PackageWithShortestDependencyPathsAndCurations.kt
```diff
@@ -19,17 +19,22 @@
...
 package org.eclipse.apoapsis.ortserver.model.runs
 
+import org.eclipse.apoapsis.ortserver.model.runs.repository.PackageCurationData
+
 /**
  * A data class representing a package, and the shortest dependency path that the package is found in (relative to a
  * project found in a run).
  */
-data class PackageWithShortestDependencyPaths(
+data class PackageWithShortestDependencyPathsAndCurations(
```
oheger-bosch, line 28: I am afraid that this naming convention does not scale, especially if more attributes will be added. Semantically, this class adds some data to a package with is relevant in the context of an ORT run. So, should it then be named something like `PackageRunData` or `PackageInRun` ?

services/hierarchy/src/main/kotlin/PackageService.kt
```diff
@@ -129,13 +149,24 @@ class PackageService(private val db: Database) {
...
-                shortestDependencyPaths = shortestPaths
+                shortestDependencyPaths = shortestPaths,
+                curations = findCurationsForPackage(pkg.pkgId, curations)
             )
         }
 
         ListQueryResult(data, parameters, listQueryResult.totalCount)
     }
 
+    private fun findCurationsForPackage(packageId: Long, curations: List<ResultRow>): List<PackageCurationData> =
```
oheger-bosch, line 160: I am still concerned about the additional selects produced by this function.

oheger-bosch (commented):
What is still missing is the part that applies the curations to the packages.

model/src/commonMain/kotlin/runs/PackageWithShortestDependencyPathsAndCurations.kt
kamil-bielecki-bosch: Class name changed to PackageRunData.

services/hierarchy/src/main/kotlin/PackageService.kt
kamil-bielecki-bosch: Two birds with one stone. Got rid of two n+1 problems there (for shortest dependency path and curations).
Now all of data that extends packages are retrieved in two queries (one for each curations and dependency paths), then transformed to maps and then feed into packages list.

services/hierarchy/src/main/kotlin/PackageService.kt
```diff
@@ -110,30 +119,54 @@ class PackageService(private val db: Database) {
...
 
-        val data = listQueryResult.data.map { pkg ->
-            val shortestPaths = ShortestDependencyPathsTable
-                .innerJoin(PackagesTable)
-                .innerJoin(AnalyzerRunsTable)
-                .innerJoin(AnalyzerJobsTable)
-                .select(ShortestDependencyPathsTable.columns)
-                .where { (AnalyzerJobsTable.ortRunId eq ortRunId) and (PackagesTable.id eq pkg.pkgId) }
-                .map { ShortestDependencyPathDao.wrapRow(it).mapToModel() }
+        val curations = mutableMapOf<Long, List<PackageCurationData>>()
```
oheger-bosch, line 131: It should be possible to use Kotlin's `[groupBy](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/group-by.html)` function to construct the maps for curations and shortest paths.

kamil-bielecki-bosch: Changed, although not sure if it's easier to read.

kamil-bielecki-bosch:
> In general, I have problems to understand how the changes are distributed between the commits, especially the changes on the model. So, in the first commit, the `PackageRunData` is renamed and extended, but the mapping of the new attribute and the corresponding API class are introduced in the second commit.
> Maybe it is more logical to do the model changes in a separate first commit.

First commit: Change  `PackageWithShortestDependencyPaths` -> `PackageRunData`, due to new field used to return curations with package. As name changed, changes to `Mappings` class applied. But without changes to output API model `Package`, due to API endpoints contract.
Second commit: Add new, already available field to API `Package` model, contracts extension.
IMHO moving just name change to separate commit is a bit of overkill.

lamppu:
> What is still missing - and this is what @sschuberth is referring to in [his comment](https://github.com/eclipse-apoapsis/ort-server/pull/2483#issuecomment-2796082590) -, is to modify the package data according to the curations. A curation basically defines some modifications on the attributes of a package, e.g. to change the home page URL or override the license. The packages service is now supposed to perform these modifications, so that the UI can directly display the package properties and be sure that this is curated data.

Something I'd like to point out here as well is that for instance the purl can be curated, and the endpoint supports filtering and sorting by purl, so if the purl in the API response will be the curated one in case there is one, this technically would need some modifications in the sorting and filtering too.

sschuberth:
> Something I'd like to point out here as well is that for instance the purl can be curated

Good point, somewhat relates to https://github.com/eclipse-apoapsis/ort-server/issues/2340, which probably should be done first.

sschuberth:
> > Something I'd like to point out here as well is that for instance the purl can be curated
> 
> Good point, somewhat relates to #2340, which probably should be done first.

Do I remember correctly from one of the alignment meetings, that we agreed to merge this first, even if curated purls would not work yet?

It's quite crucial for us to finally see the real data in the UI, including curations.

mnonnenmacher:
> > > Something I'd like to point out here as well is that for instance the purl can be curated
> > 
> > 
> > Good point, somewhat relates to #2340, which probably should be done first.
> 
> Do I remember correctly from one of the alignment meetings, that we agreed to merge this first, even if curated purls would not work yet?
> 
> It's quite crucial for us to finally see the real data in the UI, including curations.

We found a problem with the previous implementation, because sorting, filtering, and pagination was still applied to the uncurated data which was then inconsistent with the shown data. Therefore this had to be re-implemented to do sorting, filtering, and pagination in memory after applying the curations. This could lead to bad performance for large projects and we probably also need to adapt the database schema because of that, but we decided to first do it in memory to not delay this change even more.

mnonnenmacher:
This has now been implemented in #2711.

Issue #193, opened 2024-05-07
Add an API endpoint to upload pre-created analyzer results


sschuberth:
Instead of running the analyzer on the ORT server, it is sometimes necessary to run the analyzer on-premises / locally (to avoid cloning the project in the cloud). For such a use-case the ORT server should have an endpoint to upload an existing analyzer result.

sschuberth:
@heliocastro, can you maybe share the API specs of your Optima service here for reference?

heliocastro:
I will be doing this on next Monday, need clear internal naming endpoints.

mmurto:
As discussed in #907, it could make sense to have the API work in a way that an existing analyzer result is uploaded to a specific product, and uploading it creates a new repository in that product. Users could then specify whatever run arguments for runs against that repo, and the runs would then be executed with the same analyzer result as the base without the need to upload the result file for each individual run. There should also be an update endpoint to upload a new analyzer result to overwrite the previously uploaded one, leading to the runs after that using the new analyzer result as the base.

sschuberth:
> I will be doing this on next Monday, need clear internal naming endpoints.

Any progress here BTW, @heliocastro?

mnonnenmacher:
> As discussed in #907, it could make sense to have the API work in a way that an existing analyzer result is uploaded to a specific product, and uploading it creates a new repository in that product. Users could then specify whatever run arguments for runs against that repo, and the runs would then be executed with the same analyzer result as the base without the need to upload the result file for each individual run. There should also be an update endpoint to upload a new analyzer result to overwrite the previously uploaded one, leading to the runs after that using the new analyzer result as the base.

This brings up some questions, how to handle the `repository` information in the uploaded ORT result:

* If the ORT result contains a repository provenance, this should match the URL of the repository in the DB.
* What if the ORT result has no provenance information? Currently the repository URL is a mandatory field, if there is no repository URL, we would need another column to identify the repository and to show something else than the ID in the UI.
* If there is no provenance information, do we have to ensure somehow that uploaded analyzer results for the same repository match?
* Should we even allow ORT results without provenance information? If we do, automatic repository creation would be problematic.

mmurto:
> * What if the ORT result has no provenance information? Currently the repository URL is a mandatory field, if there is no repository URL, we would need another column to identify the repository and to show something else than the ID in the UI.

If using provided analyzer results requires a lot of work arounds, we could consider creating separate database entries for them, similar to repos.

[...]

PR #2484, opened 2025-04-10
CLI: Add progress bar when using --wait flag

MarcelBochtler:
Instead of printing the run status every time the status is being polled, show a progress bar when using the `--wait` flag.
If the `--json` flag is used together with the `--wait` flag, the run status is only printed after the run finished.

image

### DIFF
--- a/cli/src/commonMain/kotlin/StartCommand.kt
+++ b/cli/src/commonMain/kotlin/StartCommand.kt
@@ -46,7 +46,7 @@ import org.eclipse.apoapsis.ortserver.cli.utils.echoMessage
 import org.eclipse.apoapsis.ortserver.cli.utils.read
 import org.eclipse.apoapsis.ortserver.client.NotFoundException
 
-internal val POLL_INTERVAL = getEnv("POLL_INTERVAL")?.toLongOrNull()?.seconds ?: 60.seconds
+internal val POLL_INTERVAL = getEnv("POLL_INTERVAL")?.toLongOrNull()?.seconds ?: 10.seconds
 
 class StartCommand : SuspendingCliktCommand(name = "start") {
     private val repositoryId by option(
--- a/cli/src/commonMain/kotlin/StartCommand.kt
+++ b/cli/src/commonMain/kotlin/StartCommand.kt
@@ -44,6 +44,7 @@ import org.eclipse.apoapsis.ortserver.cli.model.printables.toPrintable
 import org.eclipse.apoapsis.ortserver.cli.utils.createOrtServerClient
 import org.eclipse.apoapsis.ortserver.cli.utils.echoMessage
 import org.eclipse.apoapsis.ortserver.cli.utils.read
+import org.eclipse.apoapsis.ortserver.cli.utils.useJsonFormat
 import org.eclipse.apoapsis.ortserver.client.NotFoundException
 
 internal val POLL_INTERVAL = getEnv("POLL_INTERVAL")?.toLongOrNull()?.seconds ?: 10.seconds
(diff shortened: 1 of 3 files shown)


nnobelis:
Should we make this feature toggable ?
In case somebody want to parse the command output for whatever reason ?

MarcelBochtler:
> Should we make this feature toggable ? In case somebody want to parse the command output for whatever reason ?

When parsing the output, the user should use the `--json` flag.
If they insist on not using JSON, the current implementation also does not show the progress bar if the terminal does not support it, which would be the case if piping the output e.g. to grep.

cli/src/commonMain/kotlin/StartCommand.kt
```diff
@@ -100,15 +113,98 @@ class StartCommand : SuspendingCliktCommand(name = "start") {
...
+    private enum class JobType(val displayName: String) {
+        ANALYZER("Analyzing"),
+        ADVISOR("Advising"),
+        SCANNER("Scanning"),
+        EVALUATOR("Evaluating"),
+        REPORTER("Reporting"),
+        NOTIFIER("Notifying");
+
+        fun isConfigured(config: JobConfigurations) = when (this) {
+            ANALYZER -> true
```
nnobelis, line 175: `isConfigured`  returns the same thing as `isStarted`. Can't only one function be used ?

cli/src/commonMain/kotlin/StartCommand.kt
```diff
@@ -100,15 +113,98 @@ class StartCommand : SuspendingCliktCommand(name = "start") {
...
+            ADVISOR -> config.advisor != null
+            SCANNER -> config.scanner != null
+            EVALUATOR -> config.evaluator != null
+            REPORTER -> config.reporter != null
+            NOTIFIER -> config.notifier != null
+        }
+
+        fun isStarted(jobs: Jobs) = when (this) {
+            ANALYZER -> true
```
nnobelis, line 184: When a run has no Analyzer Job (for instance when the user provides an ort result), is this logic still going to work ?

cli/src/commonMain/kotlin/StartCommand.kt
MarcelBochtler: `isConfigured` returns if a job was configured by the user.
`isStarted` returns if the actual job was started by the ORT Server.

E.g. if the user enables the reporter, it will return `true` for `isConfigured` from the very beginning. But will only be `true` for `isStarted` when the worker actually started.

cli/src/commonMain/kotlin/StartCommand.kt
MarcelBochtler: At the current state, it is not possible to have an ORT Server run without an analyzer.

> for instance when the user provides an ort result

Are you sure that this feature exists in the ORT Server? Can you point me to it?

nnobelis: No it does not exist but I believe it was on our roadmap.

sschuberth: See https://github.com/eclipse-apoapsis/ort-server/issues/193.

cli/src/commonMain/kotlin/StartCommand.kt
nnobelis: I see, I misread the implementation of the functions.

cli/src/commonMain/kotlin/StartCommand.kt
MarcelBochtler: The assumption is correct for the `isConfigured` function, but not for the `isStarted` function.
FIxed.

PR #2489, opened 2025-04-11
fix: Use system temporary directory for keep-alive file

wkl3nk:
Instead to use the user's home directory to store the keep-alive, now use the system temporary directory, as this approach is more robust for all kinds of worker docker images.

### DIFF
--- a/transport/spi/src/main/kotlin/EndpointComponent.kt
+++ b/transport/spi/src/main/kotlin/EndpointComponent.kt
@@ -68,9 +68,12 @@ abstract class EndpointComponent<T : Any>(
         private val logger: Logger = LoggerFactory.getLogger(EndpointComponent::class.java)
 
         /**
-         * Get the keep-alive file in the user's home directory.
+         * Get the keep-alive file in the system's temporary directory.
          */
-        private fun getKeepAliveFile(): File = File(System.getProperty("user.home"), KEEP_ALIVE_FILE_NAME)
+        private fun getKeepAliveFile(): File {
+            val tmpDir = File(System.getProperty("java.io.tmpdir"))
+            return tmpDir.resolve(KEEP_ALIVE_FILE_NAME)
+        }
 
         /**
          * Sleep while a keep-alive file exists. This is helpful in case a user terminal session is opened in the


transport/spi/src/main/kotlin/EndpointComponent.kt
```diff
@@ -68,9 +68,22 @@ abstract class EndpointComponent<T : Any>(
...
 
         /**
-         * Get the keep-alive file in the user's home directory.
+         * Get the keep-alive file in the user's home directory. As a fallback, if the user's home directory ist not
+         * writable, use the system's temporary directory.
          */
-        private fun getKeepAliveFile(): File = File(System.getProperty("user.home"), KEEP_ALIVE_FILE_NAME)
+        private fun getKeepAliveFile(): File {
+            val keepAliveFile = File(System.getProperty("user.home"), KEEP_ALIVE_FILE_NAME)
+            return if (keepAliveFile.parentFile.canWrite()) {
```
sschuberth, line 76: How about extracting `val userHome = File(System.getProperty("user.home"))` so you don't have to "re-determine" it via `keepAliveFile.parentFile`?

sschuberth: In total, this could look like

```kotlin
private fun getKeepAliveFile(): File {
    val userHome = File(System.getProperty("user.home"))
    return if (userHome.canWrite()) {
        userHome.resolve(KEEP_ALIVE_FILE_NAME)
    } else {
        val tmpDir = System.getProperty("java.io.tmpdir")
        tmpDir.resolve(KEEP_ALIVE_FILE_NAME).also {
            logger.info(
                "Home directory is not writable. Using temporary directory for keep-alive file: " +
                    "${it.absolutePath}"
            )
        }
    }
}
```

oheger-bosch:
Just a thought: Should the lock file generally be created in the temp directory to make things simpler? Not sure whether this makes a big difference for the intended use case.

wkl3nk:
@oheger-bosch I like it when I create a shell in the pod and just have to do a `ls` command to see the file was generated.

wkl3nk:
Okay, persuaded, why have a fallback mechanism if we can store the `keep-alive.lock` file in the system's temporary directory.

PR #2491, opened 2025-04-11
Change generation of the .netrc file

oheger-bosch:
This PR changes the way the `.netrc` file is generated with the goal to keep it in sync with the infrastructure services that are used by the current worker. This should avoid conflicts with multiple services referring to the same host.

Refer to the single commits for further information.

### DIFF
--- a/workers/analyzer/src/test/kotlin/AnalyzerEndpointTest.kt
+++ b/workers/analyzer/src/test/kotlin/AnalyzerEndpointTest.kt
@@ -32,7 +32,6 @@ import io.kotest.matchers.nulls.beNull
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.shouldNot
 import io.kotest.matchers.string.shouldContain
-import io.kotest.matchers.string.shouldNotContain
 
 import io.mockk.coEvery
 import io.mockk.every
@@ -128,18 +127,6 @@ class AnalyzerEndpointTest : KoinTest, StringSpec() {
                 content shouldContain "https://center.conan.io"
                 content shouldContain "verify_ssl"
                 content shouldContain "true"
-            }
-        }
-
-        "The build environment should contain a .netrc file" {
-            runEnvironmentTest { homeFolder ->
-                val netrcFile = homeFolder.resolve(".netrc")
-                val content = netrcFile.readText()
-
-                content shouldContain "machine repo.example.org"
-                content shouldContain "login $USERNAME"
-
-                content shouldNotContain "repo2.example.org"
             }
         }
 
@@ -357,7 +344,7 @@ class AnalyzerEndpointTest : KoinTest, StringSpec() {
                     resolvedJobConfigContext = null,
                     traceId = "trace-id",
                 )
-                coEvery { setupAuthentication(any()) } just runs
+                coEvery { setupAuthentication(any(), any()) } just runs
             }
 
             val repositoryFolder = File("src/test/resources/mavenProject")
--- a/workers/common/src/main/kotlin/common/context/AuthenticationListener.kt
+++ b/workers/common/src/main/kotlin/common/context/AuthenticationListener.kt
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.workers.common.context
+
+/**
+ * A data class defining an event that is triggered when [OrtServerAuthenticator] successfully authenticates a request.
+ */
+data class AuthenticationEvent(
+    /** The name of the infrastructure service whose credentials are used for authentication. */
+    val serviceName: String
+)
+
+/**
+ * An interface for listeners that receive notifications about successful authentications performed by
+ * [OrtServerAuthenticator].
+ *
+ * An implementation of this interface can be added to [OrtServerAuthenticator]. It is then always invoked when a
+ * request is successfully authenticated using the credentials of a specific infrastructure service. The purpose
+ * behind this interface is to have a mechanism to dynamically update the environment based on the infrastructure
+ * services that are involved in the current operation. This is especially necessary when dealing with external tools
+ * that rely on the content of the _.netrc_ file for authentication. Since this file is limited and allows only one
+ * set of credentials per host, it cannot be generated beforehand based on all infrastructure services defined; this
+ * would lead to conflicts if there are multiple services for the same host. Instead, the idea is to keep track of the
+ * set of infrastructure services that are currently referenced and also update the content of the _.netrc_ file if
+ * this changes. This is, of course, not a perfect solution. It requires that the [OrtServerAuthenticator] is always
+ * called before an external tool is launched.
+ */
+interface AuthenticationListener {
+    /**
+     * Notify this listener about a successful authentication with the details provided by the given
+     * [authenticationEvent].
+     */
+    fun onAuthentication(authenticationEvent: AuthenticationEvent)
+}
--- a/workers/common/src/main/kotlin/common/context/OrtServerAuthenticator.kt
+++ b/workers/common/src/main/kotlin/common/context/OrtServerAuthenticator.kt
@@ -25,13 +25,24 @@ import java.net.URI
 import java.net.URL
 import java.util.concurrent.atomic.AtomicReference
 
+import org.ossreviewtoolkit.utils.ort.OrtAuthenticator
+import org.ossreviewtoolkit.utils.ort.UserInfoAuthenticator
+
 import org.slf4j.LoggerFactory
 
+private val logger = LoggerFactory.getLogger(OrtServerAuthenticator::class.java)
+
 /**
- * Implementation of an [Authenticator] which uses the current set of infrastructure services to authenticate requests.
+ * Implementation of an [Authenticator] which is responsible for handling authentication within ORT Server.
  *
- * This implementation extends ORT's authenticator implementation which is unable to handle multiple credentials for
- * the same host.
+ * This implementation uses special authentication logic based on infrastructure services and their associated
+ * credentials. It also uses functionality from ORT's authentication mechanism, but makes sure that the different
+ * sources of authentication information are queried in the correct order:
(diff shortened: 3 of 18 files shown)


workers/common/src/main/kotlin/common/context/OrtServerAuthenticator.kt
```diff
@@ -41,7 +41,7 @@ import org.slf4j.LoggerFactory
  */
 internal class OrtServerAuthenticator(
     /** The original authenticator that was active when this instance was installed. */
-    private val original: Authenticator? = null
+    private val original: Authenticator? = null,
```
sschuberth: Nit: Trailing comma.

workers/common/src/test/kotlin/common/context/OrtServerAuthenticatorTest.kt
```diff
@@ -122,6 +137,38 @@ class OrtServerAuthenticatorTest : WordSpec() {
                 pwd.password shouldBe PASSWORD.toCharArray()
             }
 
+            "invoke a registered authentication lister for a successful authentication" {
```
mnonnenmacher, line 148: Nit: Typo in "listener"

workers/common/src/main/kotlin/common/context/AuthenticationListener.kt
```diff
@@ -0,0 +1,51 @@
...
+ * An interface for listeners that receive notifications about successful authentications performed by
+ * [OrtServerAuthenticator].
+ *
+ * An implementation of this interface can be added to [OrtServerAuthenticator]. It is then always invoked when a
+ * request is successfully authenticated using the credentials of a specific infrastructure service. The purpose
+ * behind this interface is to have a mechanism to dynamically update the environment based on the infrastructure
+ * services that are involved in the current operation. This is especially necessary when dealing with external tools
+ * that rely on the content of the _.netrc_ file for authentication. Since this file is limited and allows only one
+ * set of credentials per host, it cannot be generated beforehand based on all infrastructure services defined; this
+ * would lead to conflicts if there are multiple services for the same host. Instead, the idea is to keep track on the
```
mnonnenmacher, line 40: Nit: "keep track **of**" (also in commit message)

workers/common/src/main/kotlin/common/context/AuthenticationListener.kt
```diff
@@ -0,0 +1,51 @@
...
+ * set of infrastructure services that are currently referenced and also update the content of the _.netrc_ file if
+ * this changes. This is, of course, not a perfect solution. It requires that the [OrtServerAuthenticator] is always
+ * called before an external tool is launched.
+ */
+interface AuthenticationListener {
+    /**
+     * Notify this listener about a successful authentication with the details provided by the given
+     * [authenticationEvent].
+     */
+    fun onAuthentication(authenticationEvent: AuthenticationEvent)
```
mnonnenmacher, line 50: This function could just get the `serviceName`, it doesn't seem that there is a need for any other event types.

workers/common/src/test/kotlin/common/context/OrtServerAuthenticatorTest.kt
```diff
@@ -122,6 +137,38 @@ class OrtServerAuthenticatorTest : WordSpec() {
...
+                    null,
+                    443,
+                    "tcp",
+                    "hello",
+                    "https",
+                    URI.create(url).toURL(),
+                    Authenticator.RequestorType.SERVER
+                )
+
+                verify {
```
mnonnenmacher, line 175: Maybe with `exactly = 1`?

workers/common/src/main/kotlin/common/env/NetRcManager.kt
```diff
@@ -0,0 +1,114 @@
...
+import org.eclipse.apoapsis.ortserver.utils.logging.runBlocking
+import org.eclipse.apoapsis.ortserver.workers.common.context.AuthenticationEvent
+import org.eclipse.apoapsis.ortserver.workers.common.context.AuthenticationListener
+import org.eclipse.apoapsis.ortserver.workers.common.context.WorkerContext
+import org.eclipse.apoapsis.ortserver.workers.common.env.definition.EnvironmentServiceDefinition
+
+import org.slf4j.LoggerFactory
+
+/**
+ * An internal helper class to dynamically update the content of the `.netrc` file when there is change in the
```
mnonnenmacher, line 38: Nit: "**a** change"
Commit message nit: "keep track **of**"

workers/common/src/main/kotlin/common/env/NetRcManager.kt
```diff
@@ -0,0 +1,114 @@
...
+    services: Collection<InfrastructureService>
+) : AuthenticationListener {
+    companion object {
+        private val logger = LoggerFactory.getLogger(NetRcManager::class.java)
+
+        /**
+         * Create a new instance of [NetRcManager] and initialize it with the given [context] and [services].
+         */
+        fun create(context: WorkerContext, services: Collection<InfrastructureService>): NetRcManager =
+            NetRcManager(context, services)
```
mnonnenmacher, line 61: Why do we need a factory method that's exactly the same as the constructor of the class?

workers/common/src/main/kotlin/common/env/NetRcManager.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Return a new instance of [ConfigFileBuilder] to be used for creating the `.netrc` file.
+     */
+    internal fun createConfigFileBuilder(): ConfigFileBuilder = ConfigFileBuilder(context)
+
+    /**
+     * Return a new instance of [NetRcGenerator] that is going to be called to generate the `.netrc` file.
+     */
+    internal fun createNetRcGenerator(): NetRcGenerator = NetRcGenerator()
```
mnonnenmacher, line 93: Unnecessary indirection, just call `NetRcGenerator()` directly.

workers/common/src/main/kotlin/common/env/NetRcManager.kt
```diff
@@ -0,0 +1,114 @@
...
+
+        netRcServices[authenticationEvent.serviceName]?.also { service ->
+            runBlocking { updateNetRcServices(service) }
+        }
+    }
+
+    /**
+     * Return a new instance of [ConfigFileBuilder] to be used for creating the `.netrc` file.
+     */
+    internal fun createConfigFileBuilder(): ConfigFileBuilder = ConfigFileBuilder(context)
```
mnonnenmacher, line 88: Unnecessary indirection, just call `ConfigFileBuilder(context)` directly.

workers/common/src/main/kotlin/common/env/NetRcManager.kt
```diff
@@ -0,0 +1,114 @@
...
+    internal fun createNetRcGenerator(): NetRcGenerator = NetRcGenerator()
+
+    /**
+     * Perform an update of the `.netrc` file after receiving an authentication event for the given [service].
+     */
+    private suspend fun updateNetRcServices(service: InfrastructureService) {
+        val serviceUri = URI.create(service.url)
+        val serviceHost = serviceUri.host
+
+        val builder = createConfigFileBuilder()
```
mnonnenmacher, line 102: Instead of the `WorkerContext` this class could just take the `builder` and `generator` as constructor arguments to be more explicit  about what it actually needs. The `WorkerContext` has more functionality (and dependencies) than needed here.
It would also simplify testing a bit as you would not have to mock the two trivial functions above.

workers/analyzer/src/test/kotlin/AnalyzerEndpointTest.kt
```diff
@@ -357,7 +344,7 @@ class AnalyzerEndpointTest : KoinTest, StringSpec() {
                     resolvedJobConfigContext = null,
                     traceId = "trace-id",
                 )
-                coEvery { setupAuthentication(any()) } just runs
+                coEvery { setupAuthentication(any(), any()) } just runs
```
mnonnenmacher, line 347: Just to clarify, does this change mean that if any package manager is using the .netrc file this would not work anymore?
Also, IIRC the scanner uses the .netrc file to download source artifacts from private repositories, does this still work?

mnonnenmacher (changes requested):

workers/common/src/main/kotlin/common/context/AuthenticationListener.kt
oheger-bosch: I would prefer keeping it this way. Having an event class is a common pattern for all kinds of listeners, and this makes the interface stable for later extensions.

workers/common/src/main/kotlin/common/env/NetRcManager.kt
oheger-bosch: The factory function simplifies testing the integration of `NetRcManager` with the environment service (see later commit). Otherwise, constructor mocks would have to be used, which are always hard to deal with.

workers/common/src/main/kotlin/common/env/NetRcManager.kt
oheger-bosch: This is used to inject mocks into test code.

workers/common/src/main/kotlin/common/env/NetRcManager.kt
oheger-bosch: A builder instance can only be used for a single file generation. Therefore, for every received event, a new builder instance has to be created, and this requires a context.

workers/analyzer/src/test/kotlin/AnalyzerEndpointTest.kt
oheger-bosch: Package managers relying on the .netrc file are indeed problematic. Are you aware of any that fall into this category? The ones we typically deal with expect their credentials in a dedicated configuration file.

The scanner uses the downloader to download source code artifacts and clone repositories. This involves the authenticator and will still work. There might be issues with other version control systems than Git which are exclusively invoked as command line tools. Working around this would require some more changes. For instance, I would like to change the ORT scanner to not create the downloader itself, but allow passing it as a parameter. Then ORT Server could use a wrapper that makes sure that credentials are properly set up before delegating to the actual download functionality.

workers/common/src/main/kotlin/common/context/AuthenticationListener.kt
mnonnenmacher: To me it's just additional code and I don't see value in keeping the interface stable when everything around it has to be changed anyway if there are changes in future.

workers/common/src/main/kotlin/common/env/NetRcManager.kt
mnonnenmacher: Ideally, the design of the feature would not require that much mocking in the first place.

workers/common/src/main/kotlin/common/env/NetRcManager.kt
mnonnenmacher: I'm unhappy about that `WorkerContext` god object, the builder actually only needs a secret provider and only uses the helper code from the context to resolve secrets in parallel, but with the context it get dependencies on other stuff like downloading config files.

workers/analyzer/src/test/kotlin/AnalyzerEndpointTest.kt
mnonnenmacher: > Package managers relying on the .netrc file are indeed problematic. Are you aware of any that fall into this category? The ones we typically deal with expect their credentials in a dedicated configuration file.

I did not check, I vaguely remember Pip using the .netrc file but I think with the python-inspector that probably works differently anyway.

mnonnenmacher (approved):

PR #2522, opened 2025-04-16
ui: Repository type labels

Etsija:
This is a follow-up to #2517.

Screenshot from 2025-04-16 08-07-36

Screenshot from 2025-04-16 08-08-20

Screenshot from 2025-04-16 08-08-58

### DIFF
--- a/ui/src/lib/types.ts
+++ b/ui/src/lib/types.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,9 +18,13 @@
  */
 
 /**
- * Some input types for creating runs are strings in the API, but they actually take a limited set
- * of values. Define the possible values here to use in the UI.
+ * Redefine or extend some data types coming from the OpenAPI Query Client for UI purposes.
+ * Also define types and constants for UI usage which are not included in the query client.
  */
+
+// Some input types for creating runs are merely strings in the API, but they actually take a
+// limited set of values. Define the possible values here to use in the UI.
+// Note: with the upcoming plugin support, some of these types may be deprecated or changed.
 
 export const advisors = [
   {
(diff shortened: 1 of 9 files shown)


ui/src/lib/types.ts
```diff
@@ -178,3 +180,16 @@ export const reportFormats = [
...
+
+const repositoryTypeLabels: Record<RepositoryType, string> = {
+  GIT: 'Git',
+  GIT_REPO: 'Git-Repo',
+  MERCURIAL: 'Mercurial',
+  SUBVERSION: 'Subversion',
+};
+
+export function getRepositoryTypeLabel(type: RepositoryType): string {
+  return repositoryTypeLabels[type] || 'Unknown';
```
sschuberth, line 194: As "Unknown Repository" could be interpreted as "[a present repository of unknown / unsupported type](https://github.com/oss-review-toolkit/ort/blob/865b49bfc58cca461c81b21d3b534794e534fc06/model/src/main/kotlin/VcsType.kt#L59-L62)", I'd prefer to omit ` || 'Unknown'` here (and thus probably get rid of the function altogether, in favor of callers using `repositoryTypeLabels` directly), and just show "Repository" instead of "Unknown Repository" in the UI.

However, as that would require a little bit of special handling to not show a leading space in case of an unknown / empty type repository, I guess we could also think about another description than "Unknown" in this case. Maybe any of "Unavailable", "Absent", "Unset"?

Etsija: I prefer the "Unset" solution (which you actually justified nicely).

Etsija:
@sschuberth @mnonnenmacher bumped again to the completions failure in the PR, which has nothing to do with the OSC.

ui/src/lib/types.ts
sschuberth: I guess I would have had a slight preference for "Unavailable" as it better matches "N/A" (where the "A" could stand for both "applicable" and "available"), but I do not care *that* strongly. Any more opinions?

sschuberth:
> @sschuberth @mnonnenmacher bumped again to the completions failure in the PR, which has nothing to do with the OSC. It's now failing constantly, even after reruns. Could one of you explain how to deal with these (it has happened to me before)?

@MarcelBochtler started debugging this as part of https://github.com/eclipse-apoapsis/ort-server/pull/2523, and [this](https://github.com/eclipse-apoapsis/ort-server/pull/2523#issuecomment-2808990187) should fix it.

ui/src/lib/types.ts
Etsija: I just thought "Unavailable Repository" looks funny.

sschuberth: Thinking about the wording more, "unavailable" in this context could also be misread as "temporarily unavailable due to network issues". "Undefined" would be yet another options, but as "Unset" is shorted and already coded, let's stick to that.

sschuberth (approved):

PR #2523, opened 2025-04-16
Core: Add endpoint to retrieve the required OpenID Connect configuration for the CLI

MarcelBochtler:


api/v1/mapping/src/commonMain/kotlin/Mappings.kt
```diff
@@ -49,6 +49,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.MailNotificationConfiguration
 import org.eclipse.apoapsis.ortserver.api.v1.model.MailServerConfiguration as ApiMailServerConfiguration
 import org.eclipse.apoapsis.ortserver.api.v1.model.NotifierJob as ApiNotifierJob
 import org.eclipse.apoapsis.ortserver.api.v1.model.NotifierJobConfiguration as ApiNotifierJobConfiguration
+import org.eclipse.apoapsis.ortserver.api.v1.model.OicdConfig as ApiOicdConfig
```
sschuberth, line 52: Commit message nit: Use passive voice instead of "I".

sschuberth:
Hint: To prevent the [bogus completion script error](https://github.com/eclipse-apoapsis/ort-server/actions/runs/14489484650/job/40642403237?pr=2523#step:6:27), `scripts/cli/generate_completion_scripts.sh` should call `./gradlew` with `-q`.

scripts/cli/generate_completion_scripts.sh
```diff
@@ -21,7 +21,7 @@
 GIT_ROOT=$(git rev-parse --show-toplevel)
 
 (cd $GIT_ROOT &&
-  ./gradlew :cli:jvmRun --args='--generate-completion=bash' -DmainClass=org.eclipse.apoapsis.ortserver.cli.OrtServerMainKt --quiet > integrations/completions/osc-completion.bash &&
-  ./gradlew :cli:jvmRun --args='--generate-completion=fish' -DmainClass=org.eclipse.apoapsis.ortserver.cli.OrtServerMainKt --quiet > integrations/completions/osc-completion.fish &&
-  ./gradlew :cli:jvmRun --args='--generate-completion=zsh' -DmainClass=org.eclipse.apoapsis.ortserver.cli.OrtServerMainKt --quiet > integrations/completions/osc-completion.zsh
+  ./gradlew --quiet :cli:jvmRun --args='--generate-completion=bash' -DmainClass=org.eclipse.apoapsis.ortserver.cli.OrtServerMainKt > integrations/completions/osc-completion.bash &&
```
sschuberth, line 24: Hmm, I overlook that `--quiet` was already present at the end. Actually, I believe it should have worked that way, too. Any idea why it didn't / have you verified that it works now as expected when specified at the beginning?

MarcelBochtler: > Any idea why it didn't

Unfortunately, I don't know.

> have you verified that it works now as expected when specified at the beginning?

Having the check now succeeding, while it was failing multiple times before, should verify that it works.

core/src/main/kotlin/api/AuthenticationRoute.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.eclipse.apoapsis.ortserver.api.v1.mapping.mapToApi
+import org.eclipse.apoapsis.ortserver.core.apiDocs.getCliOidcConfig
+import org.eclipse.apoapsis.ortserver.model.authentication.OidcConfig
+
+import org.koin.ktor.ext.inject
+
+/**
+ * Endpoints related to the authentication of users with the system.
+ */
+fun Route.authentication() = route("auth") {
```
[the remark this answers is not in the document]
MarcelBochtler: > The name authentication could raise expectations that here some authentication logic is done. 

Yes, that was my intention.
If, at some point, we decide to implement some authentication logic, it should also be implemented in this Route.
I don't want to have it to be renamed then again.

Retrieving the details / configurations of the Authentication Provider should IMO be part of the Authentication Route, and this Route should IMO be called `authentication`.

I would like to keep it as-is, but if you insist, I will rename it.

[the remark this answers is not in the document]
oheger-bosch: No, I don't insist.

oheger-bosch (approved):

PR #2536, opened 2025-04-17
chore: Remove unused imports

mnonnenmacher:
It is unclear why detekt did not report these, it also did not help to enable the alternative `NoUnusedImports` rule.


sschuberth:
> It is unclear why detekt did not report these, it also did not help to enable the alternative `NoUnusedImports` rule.

I realized this happening for some (Gradle / Java) default imports as well, which is why at some point I wrote [this script](https://github.com/sschuberth/dev-scripts/blob/main/kotlin/remove-default-gradle-build-script-imports.main.kts).

PR #2538, opened 2025-04-17
feat(ui): Tune font colors for colourful elements

Etsija:
Demo page:

Screenshot from 2025-04-17 13-58-26

Screenshot from 2025-04-17 13-58-37

For discussion: for some elements, the black font is now almost too dark in light mode. Should it be fine-tuned?

### DIFF
--- a/ui/src/helpers/get-status-class.ts
+++ b/ui/src/helpers/get-status-class.ts
@@ -46,12 +46,12 @@ const STATUS_BACKGROUND_COLOR: {
   [K in Exclude<Status, undefined>]: string;
 } = {
   CREATED: 'bg-gray-500',
-  SCHEDULED: 'bg-blue-300',
-  RUNNING: 'bg-blue-500',
-  ACTIVE: 'bg-blue-500',
-  FAILED: 'bg-traffic-light-red',
-  FINISHED: 'bg-traffic-light-green',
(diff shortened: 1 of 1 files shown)


mnonnenmacher:
> For discussion: for some elements, the black font is now almost too dark in light mode. Should it be fine-tuned?

It looks good to me, I just wonder if `CREATE` should also white in dark mode.
If you want to make the effort you could use [this tool](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=BE3450) recommended by @mmurto to check if white or black scores better for each background color.

mnonnenmacher:
> > > For discussion: for some elements, the black font is now almost too dark in light mode. Should it be fine-tuned?
> > 
> > 
> > It looks good to me, I just wonder if `CREATE` should also white in dark mode. If you want to make the effort you could use [this tool](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=BE3450) recommended by @mmurto to check if white or black scores better for each background color. Or maybe there is some JS library that can automatically calculate the contrast and chose black or white based on that?
> 
> IIUC, TailwindCSS already does some magics when deciding if it uses white or black font. We could also approve this and improve on it later, what do you think? It would be nice to get it off my hands for Easter.

Of course, I just wanted to provide feedback to your question but it is already better than before.

mnonnenmacher (approved):

PR #2539, opened 2025-04-17
Prepare enabling and disabling ORT plugins globally

mnonnenmacher:
Prepare for enabling and disabling ORT plugins globally by creating the required configuration and UI, see the commit messages for details. The workers still ignore those settings, this needs to be changed in a later PR.

This PR introduces some code duplication mainly in integration tests, this will be addressed in a follow-up.

### DIFF
--- a/components/plugin-manager/implementation/src/main/kotlin/PluginEvent.kt
+++ b/components/plugin-manager/implementation/src/main/kotlin/PluginEvent.kt
@@ -23,7 +23,7 @@ import kotlinx.datetime.Instant
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 
-internal data class PluginEvent(
+data class PluginEvent(
     val pluginType: PluginType,
     val pluginId: String,
     val version: Long,
@@ -34,7 +34,7 @@ internal data class PluginEvent(
 
 /** The base class for all plugin event payloads. */
 @Serializable
-internal sealed class PluginEventPayload
+sealed class PluginEventPayload
 
 /** The payload for a plugin enabled event. */
 @Serializable
--- a/components/plugin-manager/implementation/src/main/kotlin/PluginEvent.kt
+++ b/components/plugin-manager/implementation/src/main/kotlin/PluginEvent.kt
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.components.pluginmanager
+
+import kotlinx.datetime.Instant
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+internal data class PluginEvent(
+    val pluginType: PluginType,
+    val pluginId: String,
+    val version: Long,
+    val payload: PluginEventPayload,
+    val createdAt: Instant,
+    val createdBy: String
+)
+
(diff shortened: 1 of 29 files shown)


core/src/main/kotlin/di/Module.kt
```diff
@@ -91,10 +91,9 @@ import org.koin.dsl.module
 
 /**
  * Creates the Koin module for the ORT server. The [config] is used to configure the application and the database. For
- * integration tests, the [setupDatabase] flag can be set to `false` as the data source will be provided by the
- * testcontainer there.
+ * integration tests, the [database][db] from the testcontainer can be provided directly.
  */
-fun ortServerModule(config: ApplicationConfig, setupDatabase: Boolean = true) = module {
+fun ortServerModule(config: ApplicationConfig, db: Database?) = module {
```
sschuberth, line 97: Should `db` have a default value of `null` as `setupDatabase` also had a default value before?

components/plugin-manager/implementation/src/main/kotlin/PluginEvent.kt
```diff
@@ -0,0 +1,47 @@
...
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.components.pluginmanager
+
+import kotlinx.datetime.Instant
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+internal data class PluginEvent(
```
sschuberth: Just double-checking: Doesn't this need to be `@Serializable`?

components/plugin-manager/implementation/src/main/kotlin/PluginEvent.kt
```diff
@@ -0,0 +1,47 @@
...
+    val createdBy: String
+)
+
+/** The base class for all plugin event payloads. */
+@Serializable
+internal sealed class PluginEventPayload
+
+/** The payload for a plugin enabled event. */
+@Serializable
+@SerialName("PluginEnabled")
```
sschuberth, line 41: Why is this necessary?

components/plugin-manager/implementation/src/main/kotlin/PluginEvent.kt
```diff
@@ -23,7 +23,7 @@ import kotlinx.datetime.Instant
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 
-internal data class PluginEvent(
+data class PluginEvent(
```
sschuberth, line 26: Is is really necessary to remove the `internal` classifiers?

sschuberth: Similar below.

core/src/main/kotlin/di/Module.kt
mnonnenmacher: I left it out because there is only one caller which always provides a value for that argument.

components/plugin-manager/implementation/src/main/kotlin/PluginEvent.kt
mnonnenmacher: No, only the payload is serialized when storing the event to the database.

components/plugin-manager/implementation/src/main/kotlin/PluginEvent.kt
mnonnenmacher: This makes the serialization more stable, because by default the whole qualified name is used, which means that if the package of the class is changed old entries in the database could not be deserialized anymore.

components/plugin-manager/implementation/src/main/kotlin/PluginEvent.kt
mnonnenmacher: Yes, because the event and payload classes are part of the `PluginEventStore` API which has to be public.

Etsija (approved):
I tested this by applying the branch, updating the API and running the server locally:
- UI code looks good to me
- switches work and their state is persisted properly
- I get enable/disable messages from queries when flipping the switches
My approval is for the UI part, but would like an approval for the back-end parts.

MarcelBochtler (approved):

ui/src/routes/admin/plugins/index.tsx
```diff
@@ -41,24 +50,92 @@ const PluginListCard = ({
...
+          onSuccess: () => {
+            toast.info('Disable Plugin', {
+              description: `Plugin "${plugin.displayName}" disabled successfully.`,
+            });
+            queryClient.invalidateQueries({
+              queryKey: [usePluginsServiceGetApiV1AdminPluginsKey],
+            });
+          },
+          onError(error: unknown) {
+            const apiError = error as ApiError;
```
mnonnenmacher, line 73: @Etsija This was the one tiny bit I didn't like about the UI code, everywhere else it's `onError(error: ApiError)` but I couldn't figure out how to make this work here, hence the cast. So if you have an idea how to improve it I would be interested.

PR #2567, opened 2025-04-23
Show only relevant issues in ORT Server run view

wkl3nk:
Previously, the ORT Server UI **could** display issues unrelated to the current run. This fix in the backend REST endpoint implementation for _issues_ ensures that only issues belonging to the actual run are shown.

### DIFF
--- a/services/hierarchy/src/main/kotlin/IssueService.kt
+++ b/services/hierarchy/src/main/kotlin/IssueService.kt
@@ -48,7 +48,6 @@ import org.jetbrains.exposed.sql.Op
 import org.jetbrains.exposed.sql.Query
 import org.jetbrains.exposed.sql.SortOrder
 import org.jetbrains.exposed.sql.alias
-import org.jetbrains.exposed.sql.count
 import org.jetbrains.exposed.sql.innerJoin
 
 /**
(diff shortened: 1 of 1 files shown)


wkl3nk:
@Etsija I wanted to express that this bug only shows up in the UI, while in the report everything is correct.

kamil-bielecki-bosch (approved):

Etsija:
> @Etsija I wanted to express that this bug only shows up in the UI, while in the report everything is correct.

From earlier PR comments I have understood that the `scope` in `<type>(<scope>): <description>` refers to a Gradle module (or UI project), not to where a bug is shown. But I'm not blocking this blocks the PR, I'm just saying we should stick to a policy in naming the commits, because every time I see `fix(ui)` in the title, I am expecting to be reviewing UI code.

Maybe the usage of the `scope` warrants some discussion and alignment between the teams. Any opinions, @mnonnenmacher @sschuberth ?

sschuberth:
> @Etsija I wanted to express that this bug only shows up in the UI, while in the report everything is correct.

I also still believe this should probably be `fix(services)` instead.

wkl3nk:
@Etsija @sschuberth Changed title of commit to "fix(services): ..."

PR #2569, opened 2025-04-23
ui: Show total run duration in the runs duration chart tooltips

Etsija:
Run not started:
Screenshot from 2025-04-24 07-42-14

Run ongoing (showing time counting up by seconds and pulsing):
Screenshot from 2025-04-24 07-42-20

Run finished (showing the total run time):
Screenshot from 2025-04-24 07-42-33

### DIFF
--- a/ui/src/components/charts/job-durations.tsx
+++ b/ui/src/components/charts/job-durations.tsx
@@ -17,10 +17,14 @@
  * License-Filename: LICENSE
  */
 import { useNavigate } from '@tanstack/react-router';
+import { Sigma } from 'lucide-react';
 import { useState } from 'react';
 import { Bar, BarChart, CartesianGrid, XAxis } from 'recharts';
 
 import { useRepositoriesServiceGetApiV1RepositoriesByRepositoryIdRuns } from '@/api/queries';
+import { LoadingIndicator } from '@/components/loading-indicator';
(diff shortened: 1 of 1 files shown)


ui/src/components/charts/job-durations.tsx
```diff
@@ -142,8 +143,28 @@ export const JobDurations = ({
...
+      reporterDuration,
+    ].filter((duration) => duration !== null).length;
+
+    // Calculate the total duration
+    const totalDuration =
+      (analyzerDuration ?? 0) +
+      (advisorDuration ?? 0) +
+      (scannerDuration ?? 0) +
+      (evaluatorDuration ?? 0) +
+      (reporterDuration ?? 0);
```
mnonnenmacher, line 162: You should rather use the `createdAt` and `finishedAt` properties of the run object itself to get the total run duration.

sschuberth: Oh, that's an interesting point. In that regard also the stacked-bar-chart visualization is wrong / misleading, as times do not all add up...

Etsija:
eclipsefdn/eca check seems to be failing for every PR (even dependency updates) now.

sschuberth:
> eclipsefdn/eca check seems to be failing for every PR (even dependency updates) now.

However, the ECA check is not marked as required (currently), so we may proceed.

PR #2578, opened 2025-04-24
ui: Align organization, product and repository-level menus with the UI

Etsija:
Screenshot from 2025-04-24 11-07-58

Screenshot from 2025-04-24 11-08-19

Screenshot from 2025-04-24 11-08-44


sschuberth:
Commit message nit: All your commits have "feat" as the topic (which is ok), but say "Refactor ..." in the title, which could trigger the question why the topic is not "refactor". To avoid that confusion, I'd simply change "Refactor ..." to "Rearrange ..." or "Group ...".

ui/src/routes/organizations/$orgId/route.tsx
```diff
@@ -37,56 +37,79 @@ const Layout = () => {
...
+        {
+          title: 'Settings',
+          to: '/organizations/$orgId/settings',
+          icon: () => <Settings className='h-4 w-4' />,
+          visible: user.hasRole([
+            'superuser',
+            `role_organization_${orgId}_admin`,
+          ]),
+        },
+      ],
```
lamppu, line 105: I suppose a user with only read access would now see the "Organization" label but nothing underneath it (and similar for the "Product" and "Repository" labels in the following commits). This probably should actually be changed in the Sidebar component itself, you can at the moment give a `visible` prop for the whole section, but it should probably be determined based on the visibility of the items instead. I can also do this in a follow-up PR if you'd like.

Etsija: Would be nice if you could do it, as you're more familiar with Sidebar component, as it's your handiwork.

PR #2600, opened 2025-04-28
Tune Graphite's retention policies

Etsija:
The commit message explains what is done in this PR, as per advice from @mnonnenmacher, but a sanity check is needed. 

I did verify that the changes to the files mounted from `scripts/compose/graphite/conf` are mirrored properly to the container, and the local Docker Compose environment starts properly and seems to be working. Some verification of the new retention policies would need to be done.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -130,6 +130,12 @@ services:
     - type: bind
       source: ./scripts/compose/graphite/storage
       target: /opt/graphite/storage
+    - type: bind
+      source: ./scripts/compose/graphite/conf/storage-schemas.conf
+      target: /opt/graphite/conf/storage-schemas.conf
+    - type: bind
+      source: ./scripts/compose/graphite/conf/storage-aggregation.conf
+      target: /opt/graphite/conf/storage-aggregation.conf
     logging:
       driver: gelf
       options:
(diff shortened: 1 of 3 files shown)


mnonnenmacher (commented):
@Etsija Did you also verify that changing the settings removes disk usage?

Etsija:
> @Etsija Did you also verify that changing the settings reduces disk usage?

No, hence the "Some verification of the new retention policies would need to be done". Haven't quite wrapped my head around how to do it cleverly, but I have some ideas. What I could do is to measure the amount of data written for, say, 5 runs for a test repository / with and without the retention change.

mnonnenmacher:
> > @Etsija Did you also verify that changing the settings reduces disk usage?
> 
> No, hence the "Some verification of the new retention policies would need to be done". Haven't quite wrapped my head around how to do it cleverly, but I have some ideas. What I could do is to measure the amount of data written for, say, 5 runs for a test repository / with and without the retention change. But I am not sure how the retention policy affects the long-standing data, so not sure this is the right way of doing it.

I did a quick search on the topic and it seems that the changes to the retention policy only affect new files. If you want to resize existing files there is a tool for it: https://github.com/graphite-project/whisper/tree/master?tab=readme-ov-file#whisper-resizepy
But if you just need to free your disk space you could also just remove the existing files.

mnonnenmacher (approved):

PR #2605, opened 2025-04-28
ui: Prevent the UI from breaking with Gateway Timeout

Etsija:
This PR fixes a bug that was found for a huge organization (over thousand repositories). Moreover, it introduces a general error boundary pattern that can be used elsewhere in the UI whenever a query response breaks the whole UI.

The 504 error was simulated by setting the organization statistics endpoint to return a 504 after a delay instead of an OK.

Loading state when fetching data:

Screenshot from 2025-04-28 13-46-44

Error fallback component when 504 happens:

Screenshot from 2025-04-28 13-51-28

Yes, the error title is a bit cheeky. It could be made as a prop, to generalize the component a bit.

### DIFF
--- a/ui/src/components/error-component.tsx
+++ b/ui/src/components/error-component.tsx
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


ui/src/components/error-component.tsx
```diff
@@ -0,0 +1,42 @@
...
+  CardTitle,
+} from '@/components/ui/card';
+
+export const ErrorComponent = ({ error }: ErrorComponentProps) => {
+  return (
+    <Card className='flex h-full gap-4'>
+      <CardHeader>
+        <CardTitle className='flex gap-2 align-baseline'>
+          <TriangleAlert className='text-traffic-light-red size-5' />
+          <div>Oops, something went wrong...</div>
```
wkl3nk, line 36: This is a great improvement that we can do the error handling for parts of the page in a separate way. If you don't mind, make the error message a property of the error component, so that it can be adapted. For instance, an error for statistics is not such a critical thing and we could also state this in an error message like "Currently, loading of statistics takes too long and was aborted."

wkl3nk (approved):

PR #2608, opened 2025-04-29
ui: Extend the error component

Etsija:
#2605 introduced a simple way to isolate UI errors in the DOM by setting error boundaries. If an error happens inside an error boundary, an error component is rendered in place of the erroring sub-DOM, instead of breaking the UI.

This PR extends the error component to provide both "user-friendly errors" (for production) and "technical errors" (mainly used for debugging purposes). Some examples:

1. Explicit user-friendly error title (this is now set for the organization overview page):
Screenshot from 2025-04-29 13-08-52

2. Error message with default props:
Screenshot from 2025-04-29 07-17-40

3. The error component can even show the stack trace of the error:
Screenshot from 2025-04-29 07-20-00

### DIFF
--- a/ui/src/routes/organizations/$orgId/index.tsx
+++ b/ui/src/routes/organizations/$orgId/index.tsx
@@ -87,7 +87,9 @@ const OrganizationComponent = () => {
       </div>
       <CatchBoundary
         getResetKey={() => 'reset'}
-        errorComponent={ErrorComponent}
+        errorComponent={(props) => (
+          <ErrorComponent {...props} title='Could not load statistics' />
+        )}
       >
         <div className='grid grid-cols-4 gap-2'>
           <Link
--- a/ui/src/components/error-component.tsx
+++ b/ui/src/components/error-component.tsx
@@ -26,16 +26,35 @@ import {
   CardHeader,
   CardTitle,
 } from '@/components/ui/card';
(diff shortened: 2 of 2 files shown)


ui/src/routes/organizations/$orgId/index.tsx
```diff
@@ -87,7 +87,13 @@ const OrganizationComponent = () => {
...
       <CatchBoundary
         getResetKey={() => 'reset'}
-        errorComponent={ErrorComponent}
+        errorComponent={(props) => (
+          <ErrorComponent
+            {...props}
+            title='Unable to load statistics'
+            explicitMessage='Currently, loading of statistics takes too long and was aborted.'
+          />
+        )}
```
lamppu, line 96: But this will now show this message with any error, not just the 504. So is there some way to process the actual error here and conditionally show a specific message?

lamppu (changes requested):

ui/src/routes/organizations/$orgId/index.tsx
Etsija: Oh, you're right. Will have to think.

Etsija: Fixed this as per the discussion in alignment meeting.

lamppu (approved):

PR #2613, opened 2025-04-29
UI: Set and show repository description

mnonnenmacher:
Extend the UI to:

Add inputs for the description to the forms for creating and editing repositories:
image

Show the description in the repository table:
image

Show the description in the repository overview:
image

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/-components/product-repository-table.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/-components/product-repository-table.tsx
@@ -63,6 +63,7 @@ const columns = [
           </Link>
           <div className='text-muted-foreground text-sm md:inline'>
             {row.original.type}
+            {row.original.description ? ` | ${row.original.description}` : ''}
           </div>
         </>
       ),
--- a/ui/src/routes/organizations/$orgId/products/$productId/create-repository/index.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/create-repository/index.tsx
@@ -25,6 +25,8 @@ import { z } from 'zod';
 
 import { useProductsServicePostApiV1ProductsByProductIdRepositories } from '@/api/queries';
 import { $RepositoryType, ApiError } from '@/api/requests';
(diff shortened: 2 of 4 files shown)


Etsija (commented):
Code LGTM.

mnonnenmacher:
> While at it, what do you think about dropping the repository type from various views? For most URLs, the "GIT" type does not add any value as ".git" is present in the URL. And also otherwise I guess that many users to not care too much about the repository type once the repository was added.

I think I would remove it from the table but not from the repo overview. I also don't like the way I show the type now in the overview very much, but I have no better idea right now, so maybe this could be improved separately?

sschuberth:
> I also don't like the way I show the type now in the overview very much, but I have no better idea right now,

How about simply omitting the type and URL, showing only the description? The type is not very valuable as discussed, and the URL is visible from the breadcrumb bar, so showing only the description should be sufficient.

> so maybe this could be improved separately?

Generally yes, sure, but if it can be improved by just removing something that's added here, I believe we should remove it straight here.

Etsija:
> > I also don't like the way I show the type now in the overview very much, but I have no better idea right now,
> 
> How about simply omitting the type and URL, showing only the description? The type is not very valuable as discussed, and the URL is visible from the breadcrumb bar, so showing only the description should be sufficient.
> 

The description is optional, so if not given, an empty title card would be rendered, so I'd have to object on that. I'm leaning more towards showing all three (if they exist) properties in the top title card. That would always show at least the URL and type.

Etsija:
How about the first line being "Git repository http://..."? I implemented a while ago "user-friendly" implementations of the repository types, and they should anyway be used here, instead of the  `RepositoryType` that you are now showing. The conversion is [implemented here](https://github.com/eclipse-apoapsis/ort-server/blob/3a6f069e77b8f3579fdcd8e8709178a1fd5f7b30/ui/src/lib/types.ts#L184-L191).

sschuberth (approved):
Let's merge this for the feature and improve the UI as a follow-up.

PR #2618, opened 2025-04-30
feat(schemas): Add JSON schemas for repository environment configuration

MarcelBochtler:
See commits for more details.

image


sschuberth:
PS: I'm not sure how you're generating the schema, but maybe https://github.com/Ricky12Awesome/json-schema-serialization could help in the future, even if currently unmaintained.

sschuberth:
> PS: I'm not sure how you're generating the schema, but maybe https://github.com/Ricky12Awesome/json-schema-serialization could help in the future, even if currently unmaintained.

Or actually, I guess https://github.com/SMILEY4/schema-kenerator can do the same by now.

workers/common/src/test/kotlin/common/JsonSchemaTest.kt
```diff
@@ -0,0 +1,77 @@
...
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldNot
+
+import java.io.File
+
+import org.ossreviewtoolkit.model.yamlMapper
+
+class JsonSchemaTest : WordSpec({
+    val validConfigs = File("src/test/resources")
+        .listFiles()
```
sschuberth, line 38: For portability, prefer Kotlin's `walk().maxDepth(1)` over Java's `listFiles()`. Same below.

sschuberth: That also allows to drop the `toList()` calls.

sschuberth (approved):

workers/common/src/test/kotlin/common/JsonSchemaTest.kt
```diff
@@ -0,0 +1,76 @@
...
+import io.kotest.inspectors.forAll
+import io.kotest.matchers.collections.beEmpty
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldNot
+
+import java.io.File
+
+import org.ossreviewtoolkit.model.yamlMapper
+
+@Suppress("IgnoredReturnValue")
```
MarcelBochtler, line 36: This seems to be required for `Sequence.forAll()`.
For `Collection.forAll()` it isn't required 🤷.

sschuberth: Could you please put these directly above each `forAll` call to highlight where the suppression is needed?

sschuberth (approved):

PR #2624, opened 2025-04-30
feat(dao): Retain at least one ORT run per repository

wkl3nk:
Ensure the housekeeping logic keeps at least one ORT run for each repository, even if it is older than the defined retention period. This improves continuity of scanning activities after long pauses.

Please check the respective commits for details.


mnonnenmacher:
The same could be achieved by adapting the `deleteRunsCreatedBefore` function to ignore the latest run by removing the highest value from the `runIds` list, this would be a much simpler implementation.

wkl3nk:
@mnonnenmacher You are right: To apply the changes only on the DAO for retrieving the Run IDs to be deleted has much less impact on the codebase, although the implementation is perhaps a little bit trickier than expected.

mnonnenmacher:
@wkl3nk Code looks good to me but there is one test to fix.

mnonnenmacher (approved):

PR #2625, opened 2025-04-30
Improve handling of plugin IDs

mnonnenmacher:

### DIFF
--- a/components/plugin-manager/implementation/src/main/kotlin/Utils.kt
+++ b/components/plugin-manager/implementation/src/main/kotlin/Utils.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.components.pluginmanager
+
+import org.ossreviewtoolkit.advisor.AdviceProviderFactory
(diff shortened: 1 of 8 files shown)


components/plugin-manager/implementation/src/main/kotlin/Utils.kt
```diff
@@ -0,0 +1,42 @@
...
+ */
+internal fun getInstalledPlugins(pluginType: PluginType): List<PluginDescriptor> =
+    when (pluginType) {
+        PluginType.ADVISOR -> AdviceProviderFactory.ALL.values.map { it.descriptor }
+        PluginType.PACKAGE_CONFIGURATION_PROVIDER ->
+            PackageConfigurationProviderFactory.ALL.values.map { it.descriptor }
+        PluginType.PACKAGE_CURATION_PROVIDER -> PackageCurationProviderFactory.ALL.values.map { it.descriptor }
+        PluginType.PACKAGE_MANAGER -> PackageManagerFactory.ALL.values.map { it.descriptor }
+        PluginType.REPORTER -> ReporterFactory.ALL.values.map { it.descriptor }
+        PluginType.SCANNER -> ScannerWrapperFactory.ALL.values.map { it.descriptor }
```
sschuberth, line 41: This overview reminds me: Aren't VCS plugins missing?

components/plugin-manager/implementation/src/main/kotlin/Utils.kt
```diff
@@ -40,3 +40,10 @@ internal fun getInstalledPlugins(pluginType: PluginType): List<PluginDescriptor>
         PluginType.REPORTER -> ReporterFactory.ALL.values.map { it.descriptor }
         PluginType.SCANNER -> ScannerWrapperFactory.ALL.values.map { it.descriptor }
     }
+
+/**
+ * Returns the normalized plugin ID for the given [pluginId] that exactly equals the case-sensitive ID of the plugin
+ * descriptor. If no match is found, `null` is returned.
+ */
+internal fun normalizePluginId(pluginType: PluginType, pluginId: String): String? =
```
sschuberth, line 48: Hmm. I guess the need for this is good input for https://github.com/oss-review-toolkit/ort/pull/10190, and we should adjust that PR so that this is not needed, or?

mnonnenmacher: The function would still be needed to ensure that entries in the database always use the same spelling.

components/plugin-manager/implementation/src/main/kotlin/Utils.kt
sschuberth: Can we somehow document this in the code? Otherwise core areas like this look like VCS plugins very accidentally omitted.

mnonnenmacher: I have made a separate PR #2631.

nnobelis (approved):

PR #2630, opened 2025-05-02
ui: Minor tweaks

Etsija:
This PR contains minor fixes and tweaks for the UI.

Repository URL in the title when not hovering over it:
Screenshot from 2025-05-02 07-24-24

Repository URL is rendered as blue when hovering over:
Screenshot from 2025-05-02 07-24-42

"Revision:" and the revision rendered on the same line. Moreover, "Created by:" falls back to showing the username in case full name doesn't exist:
Screenshot from 2025-05-02 07-25-30

For discussion: when changing the repository URL into a link, I decided to _not_ render it always as blue, to retain some consistency with the different titles in the UI. The link now turns blue only when hovering over with a mouse. This can easily be changed, however, if so decided.


sschuberth:
> For discussion: when changing the repository URL into a link, I decided to _not_ render it always as blue, to retain some consistency with the different titles in the UI.

While you could argue the other way around and say this is *in*consistent with other links in the UI, I'm fine with it.

sschuberth (approved):

PR #2634, opened 2025-05-02
ui: Small tweaks for the header

Etsija:
Show ORT logo as a "Home" indicator":
Screenshot from 2025-05-02 12-23-41

Show the gravatar also as the menu indicator:
Screenshot from 2025-05-02 12-49-26

### DIFF
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -27,8 +27,14 @@ import { z } from 'zod';
 import { ModeToggle } from '@/components/mode-toggle';
 import { Avatar, AvatarFallback } from '@/components/ui/avatar';
 import { Form, FormControl, FormField, FormItem } from '@/components/ui/form';
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from '@/components/ui/tooltip';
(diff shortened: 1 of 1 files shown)


mnonnenmacher:
>Show the gravatar also as the menu indicator (note that this screenshot is taken from Compose environment, with admin login, and the admin account doesn't have full name available, that's why it is showing "??". Should we change that?

If the full name is not available the first two letters of the username could be used, that would be nicer than "??".

ui/src/components/header.tsx
```diff
@@ -19,7 +19,7 @@
 
 import { zodResolver } from '@hookform/resolvers/zod';
 import { Link, useNavigate, useRouterState } from '@tanstack/react-router';
-import { CircleUser, Home, Menu } from 'lucide-react';
+import { Home, Menu } from 'lucide-react';
```
sschuberth, line 22: Commit message: This is not a "gravatar" in the sense of https://gravatar.com/. I'd say "initials icon" or so instead.

sschuberth (changes requested):

sschuberth (approved):

PR #2647, opened 2025-05-05
fix(ui): Import ORT logo from inside UI project

Etsija:
This is a fixup to #2634.


Etsija:
> Although I've been reading the commit message, it's unclear to me what issues this actually solved. The original code seems to have build fine, and the ORT logo was shown in the production deployment, so what does this actually fix? Is it purely a developer experience thing?

It is fixing a very annoying warning message coming from Vite hot reload: in VS Code, the 5-line warning message was shown every single time I saved a file (which triggers the hot reload).

sschuberth:
Thanks @Etsija, your explanation makes this now much clearer to me. Can we extract the gist of this, and replace the commit message with it?

Etsija:
> Thanks @Etsija, your explanation makes this now much clearer to me. Can we extract the gist of this, and replace the commit message with it?

I'll continue with this tomorrow.

REUSE.toml
```diff
@@ -200,6 +200,11 @@ path = "ui/src/assets/spdx-logo-color.svg"
 SPDX-FileCopyrightText = "2011 The Linux Foundation"
 SPDX-License-Identifier = "CC-BY-3.0"
 
+[[annotations]]
+path = "ui/src/assets/favicon.svg"
```
sschuberth, line 204: Please move this to before "ui/src/assets/spdx-logo-color.svg" to maintain alphabetical sorting by path.

sschuberth (changes requested):

sschuberth:
So, just to double-check: `main` already has `favicon.*` at

```
ui/public/favicon.svg
website/static/img/favicon.ico
```

and your PR now copies it to a third location at

```
ui/src/assets/favicon.svg
```

Do we really need to have basically the same file at all of these three locations, or can at least some of these be shared?

Etsija:
> So, just to double-check: `main` already has `favicon.*` at
> 
> ```
> ui/public/favicon.svg
> website/static/img/favicon.ico
> ```
> 
> and your PR now copies it to a third location at
> 
> ```
> ui/src/assets/favicon.svg
> ```
> 
> Do we really need to have basically the same file at all of these three locations, or can at least some of these be shared / `*.ico` vs `*.svg` formats be consolidated?

No, sorry, that seems to break Vite completely, and the error message wasn't at all helpful.

sschuberth:
> I tried to reference the favicon differently in the ui's root index.html and delete the public folder altogether, but that seems to break Vite completely, and the error message wasn't at all helpful.

Ok, so let's sum up:

- `website/static/img/favicon.ico` is need for the site that's currently hosted at https://eclipse-apoapsis.github.io/ort-server/, and not related to an ORT Server deployment at all, so it's quite understandable that this requires its own favicon.
- `ui/public/favicon.svg` is the *real favicon* used by an ORT Server deployment.
- `ui/src/assets/favicon.svg` actually is not a favicon (in the browser / bookmark sense), but the home icon used by the ORT Server UI, correct? So, to avoid confusion, should the copy introduced by this PR also be renamed, e.g. to `ort-server-icon.svg`?

Etsija:
> > I tried to reference the favicon differently in the ui's root index.html and delete the public folder altogether, but that seems to break Vite completely, and the error message wasn't at all helpful.
> 
> Ok, so let's sum up:
> 
>     * `website/static/img/favicon.ico` is need for the site that's currently hosted at https://eclipse-apoapsis.github.io/ort-server/, and not related to an ORT Server deployment at all, so it's quite understandable that this requires its own favicon.
> 
>     * `ui/public/favicon.svg` is the _real favicon_ used by an ORT Server deployment.
> 
>     * `ui/src/assets/favicon.svg` actually is not a favicon (in the browser / bookmark sense), but the home icon used by the ORT Server UI, correct? So, to avoid confusion, should the copy introduced by this PR also be renamed, e.g. to `ort-server-logo.svg`?

I agree. Should I do a separate commit for this renaming, or amend changes to the only commit in this PR?

sschuberth:
> I agree. @sschuberth should I do a separate commit for this renaming, or amend changes to the only commit in this PR?

I'll to a follow-up, as I discovered more icon / assets related things I's like to fix, so I'll just include this part into my PR.

sschuberth:
One more thing: I'm wonder why you chose to use the [favicon](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/public/favicon.svg) as the home icon instead of using the ORT Server [logo](https://github.com/eclipse-apoapsis/ort-server/blob/main/website/static/img/logo.svg). Is that because the "circle", which basically is the only difference between the two images, looks bad in dark mode?

Etsija:
> One more thing: I'm wonder why you chose to use the [favicon](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/public/favicon.svg) as the home icon instead of using the ORT Server [logo](https://github.com/eclipse-apoapsis/ort-server/blob/main/website/static/img/logo.svg). Is that because the "circle", which basically is the only difference between the two images, looks bad in dark mode?

Exactly.

REUSE.toml
```diff
@@ -196,7 +196,7 @@ SPDX-FileCopyrightText = "2018 OWASP Foundation, Inc (see <https://cyclonedx.org
 SPDX-License-Identifier = "CC-BY-ND-4.0"
 
 [[annotations]]
-path = "ui/src/assets/favicon.svg"
+path = "ui/src/assets/ort-server-logo.svg"
```
sschuberth, line 199: This now clashed with the renaming I propose in https://github.com/eclipse-apoapsis/ort-server/pull/2664. I realized too late that we have both a favicon and a logo, and that these actually are different. So, as this is not the "real" ORT Server logo with the circle, but a copy of the favison, I believe this shoul finally be named "home-icon.svg", and this renaming should already be part of the first commit which creates the copy.

sschuberth (changes requested):

sschuberth:
> > One more thing: I'm wondering why you chose to use the [favicon](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/public/favicon.svg) as the home icon instead of using the ORT Server [logo](https://github.com/eclipse-apoapsis/ort-server/blob/main/website/static/img/logo.svg). Is that because the "circle", which basically is the only difference between the two images, looks bad in dark mode?
> 
> Exactly.

This motivation should also be mentioned in the commit message.

REUSE.toml
Etsija: A bit of a hassle. So best to amend the two commits and rename once again?

Etsija: I have now amended the two commits, as I saw no use splitting this work into two, as all renaming and reasoning can well be kept in one commit.

sschuberth:
In the commit message, you now says "copying the ORT logo file" but later "The ORT Server favicon is used", which can be confusing. To resolve this, change the former to "copying the ORT Server favicon" (also adding the word "Server" here).

sschuberth (approved):

PR #2650, opened 2025-05-05
ui: Download links to OSC

Etsija:
Screenshot from 2025-05-05 15-40-06

### DIFF
--- a/ui/src/lib/constants.ts
+++ b/ui/src/lib/constants.ts
@@ -24,3 +24,7 @@
 // This probably needs to be changed in the future to a more sophisticated solution, e.g., by using a separate query
 // to get the distribution of items without fetching all items from back-end, which is costly.
 export const ALL_ITEMS = 100000;
+
+// The base download URL for the downloadable assets of the latest ORT Server release.
+export const GITHUB_LATEST_RELEASE_DOWNLOAD_URL =
+  'https://github.com/eclipse-apoapsis/ort-server/releases/latest/download';
--- a/ui/src/routes/about/index.tsx
+++ b/ui/src/routes/about/index.tsx
@@ -23,9 +23,23 @@ import { Fragment } from 'react';
 import { useVersionsServiceGetApiV1VersionsSuspense } from '@/api/queries/suspense';
 import { Card, CardContent, CardHeader } from '@/components/ui/card';
 import { Separator } from '@/components/ui/separator';
+import { GITHUB_LATEST_RELEASE_DOWNLOAD_URL } from '@/lib/constants';
 
 export const About = () => {
   const { data: versionData } = useVersionsServiceGetApiV1VersionsSuspense();
+
+  // Check if the ORT Server version is a release candidate (RC) version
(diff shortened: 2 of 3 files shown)


ui/src/routes/about/index.tsx
```diff
@@ -44,6 +45,48 @@ export const About = () => {
...
+            <div className='text-muted-foreground font-semibold'>Linux</div>
+            <a
+              href={`${GITHUB_DOWNLOAD_URL}/osc-cli-linux-x64.tar.gz`}
+              className='font-semibold text-blue-400 hover:underline'
+              target='_blank'
+              rel='noopener noreferrer'
+            >
+              osc-cli-linux-x64
+            </a>
+            <div className='text-muted-foreground font-semibold'>MacOS</div>
```
sschuberth, line 61: I believe the official spelling is "macOS". And we should have separate entries per macOS architecture, i.e. have "macOS (arm64)" and "macOS (x64)" entries with one link each.

ui/src/lib/constants.ts
```diff
@@ -24,3 +24,5 @@
 // This probably needs to be changed in the future to a more sophisticated solution, e.g., by using a separate query
 // to get the distribution of items without fetching all items from back-end, which is costly.
 export const ALL_ITEMS = 100000;
+export const GITHUB_DOWNLOAD_URL =
+  'https://github.com/eclipse-apoapsis/ort-server/releases/latest/download';
```
sschuberth, line 28: Now this is *always* using the latest version. Wouldn't it make more sense to use the version belonging to the release, if there is one, and only the latest version otherwise?

To simplify the logic I'd implement it was follows: If the version contains "_", assume it to be a pre-release and use the latest version. Otherwise, use the assets for the version.

ui/src/routes/about/index.tsx
Etsija: > ...And we should have separate entries per macOS architecture, i.e. have "macOS (arm64)" and "macOS (x64)" entries with one link each.

I would question why, as it's opposite to "streamlining", but OK.

ui/src/lib/constants.ts
Etsija: I have yet to see a case when the release tag coming from the back-end does _not_ end in RC and sha.  @mnonnenmacher what's your take on this?

sschuberth: > I have yet to see a case when the release tag coming from the back-end does not end in RC and sha.

Well, that would happen if we deployed a version that matches the commit for a release. But as we always deploy from `main`, we barely hit a tagged commit.

mnonnenmacher: >@mnonnenmacher what's your take on this?

We only deploy releases to our production environment, so it would be good if the download link would match the installed version.

ui/src/routes/about/index.tsx
sschuberth: To me, this breaks the visual appearance compared to the other entries, which strictly have only one value each.

sschuberth (approved):

PR #2654, opened 2025-05-05
test: Add an abstract integration test class to ktor-utils

mnonnenmacher:
Reduce code duplication in the integration tests of the plugin manager by extracting common code into an abstract class. The `AbstractIntegrationTest` is added as a test fixture to the ktor-utils so that it can be used in all components.

### DIFF
--- a/components/plugin-manager/implementation/build.gradle.kts
+++ b/components/plugin-manager/implementation/build.gradle.kts
@@ -68,15 +68,14 @@ dependencies {
 
     testImplementation(testFixtures(projects.clients.keycloak))
     testImplementation(testFixtures(projects.dao))
+    testImplementation(testFixtures(projects.shared.ktorUtils))
 
     testImplementation(libs.kotestAssertionsCore)
     testImplementation(libs.kotestAssertionsKtor)
     testImplementation(libs.kotestRunnerJunit5)
     testImplementation(libs.kotlinxSerializationJson)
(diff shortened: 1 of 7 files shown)


oheger-bosch:
The amount of redundant code that can be avoided is great.

I am just not sure about the inheritance approach which is rather strict and forces the same set of functionality to all derived classes. In [ScalaTest](https://www.scalatest.org/) which I have worked with in the past, there is a concept of mixins for different purposes that can be mixed into a test class to make specific functionality available in a flexible way. Now, Kotlin is not Scala, but with the Kotest test extensions maybe something similar can be achieved? So, the basic idea would be to not have a one-fits-all base class, but multiple test extensions for special purposes that can be combined in test classes. Just an idea?

sschuberth:
> Now, Kotlin is not Scala, but with the Kotest test extensions maybe something similar can be achieved?

Or use Kotlin class delegation, which seems to be the closest equivalent to Scala mixins (via traits)?

mnonnenmacher:
I have tried to turn it into a test extension, but I could not find a way to add the `Integration` tag from the extension, as all Kotest listeners are invoked after the spec was instantiated and then it's not allowed anymore to add tags. Therefore I'd like to keep the current inheritance approach (which we also use for all integration tests in `core`).

oheger-bosch (approved):

PR #2667, opened 2025-05-06
feat(ui): Drop the border and background color from the footer

Etsija:
This was a decision of today's alignment meeting.

Screenshot from 2025-05-06 16-01-19

Screenshot from 2025-05-06 16-01-26

### DIFF
--- a/ui/src/components/footer.tsx
+++ b/ui/src/components/footer.tsx
@@ -53,7 +53,7 @@ export function Footer() {
   const columnsAlignedRight = extractColumns(markdown, 'right');
   return (
     enabled && (
-      <footer className='bg-muted text-muted-foreground w-full border-t text-sm'>
+      <footer className='text-muted-foreground w-full text-sm'>
         <div className='mx-auto flex max-w-screen-xl flex-col items-center justify-between gap-4 p-4 sm:flex-row md:gap-8 md:p-8'>
           <nav className='flex flex-col gap-4 sm:flex-row sm:items-start'>
             {columnsAlignedLeft.map((content, index) => (


sschuberth:
> This was a decision of today's alignment meeting.

Can we get a screenshot for those who missed the decision? 😬

Etsija:
> > This was a decision of today's alignment meeting.
> 
> Can we get a screenshot for those who missed the decision? 😬

I deliberately left it out, as this is so trivial.

sschuberth:
> I deliberately left it out, as this is so trivial.

Thanks for still adding it. I just wanted to be super-sure to get the new look right. And for the record, I liked the original look quite a bit better, but I accept to be outvoted.

Etsija:
> > I deliberately left it out, as this is so trivial.
> 
> Thanks for still adding it. I just wanted to be super-sure to get the new look right. And for the record, I liked the original look quite a bit better, but I accept to be outvoted. For example, to me it looks weird that we now have no separator for the footer, but for the header.

As you missed the discussion: we decided to follow GitHub's convention on this for the footer.

sschuberth:
> As you missed the discussion: we decided to follow GitHub's convention on this for the footer.

Well, GitHub's footer is single-line and uses a smaller font than other content. So it's quite different.

Anyway, maybe we can later add a feature to the customizable footer to allow add a horizontal rule.

mnonnenmacher:
> Thanks for still adding it. I just wanted to be super-sure to get the new look right. And for the record, I liked the original look quite a bit better, but I accept to be outvoted. For example, to me it looks weird that we now have no separator for the footer, but for the header.

To me it's different, I found the background made the footer too flashy. In the rest of the UI grey is only used for highlighting elements. I find the new layout much cleaner.

But there is no need to rush if you want to test some tweaks, like adding a ruler or maybe increasing the distance between content and footer a bit. Or mabye also making the font a bit smaller.

One thing that bugs me is that the footer and also the run details bar can both be wider than the content, and I wonder why they are allowed to take more space than the content.

sschuberth:
> But there is no need to rush if you want to test some tweaks

There's also no need to block this, we can tweak things later.

> One thing that bugs me is that the footer and also the run details bar can both be wider than the content, and I wonder why they are allowed to take more space than the content.

This, on the other hand, doesn't bug me at all. People are different 🤷🏻‍♂️

mnonnenmacher (approved):

PR #2670, opened 2025-05-06
chore(ui): Remove a redundant `await` on a non-promise result

sschuberth:


sschuberth:
I was frankly just relying on the IntelliJ inspection hint without checking further details 😅 

image

PR #2679, opened 2025-05-08
feat(scanner): Add submodule fetch strategy for nested repositories

wkl3nk:
Introduce `submoduleFetchStrategy` config to control how the Scanner fetches Git submodules. When set to `TOP_LEVEL_ONLY`, only top-level submodules are fetched, avoiding timeouts on deeply nested repos.

This mirrors the behavior already available in the Analyzer and allows to resolve nested provenances even in this kind of repositories with a vast amount of nested submodules.

If activated, in the logs you will no longer see the `--recursive` flag in the `gib submodule update` command then:
````
Running 'git submodule update --init --depth 50' in '/tmp/ort-DefaultWorkingTreeCache13286267791034354700'..."
````

If particular VCS plugin configurations are active during a scan (like `submoduleFetchStrategy=TOP_LEVEL_ONLY`),
ensure that VCS plugin configurations are stored alongside nested provenance data. This prevents reuse of cache entries across scans with differing VCS plugin settings, ensuring correctness and reliability of scan results.

For storing the VCS plugin configuration, a canonical string like `Git/updateNestedSubmodules/false` is created and stored alogside with the nested provenance data.

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
+++ b/api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
@@ -187,7 +187,14 @@ data class ScannerJobConfiguration(
      * Keep the worker alive after it has finished. This is useful for manual problem analysis directly
      * within the pod's execution environment.
      */
-    val keepAliveWorker: Boolean = false
+    val keepAliveWorker: Boolean = false,
+
+    /**
+     * Specifies how submodules are fetched when resolving provenances. Currently supported only for Git repositories.
+     * If set to [SubmoduleFetchStrategy.FULLY_RECURSIVE] (default), all submodules are fetched recursively. If set
+     * to [SubmoduleFetchStrategy.TOP_LEVEL_ONLY], only the top-level submodules are fetched.
+     */
+    val submoduleFetchStrategy: SubmoduleFetchStrategy = SubmoduleFetchStrategy.FULLY_RECURSIVE
 )
 
 /**
--- a/model/src/commonMain/kotlin/JobConfigurations.kt
+++ b/model/src/commonMain/kotlin/JobConfigurations.kt
@@ -196,7 +196,14 @@ data class ScannerJobConfiguration(
      * Keep the worker alive after it has finished. This is useful for manual problem analysis directly
      * within the pod's execution environment.
      */
-    val keepAliveWorker: Boolean = false
+    val keepAliveWorker: Boolean = false,
+
+    /**
+     * Specifies how submodules are fetched when resolving provenances. Currently supported only for Git repositories.
+     * If set to [SubmoduleFetchStrategy.FULLY_RECURSIVE] (default), all submodules are fetched recursively. If set
+     * to [SubmoduleFetchStrategy.TOP_LEVEL_ONLY], only the top-level submodules are fetched.
+     */
+    val submoduleFetchStrategy: SubmoduleFetchStrategy = SubmoduleFetchStrategy.FULLY_RECURSIVE
 )
 
 /**
--- a/workers/scanner/src/main/kotlin/scanner/ScannerRunner.kt
+++ b/workers/scanner/src/main/kotlin/scanner/ScannerRunner.kt
@@ -52,6 +52,18 @@ class ScannerRunner(
     private val fileArchiver: FileArchiver,
     private val fileListStorage: OrtServerFileListStorage
 ) {
+    companion object {
+        /**
+         * Convert the VCS plugin configurations to a canonical string representation. If there are no VCS plugin
+         * configurations, return null.
+         */
+        fun createCanonicalVcsPluginConfigs(vcsPluginConfigs: Map<String, PluginConfig>) =
+            vcsPluginConfigs.keys.sorted().joinToString(separator = "&") { vcs ->
+                vcsPluginConfigs[vcs]?.options.orEmpty()
+                    .toSortedMap().entries.joinToString(separator = "&") { (key, value) -> "$vcs/$key/$value" }
+            }.ifEmpty { null }
+    }
+
     suspend fun run(
         context: WorkerContext,
         ortResult: OrtResult,
@@ -78,9 +90,14 @@ class ScannerRunner(
             )
         }
 
+        val canonicalVcsPluginConfigs = createCanonicalVcsPluginConfigs(vcsPluginConfigs)
         val packageProvenanceCache = PackageProvenanceCache()
         val packageProvenanceStorage = OrtServerPackageProvenanceStorage(db, scannerRunId, packageProvenanceCache)
-        val nestedProvenanceStorage = OrtServerNestedProvenanceStorage(db, packageProvenanceCache)
+        val nestedProvenanceStorage = OrtServerNestedProvenanceStorage(
+            db,
+            packageProvenanceCache,
+            canonicalVcsPluginConfigs
+        )
         val scanResultStorage = OrtServerScanResultStorage(db, scannerRunId)
 
         val scanStorages = ScanStorages(
--- a/workers/scanner/src/main/kotlin/scanner/ScannerRunner.kt
+++ b/workers/scanner/src/main/kotlin/scanner/ScannerRunner.kt
@@ -20,6 +20,7 @@
 package org.eclipse.apoapsis.ortserver.workers.scanner
 
 import org.eclipse.apoapsis.ortserver.model.ScannerJobConfiguration
+import org.eclipse.apoapsis.ortserver.model.SubmoduleFetchStrategy
 import org.eclipse.apoapsis.ortserver.workers.common.OrtServerFileListStorage
 import org.eclipse.apoapsis.ortserver.workers.common.context.WorkerContext
 import org.eclipse.apoapsis.ortserver.workers.common.mapToOrt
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: workers/scanner/src/main/kotlin/scanner/ScannerRunner.kt [hunk 1]
```diff
@@ -85,7 +87,20 @@ class ScannerRunner(
                 ?: listOf(SourceCodeOrigin.ARTIFACT, SourceCodeOrigin.VCS)
         )
 
-        val workingTreeCache = DefaultWorkingTreeCache()
+        // If the submodule fetch strategy is set to TOP_LEVEL_ONLY, for git use a plugin config that prevents that
```
oheger-bosch, line 90: Is there some way to test this? Maybe with a constructor mock and a verification that the Git-specific plugin options have actually been set?

workers/scanner/src/main/kotlin/scanner/ScannerRunner.kt
```diff
@@ -85,7 +87,20 @@ class ScannerRunner(
...
+            mapOf(
+                VcsType.GIT.toString() to PluginConfig(
+                    options = mapOf("updateNestedSubmodules" to "false")
+                )
+            )
+        } else {
+            emptyMap()
+        }
+
+        val workingTreeCache = DefaultWorkingTreeCache().addVcsPluginConfigs(vcsPluginConfigs)
```
mnonnenmacher, line 102: The VCS options are not part of the storage key for the nested provenance storage. This means that changing this setting has no effect for repositories where there is already a stored resolution result which could lead to unexpected results. I think to implement this correctly, the storage would have to be adapted as well which might also require changes in ORT.

mnonnenmacher: To be more precise, in the analyzer this option only affects the project repository, but here it affects also repositories of dependencies which might also be dependencies of other projects which do not use the `TOP_LEVEL_ONLY` strategy.

wkl3nk: To my understanding: You are talking about `ScanSummary`s that are persisted to storage, including things like `licenseFindings`, `copyrightFindings`, `snippetFindings` and `issues` ? But does not each Scanner (like ScanCode) check out the respective package to some "private" space before scanning it? Or am I wrong and all the scanners use this `workingTreeCache` ?

mnonnenmacher: No, I mean the `NestedProvenanceStorage`. The option you introduce influences the result when the scanner checks for submodules in a repository. The result will be stored in the `NestedProvenanceStorage` and then reused. But it will be reused independent of the configured strategy.
So let's say the strategy is set to `TOP_LEVEL_ONLY`, then the result will only contain direct submodules. Otherwise, it would also contain transitive submodules. Whatever strategy is configured when a dependency is scanned for the first time will then be reused even if a future scan uses a different strategy.
This is an edge case of course, because it's not so common that you have deeply nested submodule structures, but if it happens, the issue will be very difficult to find and debug.
Also, for the problem you are trying to solve the setting should only affect the scan of project repositories, not the scan of dependencies.
If you want we could do a brainstorming on Monday to check if there is maybe an easy way to avoid this issue.

workers/scanner/src/main/kotlin/scanner/ScannerRunner.kt
wkl3nk: @oheger-bosch This additional functionality was implemented as part of `DefaultWorkingTreeCache`, which is a class of ORT and not ORT Server. If it is to be tested, the test should be in ORT and not in ORT Server.

oheger-bosch: What about testing that the correct options are passed when doing the checkout?

wkl3nk: @oheger-bosch I tried it out to add a test, but the functionality to add the right parameters to the `git submodule update` command is buried deep into the VCS implementation (in this case: Git), and also just querying the plugin configuration of the Git (VCS) instance is not possible, as the configuration is a private member. And all of this is happing in ORT, and not ORT Server. Whatever, the functionality in the Git (VCS) is the same as it is already used by the Analyzer parameter `submoduleFetchStrategy=TOP_LEVEL_ONLY` so it is quite sure that the appropriate `git submodule update` command is used when the repository is checked out.

oheger-bosch: I would not go so far to test the VCS logic in ORT. What I mean is, there is the line
```
val workingTreeCache = DefaultWorkingTreeCache().addVcsPluginConfigs(vcsPluginConfigs)
```
in `ScannerRunner`. When using a constructor mock for `DefaultWorkingTreeCache`, it should be possible to verify that `addVcsPluginConfigs` was called on the object and to inspect the passed configuration.

workers/scanner/src/main/kotlin/scanner/ScannerRunner.kt
mnonnenmacher: This solves the storage problem, but what about this part of my comment:
> Also, for the problem you are trying to solve the setting should only affect the scan of project repositories, not the scan of dependencies.
Should the setting really affect the scan of dependencies?

wkl3nk:
@mnonnenmacher I extended the commit message including a statement about the remaining small risk of this solution.

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
```diff
@@ -187,7 +187,14 @@ data class ScannerJobConfiguration(
      * Keep the worker alive after it has finished. This is useful for manual problem analysis directly
      * within the pod's execution environment.
      */
-    val keepAliveWorker: Boolean = false
+    val keepAliveWorker: Boolean = false,
+
+    /**
+     * Specifies how submodules are fetched when resolving provenances. Currently supported only for Git repositories.
+     * If set to [SubmoduleFetchStrategy.FULLY_RECURSIVE] (default), all submodules are fetched recursively. If set
+     * to [SubmoduleFetchStrategy.TOP_LEVEL_ONLY], only the top-level submodules are fetched.
```
mnonnenmacher, line 195: Nit: docs are missing `SubmoduleFetchStrategy.DISABLED`. But to me different options don't need to be explained here at all as they are already documented in the enum.

model/src/commonMain/kotlin/JobConfigurations.kt
```diff
@@ -196,7 +196,14 @@ data class ScannerJobConfiguration(
      * Keep the worker alive after it has finished. This is useful for manual problem analysis directly
      * within the pod's execution environment.
      */
-    val keepAliveWorker: Boolean = false
+    val keepAliveWorker: Boolean = false,
+
+    /**
+     * Specifies how submodules are fetched when resolving provenances. Currently supported only for Git repositories.
+     * If set to [SubmoduleFetchStrategy.FULLY_RECURSIVE] (default), all submodules are fetched recursively. If set
+     * to [SubmoduleFetchStrategy.TOP_LEVEL_ONLY], only the top-level submodules are fetched.
```
mnonnenmacher, line 204: Nit: docs are missing SubmoduleFetchStrategy.DISABLED. But to me different options don't need to be explained here at all as they are already documented in the enum.

workers/scanner/src/main/kotlin/scanner/ScannerRunner.kt
```diff
@@ -58,6 +58,19 @@ class ScannerRunner(
     ): OrtScannerResult {
         val pluginConfigs = context.resolvePluginConfigSecrets(config.config)
 
+        // If the submodule fetch strategy is set to TOP_LEVEL_ONLY, for git use a plugin config that prevents that
+        // submodules are fetched recursively.
+        val vcsPluginConfigs = if (config.submoduleFetchStrategy == SubmoduleFetchStrategy.TOP_LEVEL_ONLY) {
```
mnonnenmacher, line 77: `SubmoduleFetchStrategy.DISBALED` should be handled as well, otherwise the results are confusing when a user sets it. If it's not supported to pass it ORT, it should at least fail.

workers/scanner/src/main/kotlin/scanner/ScannerRunner.kt
```diff
@@ -58,6 +58,19 @@ class ScannerRunner(
...
+            mapOf(
+                VcsType.GIT.toString() to PluginConfig(
+                    options = mapOf("updateNestedSubmodules" to "false")
+                )
+            )
+        } else {
+            emptyMap()
+        }
+        val canonicalVcsPluginConfigs = createCanonicalVcsPluginConfigs(vcsPluginConfigs)
```
mnonnenmacher, line 86: This line belongs to the next commit.

wkl3nk: True. But IntelliJ only allowed me to check the code block as a whole.

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
wkl3nk: For this particular case of this particular big repository I found it important to give a hint that TOP_LEVEL_ONLY could make sense.

workers/scanner/src/main/kotlin/scanner/ScannerRunner.kt
wkl3nk: Okay, instead to silently ignore it, throw an exception now.

mnonnenmacher (approved):

PR #2684, opened 2025-05-08
chore(ui): Tweak profile menu items a bit

sschuberth:
Make the conditionally shown administration item go frist and rename it to make clear it is not about an "admin" user but server "administration".

Similarly, make clear that "settings" are about the user, not the server.

### DIFF
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -270,17 +270,20 @@ export const Header = () => {
                 <span>{user.username}</span>
               </div>
             </DropdownMenuItem>
+            {user.hasRole(['superuser']) && (
+              <>
+                <DropdownMenuSeparator />
+                <Link to='/admin'>
+                  <DropdownMenuItem>Administration</DropdownMenuItem>
+                </Link>
+              </>
+            )}
             <DropdownMenuSeparator />
+            <Link to='/settings'>
+              <DropdownMenuItem>User Settings</DropdownMenuItem>
+            </Link>
             <Link to='/about'>
               <DropdownMenuItem>About</DropdownMenuItem>
-            </Link>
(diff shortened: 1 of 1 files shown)


ui/src/components/header.tsx
```diff
@@ -270,17 +270,20 @@ export const Header = () => {
                 <span>{user.username}</span>
               </div>
             </DropdownMenuItem>
+            {user.hasRole(['superuser']) && (
```
mnonnenmacher, line 273: Commit message:
* Typo in "a bit" and "go first"
* "make clear that it is"
* "but about server"

ui/src/components/header.tsx
```diff
@@ -270,17 +270,20 @@ export const Header = () => {
...
+              <>
+                <DropdownMenuSeparator />
+                <Link to='/admin'>
+                  <DropdownMenuItem>Server Administration</DropdownMenuItem>
```
mnonnenmacher, line 277: To me just "Administration" would be sufficient.

ui/src/components/header.tsx
```diff
@@ -270,17 +270,20 @@ export const Header = () => {
...
               <DropdownMenuItem>About</DropdownMenuItem>
             </Link>
-            {user.hasRole(['superuser']) && (
-              <Link to='/admin'>
-                <DropdownMenuItem>Admin</DropdownMenuItem>
-              </Link>
-            )}
             <Link to='/settings'>
-              <DropdownMenuItem>Settings</DropdownMenuItem>
+              <DropdownMenuItem>User Settings</DropdownMenuItem>
```
mnonnenmacher, line 286: I would put this also above "About".

mnonnenmacher (changes requested):

PR #2696, opened 2025-05-12
Add support for a custom "home" icon

Etsija:
This PR adds a table for storing key-value pairs for UI configuration, and endpoints for getting, inserting and updating the stored configuration values.  A minimal UI is implemented as the first use case of the table: 

1. No entries in the configuration table of the database. The UI falls back to showing the "hardcoded" ORT Server favicon as the Home icon:

Screenshot from 2025-05-12 10-42-45

2. The admin has configured the Home icon by providing a valid URL to it, and toggling the showing of the tailored icon on:

Screenshot from 2025-05-12 10-43-38

3. The admin can still  toggle between the tailored and hardcoded icons in the UI for comparison - this might be important later on, when we add support for more branding elements to the UI:

Screenshot from 2025-05-12 10-43-52


Etsija:
@mnonnenmacher I ran out of skills to implement the endpoint authorization integration tests properly, so I am seeking for your guidance here.

components/admin-config/backend/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+
+        filter {
+            includeGroup("org.gradle")
+        }
+    }
+}
+
+dependencies {
+    implementation(projects.components.authorization.implementation)
+    implementation(projects.dao)
```
sschuberth, line 44: Looks like this can be removed.

components/admin-config/backend/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+    implementation(projects.shared.ktorUtils)
+
+    implementation(libs.exposedCore)
+    implementation(libs.exposedJson)
```
sschuberth, line 48: Looks like this can be removed.

components/admin-config/backend/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+    implementation(libs.exposedKotlinDatetime)
+    implementation(libs.ktorOpenApi)
+    implementation(libs.ktorServerAuth)
+    implementation(libs.ktorServerCore)
+
+    testImplementation(testFixtures(projects.clients.keycloak))
+    testImplementation(testFixtures(projects.dao))
+    testImplementation(testFixtures(projects.shared.ktorUtils))
+
+    testImplementation(libs.kotestAssertionsCore)
```
sschuberth, line 58: Looks like this can be removed.

components/admin-config/backend/build.gradle.kts
Etsija: Do you mean both lines? That breaks the module. @mnonnenmacher promised in the alignment meeting to go through the dependencies and suggest removals.

sschuberth: No, only line 44. Similar for the other comment, they apply to the bottom-most quoted line.

Etsija: Removed.

core/build.gradle.kts
```diff
@@ -63,6 +63,7 @@ repositories {
 dependencies {
     implementation(projects.api.v1.apiV1Mapping)
     implementation(projects.clients.keycloak)
+    implementation(projects.components.adminConfig.backend)
```
mnonnenmacher, line 66: Nit: This dependency should be added in the commit where `core` actually starts depending on the new module.

components/admin-config/backend/src/main/kotlin/endpoints/GetConfigByKey.kt
```diff
@@ -0,0 +1,78 @@
...
+            description = "The configuration key."
+            required = true
+        }
+    }
+
+    response {
+        HttpStatusCode.OK to {
+            description = "Success"
+            body<Config> {
+                mediaTypes = setOf(ContentType.Application.Json)
```
mnonnenmacher, line 56: That should not be required.

components/admin-config/backend/src/main/kotlin/endpoints/GetConfigByKey.kt
```diff
@@ -0,0 +1,78 @@
...
+                    {
+                        "isEnabled": false,
+                        "value": "http://example.com/icon.png"
+                    }
+                """.trimIndent()
+                }
+            }
+        }
+
+        HttpStatusCode.InternalServerError to {
```
mnonnenmacher, line 68: This should return `BadRequest` instead.

components/admin-config/backend/src/main/kotlin/endpoints/InsertOrUpdateConfig.kt
```diff
@@ -0,0 +1,86 @@
...
+        "If the key does not exist, insert it."
+    tags = listOf("Admin")
+
+    request {
+        pathParameter<ConfigKey>("key") {
+            description = "The configuration key."
+            required = true
+        }
+        body<Config> {
+            mediaTypes = setOf(ContentType.Application.Json)
```
mnonnenmacher, line 52: This should not be required.

components/admin-config/backend/src/main/kotlin/endpoints/InsertOrUpdateConfig.kt
```diff
@@ -0,0 +1,86 @@
...
+                """.trimIndent()
+            }
+        }
+    }
+
+    response {
+        HttpStatusCode.OK to {
+            description = "The configuration entry was successfully inserted or updated."
+        }
+        HttpStatusCode.InternalServerError to {
```
mnonnenmacher, line 69: This should return `BadRequest` instead.

components/admin-config/backend/src/test/kotlin/endpoints/GetConfigByKeyIntegrationTest.kt
```diff
@@ -0,0 +1,104 @@
...
+                        authenticate("test") {
+                            insertOrUpdateConfig(dbExtension.db)
+                            getConfigByKey(dbExtension.db)
+                        }
+                    }
+                }
+
+                // Insert a test configuration key to database
+                client.post("/admin/config/HOME_ICON_URL") {
+                    contentType(ContentType.Application.Json)
```
mnonnenmacher, line 53: This should not be required.

components/admin-config/backend/src/test/kotlin/endpoints/GetConfigByKeyIntegrationTest.kt
```diff
@@ -0,0 +1,104 @@
...
+
+                val response = client.get("/admin/config/HOME_ICON_URL")
+
+                response shouldHaveStatus HttpStatusCode.OK
+                val config = response.body<Config>()
+                config.value shouldBe "https://example.com/existing_icon.png"
+                config.isEnabled shouldBe true
+            }
+        }
+        "return the default value if the configuration key does not exist in db" {
```
mnonnenmacher, line 70: Add an empty line between test cases.

components/admin-config/backend/src/test/kotlin/endpoints/GetConfigByKeyIntegrationTest.kt
```diff
@@ -0,0 +1,104 @@
...
+
+                val response = client.get("/admin/config/HOME_ICON_URL")
+
+                response shouldHaveStatus HttpStatusCode.OK
+                val config = response.body<Config>()
+                config.value shouldBe "https://example.com/icon.png"
+                config.isEnabled shouldBe false
+            }
+        }
+        "return InternalServerError if the configuration key is invalid" {
```
mnonnenmacher, line 88: Add an empty line between test cases.

components/admin-config/backend/src/test/kotlin/endpoints/InsertOrUpdateConfigIntegrationTest.kt
```diff
@@ -0,0 +1,131 @@
...
+
+                val response = client.get("/admin/config/HOME_ICON_URL")
+
+                response shouldHaveStatus HttpStatusCode.OK
+                val config = response.body<Config>()
+                config.value shouldBe "https://example.com/icon.png"
+                config.isEnabled shouldBe true
+            }
+        }
+        "update the value and isEnabled status of an existing configuration key" {
```
mnonnenmacher, line 69: Add an empty line between test cases.

components/admin-config/backend/src/test/kotlin/endpoints/InsertOrUpdateConfigIntegrationTest.kt
```diff
@@ -0,0 +1,131 @@
...
+
+                val response2 = client.get("/admin/config/HOME_ICON_URL")
+
+                response2 shouldHaveStatus HttpStatusCode.OK
+                val config2 = response2.body<Config>()
+                config2.value shouldBe "https://changed/example.com/explicit_icon.png"
+                config2.isEnabled shouldBe true
+            }
+        }
+        "return InternalServerError if the configuration key is invalid" {
```
mnonnenmacher, line 115: Add an empty line between test cases.

components/admin-config/backend/src/test/kotlin/endpoints/InsertOrUpdateConfigIntegrationTest.kt
```diff
@@ -0,0 +1,131 @@
...
+
+                val response = client.get("/admin/config/HOME_ICON_URL")
+
+                response shouldHaveStatus HttpStatusCode.OK
+                val config = response.body<Config>()
+                config.value shouldBe "https://example.com/icon.png"
+                config.isEnabled shouldBe true
+
+                client.post("/admin/config/HOME_ICON_URL") {
+                    contentType(ContentType.Application.Json)
```
mnonnenmacher, line 98: This should not be required.

components/admin-config/backend/src/test/kotlin/endpoints/InsertOrUpdateConfigIntegrationTest.kt
```diff
@@ -0,0 +1,131 @@
...
+                    routing {
+                        authenticate("test") {
+                            insertOrUpdateConfig(dbExtension.db)
+                            getConfigByKey(dbExtension.db)
+                        }
+                    }
+                }
+
+                client.post("/admin/config/HOME_ICON_URL") {
+                    contentType(ContentType.Application.Json)
```
mnonnenmacher, line 52: This should not be required.

ui/src/routes/admin/content-management/home-icon/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  return (
+    <Card>
+      <CardHeader>
+        <CardTitle>Home Icon</CardTitle>
+        <CardDescription>
+          Customise the home icon by providing a URL to an icon image. The
+          toggle specifies whether the default (ORT Server) icon should be used
+          or the icon from the provided URL.
```
mnonnenmacher, line 140: I would remove the parenthesis around "ORT Server".

mnonnenmacher (changes requested):

components/admin-config/backend/src/main/kotlin/endpoints/GetConfigByKey.kt
Etsija: _What_ shouldn't be required, exactly?

components/admin-config/backend/src/main/kotlin/endpoints/GetConfigByKey.kt
Etsija: But the endpoint returns `InternalServerError` when a non-existing key is requested for, so I don't quite understand how I can change that behavior.

components/admin-config/backend/src/main/kotlin/endpoints/InsertOrUpdateConfig.kt
Etsija: _What_, exactly? You are quoting four lines of code here.

core/build.gradle.kts
Etsija: To clarify: this is the second commit, where the endpoints are defined?

components/admin-config/backend/src/main/kotlin/endpoints/InsertOrUpdateConfig.kt
Etsija: Same comment as before.

components/admin-config/backend/src/test/kotlin/endpoints/InsertOrUpdateConfigIntegrationTest.kt
Etsija: `contentType` _is_ needed, otherwise I get 
```
Fail to prepare request body for sending. 
The body type is: class org.eclipse.apoapsis.ortserver.components.adminconfig.Config, with Content-Type: null.

If you expect serialized body, please check that you have installed the corresponding plugin(like `ContentNegotiation`) and set `Content-Type` header.
```

components/admin-config/backend/src/main/kotlin/endpoints/InsertOrUpdateConfig.kt
sschuberth: As for my comments, these comments always refer to the bottom-most line, i.e. the content type. GitHub just renders the other lines for context.

components/admin-config/backend/src/test/kotlin/endpoints/GetConfigByKeyIntegrationTest.kt
Etsija: `contentType` _is_ needed, otherwise I get 
```
Fail to prepare request body for sending. 
The body type is: class org.eclipse.apoapsis.ortserver.components.adminconfig.Config, with Content-Type: null.

If you expect serialized body, please check that you have installed the corresponding plugin(like `ContentNegotiation`) and set `Content-Type` header.
```

components/admin-config/backend/src/test/kotlin/endpoints/InsertOrUpdateConfigIntegrationTest.kt
Etsija: Same comment as above.

components/admin-config/backend/src/main/kotlin/endpoints/InsertOrUpdateConfig.kt
Etsija: Nice to know...or, would've been nice to know long time ago :smirk: I already deduced that must be it, from Martin's comments, otherwise, the code excerpt wouldn't have made sense.

components/admin-config/backend/src/test/kotlin/endpoints/GetConfigByKeyIntegrationTest.kt
mnonnenmacher: Ok, good to know, I'll address this in a follow-up because we don't need it in the integration tests in `core`.

components/admin-config/backend/src/main/kotlin/endpoints/GetConfigByKey.kt
mnonnenmacher: This happens because `enumValueOf<ConfigKey>` below throws an exception when the key is invalid and you do not catch it, so the catch-all hanlder in `StatusPages.kt` in `core` turns it into an internal server error.
Instead, you should catch the exception and return a meaningful error response like so:
```kotlin
val keyParameter = call.requireParameter("key")
val key = runCatching { 
  enumValueOf<ConfigKey>(keyParameter)
}.getOrElse {
  call.respond(HttpStatusCode.BadRequest, ErrorResponse("Invalid config key $keyParameter.", "Allowed keys are: ..."))
  return@get
}
```

You could also throw an exception that is mapped to `BadRequest` in `StatusPages.kt`, but this global error handling does not play well with the idea that component should be self-contained.

Etsija: Trying to use `ErrorResponse` in the endpoint leads to a circular dependency warning:
Screenshot from 2025-05-14 06-19-10

Etsija: I can't think of another way to pass in an error response properly than to specify a serializable `ErrorResponse` data class locally, but that sounds like a dumb workaround. Isn't core/api's `ErrorResponse` sitting in wrong location for the components to stay self-contained?

Etsija: I avoided the circular dependency problem by mapping the error response locally to similar than the `ErrorResponse` class. Also verified that the error toast is rendered properly. While this isn't as elegant as using `ErrorResponse` for the type of the response, I hope it is enough for now.

mnonnenmacher: I would suggest to move `ErrorResponse` to `:shared:ktor-utils` but it can be done in a follow-up.

components/admin-config/backend/src/test/kotlin/endpoints/GetConfigByKeyIntegrationTest.kt
sschuberth: @mnonnenmacher, will you then also do the follow-up you've mentioned [here](https://github.com/eclipse-apoapsis/ort-server/pull/2696#discussion_r2088105104)? I just don't want to lose track of planned follow-ups.

Issue #2668, opened 2025-05-06
Some endpoints do not return error message and cause


sschuberth:
When clicking on "Create" to create a new run with invalid configuration settings, like using a reporter plugin that's disabled, a "Bad Gateway" toast is shown (whose title already is a bit weird), but more importantly the toast misses the details of the error response. See:

![Image](https://github.com/user-attachments/assets/ce0853f3-b7c7-4a78-b947-ed9705f3811f)

sschuberth:
@Etsija, to reproduce this, disable e.g. the "PDF Template Reporter" in the admin section, and then create / rerun any run and enable the "ORT PDF Reports".

Etsija:
The error toast is designed to [always display the error `message` and `cause` when the error is an `ApiError` and those exist](https://github.com/eclipse-apoapsis/ort-server/blob/770d479c2b199be4729f8d10f0e775a58f7cf824/ui/src/components/toast-error.tsx#L34-L44). I reproduced an endpoint failure by specifically setting an endpoint in my newest PR to fail. Because the endpoint is designed to catch this, it does produce an `ApiError` and the error response is proper, and the toast shows the response:

![Image](https://github.com/user-attachments/assets/530d46e9-339d-4433-a961-ddd563bab9cf)

Then I reproduced your error situation, and saw a curious response:

![Image](https://github.com/user-attachments/assets/c97d821a-af63-4b51-8e58-7d8e1d07640d)

The error response is totally different, and specifically, it completely misses the message and cause of the error.

I went to look for the "create run" endpoint's implementation. I am suspecting this happens, because [the endpoint only seems to specify implementation for the 200 OK case](https://github.com/eclipse-apoapsis/ort-server/blob/770d479c2b199be4729f8d10f0e775a58f7cf824/core/src/main/kotlin/apiDocs/RepositoriesDocs.kt#L436-L464), and none of the failing cases.

Because I am suspecting the endpoint error handling needs more implementation, I also labeled this with `api`.

Etsija:
This could be improved if I could get my hands on the "payload" of the error. However, I do not know of a way to do that.

mnonnenmacher:
@Etsija Could you please verify if #2703 fixes this issue?

Etsija:
> [@Etsija](https://github.com/Etsija) Could you please verify if [#2703](https://github.com/eclipse-apoapsis/ort-server/pull/2703) fixes this issue?

It does, thank you.

PR #2703, opened 2025-05-12
fix(core): Return `ErrorResponse` when disabled plugins are used

mnonnenmacher:
This aligns the error response with other endpoints.


Etsija:
I verified this by
- disabling the PDF reporter plugin
- trying to create a run for a repository, using the "ORT PDF Reports" option in the Reporters section
- now the error message and cause are shown properly:

Screenshot from 2025-05-13 06-55-47

The PR seems to fix #2668 (at least from this part; there might still be some endpoints which do not return `ErrorResponse` properly).

mnonnenmacher:
> The PR seems to fix #2668 (at least from this part; there might still be some endpoints which do not return `ErrorResponse` properly).

Thanks for checking, I went through the code, all other endpoints should be fine.

sschuberth (approved):

PR #2711, opened 2025-05-13
Apply curations in packages endpoint

mnonnenmacher:
Apply curations to packages and include the applied curations in the result. See the commit messages for details.

### DIFF
--- a/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
+++ b/api/v1/mapping/src/commonMain/kotlin/Mappings.kt
@@ -58,6 +58,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunFilters as ApiOrtRunFil
 import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunStatus as ApiOrtRunStatus
 import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunSummary as ApiOrtRunSummary
 import org.eclipse.apoapsis.ortserver.api.v1.model.Package as ApiPackage
+import org.eclipse.apoapsis.ortserver.api.v1.model.PackageCurationData as ApiPackageCurationData
 import org.eclipse.apoapsis.ortserver.api.v1.model.PackageFilters as ApiPackageFilters
 import org.eclipse.apoapsis.ortserver.api.v1.model.PackageManagerConfiguration as ApiPackageManagerConfiguration
 import org.eclipse.apoapsis.ortserver.api.v1.model.PagedResponse as ApiPagedResponse2
@@ -88,6 +89,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.User as ApiUser
 import org.eclipse.apoapsis.ortserver.api.v1.model.UserDisplayName as ApiUserDisplayName
 import org.eclipse.apoapsis.ortserver.api.v1.model.UserGroup as ApiUserGroup
 import org.eclipse.apoapsis.ortserver.api.v1.model.VcsInfo as ApiVcsInfo
+import org.eclipse.apoapsis.ortserver.api.v1.model.VcsInfoCurationData as ApiVcsInfoCurationData
 import org.eclipse.apoapsis.ortserver.api.v1.model.Vulnerability as ApiVulnerability
 import org.eclipse.apoapsis.ortserver.api.v1.model.VulnerabilityRating as ApiVulnerabilityRating
 import org.eclipse.apoapsis.ortserver.api.v1.model.VulnerabilityReference as ApiVulnerabilityReference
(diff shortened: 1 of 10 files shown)


api/v1/mapping/src/commonMain/kotlin/Mappings.kt
```diff
@@ -58,6 +58,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunFilters as ApiOrtRunFil
 import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunStatus as ApiOrtRunStatus
 import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunSummary as ApiOrtRunSummary
 import org.eclipse.apoapsis.ortserver.api.v1.model.Package as ApiPackage
+import org.eclipse.apoapsis.ortserver.api.v1.model.PackageCurationData as ApiPackageCurationData
```
sschuberth, line 61: Commit message:

> include the applied packages in the result

Should this say "original packages"? Or "applied curations"?

BTW, in case we'd want to show original / uncurated data in the UI, we should be able to get it. I.e. our previous approach in ORT to "unapply" curations did not work, so also here we should probably get the original package plus any curations, and maybe for convenience also the curated package.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
```diff
@@ -878,7 +882,31 @@ fun PackageRunData.mapToApi() = ApiPackage(
     pkg.vcsProcessed.mapToApi(),
     pkg.isMetadataOnly,
     pkg.isModified,
-    shortestDependencyPaths.map { it.mapToApi() }
+    shortestDependencyPaths.map { it.mapToApi() },
+    curations.map { it.mapToApi() }
+)
+
+fun PackageCurationData.mapToApi() = ApiPackageCurationData(
+comment = comment,
```
sschuberth, line 890: Wrong indentation.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
```diff
@@ -878,7 +882,31 @@ fun PackageRunData.mapToApi() = ApiPackage(
...
+    authors = authors,
+    concludedLicense = concludedLicense,
+    description = description,
+    homepageUrl = homepageUrl,
+    binaryArtifact = binaryArtifact?.mapToApi(),
+    sourceArtifact = sourceArtifact?.mapToApi(),
+    vcs = vcs?.mapToApi(),
+    isMetadataOnly = isMetadataOnly,
+    isModified = isModified,
+    declaredLicenseMapping = declaredLicenseMapping
```
sschuberth, line 902: This lacks the `sourceCodeOrigins` and `labels` properties.

services/ort-run/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -495,6 +501,63 @@ class PackageServiceTest : WordSpec() {
...
+                    provider = PackageCurationProviderConfig("test"),
+                    curations = listOf(curation1, curation2)
+                )
+
+                fixtures.resolvedConfigurationRepository.addPackageCurations(ortRunId, listOf(resolvedPackageCurations))
+
+                val packages = service.listForOrtRunId(ortRunId)
+                packages.data shouldHaveSize 2
+
+                with (packages.data.single { it.pkg.identifier == pkg1.identifier }) {
```
sschuberth, line 545: Use no space after `with`.

services/ort-run/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -495,6 +501,63 @@ class PackageServiceTest : WordSpec() {
...
+                    pkg.authors should containExactly(*curation1.data.authors.orEmpty().toTypedArray())
+                    concludedLicense shouldBe curation1.data.concludedLicense
+                    curations shouldHaveSize 1
+                    curations.first() shouldBe curation1.data
+                }
+
+                with (packages.data.single { it.pkg.identifier == pkg2.identifier }) {
```
sschuberth, line 552: Use no space after `with`.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
mnonnenmacher: Fixed the commit message. The package is already curated, if you think we want to show the uncurated package in the UI I can make another PR to add the uncurated package.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
mnonnenmacher: Both of them have not been implemented in the server after they were introduced in ORT, so currently they are simply ignored.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
sschuberth: > if you think we want to show the uncurated package in the UI I can make another PR to add the uncurated package.

I don't have an immediate need for that *right now*, but I remember you saying you'd want a similar step-by-step view of applied curations as in the workbench. And I still like that idea.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
sschuberth: I've created https://github.com/eclipse-apoapsis/ort-server/issues/2773 to track that.

api/v1/mapping/src/commonMain/kotlin/Mappings.kt
mnonnenmacher: I think the step-by-step view would require re-implementing ORT's logic to apply curations in the UI, but I will add the uncurated package, then the UI could at least show a diff between curated and uncurated package.

PR #2717, opened 2025-05-14
CLI: Provide some default values for the download commands 

MarcelBochtler:


cli/src/commonMain/kotlin/utils/FileHelpers.kt
```diff
@@ -24,11 +24,18 @@ import io.ktor.utils.io.readAvailable
...
+import okio.Path.Companion.toPath
 import okio.SYSTEM
 import okio.Source
 import okio.buffer
 import okio.use
 
+/**
+ * Get the current working directory as a [Path].
+ */
+internal val cwd: Path
```
mnonnenmacher, line 36: Please call this `currentWorkingDir` for readability.

cli/src/commonMain/kotlin/ReportsCommand.kt
```diff
@@ -86,7 +85,13 @@ class ReportsCommand : SuspendingCliktCommand(name = "reports") {
 
         outputDir.mkdirs()
 
-        fileNames.forEach { fileName ->
+        val reportNames = fileNames ?: run {
+            val ortRun = client.runs.getOrtRun(resolvedOrtRunId)
+
+            ortRun.jobs.reporter?.reportFilenames
+        } ?: emptyList()
```
mnonnenmacher, line 92: Nit: You could make it a bit shorter:
```kotlin
val reportNames = fileNames
  ?: client.runs.getOrtRun(resolvedOrtRunId).jobs.reporter?.reportFilenames
  ?: emptyList()
```

mnonnenmacher (changes requested):

cli/src/commonMain/kotlin/ReportsCommand.kt
sschuberth: It becomes yet shorter by using `orEmpty()` 😸

MarcelBochtler: Yes, it is shorter.
But IMO the readability gets worse:
```kt
val reportNames = fileNames ?: client.runs.getOrtRun(resolvedOrtRunId).jobs.reporter?.reportFilenames.orEmpty()
```

I would like to keep it as is
(I replaced `?: emptyList()` with `.orEmpty()`)

Alternatively, I would extract the ORT run fallback to a private function.

mnonnenmacher (approved):

PR #2719, opened 2025-05-14
feat(core): Improve error message for unavailable plugins

mnonnenmacher:
Differentiate between plugins which are not installed and plugins which are disabled when checking if the plugins used in an ORT run are available. Previously, also not installed plugins were reported as disabled.

### DIFF
--- a/core/src/main/kotlin/api/ProductsRoute.kt
+++ b/core/src/main/kotlin/api/ProductsRoute.kt
@@ -71,7 +71,7 @@ import org.eclipse.apoapsis.ortserver.core.apiDocs.postRepository
 import org.eclipse.apoapsis.ortserver.core.apiDocs.postSecretForProduct
 import org.eclipse.apoapsis.ortserver.core.apiDocs.putUserToProductGroup
 import org.eclipse.apoapsis.ortserver.core.services.OrchestratorService
-import org.eclipse.apoapsis.ortserver.core.utils.getDisabledPlugins
+import org.eclipse.apoapsis.ortserver.core.utils.getUnavailablePlugins
 import org.eclipse.apoapsis.ortserver.core.utils.pagingOptions
 import org.eclipse.apoapsis.ortserver.model.Repository
 import org.eclipse.apoapsis.ortserver.model.Secret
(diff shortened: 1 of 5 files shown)


sschuberth:
To me, "invalid plugin" sounds odd in this context, like a corrupted plugin. How about using "unavailable plugin" instead as the umbrella term for not installed and disabled plugins?

core/src/main/kotlin/api/RepositoriesRoute.kt
```diff
@@ -150,15 +150,14 @@ fun Route.repositories() = route("repositories/{repositoryId}") {
                     UserDisplayName(principal.getUserId(), principal.getUsername(), principal.getFullName())
                 }
 
-                // Check if disabled plugins are used.
-                val disabledPlugins = createOrtRun.getDisabledPlugins(pluginService)
-                if (disabledPlugins.isNotEmpty()) {
+                // Check if invalid plugins are used.
```
sschuberth, line 153: This still says "invalid".

core/src/test/kotlin/api/ProductsRouteIntegrationTest.kt
```diff
@@ -1579,6 +1579,62 @@ class ProductsRouteIntegrationTest : AbstractIntegrationTest({
...
+                    productId = productId,
+                    description = null
+                ).id
+
+                val advisorPluginId = "invalidAdvisor"
+                val packageConfigurationProviderPluginId = "invalidPackageConfigurationProvider"
+                val packageCurationProviderPluginId = "invalidPackageCurationProvider"
+                val packageManagerPluginId = "invalidPackageManager"
+                val reporterPluginId = "invalidReporter"
+                val scannerPluginId = "invalidScanner"
```
sschuberth, line 1597: Should these be changed to also say "unavailable"?

core/src/test/kotlin/api/RepositoriesRouteIntegrationTest.kt
```diff
@@ -756,6 +756,57 @@ class RepositoriesRouteIntegrationTest : AbstractIntegrationTest({
...
+        "respond with 'BadRequest' if not installed plugins are used" {
+            integrationTestApplication {
+                val createdRepository = createRepository()
+
+                val advisorPluginId = "invalidAdvisor"
+                val packageConfigurationProviderPluginId = "invalidPackageConfigurationProvider"
+                val packageCurationProviderPluginId = "invalidPackageCurationProvider"
+                val packageManagerPluginId = "invalidPackageManager"
+                val reporterPluginId = "invalidReporter"
+                val scannerPluginId = "invalidScanner"
```
sschuberth, line 768: Should these be changed to also say "unavailable"?

core/src/main/kotlin/api/ProductsRoute.kt
```diff
@@ -410,14 +410,13 @@ fun Route.products() = route("products/{productId}") {
             }
 
             // Check if disabled plugins are used.
```
sschuberth, line 412: I guess the comment here should be updated as well.

core/src/main/kotlin/utils/Utils.kt
```diff
@@ -24,19 +24,25 @@ import org.eclipse.apoapsis.ortserver.components.pluginmanager.PluginService
...
-                    add(pluginType to pluginId)
-                }
+fun CreateOrtRun.getUnavailablePlugins(pluginService: PluginService): UnavailablePlugins {
+    val notInstalled = mutableListOf<Pair<PluginType, String>>()
+    val disabled = mutableListOf<Pair<PluginType, String>>()
+
+    getPlugins().forEach { (pluginType, pluginIds) ->
+        pluginIds.forEach { pluginId ->
+            if (!pluginService.isInstalled(pluginType, pluginId)) {
+                notInstalled.add(pluginType to pluginId)
```
sschuberth, line 36: Why not use `+=` here and below?

sschuberth (changes requested):

wkl3nk (approved):

sschuberth (approved):

PR #2727, opened 2025-05-16
refactor(worker): Make createContext public again

wkl3nk:
Change the visibility of `WorkerContextFactory.createContext` back to public again, because the ORT Server Config script test must be able to call it to validate the parameter script.


oheger-bosch:
It is not necessary for this use case that the function is public. Tests can instead use the `withContext` function.

PR #2740, opened 2025-05-19
refactor: Do not add storages to the Koin context

mnonnenmacher:
Different `Storage`s could be use by different services, so do not add them to the Koin context to avoid that a service accidentally uses the wrong `Storage`. This also is more consistent because in modules that define multiple `Storage`s only one of them was added to the context.


oheger-bosch:
Koin supports giving bindings a [name](https://insert-koin.io/docs/reference/koin-core/definitions/#definition-naming--default-bindings). This allows having multiple beans of the same type and distinguishing between them. I would actually prefer using this alternative for storages, as it is more in-line with the DI-based programming model used in ORT Server.

oheger-bosch (approved):

PR #2742, opened 2025-05-19
build: Replace printVersion Gradle task with an ORT specific version

MarcelBochtler:
Fix the `printVersion` task by replacing it with an custom variant.
This is helpful for automated tasks where the exact version is required.

Before: 0.7.0-RC
After: 0.7.0-RC.040.sha.8baa883

Signed-off-by: Marcel Bochtler <marcel.bochtler@bosch.com>


sschuberth (approved):

build.gradle.kts
```diff
@@ -88,6 +90,15 @@ tasks.register("allDependencies") {
     }
 }
 
+open class OrtServerPrintTask : PrintTask({ "" }, "Prints the current project version") {
```
mnonnenmacher, line 93: `VersionPrintTask` would be a better more specific name.

sschuberth: @MarcelBochtler, if you care to change it, I'd be fine with the same name in ORT.

PR #2743, opened 2025-05-19
test: Use valid test data

mnonnenmacher:
Use valid values in test data that would not cause exceptions when importing the data into an `OrtResult`. This is a preparation for using the `OrtResult` class in more places in a later change.


core/src/test/kotlin/api/OrganizationsRouteIntegrationTest.kt
```diff
@@ -1782,12 +1782,12 @@ class OrganizationsRouteIntegrationTest : AbstractIntegrationTest({
                     homepageUrl = "https://example.com",
                     binaryArtifact = RemoteArtifact(
                         "https://example.com/example-1.0.jar",
-                        "sha1:value",
+                        "0123456789abcdef0123456789abcdef01234567",
```
sschuberth, line 1785: BTW, as I remember having discussions in the past in ORT core what constant string to use for dummy hash values in tests (all zeros vs something else) so they are easily identifiable as dummy values, maybe also `SHA1.emptyValue` could be used.

oheger-bosch: Isn't there a high risk that when adding new test cases again invalid values come in? Not sure how to fully avoid this (maybe by using dedicated value classes in the model?), but couldn't there be at least some functions somewhere that support the generation of test data for hashes, licenses, and whatever might be problematic?

sschuberth: > Isn't there a high risk that when adding new test cases again invalid values come in?

Well, they wouldn't be strictly *invalid*, but just violating verbal conventions, so it's not that bad.

> Not sure how to fully avoid this

A custom detect rule comes to mind. We could use and publish such a rule as part of ORT core.

> couldn't there be at least some functions somewhere that support the generation of test data for hashes, licenses, and whatever might be problematic?

Instead of having individual helper functions for these, I'd rather wish could extend the [test data DSL](https://github.com/oss-review-toolkit/ort/issues/755#issuecomment-1441539863) in ORT.

sschuberth (approved):

PR #2747, opened 2025-05-19
feat(dao): Improve performance of loading analyzer runs

mnonnenmacher:
Add queries to bulk load the contents of an `AnalyzerRun` and use that instead of the `AnalyzerRunDao` in the `DaoAnalyzerRunRepository`.

The DAO suffers from the N + 1 problem because an `AnalyzerRun` contains several nested objects and the Exposed DAO API creates lots of separate queries to load them. Depending on the amount of projects and packages in an `AnalyzerRun` this can easily result in several thousand executed queries.

The new approach solves the N + 1 problem by bulk loading the data from the one-to-many relations: For example, the authors for all packages are loaded in a single query and then associated to the packages in memory. This results in a constant amount of executed queries, independent of the number of projects or packages in the `AnalyzerRun`.

Loading all data in a single query would not have been feasible, because SQL builds a cross product when multiple one-to-many relations are involved in a joined query, exploding the number of result rows.

Tests have shown that this change reduces the duration for loading an `AnalyzerRun` by a factor of 50-100, for example, a run with ~1,000 packages now loads in ~200ms instead of ~10s. This affects all workers that run after the analyzer, because all of them have to build an `OrtResult` object from the database.

Similar changes should also be made for loading the other parts of the ORT result.

### DIFF
--- a/dao/src/main/kotlin/repositories/analyzerjob/AnalyzerJobsTable.kt
+++ b/dao/src/main/kotlin/repositories/analyzerjob/AnalyzerJobsTable.kt
@@ -48,6 +48,13 @@ object AnalyzerJobsTable : LongIdTable("analyzer_jobs") {
     val finishedAt = timestamp("finished_at").nullable()
     val configuration = jsonb<AnalyzerJobConfiguration>("configuration")
     val status = enumerationByName<JobStatus>("status", 128)
+
+    /** Get the ORT run ID for the given job [id]. Returns `null` if no job is found. */
+    fun getOrtRunIdById(id: Long): Long? =
+        select(ortRunId)
+            .where { AnalyzerJobsTable.id eq id }
+            .map { it[ortRunId].value }
+            .firstOrNull()
 }
 
 class AnalyzerJobDao(id: EntityID<Long>) : LongEntity(id) {
--- a/dao/src/main/kotlin/repositories/analyzerrun/AnalyzerConfigurationsPackageManagerConfigurationsTable.kt
+++ b/dao/src/main/kotlin/repositories/analyzerrun/AnalyzerConfigurationsPackageManagerConfigurationsTable.kt
@@ -19,7 +19,10 @@
 
 package org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun
 
+import org.eclipse.apoapsis.ortserver.model.runs.PackageManagerConfiguration
+
 import org.jetbrains.exposed.sql.Table
+import org.jetbrains.exposed.sql.selectAll
 
 /**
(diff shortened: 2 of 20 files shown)


kamil-bielecki-bosch:
As a general observation: adding custom queries to classes that are named as *Table may be bit misleading.
Maybe, instead of adding more and more stuff into *Table class, creating new classes named i.e. *ReadModel would be a better and cleaner solution? It also will make future refactors towards CQRS way easier.

oheger-bosch:
First, it is great that you achieved such a significant performance improvement!

However, I also have to admit that I am quite overwhelmed by these changes. The efficient loading of an Analyzer result now requires non-trivial query logic spread over more than a dozen of table classes; I am very concerned how this can be maintained. This gives the impression that we do not use the ORM framework, but fight against it.

I have no better solution either. Could it help to have the query logic in a more central place, for instance in an `OrtResultLoader` class? This could make it easier to keep the overview over the tables that are queried. (Maybe that goes a bit in the same direction as what @kamil-bielecki-bosch says.)

Probably some radical simplifications in the data model are required on the long run.

mnonnenmacher:
@kamil-bielecki-bosch @oheger-bosch To me it seemed convenient to add the queries to the tables that serve as "entry points" to the queries, it also makes the code a bit shorter as I can often skip the table names (e.g., `allowDynamicVersions` instead of `AnalyzerConfigurationsTable.allowDynamicVersions`). We could also collect them in a central place, of course. How about discussing this in the community meeting today and applying it as a follow-up?

kamil-bielecki-bosch:
@mnonnenmacher @oheger-bosch What I meant was: Single responsibility for a single class. If class is called i.e. "AnalyzerRunTable" I would look there just for stuff directly associated with this table and only this table. Maybe, with references necessary for this table to "live". In case of multiple JOINs / custom query - It's no longer responsibility of table, but it's more like service / read model to read bunch of data in some specified and structurized way. In fact that approach simplifies things in long run.
About using long names - maybe Kotlin's import-as would be useful there?
And I agree with @oheger-bosch: DB structure is way too complicated. IMHO we should consider some document database, which is by design great fit for keeping stuff like reports / results "all-in-one" (also to delete it as single document).

Other than that: great job!

mnonnenmacher:
> Single responsibility for a single class. If class is called i.e. "AnalyzerRunTable" I would look there just for stuff directly associated with this table and only this table.

I agree with that, but my question remains if it is ok to move the queries to another location later on.

kamil-bielecki-bosch:
@mnonnenmacher Fine for me. Just please, create some of ticket for that, to not to forget that.

oheger-bosch (approved):

kamil-bielecki-bosch (approved):

PR #2757, opened 2025-05-20
fix(ui): Render empty breadcrumb items if no access rights to entities

Etsija:
Screenshot from 2025-05-20 13-02-06

### DIFF
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -198,30 +198,37 @@ export const Header = () => {
         </Sheet>
         <Breadcrumb>
           <BreadcrumbList>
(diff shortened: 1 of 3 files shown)


Etsija:
> A few improvement suggestions:
> 
>     1. Use null instead of an empty string when you don't want to show the breadcrumb. This might also automatically hide the separators.
> 
>     2. If the above doesn't hide the separators, change the breadcrumb component to not show them,
> 
>     3. In the try-catch for the queries, only use null for 403 errors.

`null` wasn't allowed, I used `undefined`. Addressed all of your comments, now the separators won't be showing.

mmurto (approved):

PR #2758, opened 2025-05-20
test: Add `:utils:logging` as test dependency to all JVM modules

mnonnenmacher:
This fixes the issue that the custom `MdcConverter` was not on the classpath when running tests in several modules resulting in an exception from Logback.

### DIFF
--- a/buildSrc/src/main/kotlin/ort-server-kotlin-jvm-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/ort-server-kotlin-jvm-conventions.gradle.kts
@@ -37,6 +37,7 @@ plugins {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/ort-server-kotlin-jvm-conventions.gradle.kts
```diff
@@ -37,6 +37,7 @@ plugins {
 }
 
 dependencies {
+    testImplementation(project(":utils:logging"))
```
sschuberth, line 40: Have you checked whether this in turn should be removed from `testImplementation` in the actual JVM projects?

sschuberth (approved):

PR #2761, opened 2025-05-20
feat(keycloak): Get Keycloak groups in chunks

kamil-bielecki-bosch:
This commit changes way that groups are retrieved from Keycloak. As group number increases with new clients, in short time it's number can reach numbers, that cause problems with request timeouts. To prevent that, groups are now retrieved in chunks of 5000.


sschuberth:
> it's number can reach numbers

That reads a bit odd. How about something like "With the groups count growing large, requesting all of them at once can cause request timeouts"?

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
```diff
@@ -155,10 +159,25 @@ class DefaultKeycloakClient(
...
+            val groupNumber = httpClient.get("$apiUrl/groups/count").body<GroupCount>().count
+            val groups = mutableSetOf<Group>()
+            val groupRequests = ceil(groupNumber.toDouble() / GROUP_GET_CHUNK_SIZE).toInt()
+
+            for (chunkNo in 0 until groupRequests) {
+                groups.addAll(
+                    httpClient.get("$apiUrl/groups") {
+                        url {
+                            parameters.append("first", (chunkNo * GROUP_GET_CHUNK_SIZE).toString())
+                            parameters.append("max", (((chunkNo + 1) * GROUP_GET_CHUNK_SIZE) - 1).toString())
```
sschuberth, line 171: `max` is not the last index, but "the *maximum* number of results that are to be returned". So this can be hard-coded to `GROUP_GET_CHUNK_SIZE`.

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
```diff
@@ -155,10 +159,25 @@ class DefaultKeycloakClient(
...
+                        }
+                    }.body<List<Group>>()
+                )
+            }
+
+            return groups
```
sschuberth, line 177: The `return` is superfluous here.

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
```diff
@@ -155,10 +159,25 @@ class DefaultKeycloakClient(
...
         }.getOrElse {
             throw KeycloakClientException("Failed to load groups.", it)
-        }.body()
+        }
```
sschuberth, line 180: All in all, I'd propose instead something like

    override suspend fun getGroups(): Set<Group> =
        runCatching {
            val groupCount = httpClient.get("$apiUrl/groups/count").body<GroupCount>().count

            buildSet(groupCount) {
                var first = 0

                while (first < groupCount) {
                    val chunk = httpClient.get("$apiUrl/groups") {
                        url {
                            parameters.append("first", first.toString())
                            parameters.append("max", GROUP_GET_CHUNK_SIZE.toString())
                        }
                    }.body<Set<Group>>()

                    addAll(chunk)

                    first += chunk.size
                }
            }
        }.getOrElse {
            throw KeycloakClientException("Failed to load groups.", it)
        }

sschuberth (changes requested):

wkl3nk:
Such a pity that the API does not directly supporting filtering by some kind of prefix 😭

mnonnenmacher:
> > Such a pity that the API does not directly supporting filtering by some kind of prefix 😭
> 
> BTW, why does ORT Server use its own Keycloak client instead of [the official one](https://github.com/keycloak/keycloak-client)? 

@bs-ondem Can you tell?

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
kamil-bielecki-bosch: Changes made

sschuberth: I believe that someone else than me should review this now.

kamil-bielecki-bosch:
@mnonnenmacher @wkl3nk 

> > Such a pity that the API does not directly supporting filtering by some kind of prefix 😭
> 
> Not by prefix, but it support search, so it could search for `PREFIX_` and then manually filter groups that do not start with "PREFIX_" but have the string somewhere in the middle of the name. However, the last part would likely never be required because group names follow the pattern `(ORGANIZATION|PRODUCT|REPOSITORY)_{ID}_(ADMINS|READERS|WRITERS)`. So if the refix is `PROD`, searching for `PROD_` should already filter out all groups that belong to other instances.

Tried to search by "prod" (query: `{{URL}}/groups?search=prod_&max=50`) and looks like Keycloak ignores `_` sign, so all groups that contains `prod` are included in results, including ones like `dev_PRODUCT_1_READERS`. Tried with wildcards, quotas, with `q` parameter (`/groups?q='name:prod_'` etc). Such queries either returns everything that contains `prod` or returns nothing.

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
kamil-bielecki-bosch: You're right. Anyway, I prefer to keep it as constant (to avoid "magic numbers") inside code.

mnonnenmacher:
> @mnonnenmacher @wkl3nk
> 
> > > Such a pity that the API does not directly supporting filtering by some kind of prefix 😭
> > 
> > 
> > Not by prefix, but it support search, so it could search for `PREFIX_` and then manually filter groups that do not start with "PREFIX_" but have the string somewhere in the middle of the name. However, the last part would likely never be required because group names follow the pattern `(ORGANIZATION|PRODUCT|REPOSITORY)_{ID}_(ADMINS|READERS|WRITERS)`. So if the refix is `PROD`, searching for `PROD_` should already filter out all groups that belong to other instances.
> 
> Tried to search by "prod" (query: `{{URL}}/groups?search=prod_&max=50`) and looks like Keycloak ignores `_` sign, so all groups that contains `prod` are included in results, including ones like `dev_PRODUCT_1_READERS`. Tried with wildcards, quotas, with `q` parameter (`/groups?q='name:prod_'` etc). Such queries either returns everything that contains `prod` or returns nothing.

It's a pity that Keycloak ignores the underscore but it would still reduce the size of the result significantly, especially for prefixes that are not part of the generated group names.

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
sschuberth: That's what I meant: Just use `GROUP_GET_CHUNK_SIZE` without any math. I wasn't referring to inlining the `GROUP_GET_CHUNK_SIZE` constant.

kamil-bielecki-bosch:
@mnonnenmacher Added basic filtering for groups - it should decrease number of records significantly.

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
```diff
@@ -153,12 +155,36 @@ class DefaultKeycloakClient(
         internalClientId ?: findClientId(clientId).also { internalClientId = it }
     }
 
-    override suspend fun getGroups(): Set<Group> =
+    override suspend fun getGroups(groupNameFilter: String?): Set<Group> =
         runCatching {
-            httpClient.get("$apiUrl/groups")
+            val groupCount = httpClient.get("$apiUrl/groups/count") {
+                url {
+                    groupNameFilter?.let {
```
sschuberth, line 162: As the return value is unused, prefer `also` over `let`.

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
```diff
@@ -153,12 +155,36 @@ class DefaultKeycloakClient(
...
+                        parameters.append("search", groupNameFilter)
```
sschuberth, line 163: Here, `it` should be used instead of `groupNameFilter`.

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
```diff
@@ -153,12 +155,36 @@ class DefaultKeycloakClient(
...
+            buildSet(groupCount) {
+                var first = 0
+
+                while (first < groupCount) {
+                    val chunk = httpClient.get("$apiUrl/groups") {
+                        url {
+                            parameters.append("first", first.toString())
+                            parameters.append("max", GROUP_GET_CHUNK_SIZE.toString())
+                            groupNameFilter?.let {
+                                parameters.append("search", groupNameFilter)
```
sschuberth, line 177: See above.

sschuberth (dismissed):

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
```diff
@@ -153,12 +155,36 @@ class DefaultKeycloakClient(
         internalClientId ?: findClientId(clientId).also { internalClientId = it }
     }
 
-    override suspend fun getGroups(): Set<Group> =
+    override suspend fun getGroups(groupNameFilter: String?): Set<Group> =
         runCatching {
-            httpClient.get("$apiUrl/groups")
+            val groupCount = httpClient.get("$apiUrl/groups/count") {
```
oheger-bosch, line 160: Idea for a further optimization: Is it possible to drop this call to get the group count by counting the number of groups returned for a chunk? If a request for a chunk of groups yields less than `GROUP_GET_CHUNK_SIZE` elements, the last chunk has been reached.

sschuberth: Note that this is the *total* number of groups, which is used to optimize the capacity for the set in line 168. Esp. for a large number of groups / chunks properly setting the initial capacity can also yield significant performance gains. I'm not sure whether the same would work with your approach.

clients/keycloak/src/main/kotlin/KeycloakClient.kt
```diff
@@ -28,7 +28,7 @@ interface KeycloakClient {
     /**
      * Return a set of all [groups][Group], which currently exist in the Keycloak realm.
      */
-    suspend fun getGroups(): Set<Group>
+    suspend fun getGroups(groupNameFilter: String? = null): Set<Group>
```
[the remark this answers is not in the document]
kamil-bielecki-bosch: Good point. Done.

clients/keycloak/src/main/kotlin/DefaultKeycloakClient.kt
```diff
@@ -153,12 +155,36 @@ class DefaultKeycloakClient(
...
+                            parameters.append("first", first.toString())
+                            parameters.append("max", GROUP_GET_CHUNK_SIZE.toString())
+                            groupNameFilter?.also { filter ->
+                                parameters.append("search", filter)
+                            }
+                        }
+                    }.body<List<Group>>()
+                    addAll(chunk)
+                    first += chunk.size
+                }
```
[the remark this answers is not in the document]
mnonnenmacher: I think a constructor argument for the chunk size would be better as the function caller should not have to care about it and we might want to make it configurable.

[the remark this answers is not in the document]
kamil-bielecki-bosch: Tests added, new param to config added to accomodate chunk size

bs-ondem:
> > Such a pity that the API does not directly supporting filtering by some kind of prefix 😭
> 
> BTW, why does ORT Server use its own Keycloak client instead of [the official one](https://github.com/keycloak/keycloak-client)? Would it make sense to switch to reduce maintenance effort?

I honestly wasn’t aware of the existence of the official Keycloak Java client when I started implementing the requests. There was no specific reason for not using it. It simply didn’t come up during the initial research and IMO the official Keycloak documentation doesn’t clearly point to the Java client.

sschuberth:
> I honestly wasn’t aware of the existence of the official Keycloak Java client when I started implementing the requests. There was no specific reason for not using it.

Ok, then let me check switching to it in the context of https://github.com/eclipse-apoapsis/ort-server/issues/2778.

PR #2763, opened 2025-05-20
Load packages from ORT Result

mnonnenmacher:
Load packages from an `OrtResult` instead of directly from the database to prepare for upcoming features. See the commit messages for details.


core/src/main/kotlin/di/Module.kt
```diff
@@ -161,7 +192,29 @@ fun ortServerModule(config: ApplicationConfig, db: Database?) = module {
...
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get()
```
sschuberth, line 215: 🤣

oheger-bosch: Can the `singleOf(::OrtRunService)` declaration style be applied here?

services/ort-run/src/main/kotlin/OrtRunService.kt
```diff
@@ -375,13 +375,19 @@ class OrtRunService(
      * If [failIfRepoInfoMissing] is *true*, throw an [IllegalArgumentException] if the repository information
      * is missing; otherwise, return an empty [Repository] object.
      */
-    fun generateOrtResult(ortRun: OrtRun, failIfRepoInfoMissing: Boolean = true): OrtResult {
+    fun generateOrtResult(
+        ortRun: OrtRun,
+        loadAdvisorRun: Boolean = true,
```
oheger-bosch, line 380: I like this feature to select the parts that should be loaded. This should also be reflected in the docs of this function.

services/ort-run/src/main/kotlin/OrtRunService.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
```
oheger-bosch: I am fine with moving this functionality from `workers:common` to `services`, but the existing two `OrtRunService`s seem rather disjunct to me. One provides functions dealing with multiple runs, the other one deals with the data of a single run (and maybe should be better named `OrtResultService`?). So, is it really beneficial to merge the two? The resulting class becomes quite huge with a long list of constructor parameters, and I don't see much synergy of functionality that can be shared.

mnonnenmacher: I planned to split the big class by run type, so move all functions related to an analyzer run to an `AnalyzerRunService` and so on, and keep only functions that deal with the whole run in the `OrtRunService`, like `generateOrtResult`, or functions that do not belong to a specific run, like `storeResolvedPackageCurations` or `storeIssues`. What do you think?

oheger-bosch: Ok, it would probably be possible to have a common `WorkerRunService` base class defining a common interface and providing functionality based on the `WorkerJobRepository` interface. The workers would then use their dedicated `XXXRunService`. Is this the plan?

oheger-bosch (approved):

PR #2771, opened 2025-05-21
docs(ui): Remove "ORT" from "ORT Run"

sschuberth:
Simplify and align the wording to just refer to a "run". Having "ORT" here does not really add any value, and the distinction between the similar terms "run ID" and "run index" is already confusing enough without "ORT".

### DIFF
--- a/ui/src/routes/admin/plugins/index.tsx
+++ b/ui/src/routes/admin/plugins/index.tsx
@@ -178,7 +178,7 @@ const PluginsComponent = () => {
       />
       <PluginListCard
         title='Reporters'
-        description='Reporters generate reports like SBOMs from the ORT run data.'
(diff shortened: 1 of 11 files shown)


ui/src/routes/admin/plugins/index.tsx
```diff
@@ -178,7 +178,7 @@ const PluginsComponent = () => {
       />
       <PluginListCard
         title='Reporters'
-        description='Reporters generate reports like SBOMs from the ORT run data.'
+        description='Reporters generate reports like SBOMs from the data of a run.'
```
mnonnenmacher, line 181: Maybe "from the data collected in a run"?

mnonnenmacher (approved):

PR #2779, opened 2025-05-22
fix(cli): Properly configure the release build

sschuberth:
Just enabling `optimized` clashes with the default of `debuggable` also being enabled, resulting in a compiler warning saying

    Unsupported combination of flags: -opt and -g. Please pick one.

Instead of disabling `debuggable`, use the proper way to configure the build type(s) [1], which takes care of mutually setting these options.

### DIFF
--- a/cli/build.gradle.kts
+++ b/cli/build.gradle.kts
@@ -19,6 +19,7 @@
 
 import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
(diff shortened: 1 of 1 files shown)


nnobelis (approved):

nnobelis:
Approved. Feel free to wait for @MarcelBochtler feedback.

MarcelBochtler:
> Is it acceptable to now _only_ build the release type, @MarcelBochtler? At least that should reduce build times, I guess.

Yes. I've never used the debuggable build anyways.
Thanks.

MarcelBochtler (approved):

PR #2788, opened 2025-05-23
fix(cli): Enable path completion for `--parameters-file` option

MarcelBochtler:
Because Clikt doesn't support the `file` and `path` parameter types on
Kotlin native [1], we have to use a String parameter.
This means that the `--parameter-file` option doesn't offer any
tab-completion options.

Fix this by explicitly defining paths as completion candidates.

[1]: https://ajalt.github.io/clikt/advanced/#jvm

### DIFF
--- a/cli/src/commonMain/kotlin/StartCommand.kt
+++ b/cli/src/commonMain/kotlin/StartCommand.kt
@@ -20,6 +20,7 @@
 package org.eclipse.apoapsis.ortserver.cli
 
 import com.github.ajalt.clikt.command.SuspendingCliktCommand
+import com.github.ajalt.clikt.completion.CompletionCandidates
 import com.github.ajalt.clikt.core.Context
 import com.github.ajalt.clikt.core.terminal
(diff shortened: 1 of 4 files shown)


cli/src/commonMain/kotlin/StartCommand.kt
```diff
@@ -80,7 +81,8 @@ class StartCommand : SuspendingCliktCommand(name = "start") {
             "--parameters-file",
             envvar = "OSC_RUNS_START_PARAMETERS_FILE",
             help = "The path to a JSON file containing the run configuration " +
-                    "(see https://eclipse-apoapsis.github.io/ort-server/api/post-ort-run)."
+                    "(see https://eclipse-apoapsis.github.io/ort-server/api/post-ort-run).",
+            completionCandidates = CompletionCandidates.Path
```
sschuberth, line 85: Could you briefly explain in the commit message why it was not working before, and what setting `completionCandidates` changes to fix this?

sschuberth (approved):

PR #2790, opened 2025-05-23
fix(ui): Do not show the main package's purl in the shortest path

sschuberth:
If the user configured to show purls instead of identifiers, the previous implementation erronously used the main package's purl for all packages on the shortest path.

As the shortest paths endpoint currently only returns identifiers, no purls, create purls based on the identifier as a work-around. Note that this will yield wrong data in case the purl of a package on the path has been curated. However, this is better than always showing wrong purls.


mnonnenmacher (approved):

sschuberth:
@mnonnenmacher, I just discovered another bug for which I'll add a fix to this PR, stay tuned!

mnonnenmacher (approved):

PR #2791, opened 2025-05-23
refactor(dao): Move queries from table classes into query classes

mnonnenmacher:
Add a new `Query` interface to define reusable queries and migrate the queries from table classes to implementations of the new interface.

This provides better separation of concerns as the table classes now only describe the table structure, and improves cohesion as closely related query code is now grouped within the Query classes.


dao/src/main/kotlin/queries/GetAnalyzerConfigurationForAnalyzerRun.kt
```diff
@@ -0,0 +1,105 @@
...
+import org.eclipse.apoapsis.ortserver.model.runs.AnalyzerConfiguration
+import org.eclipse.apoapsis.ortserver.model.runs.PackageManagerConfiguration
+
+import org.jetbrains.exposed.sql.selectAll
+
+/**
+ * A query to get the [AnalyzerConfiguration] for a given [analyzerRunId]. Returns `null` if the configuration is not
+ * found.
+ */
+class GetAnalyzerConfigurationForAnalyzerRun(
```
oheger-bosch, line 40: Since we can expect a large number of classes, should we already start to group them in sub packages? E.g. an `analyzer` package for Analyzer-related queries.

dao/src/main/kotlin/repositories/analyzerrun/AnalyzerRunsTable.kt
```diff
@@ -63,12 +68,12 @@ object AnalyzerRunsTable : LongIdTable("analyzer_runs") {
     private fun loadAnalyzerRun(resultRow: ResultRow): AnalyzerRun {
         val analyzerRunId = resultRow[id].value
         val analyzerJobId = resultRow[analyzerJobId].value
-        val environment = checkNotNull(EnvironmentsTable.getById(resultRow[environmentId].value))
-        val config = checkNotNull(AnalyzerConfigurationsTable.getByAnalyzerRunId(analyzerRunId))
-        val projects = ProjectsAnalyzerRunsTable.getProjectsByAnalyzerRunId(analyzerRunId)
-        val packages = PackagesAnalyzerRunsTable.getPackagesByAnalyzerRunId(analyzerRunId)
-        val ortRunId = checkNotNull(AnalyzerJobsTable.getOrtRunIdById(analyzerJobId))
-        val issues = OrtRunsIssuesTable.getIssuesByOrtRunId(ortRunId, AnalyzerRunDao.ISSUE_WORKER_TYPE)
+        val environment = checkNotNull(GetEnvironment(resultRow[environmentId].value).execute())
```
oheger-bosch, line 71: I would it find slightly more readable if the query classes ended on `Query`, e.g. `GetEnvironmentQuery(...).execute()` rather than `GetEnvironment(...).execute()`. But maybe this is just personal taste.

dao/src/main/kotlin/repositories/analyzerrun/AnalyzerRunsTable.kt
```diff
@@ -63,12 +68,12 @@ object AnalyzerRunsTable : LongIdTable("analyzer_runs") {
     private fun loadAnalyzerRun(resultRow: ResultRow): AnalyzerRun {
```
oheger-bosch, line 68: To be consequent, shouldn't there by a `GetAnalyzerRun` query?

mnonnenmacher:
> Just as a side note, I am bit surprised that this refactoring does not require any changes on tests. So, there are no tests for database queries in isolation?

Correct, I did not add tests during the previous refactoring because the function from `DaoAnalyzerRunRepository` covered all queries, but I would agree that adding tests for individual queries would make sense, and actually it's easy to do now.

PR #2846, opened 2025-05-30
ui: Change document title for repository and run levels

Etsija:
Screenshot from 2025-05-30 12-00-12

Screenshot from 2025-05-30 12-00-27

### DIFF
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -120,9 +120,14 @@ export const Header = () => {
     const organization = organizationMatch?.context.breadcrumbs.organization;
     const product = productMatch?.context.breadcrumbs.product;
     const repository = repoMatch?.context.breadcrumbs.repo;
+    // Extract only the repository name from the full URL to show in the title.
+    const repoName = repository?.split('/').pop()?.split('.git')[0];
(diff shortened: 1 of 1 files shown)


ui/src/components/header.tsx
```diff
@@ -120,17 +120,22 @@ export const Header = () => {
     const organization = organizationMatch?.context.breadcrumbs.organization;
     const product = productMatch?.context.breadcrumbs.product;
     const repository = repoMatch?.context.breadcrumbs.repo;
+    // Extract only the repository name from the full URL to show in the title.
+    const repoName = repository?.split('/').pop()?.split('.')[0];
```
sschuberth, line 124: I wonder whether we should be more strict about the `.split('.')[0]` part and only remove ".git" suffix to prevent names with dots to be truncated.

Etsija: What about Gitrepo, Mercurial, Subversion names, do they follow a pattern?

sschuberth: No, they usually do not have such URL suffixes.

sschuberth (approved):

PR #2861, opened 2025-06-02
ui: Add support for a tailored product name

Etsija:
Screenshot from 2025-06-02 11-54-59

Screenshot from 2025-06-02 11-55-49


sschuberth:
We're becoming a bit too fine-granular regarding the menu structure for my taste. How about combining all of "Product Name", "Home Icon" and "Footer" under a new "Branding" menu item?

Maybe omit "Foot" from there as it requires a rather large text field, but at least "Product Name" and "Home Icon" should be grouped, IMO, as otherwise their right-hand content panel only shows a single setting each.

ui/src/routes/admin/content-management/product-name/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+    });
+    return;
+  }
+
+  return (
+    <Card>
+      <CardHeader>
+        <CardTitle>Product Name</CardTitle>
+        <CardDescription>
+          Customise the product name. The toggle specifies whether the default
```
lamppu, line 138: Itty bitty nitty gritty: Customize :grinning:

Etsija: A bit of a petrol/gasoline opinion, but OK. :smirk:

sschuberth: It's a BE vs. AE thing, and we should stick to AE.

Etsija:
> > I'm not sure I get what you want. I'll provide the side menu here for reference. So, add another top-level section "Branding" to the menu?
> 
> No. _Combine_ "Product Name", "Home Icon" and "Footer" (maybe) so that it looks like
> 
> ```
> Content Management
>   Branding
>   Colors
>   (Footer)
> ```
> 
> and if you click in "Branding", the right-hand content panel shows all of "Product Name", "Home Icon" and (maybe) "Footer" cards below each other.

Well that's not possible with the current side menu component I think, or what do you say @lamppu ?  It has no support for opening submenus. There is a new `shadcn-ui` component for side menus which I demoed to you a while ago, but I don't think it's in the context of this small PR. Maybe do the component update separately?

Etsija:
Ah, now I get it, I misunderstood you @sschuberth. Now I understand.

Issue #2878, opened 2025-06-03
Issues, rule violations endpoints should return whether they were triggered by a package or project


Etsija:
As discussed in #2340, showing PURLs / ORT IDs (user preference, configured in the UI's Preferences section) for issues and rule violations found in an ORT run need a way to know whether the issue / rule violation was triggered by a package or a project. Example:

![Image](https://github.com/user-attachments/assets/225296a9-e8fe-4301-b095-44833c34a97e)

Here the first three rule violations were caused by packages, and the five latter by projects.

When the endpoints return this project/package information, we will probably add a column telling whether the item was coming from a project or package, and then we can also apply logic to decide whether we want to show a PURL or an ORT ID for the corresponding item.

sschuberth:
Quoting our internal notes for this issue:

> This was promised to be done soon by Bosch people in the alignment meeting 3.6.2025.

@mnonnenmacher, is this something easy that could be added by you this week still?

lamppu:
Closed by https://github.com/eclipse-apoapsis/ort-server/pull/3425.

PR #2879, opened 2025-06-03
ui: Show PURL for vulnerabilities if so configured

Etsija:
This PR is a temporary means to advance towards the goal of showing either PURLs or ORT IDs everywhere throughout the UI. Once the endpoints return PURLs (and possibly, whether the item was triggered by a project or package), we will follow-up with a proper implementation in the UI.

Relates to #2176, #2340, #2878.

Screenshot from 2025-06-03 12-30-41

Screenshot from 2025-06-03 12-30-55


Etsija:
@mnonnenmacher once again I am stupidified by why commit-lint check fails: I have checked my commit message three times and verified that in the editor, the sign-off is preceded with an empty line.

PR #2893, opened 2025-06-05
feat(api): Add applicable resolutions to list of rule violations

MarcelBochtler:


services/ort-run/src/main/kotlin/RuleViolationService.kt
```diff
@@ -23,42 +23,71 @@ import org.eclipse.apoapsis.ortserver.dao.dbQuery
...
-import org.eclipse.apoapsis.ortserver.dao.repositories.evaluatorrun.RuleViolationDao
 import org.eclipse.apoapsis.ortserver.dao.repositories.evaluatorrun.RuleViolationsTable
-import org.eclipse.apoapsis.ortserver.dao.tables.shared.IdentifiersTable
-import org.eclipse.apoapsis.ortserver.dao.utils.listCustomQuery
 import org.eclipse.apoapsis.ortserver.model.CountByCategory
 import org.eclipse.apoapsis.ortserver.model.Severity
 import org.eclipse.apoapsis.ortserver.model.runs.OrtRuleViolation
 import org.eclipse.apoapsis.ortserver.model.util.ListQueryParameters
 import org.eclipse.apoapsis.ortserver.model.util.ListQueryResult
+import org.eclipse.apoapsis.ortserver.model.util.OrderDirection
```
sschuberth, line 33: Commit message: Drop "us" for sticking to passive voice.

services/ort-run/src/main/kotlin/RuleViolationService.kt
```diff
@@ -23,42 +23,71 @@ import org.eclipse.apoapsis.ortserver.dao.dbQuery
...
-
-                join.select(
-                    RuleViolationsTable.columns + IdentifiersTable.columns
-                ).where { EvaluatorJobsTable.ortRunId eq ortRunId }
+    ): ListQueryResult<OrtRuleViolation> {
+        val ortRun = ortRunService.getOrtRun(ortRunId)
+
+        if (ortRun == null) {
+            return ListQueryResult(emptyList(), parameters, 0)
+        }
```
sschuberth, line 53: IDE inspection (and myself 😄) suggests to combine this to:

```
val ortRun = ortRunService.getOrtRun(ortRunId) ?: return ListQueryResult(emptyList(), parameters, 0)
```

services/ort-run/src/main/kotlin/RuleViolationService.kt
```diff
@@ -92,10 +96,29 @@ class RuleViolationService(private val db: Database, private val ortRunService:
...
+    ): List<OrtRuleViolation> = when (ruleViolationFilter.resolved) {
+        true -> {
+            filter { violation ->
+                resolutions.any { it.matches(violation.mapToOrt()) }
+            }
+        }
+
+        false -> {
+            filter { violation ->
+                !resolutions.any { it.matches(violation.mapToOrt()) }
```
sschuberth, line 115: Prefer `none` over `!any`.

api/v1/model/src/commonMain/kotlin/RuleViolation.kt
```diff
@@ -32,3 +32,15 @@ data class RuleViolation(
...
+/**
+ * Filters to apply when querying for rule violations.
+ */
+@Serializable
+data class RuleViolationFilters(
+    /**
+     * Filter to only return resolved rule violations. Null if both, resolved an unresolved rule violations should be
+     * returned.
+     */
+    val resolved: Boolean? = null
```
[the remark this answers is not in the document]
MarcelBochtler: After implementing this with an enum, I think the `Boolean?` design is the cleanest REST API design.

Boolean?:
```
?resolved=true
?resolved=false
``` 

ResolutionStatus enum:
```
?resolutionStatus=resolved
?resolutionStatus=unresolved
?resolutionStatus=undefined
```

For the `ResolutionStatus` implementation, the possible options have to be documented, while they are self-explanatory for `Boolean?`.
For REST API users, `Boolean?` is not a ternary boolean logic, because `null` is only used for the default. They will never set `?resolved=null`.

I could work around this issue by defining an enum with the values `TRUE`, `FALSE` and `UNDEFINED`. But this seems overengineered and wrong.

Therefore, I would like to keep / revert to `Boolean?`, what do you think?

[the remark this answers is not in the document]
MarcelBochtler: I decided to revert to `Boolean?`, as this also solves the issue that users might expect these filters to work as well, which is not intended to support:
```
?resolutionStatus=resolved,unresolved
```

So IMO only providing `true`, `false` and no filter at all is the most intuitive thing to support.

oheger-bosch (approved):

PR #2898, opened 2025-06-06
ui: Sibling navigation in breadcrumb

Etsija:
This PR introduces a "shortcut" navigation to sibling organizations, products, repositories and runs to the breadcrumb. 

Screenshot from 2025-06-26 11-22-39

Screenshot from 2025-06-26 11-22-51

Screenshot from 2025-06-26 11-23-05

@sschuberth I decided to keep the down carets in the breadcrumb in case there are siblings available on server, to simplify the logic of this feature combined with the earlier feature of not showing upper-level entities in breadcrumb for technical users that do not have access rights to upper-level entities.

I also tested that the breadcrumb works as intended also for the said technical users, which only have access to some lower-level entities such as product/repository.

1. This user only has access rights to one particular product (Semver).  No sibling products are shown in the breadcrumb, as intended.

Screenshot from 2025-06-26 11-30-40

2. This user only has access rights to one particular repository (harmony-access-point).  No sibling repositories are shown in the breadcrumb, as intended.

Screenshot from 2025-06-24 12-20-57

Note on performance: extra care has been taken to not introduce a visible degradation into navigation performance:
- stale time of 2 hours to all "entity" queries, to retrieve the entities (org, prod, repo) efficiently from cache instead of an endpoint
- clicking open a "sibling" item which opens up a dropdown only runs the query for getting the list of that particular entity (and only if stale time is passed)

### DIFF
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -27,6 +27,7 @@ import { z } from 'zod';
 import { useAdminServiceGetApiV1AdminConfigByKey } from '@/api/queries';
 import homeIcon from '@/assets/home-icon.svg';
 import { ModeToggle } from '@/components/mode-toggle';
+import { Siblings } from '@/components/siblings';
 import { Avatar, AvatarFallback } from '@/components/ui/avatar';
 import { Form, FormControl, FormField, FormItem } from '@/components/ui/form';
 import {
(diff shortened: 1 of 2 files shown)


Etsija:
> > No, the items in the dropdown list are siblings to Semver, ie. they are products indeed, not organizations.
> 
> So how can I switch to an org-sibling of "My test organization" then?

I thought it was obvious: by clicking the down-caret after the org. name.

PR #2927, opened 2025-06-10
fix(services): Respect group prefix when querying Keycloak members

wkl3nk:
Fix user membership queries to include the Keycloak group prefix when listing users in _Organization_, _Product_, or _Repository_ groups. This prevents incorrect results when multiple ORT Server instances share the same Keycloak realm.

This bug came up in an environment where a productive ORT Server instance and a development ORT Server instance shared the same Keycloak realm. In the ORT Server UI's User page, users showed up in the list that actually did not have an group membership on the production environment.


sschuberth:
FYI @lamppu, maybe of interest to you.

mnonnenmacher (approved):

PR #2947, opened 2025-06-12
fix(analyzer): Set correct attributes for the `settings.xml` root ele…

oheger-bosch:
…ment

The declaration of the namespace and the schema in the generated document was incomplete. This did not cause issues with Maven; obviously it ignores such declarations when loading the settings file. However, some text editors were complaining when opening the file. To fix this, use the declaration from [1].

[1]: https://maven.apache.org/settings.html


workers/common/src/main/kotlin/common/env/MavenSettingsGenerator.kt
```diff
@@ -37,9 +37,11 @@ class MavenSettingsGenerator : EnvironmentConfigGenerator<MavenDefinition> {
...
         private const val TARGET_NAME = ".m2/settings.xml"
 
-        /** The schema location for Maven's settings.xml. */
-        private const val SCHEMA = "\"http://maven.apache.org/SETTINGS/1.1.0 " +
-                "http://maven.apache.org/xsd/settings-1.1.0.xsd\""
+        /** The attributes of the `settings` root element. */
+        private const val ROOT_ATTRIBUTES = "xmlns=\"http://maven.apache.org/SETTINGS/1.0.0\" " +
+                "xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" " +
+                " xsi:schemaLocation=\"http://maven.apache.org/SETTINGS/1.0.0 " +
+                "https://maven.apache.org/xsd/settings-1.0.0.xsd\""
```
MarcelBochtler, line 44: This changes the expected settings version to 1.0.0.
[Here](https://maven.apache.org/ref/3.2.1/maven-settings/settings.html) is an example with the version 1.1.0.

MarcelBochtler (changes requested):

PR #2951, opened 2025-06-12
feat(services): Mark secrets as deleted instead of removing them

wkl3nk:
Secrets are now soft-deleted by setting a deletion flag instead of being physically removed from the database. This preserves foreign key references from InfrastructureService entities while preventing deleted secrets from being used via the ORT Server UI.

Before this change, it was not possible to delete secrets when they were at least used once by one scan.


oheger-bosch:
I am not sure about this approach for multiple reasons.
* It increases complexity, as now everywhere a condition has to be checked. 
* It is strange to have a special soft-delete mode just for secrets, while other entities are deleted immediately.
* Does this also affect the relation to secret values? Are the values of such secrets deleted from the underlying storage while the entities still exist?

PR #2957, opened 2025-06-13
feat(services): Add support for dynamic infrastructure services

wkl3nk:
Persist **Infrastructure Service Declarations**. These are
* created dynamically on the fly from _.ort.env.yml_ files in the source repository and associated
directly to the current ORT run. 
* referencing secrets **by name** (unlike existing infrastructure services which have a foreign key relationship to secrets). This allows that Secrets can be deleted, even when referenced. An ORT Run will fail if Secrets cannot be resolved by name at runtime.
* using some logic to look up the Secrets by name, searching on levels Repository --> Product --> Organization and writing warnings to the log if there is ambiguity detected
* not having attributes _organization_ or _product_, because the related ORT run specifies to which Organization or Product they belong to.

In contract to this, the existing (static) Infrastructure Services
* are managed by the user using the REST API or the ORT Server UI, 
* are assigned to a hierarchy level
* have strong references to their Secrets (database foreign key contraint). Not possible to delete a Secret as long it is still references in a (static) Infrastructure Service

### DIFF
--- a/workers/common/src/main/kotlin/common/env/EnvironmentService.kt
+++ b/workers/common/src/main/kotlin/common/env/EnvironmentService.kt
@@ -28,7 +28,15 @@ import kotlinx.coroutines.withContext
 
 import org.eclipse.apoapsis.ortserver.model.EnvironmentConfig
 import org.eclipse.apoapsis.ortserver.model.InfrastructureService
+import org.eclipse.apoapsis.ortserver.model.InfrastructureServiceDeclaration
+import org.eclipse.apoapsis.ortserver.model.OrganizationId
+import org.eclipse.apoapsis.ortserver.model.OrtRun
+import org.eclipse.apoapsis.ortserver.model.ProductId
+import org.eclipse.apoapsis.ortserver.model.RepositoryId
+import org.eclipse.apoapsis.ortserver.model.Secret
+import org.eclipse.apoapsis.ortserver.model.repositories.InfrastructureServiceDeclarationRepository
 import org.eclipse.apoapsis.ortserver.model.repositories.InfrastructureServiceRepository
+import org.eclipse.apoapsis.ortserver.model.repositories.SecretRepository
 import org.eclipse.apoapsis.ortserver.workers.common.context.WorkerContext
 import org.eclipse.apoapsis.ortserver.workers.common.env.config.EnvironmentConfigLoader
 import org.eclipse.apoapsis.ortserver.workers.common.env.config.ResolvedEnvironmentConfig
@@ -51,6 +59,12 @@ private val logger = LoggerFactory.getLogger(EnvironmentService::class.java)
 class EnvironmentService(
     /** The repository for accessing infrastructure services. */
     private val infrastructureServiceRepository: InfrastructureServiceRepository,
+
+    /** The repository for accessing dynamic infrastructure services. */
+    private val infrastructureServiceDeclarationRepository: InfrastructureServiceDeclarationRepository,
+
+    /** The repository for secrets. This is used to resolve secret references. */
+    private val secretRepository: SecretRepository,
 
     /** A collection with the supported generators for configuration files. */
     private val generators: Collection<EnvironmentConfigGenerator<*>>,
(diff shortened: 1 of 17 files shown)


core/src/test/kotlin/api/OrganizationsRouteIntegrationTest.kt
```diff
@@ -988,8 +988,8 @@ class OrganizationsRouteIntegrationTest : AbstractIntegrationTest({
                         service.name,
                         service.url,
                         service.description,
-                        service.usernameSecret.name,
-                        service.passwordSecret.name,
+                        service.usernameSecret?.name ?: "",
+                        service.passwordSecret?.name ?: "",
```
sschuberth, line 992: Consider using `.orEmpty()` here and below.

oheger-bosch:
Could you maybe elaborate a bit why you chose this approach and ignored the comment [here](https://github.com/eclipse-apoapsis/ort-server/pull/2951#issuecomment-2969159076)?

mnonnenmacher:
> Could you maybe elaborate a bit why you chose this approach and ignored the comment [here](https://github.com/eclipse-apoapsis/ort-server/pull/2951#issuecomment-2969159076)?

Independent of the effort I would also have preferred to reference secrets by name, because that better resembles how I as a user would expect the feature to work.

oheger-bosch:
I like the new approach, but I am not sure about the naming. The "dynamic" is not really meaningful. These services are declared in the repository in the environment configuration file, so maybe something like `RepositoryInfrastructureService` or `EnvironmentInfrastructureService`?

workers/common/src/main/kotlin/common/env/EnvironmentService.kt
```diff
@@ -164,11 +178,38 @@ class EnvironmentService(
...
-     * which has initialized the required information.
+     * the database. Given the secret names in [DynamicInfrastructureService.usernameSecretName] and
+     * [DynamicInfrastructureService.passwordSecretName], find these secrets and continue to process them.
+     * This function can be used by workers running after the Analyzer, which has initialized the required information.
      */
     suspend fun setupAuthenticationForCurrentRun(context: WorkerContext) {
-        setupAuthentication(context, infrastructureServiceRepository.listForRun(context.ortRun.id))
+        val dynamicInfrastructureServices = dynamicInfrastructureServiceRepository.listForRun(context.ortRun.id)
+
+        val infrastructureServices = dynamicInfrastructureServices.map { service ->
```
oheger-bosch, line 188: Doing this resolving one by one may be inefficient, also if there are services that share secrets (e.g. for username). In the worker context, functionality is available to resolve secrets in parallel - maybe something similar could be done here as well?

workers/common/src/main/kotlin/common/env/EnvironmentService.kt
```diff
@@ -164,11 +178,38 @@ class EnvironmentService(
...
+                service.usernameSecretName,
+                context.ortRun,
+                service.name
+            ) ?: error("Username secret ${service.usernameSecretName} not found for service '${service.name}'.")
+
+            val passwordSecret = resolveSecretByName(
+                service.passwordSecretName,
+                context.ortRun,
+                service.name
+            ) ?: error("Password secret ${service.passwordSecretName} not found for service '${service.name}'.")
```
oheger-bosch, line 199: How will such an error appear to the user? Is it clear what the problem is? This could be a frequent issue and we should try to reduce requests to support.

workers/common/src/main/kotlin/common/env/EnvironmentService.kt
```diff
@@ -177,9 +218,72 @@ class EnvironmentService(
...
+            secretRepository.getByIdAndName(RepositoryId(ortRun.repositoryId), secretName)
+        val secretProductLevel =
+            secretRepository.getByIdAndName(ProductId(ortRun.productId), secretName)
+        val secretOrganizationLevel =
+            secretRepository.getByIdAndName(OrganizationId(ortRun.organizationId), secretName)
+
+        secretRepositoryLevel?.let {
+            secretProductLevel?.let {
+                secretOrganizationLevel?.let {
+                    logger.warn(
```
oheger-bosch, line 246: It is certainly helpful to log from which level the secret was used, but I would not use warning level. Overriding a secret may be a normal use case.

model/src/commonMain/kotlin/DynamicInfrastructureService.kt
```diff
@@ -0,0 +1,82 @@
...
+    val name: String,
+
+    /** The URL of this service. */
+    val url: String,
+
+    /** The name of the [Secret] that contains the username of the credentials for this infrastructure service. */
+    val usernameSecretName: String,
+
+    /** The name of the [Secret] that contains the password of the credentials for this infrastructure service. */
+    val passwordSecretName: String,
```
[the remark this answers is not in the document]
wkl3nk: Oh thanks. I was not aware of this. I think I will adapt that immediately.

workers/common/src/main/kotlin/common/env/EnvironmentService.kt
wkl3nk: I think that is not happening **very infrequently**. This only happens if **during the run a secret is deleted**, and a new worker won't find it. Most of the time, already in the Analyzer worker that merges together all kind of environmental entries you will get an Issue like this:

image

workers/common/src/test/kotlin/common/env/EnvironmentServiceTest.kt
```diff
@@ -513,10 +650,157 @@ class EnvironmentServiceTest : WordSpec({
             result.environmentVariables shouldContainExactlyInAnyOrder listOf(variable2)
         }
     }
+
+    "resolveServiceByName" should {
```
oheger-bosch, line 654: Nit: This should probably be `resolveSecretByName`.

oheger-bosch (approved):
I have one suggestion and a minor nit. Already approving and leaving it to you if you want to address them.

Issue #2901, opened 2025-06-06
Move admin config from job configs to config file


mnonnenmacher:
The job configurations currently contain several properties which should only be configurable by admins. These properties should be removed from the job configs and instead be read from a new config file (e.g., `ort-server.config.yml`).

To be able to provide different configurations for different organizations, for example, a different evaluator rules, the config file must support profiles (or "rule sets" as an alternative name).

Affected properties are:

* `EvaluatorJobConfig`
  * All existing options refer to file paths in the config repository. All of them should be removed and be replaced with a single `profile` (or `ruleSet`) property which selects the config from the config file.
* `NotifierJobConfig`
  * All existing options should only be configurable for admins, except for the recipient list for the mail notifier. They should be replaced by new options:
    * `emailRecipient` to set the recipients of email notifications.
    * `ruleSet` (or `profile`) to select which resolutions file should be used (it must be consistent with the one used for the evaluator and reporter jobs).
* `ReporterJobConfig`
  * The problematic properties are: `assetDirectories`, `assetFiles`, `copyrightGarbageFile`, `customLicenseTextDir`, `howToFixTextProviderFile`, `licenseClassificationsFile`, `nameMappings`, and `resolutionsFile`. All of them should be removed and replaced with a `profile` (or `ruleSet`) property.
  * An unsolved issue is currently how users can select templates for template reporters, as they cannot know which template files are available. This needs to be addressed after the implementation of plugin templates in the plugin manager.
* `ScannerJobConfig`
  * The problematic properties are: `detectedLicenseMapping`, `ignorePatterns`, `sourceCodeOrigins`. For us it would be fine to move all of them to the config file, but I would like to have feedback from Double Open (@sschuberth) to know if they would like their users to be able to configure those.

All options that are removed from the job configs must also be removed from the UI to trigger a new run (if they are available there).

In the long-term it could be considered to move at least part of this configuration to the admin UI.

sschuberth:
> In the long-term it could be considered to move at least part of this configuration to the admin UI.

However, I believe this should not be done long-term, but short-term. For us, one of the major reasons to use ORT Server of over ORT is to make configuration easier and more accessible. A crucial element of this is to make configuration accessible in the UI.

While I acknowledge the necessity to split admin from non-admin config, not having the config available in the UI anymore (as it was previously in the run configuration) is a severe step backward for us.

Etsija:
[...]

PR #2987, opened 2025-06-18
Introduce admin configuration

oheger-bosch:
This PR starts with moving properties from job configurations to a new admin configuration as described in https://github.com/eclipse-apoapsis/ort-server/issues/2901. It does this on all layers for the Evaluator configuration to have an initial admin configuration available and a proof of concept for this approach. Similar exercises then need to be done for the other workers.

### DIFF
--- a/services/admin-config/src/main/kotlin/RuleSet.kt
+++ b/services/admin-config/src/main/kotlin/RuleSet.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.services.config
+
+/**
+ * A data class that represents the configuration of a rule set.
+ *
+ * A rule set consists of a number of paths to files that are used by the Evaluator and partly by the Reporter. The
+ * paths are passed to the _config file provider_ to obtain the actual configuration files.
+ */
+data class RuleSet(
+    /** The path to the copyright garbage file. */
+    val copyrightGarbageFile: String,
+
+    /** The path to the license classifications file. */
+    val licenseClassificationsFile: String,
+
+    /** The path to the resolutions file. */
+    val resolutionsFile: String,
+
+    /** The path to the files with rules to use for the evaluation. */
+    val evaluatorRules: String
+)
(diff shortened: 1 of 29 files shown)


services/admin-config/src/test/kotlin/AdminConfigServiceTest.kt
```diff
@@ -0,0 +1,260 @@
...
+                service.loadAdminConfig(context, ORGANIZATION_ID).getRuleSet(nonExistentRuleSetName)
+            }
+
+            exception.message shouldContain nonExistentRuleSetName
+            exception.message shouldContain "No rule set"
+        }
+
+        "set undefined properties in rule sets from the default rule set" {
+            val config = """
+                    defaultRuleSet {
```
mnonnenmacher, line 148: For the user it might be nicer to have this as a rule set with the name "default" in the list below, and to treat it like any other rule set just with a reserved name, what do you think?

workers/config/src/main/kotlin/ConfigValidator.kt
```diff
@@ -92,9 +107,39 @@ class ConfigValidator private constructor(private val context: WorkerContext) :
...
+            validationIssues += Issue(
+                Clock.System.now(),
+                ADMIN_CONFIG_VALIDATION_SOURCE,
+                "Invalid rule set '$ruleSet'. " +
+                        "Available rule sets are: ${adminConfig.ruleSetNames.joinToString(", ")}.",
+                Severity.ERROR
+            )
+        }
+
+        return takeIf { validationIssues.isEmpty() } ?: ConfigValidationResultFailure(issues + validationIssues)
```
mnonnenmacher, line 143: Should this also check if the files defined in the rule set do actually exist?

services/admin-config/src/main/kotlin/RuleSet.kt
```diff
@@ -0,0 +1,40 @@
...
+    val copyrightGarbageFile: String,
+
+    /** The path to the license classifications file. */
+    val licenseClassificationsFile: String,
+
+    /** The path to the resolutions file. */
+    val resolutionsFile: String,
+
+    /** The path to the files with rules to use for the evaluation. */
+    val evaluatorRules: String
```
mnonnenmacher, line 39: Have you considered adding the how-to-fix text provider script as well?

services/admin-config/src/test/kotlin/AdminConfigServiceTest.kt
oheger-bosch: The default rule set is special, as its properties are inherited by other rule sets unless they override them. I thought this would be more obvious if it was declared in a more prominent place in the configuration. If you have a longer list of rule sets and the default is just one of them, it is hard to see what are the default values.

services/admin-config/src/main/kotlin/RuleSet.kt
oheger-bosch: No, this did no occur to me, since this script was not part of the original evaluator job configuration. Can we add it later, when dealing with the reporter configuration?

services/admin-config/src/test/kotlin/AdminConfigServiceTest.kt
mnonnenmacher: I have no clear preference, it just came to my mind during the review.

mnonnenmacher (approved):

Etsija:
@oheger-bosch @mnonnenmacher I have some issues with this merged PR:

1. Where are some examples of this ruleset file? I don't see anything added to documentation.

Screenshot from 2025-06-26 07-47-46

2. Why are we adding yet another configuration file to ORT Server? I was away, so couldn't take part to the discussion, but as we do have the UI for the server, shouldn't these admin configurations rather be set in the Administration section?

3. The Evaluator configuration is now just empty in the UI, which is certainly not good UX.

Screenshot from 2025-06-26 07-48-08

oheger-bosch:
> > which should explain the reasoning behind the way it was implemented.
> 
> Still, I guess breaking the UI was not intended 😉

Hey, hey - what does "breaking the UI" mean? For my first UI contribution, I think the results are not that bad.

sschuberth:
> Hey, hey - what does "breaking the UI" mean? For my first UI contribution, I think the results are not that bad.

More seriously, how are we going to proceed here? @oheger-bosch, are you going to fixup the UI, or should @Etsija do a follow-up?

PR #3029, opened 2025-06-25
fix(service): Prevent duplicating infrastructure-service

kamil-bielecki-bosch:
This commit adds validation for infrastructure services creation that prevents adding infrastructure service with duplicate name for organization.

### DIFF
--- a/core/src/test/kotlin/api/OrganizationsRouteIntegrationTest.kt
+++ b/core/src/test/kotlin/api/OrganizationsRouteIntegrationTest.kt
@@ -880,6 +880,40 @@ class OrganizationsRouteIntegrationTest : AbstractIntegrationTest({
                 ).shouldBeNull()
             }
         }
+
+        "respond with 'Conflict' if service with same name and orgId already exists" {
+            integrationTestApplication {
+                val orgId = createOrganization().id
+
+                val userSecret = createSecret(orgId, path = "user", name = "user")
+                val passSecret = createSecret(orgId, path = "pass", name = "pass")
+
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/api/OrganizationsRoute.kt
```diff
@@ -211,6 +211,16 @@ fun Route.organizations() = route("organizations") {
                 val organizationId = call.requireIdParameter("organizationId")
                 val createService = call.receive<CreateInfrastructureService>()
 
+                infrastructureServiceService.getByOrganizationAndName(organizationId, createService.name.trim())
```
oheger-bosch, line 214: I think this validation would be better done in the `InfrastructureServiceService`, in the create operation. It is then not necessary to expose another function to do the name check. If the service detects a duplicate name, it can throw a `UniqueConstraintException` which gets mapped automatically to response status 409 Conflict.

kamil-bielecki-bosch: That's a good idea, but `UniqueConstraintException` is SQL-ish tailored exception, that:
- IMHO should not be used out of DAO/repository.
- Requires 2 parameters, where second one is "original exception" that is intended to be DB-realted one.

So, there I can see 2 solutions:
1. Create variant of `UniqueConstraintException` that requires only message and carry on.
2. Create another exception service-related, that can be thrown from service, that will map to same 409 Conflict error handler.

oheger-bosch: I am fine with both options.

oheger-bosch (approved):

PR #3064, opened 2025-06-30
Notifier admin config

oheger-bosch:
This PR reworks the job configuration for the Notifier to move the properties that cannot be set by end users to the admin configuration. Refer to the single commits for further information.


workers/notifier/src/main/kotlin/notifier/NotifierRunner.kt
```diff
@@ -20,7 +20,8 @@
 package org.eclipse.apoapsis.ortserver.workers.notifier
```
bs-ondem, line 20: Could you update the commit message to reflect that this change is not only about using the mail and jira configuration the `AdminConfig`, but also about using the `notifierRules` and `rulesSet` from the `AdminConfig` to read the corresponding files? It's not clear from the commit message.

bs-ondem (changes requested):

PR #3071, opened 2025-07-01
feat(ui): Add repository ID to repository overview page

Etsija:
Screenshot from 2025-07-03 11-32-57

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/runs/index.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/runs/index.tsx
@@ -78,23 +78,32 @@ const RepositoryRunsComponent = () => {
 
   return (
     <div className='flex flex-col gap-4'>
-      <Card>
-        <CardHeader>
-          <CardTitle>
-            <span className='font-normal'>
-              {getRepositoryTypeLabel(repo.type)} repository
-            </span>{' '}
-            <Link
-              className='font-semibold break-all hover:text-blue-400 hover:underline'
-              to={repo.url}
-              target='_blank'
-            >
-              {repo.url}
-            </Link>
-          </CardTitle>
-          <CardDescription>{repo.description}</CardDescription>
-        </CardHeader>
-      </Card>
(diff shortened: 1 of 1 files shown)


sschuberth:
I know this is just a quick fix to address @hanna-modica's request, but I'm still wondering whether we could make better use of screen space, e.g. by using the horizontal space to the right of the repository URL.

Just thinking out aloud: I could imagine to split the top-level card in two, keeping a left card with the current information (URL and description), and introducing a right card with the ID, and probably *some* other meaningful info. Examples could be the number of failed vs. successful runs, similar to like one of [these Jenkins plugins](https://stackoverflow.com/a/25331828/1127485) show.

Etsija:
> I know this is just a quick fix to address @hanna-modica's request, but I'm still wondering whether we could make better use of screen space, e.g. by using the horizontal space to the right of the repository URL.
> 
> Just thinking out aloud: I could imagine to split the top-level card in two, keeping a left card with the current information (URL and description), and introducing a right card with the ID, and probably _some_ other meaningful info. Examples could be the number of failed vs. successful runs, similar to like one of [these Jenkins plugins](https://stackoverflow.com/a/25331828/1127485) show.

Sure, I can work on this and change the PR title accordingly when I'm done.

Etsija:
@mnonnenmacher do you have an idea for what other info than the repository ID to put to the top-right Card in the overview page that @sschuberth refers to?  I want to avoid putting any data that involves running the "Get all runs of a repository" query with "infinite" limit, as that would probably have a negative impact to the UI performance.

sschuberth (approved):

mnonnenmacher:
> @mnonnenmacher do you have an idea for what other info than the repository ID to put to the top-right Card in the overview page that @sschuberth refers to? I want to avoid putting any data that involves running the "Get all runs of a repository" query with "infinite" limit, as that would probably have a negative impact to the UI performance.
> 
> We are unfortunately missing a "statistics" endpoint for the repositories, so in practise, any data about the runs would involve running the above query, which is pretty consuming.

No, currently I would not know what else to add there, because I agree that a dedicated endpoint would be required to return som aggregated data.

PR #3073, opened 2025-07-01
feat(api): Add applicable issue resolutions to `/issues` route

MarcelBochtler:

### DIFF
--- a/services/hierarchy/src/main/kotlin/IssueService.kt
+++ b/services/hierarchy/src/main/kotlin/IssueService.kt
@@ -127,20 +127,19 @@ internal fun List<Issue>.sort(sortFields: List<OrderField>): List<Issue> {
     // Explicitly only support a single sort field
     val sortField = sortFields.first()
 
-    when (sortField.name) {
-        "timestamp" -> compareBy<Issue> { it.timestamp }
-        "source" -> compareBy<Issue> { it.source }
-        "message" -> compareBy<Issue> { it.message }
-        "severity" -> compareBy<Issue> { it.severity }
-        "affectedPath" -> compareBy<Issue> { it.affectedPath }
-        "identifier" -> compareBy<Issue> { it.identifier?.toConcatenatedString() }
-        "worker" -> compareBy<Issue> { it.worker }
+    val comparator: Comparator<Issue> = when (sortField.name) {
+        "timestamp" -> compareBy { it.timestamp }
+        "source" -> compareBy { it.source }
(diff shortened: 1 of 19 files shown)


services/hierarchy/src/main/kotlin/IssueService.kt
```diff
@@ -127,20 +127,19 @@ internal fun List<Issue>.sort(sortFields: List<OrderField>): List<Issue> {
     // Explicitly only support a single sort field
```
oheger-bosch, line 181: Commit message nit: "Extract comparator creation into *an* explicitly typed variable..."

services/ort-run/src/main/kotlin/IssueService.kt
```diff
@@ -93,6 +103,31 @@ class IssueService(private val db: Database) {
...
+            ortResult.getAnalyzerIssues() to AnalyzerRunDao.ISSUE_WORKER_TYPE,
+            ortResult.getAdvisorIssues() to AdvisorRunDao.ISSUE_WORKER_TYPE,
+            ortResult.getScannerIssues() to ScannerRunDao.ISSUE_WORKER_TYPE
+        ).flatMap { (issuesByIdentifier, workerType) ->
+            collectIssuesFromWorker(issuesByIdentifier, workerType)
+        }
+
+        // Collect issues that are not stored in the OrtResult, like issues from the config, reporter or notifier
+        // workers.
+        val dbIssues = db.dbQuery { IssueDao.createFromQuery(createOrtRunIssuesQuery(ortRunId)) }
```
oheger-bosch, line 136: I would assume that most issues are already stored in the database; there should only be a few exceptions (e.g. the ones from the dependency graph). Could you please comment which exceptions these are? (The comment is currently the other way round, which does not answer this question.)

MarcelBochtler: > I would assume that most issues are already stored in the database

This is correct

> there should only be a few exceptions (e.g. the ones from the dependency graph). Could you please comment which exceptions these are?

The dependency graph issues are the only one that are currently known to me. I rephrased the comment to reflect this.

MarcelBochtler: Good point.
The database precision could indeed cause duplicates to not be found.
The other values should be safe for the comparison (and IMO we better show duplicates than removing issues that should have been shown).

oheger-bosch (approved):

PR #3081, opened 2025-07-02
build: Define local maven repository definition

MarcelBochtler:
The intention for using a local maven repository for ORT was to allow
changing the ORT version during development. [1] broke the current
approach to do this.

To restore the old behavior, add the local maven repository to the
dependency resolution management repositories definition.

Signed-off-by: Marcel Bochtler <marcel.bochtler@bosch.com>

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -118,6 +118,14 @@ plugins {
 
 dependencyResolutionManagement {
     repositories {
+        mavenLocal {
+            name = "localOrt"
+
+            content {
+                includeGroupByRegex("org\\.ossreviewtoolkit(\\..*)?")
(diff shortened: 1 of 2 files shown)


settings.gradle.kts
```diff
@@ -118,6 +118,14 @@ plugins {
 
 dependencyResolutionManagement {
     repositories {
+        mavenLocal {
+            name = "localOrt"
+
+            content {
+                includeGroupByRegex("org\\.ossreviewtoolkit(\\..*)?")
```
sschuberth, line 125: I guess you could use `includeGroupAndSubgroups()` (or just `includeGroup()`) instead.

PR #3083, opened 2025-07-02
Plugin template UI

mnonnenmacher:
Add the UI for managing plugin templates. This is a first version that needs improvements in presentation and error handling. Also, currently templates can not be edited, this will be added later. Note that plugin templates are still not applied when creating ORT runs, this will follow in another PR.

image

image

### DIFF
--- a/components/plugin-manager/implementation/src/main/kotlin/PluginTemplateService.kt
+++ b/components/plugin-manager/implementation/src/main/kotlin/PluginTemplateService.kt
@@ -80,6 +80,35 @@ class PluginTemplateService(
                 }
         }
 
+    /**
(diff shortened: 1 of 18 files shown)


Etsija (commented):
All the UI changes are in one huge commit, which makes it pretty hard to do code inspection. Could you consider splitting into something more logical, maybe a few commits?

mnonnenmacher:
> All the UI changes are in one huge commit, which makes it pretty hard to do code inspection. Could you consider splitting into something more logical, maybe a few commits?

I'm afraid that might take me a very long time to split the new pages into steps that make sense and still compile. How about you just check for grave mistakes or code smells and we make a dedicated session in the next weeks to review the code together? I have lots of questions anyway before I start improving the visuals.

PR #3085, opened 2025-07-02
ci(docker): Push images when no jib task is defined

MarcelBochtler:
The current implementation never pushes base worker images because they have the task `jibDockerBuild` configured. Change the condition to push images if the task does not end with `:jib` (which itself would push the image).


mnonnenmacher (changes requested):
Commit message: Please also provide a reason why the base images should be published.

.github/workflows/docker-build.yml
```diff
@@ -138,7 +138,7 @@ jobs:
         # Load the image to the Docker daemon if it is required as a base image for the Jib build.
         echo "LOAD_IMAGE=$([[ '${{ matrix.docker.task }}' != '' ]] && echo 'true' || echo 'false')" >> $GITHUB_ENV
         # Push the image to the registry if it is not a PR and no Jib build is configured.
-        echo "PUSH_IMAGE=$([[ ${{ env.IS_PR }} == 'false' && '${{ matrix.docker.task }}' == '' ]] && echo 'true' || echo 'false')" >> $GITHUB_ENV
+        echo "PUSH_IMAGE=$([[ ${{ env.IS_PR }} == 'false' && '${{ matrix.docker.task }}' != *':jib' ]] && echo 'true' || echo 'false')" >> $GITHUB_ENV
```
mnonnenmacher, line 141: I thinkt this will always be a jib task so you could also just remove the whole condition `&& '${{ matrix.docker.task }}' == ''`.

MarcelBochtler: Well, currently these are `jibDockerBuild` tasks.

mnonnenmacher (approved):

Issue #3096, opened 2025-07-03
UI does not show all issues


hanna-modica:
In the UI, we sometimes have runs that finished with issues (and sometimes runs that are even successful) and show no issues in the issues section, but in the scan-report-web-app.html there are issues shown. The shown issues should be consistent IMO.

![Image](https://github.com/user-attachments/assets/df38b4ec-22de-42de-9361-d425415b0d00)

![Image](https://github.com/user-attachments/assets/37778ae2-8949-4307-b175-62253d737a6a)

sschuberth:
@hanna-modica can you verify whether the issues are already missing from the backend's response, or are really just missing in the UI?

mnonnenmacher:
This is also weird because the scanner run in the example is green. @hanna-modica Please also create an internal ticket for this because I think it is impossible to debug without access to the actual data of the run or clear steps to reproduce.

wkl3nk:
Analyzing it

wkl3nk:
The first problematic thing I see is this Dashboard widgets that shows the number of issues found, grouped by severity by a color code. While for retrieving the issues we have changed the approach to re-generate a ORT result data model from the database, this one still directly uses the _/statistics_ endpoint of the run, which directly queries the database, what we have not found suitable recently. Both the issues detail view and the dashboard widget information should be gained by the same code.
If it is changed, it **may** have negative impact on the performance of the dashboard.

<img width="521" height="189" alt="Image" src="https://github.com/user-attachments/assets/cd985634-2c74-4e9c-bb54-8a4c6750d727" />

@MarcelBochtler What do you think, should we also change this statistics endpoint to use the approach with the ORT result data model? What about higher aggregation levels, as on product level and organization level?

mnonnenmacher:
> [@MarcelBochtler](https://github.com/MarcelBochtler) What do you think, should we also change this statistics endpoint to use the approach with the ORT result data model? What about higher aggregation levels, as on product level and organization level?

I think we should first fix the issue described here and fix the dashboard separately later because of the challenge to get the performance right (it also needs to be fixed for the product and organization dashboards, so I think it cannot be fixed without storing the statistics somewhere, because building the ORT result for the last runs of potentially hundreds of repositories is not feasible).

hanna-modica:
That proposal is fine with me. If I come across this issue again, I will report.

PR #3106, opened 2025-07-04
fix(analyzer): Take all issues into account for run status

mnonnenmacher:
Use the `AnalyzerResult.getAllIssues()` function instead of the `AnalzyerResult.issues` property when determining the analyzer job status, because that also includes issues from the dependency trees.

### DIFF
--- a/workers/analyzer/src/main/kotlin/analyzer/AnalyzerWorker.kt
+++ b/workers/analyzer/src/main/kotlin/analyzer/AnalyzerWorker.kt
@@ -135,7 +135,7 @@ internal class AnalyzerWorker(
                 ortRunService.storeAnalyzerRun(analyzerRun.mapToModel(jobId), shortestPathsByIdentifier)
             }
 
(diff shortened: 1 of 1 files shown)


sschuberth (dismissed):

sschuberth:
I guess this does not fix https://github.com/eclipse-apoapsis/ort-server/issues/3096 yet, as the issues missing there were not from the analyzer, correct?

workers/analyzer/src/main/kotlin/analyzer/AnalyzerWorker.kt
```diff
@@ -135,7 +135,7 @@ internal class AnalyzerWorker(
                 ortRunService.storeAnalyzerRun(analyzerRun.mapToModel(jobId), shortestPathsByIdentifier)
             }
 
-            if (analyzerRun.result.issues.values.flatten().any { it.severity >= Severity.WARNING }) {
+            if (analyzerRun.result.getAllIssues().values.flatten().any { it.severity >= Severity.WARNING }) {
```
sschuberth, line 138: Should line 119 also be changed?

PR #3122, opened 2025-07-07
feat(workers): Introduce a generator for Gradle's `init.gradle.kts` file

bs-ondem:


sschuberth:
I probably know too little about these generators and your use-case for them still, but is it worth considering to write `init.gradle` in Groovy to support pre-Kotlin-DSL Gradle versions?

oheger-bosch:
> I probably know too little about these generators and your use-case for them still, but is it worth considering to write `init.gradle` in Groovy to support pre-Kotlin-DSL Gradle versions?

So, IIUC, this means that the same functionality would be available for older Gradle versions? Then this should probably be done to have the highest level of compatibility.

sschuberth:
> So, IIUC, this means that the same functionality would be available for older Gradle versions?

Yes, it's purely a syntax thing.

sschuberth:
> Is it enough to just switch to Groovy or could there be any drawbacks with the generated `init.gradle` file with older versions?

Just switching to Groovy syntax should be enough, even the latest versions of Gradle still support Groovy. However, if you anyway depend on Gradle features that are only available in Gradle versions that already support Kotlin DSL, there's of course no benefit in switching to Groovy syntax.

bs-ondem:
That's the case for the generated `dependencyResolutionManagement` section in the generated file. The question is whether to keep it as is with `init.gradle.kts` (since Kotlin DSL is supported starting with Gradle 5.0), or switch to Groovy and add conditional logic to the generated init.gradle that checks the Gradle version via `Gradle.getGradleVersion()`.

oheger-bosch:
> > The question is whether to keep it as is with `init.gradle.kts` (since Kotlin DSL is supported starting with Gradle 5.0), or switch to Groovy
> 
> I'd stick to Kotlin DSL then unless the concrete need arises to switch to Groovy.

The Gradle version used by our customers is something that we cannot control. If the conditional logic is rather trivial, I would tend towards adding it. Otherwise, it needs to be documented for which Gradle versions this feature is supported.

oheger-bosch (approved):
Approving to make this feature available. If the need arises, support for older Gradle versions can be added later.

PR #3124, opened 2025-07-07
Wkl3nk/direct entering credentials in new repository

wkl3nk:
Add support to input username and password or PAT **directly on the create repository page**. This greatly improves usability compared to the previous multi-step process, which required manually creating two secrets and configuring an infrastructure service just to authenticate with the repository.

This is just an option. As before, one can also keep the input fields for username and password empty and configure the
required infrastructure service manually or reuse an existing one.

image


Etsija:
I am a bit unsure about the context of this PR, as I was probably not present when it was discussed. Didn't find any relevant topics from the Discussion section either.

sschuberth:
Username and password now look as if they were mandatory. How about hiding these fields behind a toggle that says "Authentication required" (or similar) and only show these field if the toggle is activated?

sschuberth:
> If nothing is filled out, then there is a grey background text telling "Optional".

I guess the same should go for the password field then?

wkl3nk:
@Etsija Should I give the context here, or would you prefer if I open a "discussion" post ?

Etsija:
> @Etsija Should I give the context here, or would you prefer if I open a "discussion" post ?

Either way is fine for me. Maybe a discussion would be a more permanent item

wkl3nk:
@Etsija I added a discussion of the topic here, giving background information and context: https://github.com/eclipse-apoapsis/ort-server/discussions/3135

wkl3nk:
@Etsija @lamppu @mmurto Feedback from the discussions page is that the solution proposed in this PR is seen as problematic, although there is the mentioning that you should only directly enter username and password if you have only a small number of repositories, and won't benefit this much from shared infrastructure services and secrets.  I also understand the fear that things are introduced and data generated in the database that later on is hard to change to a different solution.
What I see as problem is that the approach with secrets and infrastructure services is rather generic, possibly too generic for a user. But then we first would need to change this basic approach, e.g. make a difference between username and password/token secrets. 

However, if you say: **I temporarily can live with the approach with directly entering user and password, then please give this PR a positive review**. If not, I will have to withdraw and close it.

PR #3130, opened 2025-07-08
feat(workers): Include `GradleDefinition`s in the `GradleInitGenerator`

bs-ondem:
Extend the `GradleInitGenerator` to include repositories from `GradleDefinition`s in addition to the globally configured `MavenCentralMirror`. Override the globally configured `MavenCentralMirror` if a `GradleDefinition` uses the same URL.

### DIFF
--- a/workers/common/src/test/kotlin/common/env/GradleInitGeneratorTest.kt
+++ b/workers/common/src/test/kotlin/common/env/GradleInitGeneratorTest.kt
@@ -48,6 +48,315 @@ class GradleInitGeneratorTest : WordSpec({
             mockBuilder.homeFileName shouldBe ".gradle/init.gradle.kts"
         }
 
+        "generate repositories blocks from definitions" {
+            val secUser1 = MockConfigFileBuilder.createSecret("user1Secret")
+            val secPass1 = MockConfigFileBuilder.createSecret("pass1Secret")
+            val secUser2 = MockConfigFileBuilder.createSecret("user2Secret")
+            val secPass2 = MockConfigFileBuilder.createSecret("pass2Secret")
+
+            val definition1 = GradleDefinition(
+                MockConfigFileBuilder.createInfrastructureService("https://repo1.example.org", secUser1, secPass1),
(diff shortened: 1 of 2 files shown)


workers/common/src/test/kotlin/common/env/GradleInitGeneratorTest.kt
```diff
@@ -48,6 +48,312 @@ class GradleInitGeneratorTest : WordSpec({
...
+            val definition2 = GradleDefinition(
+                MockConfigFileBuilder.createInfrastructureService("https://repo2.example.org", secUser2, secPass2),
+                emptySet()
+            )
+
+            val mockBuilder = MockConfigFileBuilder()
+
+            GradleInitGenerator().generate(mockBuilder.builder, listOf(definition1, definition2))
+
+            val expectedLines = listOf(
```
sschuberth, line 71: How about beautifying this with the

```kotlin
"""
...
""".trimIndent()
```

approach?

sschuberth:
> Or more generally, are there use cases where it makes sense to configure repositories in the `init.gradle.kts` file rather than `build.gradle(.kts)`?

I guess this boils down to the old question of self-contained builds vs. central configuration, which means there's no clear answer. Personally, I'm pro self-contained builds, even if that might require some repetition of global configuration in some cases.

So I'd proceed as I always try in such cases: Keep it simple unless the added complexity is clearly needed.

workers/common/src/test/kotlin/common/env/GradleInitGeneratorTest.kt
sschuberth: FYI, I've done this now also for existing code in https://github.com/eclipse-apoapsis/ort-server/pull/3153.

mnonnenmacher:
> @eclipse-apoapsis/bosch Are there any plans to continue working on this?

@eclipse-apoapsis/bosch ping?

MarcelBochtler:
Closed as part of a cleanup.

PR #3163, opened 2025-07-14
Apply resolutions to vulnerablities endpoint

wkl3nk:
If resolutions exist for vulnerabilities, add them to the respective vulnerabilities in the `api/v1/runs/{id}/vulnerabilities` endpoint. 

Add a request query parameter to support filtering vulnerabilities by their resolution status, to provide the possibility for clients to only query for relevant data.

For details, please check the commit messages of the respective commits.

For review: The change was split up into several commits. 

- The first commit is pure refactoring done by the IDE, it is just about moving a class to a different package.
- The second commit changes the way that vulnerabilities are looked up (using ORTRunService instead of querying the results directly from the database)
- The third commit adds support for adding resolutions to vulnerabilities
- The fourth commit adds the query parameter used for filtering

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/VulnerabilityResolution.kt
+++ b/api/v1/model/src/commonMain/kotlin/VulnerabilityResolution.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,15 +22,12 @@ package org.eclipse.apoapsis.ortserver.api.v1.model
 import kotlinx.serialization.Serializable
 
 /**
- * A union data class to associate a [Vulnerability] with an [Identifier].
+ * Defines the resolution of a Vulnerability. This can be used to silence false positives, or vulnerabilities that
+ * have been identified as not being relevant.
  */
 @Serializable
-data class VulnerabilityWithIdentifier(
-    val vulnerability: Vulnerability,
-    val identifier: Identifier,
-
-    /** An advisory rating for the [Vulnerability], derived from the individual references of the [Vulnerability]. */
-    val rating: VulnerabilityRating,
-
-    val resolutions: List<VulnerabilityResolution> = emptyList()
+data class VulnerabilityResolution(
+    val externalId: String,
+    val reason: String,
+    val comment: String
 )
--- a/core/src/main/kotlin/api/RunsRoute.kt
+++ b/core/src/main/kotlin/api/RunsRoute.kt
@@ -80,11 +80,11 @@ import org.eclipse.apoapsis.ortserver.model.runs.Project
 import org.eclipse.apoapsis.ortserver.services.ProjectService
 import org.eclipse.apoapsis.ortserver.services.ReportStorageService
 import org.eclipse.apoapsis.ortserver.services.RepositoryService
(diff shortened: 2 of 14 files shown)


api/v1/model/src/commonMain/kotlin/VulnerabilityResolution.kt
```diff
@@ -0,0 +1,29 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.api.v1.model
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class VulnerabilityResolution(
```
oheger-bosch, line 25: Please add documentation to public API classes.

core/src/main/kotlin/api/RunsRoute.kt
```diff
@@ -209,9 +210,14 @@ fun Route.runs() = route("runs") {
                     requirePermission(RepositoryPermission.READ_ORT_RUNS.roleName(ortRun.repositoryId))
 
                     val pagingOptions = call.pagingOptions(SortProperty("externalId", SortDirection.ASCENDING))
+                    val filters = call.vulnerabilityFilters()
 
                     val vulnerabilitiesForOrtRun =
-                        vulnerabilityService.listForOrtRunId(ortRun.id, pagingOptions.mapToModel())
+                        vulnerabilityService.listForOrtRunId(
```
oheger-bosch, line 216: There should be an integration test for checking this new filter.

wkl3nk: Added a second vulnerability to the Integration Test, and added 2 new tests for testing this "resolved" query parameter.

core/src/main/kotlin/apiDocs/RunsDocs.kt
```diff
@@ -316,14 +324,6 @@ val getRuleViolationsByRunId: RouteConfig.() -> Unit = {
...
         }
 
-        queryParameter<Boolean>("resolved") {
-            description =
-                """
-                    If true, only resolved rule violations are returned. If false, only unresolved rule violations are
-                    returned. If missing, both resolved and unresolved rule violations are returned.
-                """.trimIndent()
-        }
-
```
MarcelBochtler, line 326: This should not be removed.

model/src/commonMain/kotlin/VulnerabilityWithIdentifier.kt
```diff
@@ -35,3 +37,15 @@ data class VulnerabilityWithIdentifier(
...
     val resolutions: List<VulnerabilityResolution> = emptyList()
 )
+
+/**
+ * Filters to apply when querying for vulnerabilities.
+ */
+@Serializable
+data class VulnerabilityFilters(
+    /**
+     * Filter to only return resolved vulnerabilities. Null if both, resolved an unresolved vulnerabilities should be
```
MarcelBochtler, line 47: `resolved an unresolved` -> `and`.

MarcelBochtler (changes requested):

services/ort-run/src/test/kotlin/VulnerabilityServiceTest.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
```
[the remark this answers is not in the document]
wkl3nk: I guess the IDE does this automatically :(

core/src/main/kotlin/apiDocs/RunsDocs.kt
wkl3nk: Oh. Mistakenly deleted. I was wondering why it already existed, but did not notice it is for rule violations, and not for vulnerabilities.

PR #3181, opened 2025-07-16
feat(ui): Remove unsupported credentials type

kamil-bielecki-bosch:
This commit removes no longer supported .netrc file based infrastructure service authentication option.

### DIFF
--- a/ui/src/routes/organizations/$orgId/infrastructure-services/$serviceName/edit/index.tsx
+++ b/ui/src/routes/organizations/$orgId/infrastructure-services/$serviceName/edit/index.tsx
@@ -294,16 +294,6 @@ const EditInfrastructureServicePage = () => {
                     credentials.
                   </p>
                   <p>
-                    In most cases, the desired option is to add the credentials
-                    to the Netrc file, as this allows access to the service from
-                    most external tools. Normally, Git should be able to obtain
-                    the credentials from the Netrc file, however, there are rare
-                    cases when Git is not able to authenticate against a
-                    repository based on the information in the Netrc file. In
-                    these cases, you can choose to add the credentials to the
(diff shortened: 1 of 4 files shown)


mnonnenmacher:
@kamil-bielecki-bosch @oheger-bosch @MarcelBochtler What is the plan for tools like python-inspector which get their credentials from `.netrc`?

MarcelBochtler:
> @kamil-bielecki-bosch @oheger-bosch @MarcelBochtler What is the plan for tools like python-inspector which get their credentials from `.netrc`?

I was told that this functionality was already removed from the server.
@oheger-bosch?

oheger-bosch:
> > @kamil-bielecki-bosch @oheger-bosch @MarcelBochtler What is the plan for tools like python-inspector which get their credentials from `.netrc`?
> 
> I was told that this functionality was already removed from the server. @oheger-bosch?

There is currently no direct way to force the creation of a `.netrc` file with a specific content. Under certain circumstances, such a file is generated when a service is accessed that has the credentials type `NETRC_FILE` assigned; this is needed for cloning some Git repositories that make use of git-lfs (as a work-around for what I consider a bug in git-lfs).

The plan is to make these tools support an authentication mechanism that is not limited to a single pair of credentials per host and then write specialized generator classes for their configuration.

oheger-bosch:
I am fine with the changes, it is certainly an improvement to remove a misleading description. Since this is UI code, will any of the UI developers have a look?

sschuberth:
> I am fine with the changes, it is certainly an improvement to remove a misleading description. Since this is UI code, will any of the UI developers have a look?

I think this is trivial enough to just go ahead. I'll approve.

PR #3194, opened 2025-07-18
feat(core): Add a function to extract plugin configs from `CreateOrtRun`

mnonnenmacher:
The function will be used for validating plugin configs in a follow-up commit.

### DIFF
--- a/core/src/main/kotlin/utils/Extensions.kt
+++ b/core/src/main/kotlin/utils/Extensions.kt
@@ -21,7 +21,10 @@ package org.eclipse.apoapsis.ortserver.core.utils
 
 import io.ktor.server.config.tryGetString
 
+import org.eclipse.apoapsis.ortserver.api.v1.model.CreateOrtRun
+import org.eclipse.apoapsis.ortserver.api.v1.model.PluginConfig
 import org.eclipse.apoapsis.ortserver.clients.keycloak.KeycloakClientConfiguration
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/utils/Extensions.kt
```diff
@@ -43,3 +46,85 @@ fun ConfigManager.createKeycloakClientConfiguration(): KeycloakClientConfigurati
         subjectClientId = getString("keycloak.subjectClientId")
     )
 }
+
+/**
+ * Get all [PluginConfig]s from a [CreateOrtRun] object. The result is a map from [PluginType]s to maps of plugin IDs to
+ * [PluginConfig]s. Some job configs allow to enable plugins without addintional configuration, in this case `null` is
```
oheger-bosch, line 52: Typo in "additional".

core/src/main/kotlin/utils/Extensions.kt
```diff
@@ -43,3 +46,85 @@ fun ConfigManager.createKeycloakClientConfiguration(): KeycloakClientConfigurati
...
+ * used as the value for the plugin ID in the map.
+ */
+fun CreateOrtRun.getPluginConfigs(): Map<PluginType, Map<String, PluginConfig?>> =
```
oheger-bosch, line 55: Question: Wouldn't an empty `PluginConfig` instead of a `null` value be easier to handle for consumers?

oheger-bosch (approved):

PR #3203, opened 2025-07-21
style(gradle): Consistenly prefer `=` over `set()`

sschuberth:


core/build.gradle.kts
```diff
@@ -38,7 +38,7 @@ plugins {
 group = "org.eclipse.apoapsis.ortserver"
 
 application {
-    mainClass.set("io.ktor.server.netty.EngineMain")
+    mainClass = "io.ktor.server.netty.EngineMain"
```
sschuberth, line 41: @eclipse-apoapsis/technology-apoapsis-committers is this whole line even required as there is also line 153 present? At the least the other workers do not have this line.

MarcelBochtler: Line `153` is jib / container specific. Therefore, I think that this line is still required.

PR #3204, opened 2025-07-21
build: Make the CLI command used to build containers configurable

MarcelBochtler:
Enable the usage of Docker compatible container engines like Podman [1] by making the command configurable for the `buildAllImages` task.

[1]: https://podman-desktop.io/docs/migrating-from-docker/managing-docker-compatibility


sschuberth:
As a side remark, IMO we should generally aim for building OCI containers without the need for any third-party CLI tool, be it `docker` or `podman`. https://github.com/SgtSilvio/gradle-oci looks like a promising base for that; however, it's yet a bit unclear how to add required CLI tools with that plugin. But that's something I'm looking into occasionally.

sschuberth (approved):

PR #3216, opened 2025-07-22
chore(github): Use newly created teams to define code owners

mnonnenmacher:


sschuberth:
Note that a downside of using teams might be that GitHub then only requires *one* team member to start a review, and from then on only that member is listed as a reviewer, and the team is removed. At least that's the behavior I see in ORT. The problem with this is that if the original reviewer does not complete the review / approve, then no other reviewers will be added back automatically.

sschuberth:
We actually had a discussion about this exact topic at OSS Summit Europe with @schr3gl3j and others, coming to the conclusion that keeping the explicit names is more transparent to contributors, and also allows for a more fine-granular configuration of code owners. So I'd actually propose to drop this PR, and also revert to explicit user names .in `CODEOWNERS` for ORT (core).

mnonnenmacher:
> We actually had a discussion about this exact topic at OSS Summit Europe with @schr3gl3j and others, coming to the conclusion that keeping the explicit names is more transparent to contributors, and also allows for a more fine-granular configuration of code owners. So I'd actually propose to drop this PR, and also revert to explicit user names .in `CODEOWNERS` for ORT (core).

Fine with me.

mnonnenmacher:
@sschuberth I have made the change for ORT here: https://github.com/oss-review-toolkit/ort/pull/10818

PR #3217, opened 2025-07-22
fix(logs): Ensure that logs of forked processes set the correct MDC

MarcelBochtler:


workers/common/src/main/kotlin/common/env/EnvironmentForkHelper.kt
```diff
@@ -76,9 +84,14 @@ object EnvironmentForkHelper {
     fun setupFork(pipe: InputStream) {
         logger.info("Setting up forked process...")
 
-        val authInfo = Json.decodeFromStream<SerializableAuthenticationInfo>(pipe).toAuthenticationInfo()
+        val forkData = Json.decodeFromStream<SerializableForkData>(pipe)
+        val authInfo = forkData.authInfo.toAuthenticationInfo()
+        val mdcContext = forkData.mdcContext
+
+        restoreMdcContext(mdcContext)
```
oheger-bosch, line 91: Can `EnvironmentForkHelperTest` be extended to test whether the MDC is correctly set up in the fork?

PR #3220, opened 2025-07-23
chore(docker): Set `linux/amd64` platform for images that install ScanCode

MarcelBochtler:
ScanCode does not support to be installed on `ARM64` machines, see [1], [2] and [3].
Use `linux/amd64` base-images for any all images that install ScanCode to be able to build them on `ARM64` machines.

[1]: https://scancode-toolkit.readthedocs.io/en/stable/getting-started/install.html#system-requirements
[2]: https://github.com/aboutcode-org/scancode-toolkit/issues/3141
[3]: https://github.com/aboutcode-org/scancode-toolkit/issues/3958


sschuberth:
> for any all images

for *all* images

Issue #2458, opened 2025-04-07
UI breaks when refresh token is expired


mnonnenmacher:
When opening an old browser tab with the UI where the refresh token has expired, the whole page breaks and shows the infamous "Something went wrong!" error. Refreshing the page redirects to the Keyloak login.
Instead of failing completely, the UI should redirect to the Keycloak login when the refresh token has expired.

mnonnenmacher:
Sometimes the UI does not break but shows lots of errors:

![Image](https://github.com/user-attachments/assets/edf83c74-43f6-484d-90d0-d7c3eca0b091)

Etsija:
@mnonnenmacher I will possibly start looking into this next week.  What would be a good way to make the error happen?`, so I can test?

mnonnenmacher:
@lamppu Do you plan to continue working on this?

PR #3227, opened 2025-07-24
ui: Auth improvements

lamppu:
This PR adds a solution for handling errors in case of an expired session. The problem is two-fold, sometimes the automatic silent renewal fails and results to a "Something went wrong" error, this is handled by adding a handler for such errors. In some cases, for example if the session is manually killed in Keycloak, the API requests may start failing before the automatic renewal kicks in. Handling for such cases is added to the query client.

Also a debug logger is added and some improvements for the "Unable to log in" screen.

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -21,7 +21,10 @@ import { createRouter, RouterProvider } from '@tanstack/react-router';
 import { useEffect, useState } from 'react';
 import { hasAuthParams } from 'react-oidc-context';
 
+import { CopyToClipboard } from '@/components/copy-to-clipboard.tsx';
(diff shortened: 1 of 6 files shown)


lamppu:
> Does the PR in sum actually fix #2458? If so, should this be documented in the last commit that refers to the issue by changing "Related to" to "Fixes"? (And should we anyway say "Relate_s_ to"?)

Yes it should. I wasn't sure about the wording, so whatever is the preferred way is fine by me.

mnonnenmacher:
> @mnonnenmacher, please also have a look to judge whether you believe this would fix your issue.

Yes, from looking at the changes I believe this would exactly fix the issues we are seeing.

PR #3266, opened 2025-08-01
feat(ui): Add inputs for `keepAliveWorker` flags

mnonnenmacher:
Add inputs for the `keepAliveWorker` flags for each job when creating an ORT run. The inputs are only shown for superusers because only they can use the flags.

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
@@ -65,6 +65,7 @@ export const createRunFormSchema = z.object({
       allowDynamicVersions: z.boolean(),
       skipExcluded: z.boolean(),
       environmentVariables: z.array(environmentVariableSchema).optional(),
+      keepAliveWorker: z.boolean(),
       packageManagers: z
         .object({
           Bazel: packageManagerOptionsSchema,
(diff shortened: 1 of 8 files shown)


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -360,6 +372,9 @@ export function defaultValues(
             environmentVariables:
               ortRun.jobConfigs.analyzer?.environmentConfig
                 ?.environmentVariables || undefined,
+            keepAliveWorker:
+              ortRun.jobConfigs.analyzer?.keepAliveWorker ||
+              baseDefaults.jobConfigs.analyzer.keepAliveWorker,
```
lamppu, line 377: I have one question, what happens if a superuser has enabled this in the latest run, and a regular user tries to do a rerun, will the request fail?

mnonnenmacher: Yes, and there would be now way to make it work because the inputs are not showing. So I think the rerun logic should automatically set the flags to false if the user is not a superuser, would you agree?

mnonnenmacher: I have updated the `defaultValues` function.

PR #3279, opened 2025-08-04
ui: Status column for resolved/unresolved issues

Etsija:
This PR adds a sortable and filterable Status column to the issues table. For now, the column holds information on unresolved/resolved status of the issues.

Screenshot from 2025-08-04 13-06-45

Please note that in lack of suitable test data (no resolutions in my local dev environment), the logic could need some fixes later, thus this PR only includes the status column for one table. The rest of the tables supporting resolutions status will come in later PRs.


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/issues/index.tsx
```diff
@@ -178,6 +181,46 @@ const IssuesComponent = () => {
         },
       },
     }),
+    columnHelper.accessor(
+      (issue) => {
+        return issue.resolutions && issue.resolutions.length > 0
+          ? 'Resolved'
+          : 'Unresolved';
```
sschuberth, line 188: Is there a way to only hard-code this once and re-use `itemStatusSchema` from the previous commit? Similar below.

Etsija: Good question; Zod specifies schemas for validation, and it is quite dirty to use the enum values later. I only know of a very hacky way to for example to get the "Resolved": `itemStatusSchema.options[0]`

PR #3296, opened 2025-08-07
ui: Tweak the main layout

Etsija:
This PR includes some minor modifications to the main layout and tables:
- action buttons in table headers are moved  more closely together
- main layout is made a bit wider to prevent horizontal overflow of some tables

Screenshot from 2025-08-07 08-10-40

### DIFF
--- a/ui/src/components/ui/button-variants.ts
+++ b/ui/src/components/ui/button-variants.ts
@@ -28,6 +28,7 @@ export const buttonVariants = cva(
       },
       size: {
         default: 'h-9 px-4 py-2 has-[>svg]:px-3',
+        narrow: 'h-8 rounded-md px-1.5 has-[>svg]:px-1',
         sm: 'h-8 rounded-md gap-1.5 px-3 has-[>svg]:px-2.5',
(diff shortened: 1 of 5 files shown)


ui/src/components/ui/button-variants.ts
```diff
@@ -28,6 +28,7 @@ export const buttonVariants = cva(
       },
       size: {
         default: 'h-9 px-4 py-2 has-[>svg]:px-3',
+        narrow: 'h-8 rounded-md px-1.5 has-[>svg]:px-1',
```
sschuberth, line 31: Nit: I wonder whether we should stay closer to the existing "sm" (for "small") wording. Maybe call this "xsm" for "extra small", like in t-short sizes?

Etsija: IMHO "narrow" communicates the meaning of the size better, as the size definition is mostly as with "sm" but just narrower.

sschuberth: Ok, I'm also fine with sticking to "narrow", just wanted to share the thought.

Issue #1361, opened 2024-11-04
Some top-level items of `config.yml` are not configurable in the server yet


sschuberth:
The server does not make use of ORT's `config.yml`. As such all configuration items from [OrtConfiguration](https://github.com/oss-review-toolkit/ort/blob/f8a0c3995c1095dbeb5d839f828c4f8d9e5bf3b4/model/src/main/kotlin/config/OrtConfiguration.kt) need to be mapped to something inside [JobConfigurations](https://github.com/eclipse-apoapsis/ort-server/blob/e161efe306a4236225a6659bd3147fdc17ec8373/model/src/commonMain/kotlin/JobConfigurations.kt). However, some top-level items of `config.yml` are currently missing from that model. At a quick glance, these seem to be missing

    /**
     * A flag to indicate whether authors should be considered as copyright holders.
     */
    val addAuthorsToCopyrights: Boolean = false,

    /**
     * A set with the names of environment variables that are explicitly allowed to be passed to child processes,
     * even if they are matched by one of the [deniedProcessEnvironmentVariablesSubstrings].
     * See [EnvironmentVariableFilter] for further details.
     */
    val allowedProcessEnvironmentVariableNames: Set<String> = EnvironmentVariableFilter.DEFAULT_ALLOW_NAMES,

    /**
     * A set with substrings to filter out environment variables before creating child processes to prevent that those
     * processes can access sensitive information. See [EnvironmentVariableFilter] for further details.
     */
    val deniedProcessEnvironmentVariablesSubstrings: Set<String> = EnvironmentVariableFilter.DEFAULT_DENY_SUBSTRINGS,

    /**
     * Enable the usage of project-local package configurations from the [RepositoryConfiguration]. If set to true,
     * apply package configurations from a local .ort.yml file before applying those specified via the command line i.e.
     * configurations from the .ort.yml take precedence.
     */
    val enableRepositoryPackageConfigurations: Boolean = false,

    /**
     * Enable the usage of project-local package curations from the [RepositoryConfiguration]. If set to true, apply
     * package curations from a local .ort.yml file before applying those specified via the command line i.e. curations
     * from the .ort.yml take precedence.
     */
    val enableRepositoryPackageCurations: Boolean = false,

    /**
     * The license file patterns.
     */
    val licenseFilePatterns: LicenseFilePatterns = LicenseFilePatterns.DEFAULT,

    /**
[...]

PR #3297, opened 2025-08-07
fix(scanner): Account issues from failed scans to job status

lamppu:
Take the issues from `ScannerRun` into account when determining the job status.

### DIFF
--- a/workers/scanner/src/main/kotlin/scanner/ScannerWorker.kt
+++ b/workers/scanner/src/main/kotlin/scanner/ScannerWorker.kt
@@ -94,7 +94,9 @@ class ScannerWorker(
                 )
             }
 
-            if (issues.any { it.severity >= Severity.WARNING }) {
+            val allIssues = issues + scannerRunResult.scannerRun.issues.values.flatten().map { it.mapToModel() }
+
+            if (allIssues.any { it.severity >= Severity.WARNING }) {
                 RunResult.FinishedWithIssues
             } else {
                 RunResult.Success


workers/scanner/src/main/kotlin/scanner/ScannerWorker.kt
```diff
@@ -94,7 +94,7 @@ class ScannerWorker(
                 )
             }
 
-            if (issues.any { it.severity >= Severity.WARNING }) {
+            if (issues.any { it.severity >= Severity.WARNING } || scannerRunResult.scannerRun.issues.isNotEmpty()) {
```
sschuberth, line 97: > These are errors from failing scans so checking that the issues
map is not empty should suffice here, no need to check the individual
severities.

While that might be correct right now, I wonder whether we should still make this more "future proof" by first merging the issues and then doing the check, briefly like (probably introduce a variable):

```kotlin
(issues + scannerRunResult.scannerRun.issues).any { it.severity >= Severity.WARNING }
```

Otherwise the code reads as if somehow `scannerRunResult.scannerRun.issues` were more important as we fail for any severity for them.

Also, this might get important once / if we take a configured `severeIssueThreshold` into account, see https://github.com/eclipse-apoapsis/ort-server/issues/1361.

lamppu: Yeah that makes sense. I had to do it a bit differently though, because `scannerRunResult.scannerRun.issues` is a map.

sschuberth: I see. Personally though, I'd still have a slight preference for hard-coding the severity threshold only once, like

```kotlin
val allIssues = issues + scannerRunResult.scannerRun.issues.values.flatten().map { it.mapToModel() }
if (allIssues.any { it.severity >= Severity.WARNING }) {
```

Actually, should already line 88 be modified to

```kotlin
val issues = scannerRunResult.extractIssues() +
        scannerRunResult.scannerRun.issues.values.flatten().map { it.mapToModel() }
```

so that in line 93 all `issues` are used in the `db.dbQuery` block?

mnonnenmacher: I think this code needs some refactoring to make it clear what the different kinds of issues are.

PR #3312, opened 2025-08-08
ui: Wrap Package ID around ":" and "::" where needed

Etsija:
"Normal" situation with easily word-wrappable Package IDs:

Screenshot from 2025-08-08 14-47-45

A Package ID that cannot be word-wrapped with usual supported TailwindCSS classes:

Screenshot from 2025-08-08 14-46-56

### DIFF
--- a/ui/src/components/breakable-string.tsx
+++ b/ui/src/components/breakable-string.tsx
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


ui/src/components/breakable-string.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const BreakableString = ({ text, className }: BreakableStringProps) => {
+  const parts = text.split(/(:+)/g); // Split by ':' or '::', keeping delimiters
+
+  return (
+    <span className={cn('break-words', className)}>
+      {parts.map((part, index) => (
+        <Fragment key={index}>
+          {part}
+          {part.startsWith(':') && <wbr />}
```
mnonnenmacher, line 37: If the part starts with ":", should the `<wbr/>` not be before that part?
IIUC the current implementation would create `type:namespace<wbr/>:name<wbr/>:version<wbr/>` instead of `type<wbr/>:namespace<wbr/>:name<wbr/>:version`.
And out of ignorance, if each part is in a separate `Fragment`, is `<wbr/>` even required?

mnonnenmacher: Ok, I think I got it, the ":" are individual parts, right? So the result of the split in your example is an array with the elements "Gem", "::", "google-apis-iamcredentials_v1", ":", "0.17.0", correct?

mnonnenmacher (approved):

PR #3330, opened 2025-08-11
Collect plugin information at build time

oheger-bosch:
This PR introduces a mechanism to collect information about ORT plugins already at build time. This has the advantage that information about plugins can be queried without the requirement that all plugin artifacts must be on the classpath. This reduces the size of container images and reflection overhead at runtime.

The extracted information is made available via a new module which can be referenced by interested parties. The module can also be used by the plugin manager; alternatively, the functionality to collect this information could be moved to the plugin manager module itself.

While being fully functional, this PR is still in PoC status, since currently only minimum information about plugins is exposed. However, this can be easily extended, based on the use cases that need to be fulfilled. I wanted to get feedback first before adding further functionality.


mnonnenmacher:
That is a very interesting idea, I only have concerns because it breaks use cases like #377 where plugins are added after the build. I think it was never discussed in the community if these use cases should be supported by the server, so I don't feel so comfortable to introduce something that prevents them without asking for feedback first.

> This has the advantage that information about plugins can be queried without the requirement that all plugin artifacts must be on the classpath. This reduces the size of container images and reflection overhead at runtime.

Maybe you could elaborate on that, because my impression was that container size is only an issue for the analyzer worker and even there we don't see big delays caused by image pulls, and getting plugins at runtime via service loader is also super fast. I am more concerned with classpath pollution and potential classpath conflicts between plugin dependencies where https://github.com/oss-review-toolkit/ort/issues/6798 could help if implemented (the [example plugin project](https://github.com/oss-review-toolkit/ort-package-manager-plugin) already uses the shadow plugin).

oheger-bosch:
This approach does not necessarily break use cases to add plugins dynamically. This is just an aspect that has to be dealt with when/if such use cases become more concrete.

Regarding the classpath issue: Access to the available plugins is needed by the admin config service to validate the configuration. This service is referenced by `:workers:commons`, so the dependencies would end up in all worker images. Some of these dependencies are not so trivial, e.g. the Gradle tooling API, JRuby, AsciiDoctor, the Maven resolver library, ... It just feels wrong to have all these libraries lying around just to read some manifest files. Another aspect is security: If there is a vulnerability in one of these dependencies, all ORT Server images would be affected.

And finally I think we should really aim to create an efficient application to reduce resource consumption and CO2 production. Here every MB counts.

buildSrc/src/main/kotlin/PluginInfoCollectorPlugin.kt
```diff
@@ -0,0 +1,153 @@
...
+
+                        zip.getInputStream(entry).use { input ->
+                            targetFile.outputStream().use { output ->
+                                input.copyTo(output)
+                            }
+                        }
+                    }
+            }
+        } catch (e: Exception) {
+            logger.warn("Could not extract plugin information from '{}': {}.", jarFile, e.message)
```
mnonnenmacher, line 155: Should the task not fail in this case?

buildSrc/src/main/kotlin/PluginInfoCollectorPlugin.kt
```diff
@@ -0,0 +1,153 @@
...
+    private fun extractPluginInfosFromJar(jarFile: File, extractDir: File) {
+        try {
+            ZipFile(jarFile).use { zip ->
+                zip.entries().asSequence()
+                    .filter { entry ->
+                        !entry.isDirectory && entry.name.matches(regexPluginFile)
+                    }.forEach { entry ->
+                        val entryFile = File(entry.name)
+                        val targetFile = extractDir.resolve(entryFile.name)
+                        targetFile.parentFile.mkdirs()
```
mnonnenmacher, line 145: Is it not sufficient that `extractDir.mkdirs()` is already called above?

oheger-bosch:
Here are some numbers for the Config worker:
The original size of the container image was 698477438 bytes, it used 352 libraries.
After adding references to all plugins, the size increased to 782490109 bytes (~ + 80 MB), and the number of libraries to 537.

For unknown reasons, the declaration of the plugins in the admin config service module made it necessary to add the special repository with the Gradle tooling API to many other projects; not sure whether there is a better solution (declare it globally?) to avoid this redundancy.

mnonnenmacher:
@oheger-bosch I just thought a bit about if there is a generic way to get all plugin descriptors without the actual plugin code and all the dependencies, because this would also help the plugin manager and the core module. So I thought ORT could be modified to publish these descriptors (either as JSON or as Kotlin class) in separate artifacts (e.g., with a specific classifier or a different ID) by adding them to a separate configuration, this should not be so hard to do with Gradle. And when plugins are in the same Gradle multi-module project (like our server specific plugins), the same configurations could be used when defining the project dependencies.

Would you think it's worth to spend some time investigating if this is doable?

oheger-bosch:
> It's because of the two Gradle plugin managers and because Gradle does not inherit repositories from dependencies. It could be added to a convention plugin in `buildSrc`, but if the approach from this PR is used it would not be required, right?

Correct, then this repository only needs to be declared in the plugin-info project.

buildSrc/src/main/kotlin/PluginInfoCollectorPlugin.kt
oheger-bosch: Removed the `try` construct, so that exceptions will be thrown.

buildSrc/src/main/kotlin/PluginInfoCollectorPlugin.kt
oheger-bosch: Yes, dropped. Initially, files were collected in different folder structures derived from their origins.

mnonnenmacher:
> Regarding the proposal to publish a dedicated artifact based on custom configurations, I am not sure I fully got this with my limited Gradle knowledge. So, would this basically mean that there is some mechanism in place (as a plugin provided by ORT?) that collects the KSP-generated plugin information and publishes them as a module with separate coordinates? Similar to the plugin implemented here, but it would only deal with projects from the same build?

Something in that direction, I haven't fully thought it through yet, the goal is to make it easy to consume only the plugin metadata without the implementation and transitive dependencies if we found a valid use case. I would look into this after my holidays and approve this PR to unblock the admin config change.

> In an early iteration of this PR, I was using a task that depends on the KSP tasks of all projects and collected the generated files if they were plugin descriptions. I then dropped this, as I realized that external dependencies and project dependencies could be handled in the same way to extract plugin information from their META-INF folder. But would this go in the direction you have in mind?

Not directly, the basic idea is to provide the plugin information as separate artifacts that you can simply depend on, without any custom logic on the consumer side.

mnonnenmacher (approved):

Issue #3293, opened 2025-08-06
Group IDs should be implemented as an enum


sschuberth:
> @wkl3nk your PR #753 implemented the "Groups" endpoints, for which I am now implementing the UI support. The `groupId` parameter is not specified as an `enum` class, but a `string` (although semantically it is of enum type, as there is only a finite number of groups available). This means we need to repeat the possible groups in the UI code when validating the add/remove user forms:
> 
> ```
> 
> const GroupEnum = z.enum(['admins', 'writers', 'readers']);
> 
> ```
> 
> As described in this issue, it would reduce repetitive code if the `groupId` would use an enum type. 

 _Originally posted by @Etsija in [#281](https://github.com/eclipse-apoapsis/ort-server/issues/281#issuecomment-2354513997)_

PR #3332, opened 2025-08-11
Authorization refactorings

mnonnenmacher:
Prepare for fixing #3293 with some refactorings to the authorization service and API. The main change is to remove the group semantics from the API because groups are only used in Keycloak for convenience, the actual semantics of the authorization system is that a user can have one role per hierarchy entity (organization, product, role).

The remaining work will be done in a follow-up PR to not make this PR even larger than it already is.


services/authorization/src/main/kotlin/AuthorizationService.kt
```diff
@@ -105,13 +105,22 @@ interface AuthorizationService {
     suspend fun ensureSuperuserAndSynchronizeRolesAndPermissions()
 
     /**
-     * Add the [role] for the hierarchy element with the given [hierarchyId] to the user with the given [username].
+     * Add the [role] for the hierarchy element with the given [hierarchyId] to the user with the given [username]. If
+     * the user already has another role for the same hierarchy element, it will be replaced by the new [role].
      */
-    suspend fun <ID : HierarchyId> addUserRole(username: String, hierarchyId: ID, role: Role<ID>)
+    suspend fun <TYPE : Role<TYPE, ID>, ID : HierarchyId> addUserRole(
```
oheger-bosch, line 111: If there is now only a single role, the names `addUserRole` and `removeUserRole` are maybe a bit misleading. Should these methods be renamed to something like `setUserRole` and `unsetUserRole`?

components/authorization/api-model/src/commonMain/kotlin/Roles.kt
```diff
@@ -0,0 +1,34 @@
...
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.components.authorization.api
+
+import kotlinx.serialization.Serializable
+
+/** An enum for the available organization roles. */
+@Serializable
+enum class OrganizationRole { READER, WRITER, ADMIN }
```
oheger-bosch, line 26: I am not sure whether these different enum classes actually provide some value. For the REST API, the user passes a string; so, there is no improvement in type-safety. Whether this string is converted directly to a model role or via the indirection through the mappings, does not make a big difference, does it?

core/src/main/kotlin/api/OrganizationsRoute.kt
```diff
@@ -263,16 +265,14 @@ fun Route.organizations() = route("organizations") {
...
-                put(putUserToOrganizationGroup) {
+        route("roles") {
+            route("{role}") {
+                put(putOrganizationRoleToUser) {
                     requirePermission(OrganizationPermission.MANAGE_GROUPS)
 
                     val user = call.receive<Username>()
                     val organizationId = call.requireIdParameter("organizationId")
-                    val groupId = call.requireParameter("groupId")
+                    val role = call.requireEnumParameter<OrganizationRole>("role").mapToModel()
```
oheger-bosch, line 275: Renaming the parameter is a breaking change of the REST API. Is there a way to make this more obvious? At the very least, the commit title should have a "!".

services/authorization/src/main/kotlin/AuthorizationService.kt
mnonnenmacher: Good point, if that's ok I will add this to my todo list for the follow-up PR to avoid the rebase work. I think `remove` is still ok but `add` needs to be replaced.

Etsija:
@mnonnenmacher I believe the reference to my non-related PR should still be fixed in the description.

mnonnenmacher:
> @mnonnenmacher I believe the reference to my non-related PR should still be fixed in the description.

@Etsija I have fixed it.

PR #3342, opened 2025-08-13
Commit openapi.json to repository

mnonnenmacher:
Commit the openapi.json to the repository to:
* Help frontend devs so that they do not have to generate the file manually each time the API changes.
* Make API changes more visible in PR diffs.

@eclipse-apoapsis/frontend-devs As discussed, this would remove the need for you to run Gradle most of the times.
@eclipse-apoapsis/backend-devs Please provide feedback on how much it would annoy you if you have to run `./gradlew :core:generateOpenApiSpec` when you change the API.


oheger-bosch:
Adding generated files to version control is typically considered an anti-pattern. I would therefore first like to know whether there are other options to achieve the mentioned goals:
* Can the call to gradle to generate the OpenAPI spec be automated in the frontend developer workflow, e.g. by invoking it as a script from pnpm?
* To detect changes in the REST API, there are tools to compare OpenAPI files, e.g. [oasdiff](https://github.com/oasdiff/oasdiff), which can be integrated into GitHub actions. Would this be an alternative?

sschuberth:
> Please provide feedback on how much it would annoy you if you have to run `./gradlew :core:generateOpenApiSpec` when you change the API.

I would annoy me quite a bit, and it's also easy to forget.

> Adding generated files to version control is typically considered an anti-pattern.

I agree with @oheger-bosch here.

If the goal is to "remove the need for [@eclipse-apoapsis/frontend-devs] to run Gradle most of the times", then I wonder whether the `openapi.json` generation could be integrated into a command that they need to run anyway (like `pnpm dev` maybe).

mnonnenmacher:
> Can the call to gradle to generate the OpenAPI spec be automated in the frontend developer workflow, e.g. by invoking it as a script from pnpm?

> If the goal is to "remove the need for [@eclipse-apoapsis/frontend-devs] to run Gradle most of the times", then I wonder whether the `openapi.json` generation could be integrated into a command that they need to run anyway (like `pnpm dev` maybe).

IIRC the main complaints were having to remember to run the task but also that is takes too long to run (I guess especially when you don't run Gradle builds frequently anyway which would also be an argument against including it in `pnpm dev`). Maybe @Etsija and @lamppu could comment on that.

lamppu:
> > Maybe @Etsija and @lamppu could comment on that.
> 
> @Etsija, @lamppu, you did not reply here yet. WOuld it help you if `pnpm build` would call `./gradlew :core:generateOpenApiSpec` automatically?

I'm not really that bothered about running it separately and either way run gradle tasks locally quite often, so for me it's fine the way it is. If @Etsija would like to have it in `pnpm build`, I think that's ok, I mean it might be annoying if there's some build error coming from a dependency update that you need to debug locally and run `pnpm build` multiple times, but I mean you could just remove it for the duration of the debugging so I don't really care all that much.

PR #3347, opened 2025-08-14
ui: Infrastructure durations

Etsija:
This PR was a required feature from @sschuberth: to separate and show the time spent on "infrastructure" (for example, spinning up a Kubernetes pod) from the "actual" running time of the jobs.

The main calculation principle in this PR is as follows:
```
infrastructureDuration = totalRunDuration - sum(jobDurations)
```
where `totalRunDuration` is calculated from `[createdAt, finishedAt]` of the run, and `jobDurations` is calculated from `[startedAt, finishedAt]` of each respective job. Advisor and Scanner are typically run in parallel, so the longer of those jobs is included in calculation.

As the DoubleOpen team doesn't know the deep details of the state machines of the jobs and runs, I'd like to ask @mnonnenmacher (or anyone else more familiar with the back-end implementation): is this a sane assumption at all?

Screenshot from 2025-08-14 13-28-24

Screenshot from 2025-08-14 13-28-47

As we are missing a staging environment, there really is no means to test whether this feature shows with any reliability the time spent on infrastructure, before it is merged and deployed. But I'd still like to get an answer to whether my original assumption holds at all or not.

### DIFF
--- a/ui/src/components/charts/job-durations.tsx
+++ b/ui/src/components/charts/job-durations.tsx
@@ -48,6 +48,10 @@ import {
 import { toast } from '@/lib/toast';
 
 const chartConfig = {
+  infrastructure: {
+    label: 'Infrastructure',
+    color: 'gray',
+  },
   analyzer: {
     label: 'Analyzer',
     color: 'var(--analyzer)',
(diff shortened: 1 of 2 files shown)


Etsija:
> The Advisor and Scanner jobs typically run in parallel. So, the assumption that the total run duration correlates to the sum of all job durations is not correct. This would explain the negative numbers you mentioned. (The advisor job is normally rather fast, so it does not contribute that much to the overall duration.)

Thanks, I did forget the matter of Advisor and Scanner jobs running in parallel. I have to investigate, whether that explains the negativity of the result, at least part of it

> Also, I would have to look this up, but I doubt that the start and end times of jobs and runs are computed at points of time that allow determining infrastructure costs, at least not exactly when pods are started and terminate. So, I am not sure how these numbers could be calculated in a reliable way.

I would be very valuable if you could somewhat investigate the matter. We are (probably) not trying to find a second-perfect estimate of the "infrastructure duration" but rather get it in the right ballpark. The case is, recently we tuned down the specs of some of our pods, and as a result, the run durations chart _looks like_ all of a sudden, Analyzer job takes loads more time than before, while we do know (and would like to show/filter out) that the reason is increased pod spin-up time.

Etsija:
@mnonnenmacher @oheger-bosch in case we don't hear anything contrary from you by tomorrow (when Martin goes to summer vacation), we'd like to merge this tomorrow, to see whether it reveals anything useful in our deployment environment. Showing the infrastructure costs in the UI (the checkbox) is off by default, so no harm having the PR in so far, although it might come out that it doesn't help.

lamppu (approved):

PR #3376, opened 2025-08-19
ui: Set a custom favicon

Etsija:
Screenshot from 2025-08-19 13-17-38

### DIFF
--- a/ui/src/components/header.tsx
+++ b/ui/src/components/header.tsx
@@ -54,6 +54,7 @@ import {
   TooltipTrigger,
 } from '@/components/ui/tooltip';
 import { extractInitials } from '@/helpers/extract-initials.ts';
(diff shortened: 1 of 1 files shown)


lamppu:
I guess the default favicon will briefly flash for an unauthenticated user, but perhaps that's not a huge problem, as the app quite quickly redirects to the login page.

lamppu (changes requested):
There's an accidental merge commit here.

sschuberth:
> I guess the default favicon will briefly flash for an unauthenticated user, but perhaps that's not a huge problem, as the app quite quickly redirects to the login page.

It not happens for an unauthenticated user, but also if you refresh the page as an authenticated user. Any idea how to avoid that?

Issue #1920, opened 2025-01-30
Renaming organizations / products does not update the breadcrumb bar


sschuberth:
After clicking the "Submit" button, the new name does not show up immediately in the breadcrumb bar.

Etsija:
I noticed renaming also causes the page title to go out of sync. Couldn't find any solution to this by myself, unfortunately, although tried many...there must be a reason why it hasn't been fixed earlier.

Etsija:
PR #3380 resolves the issue - in a crude way.

PR #3380, opened 2025-08-20
ui: Fix updating the UI after renaming items

Etsija:

Also do note that this is a sledgehammer way of resolving this long-pending bug - many solutions have been tried before, none of which have worked.


Etsija:
GitHub fails to show the issue #1920 in the dropdown list, so I cannot link the issue with conventional ways to this PR.

PR #3409, opened 2025-08-26
fix(cli): Ignore white space around report file names

MarcelBochtler:


cli/src/commonMain/kotlin/ReportsCommand.kt
```diff
@@ -57,7 +56,7 @@ class ReportsCommand : SuspendingCliktCommand(name = "reports") {
         envvar = "OSC_DOWNLOAD_REPORTS_FILE_NAMES",
         help = "The names of the files to download, separated by commas. If not provided, all report files will be " +
                 "downloaded."
-    ).split(",")
+    ).convert { it.split(',').map(String::trim) }
```
oheger-bosch, line 59: Would be great to extend the existing test case to cover this issue.

cli/src/jvmTest/kotlin/ReportsCommandTest.kt
```diff
@@ -70,8 +71,12 @@ class ReportsCommandTest : StringSpec({
             coVerify(exactly = 1) {
                 runsMock.downloadReport(1, "example2.txt", any())
             }
+            coVerify(exactly = 1) {
+                runsMock.downloadReport(1, "example3.txt", any())
+            }
 
-            result.output.trimEnd() shouldBe "/tmp/output/example1.txt\n/tmp/output/example2.txt"
+            result.output.trimEnd() shouldBe
+                    "/tmp/output/example1.txt\n/tmp/output/example2.txt\n/tmp/output/example3.txt"
```
sschuberth, line 79: It seems a bit arbitrary to trim whitespace from the end, but leave `\n` inside the string... besides, I'm not sure whether this would work cross-platform.

So how about using ORT's `splitOnWhitespace()` helper function and comparing to a list of strings?

sschuberth: To the point, I propose to rewrite this as

```kotlin
result.output.splitOnWhitespace().shouldContainExactly(
    "/tmp/output/example1.txt",
    "/tmp/output/example2.txt",
    "/tmp/output/example3.txt"
)
```

to avoid hard-coding platform-specific `\n` and the `trimEnd()` call.

MarcelBochtler: This would cause the test to succeed if the output is:
```
/tmp/output/example1.txt /tmp/output/example2.txt /tmp/output/example3.txt
```

Which would be incorrect.

The reason for this assertion is actually to check if every file is being printed, as well as checking that they are printed on separate lines.
So the hard coded `\n` is required and working as intended. I would like to keep this as-is.

sschuberth: > So the hard coded `\n` is required and working as intended. I would like to keep this as-is.

But would you agree that ideally `\r\n` should be supported as well in case someone runs this test on Window? So like

```kotlin
result.output.trim().lines().shouldContainExactly(
    "/tmp/output/example1.txt",
    "/tmp/output/example2.txt",
    "/tmp/output/example3.txt"
)
```

sschuberth (approved):

sschuberth (approved):

PR #3410, opened 2025-08-26
build: Enable Detekt's autoCorrect

MarcelBochtler:
This allows the developer to use `./gradle detekt --auto-correct` to fix trivial formatting issues.


buildSrc/src/main/kotlin/ort-server-kotlin-conventions.gradle.kts
```diff
@@ -61,6 +61,7 @@ detekt {
 }
 
 tasks.withType<Detekt>().configureEach {
+    autoCorrect = true
```
sschuberth, line 64: Why is this done here instead of in `.detekt.yml` like done in ORT?

PR #3438, opened 2025-09-01
build: Configure maven central for plugin management

MarcelBochtler:
By default, Gradle uses `plugins.gradle.ort/m2` in order to download plugins. Unfortunately, this might result to forwarding the request to Maven Central for artifacts that cannot be found on `plugins.gradle.org`.
This forwarding causes rate limiting for corporations trying to scan the ORT Server with the ORT Server which will then try to download artifacts from Maven Central [1].

More information can be found in [2].

Specifically, adding `mavenCentral()` as plugin repository will trigger the workaround implemented in [3] where `mavenCentral()` will be replaced by a corporate mirror of Maven Central.

[1]: https://www.sonatype.com/blog/beyond-ips-addressing-organizational-overconsumption-in-maven-central
[2]: https://blog.gradle.org/maven-central-mirror
[3]: https://github.com/eclipse-apoapsis/ort-server/pull/3304

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -25,6 +25,7 @@ plugins {
 }
 
 repositories {
+    mavenCentral()
     // Allow resolving external plugins from precompiled script plugins.
     gradlePluginPortal()
 }
--- a/buildSrc/settings.gradle.kts
+++ b/buildSrc/settings.gradle.kts
@@ -17,6 +17,13 @@
  * License-Filename: LICENSE
  */
 
+pluginManagement {
+    repositories {
+        mavenCentral()
+        gradlePluginPortal()
+    }
+}
+
 rootProject.name = "buildSrc"
 
 dependencyResolutionManagement {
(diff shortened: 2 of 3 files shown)


mnonnenmacher:
As an alternative, maybe it is possible to add Maven Central as a plugin repository in the `init.gradle.kts` template, then this should work as a generic solution for all projects?

mnonnenmacher:
> > Specifically, adding `mavenCentral()` as plugin repository will trigger the workaround implemented in [3](https://github.com/eclipse-apoapsis/ort-server/pull/3304) where `mavenCentral()` will be replaced by a corporate mirror of Maven Central.
> 
> I thought that work-around would only apply to the _runtime_ of ORT Server. But this change applies to the _build time_ of ORT Server, or?

The problem is that the [init scrip template](https://github.com/eclipse-apoapsis/ort-server/blob/main/workers/common/src/main/resources/init.gradle.kts) allows configuring a mirror for Maven Central, but this does not work for the automatic fallback of the Gradle plugin portal to Maven Central (it redirects to Maven Central if a plugin is requested which it does not contain). So adding Maven Central as a plugin repository before the plugin portal works around that because all plugins contained in Maven Central are not requested from the plugin portal anymore. This allows us to analyze this repository in our infrastructure where we would otherwise get 429 responses from Maven Central.

However, I think it would be better to find a generic solution that does not require modifying the analyzed projects, like adapting the init script to inject Maven Central as a plugin repository (if this works).

PR #3494, opened 2025-09-10
ui: Change text filter behaviour

Etsija:


ui/src/components/data-table/filter-text.tsx
```diff
@@ -17,9 +17,8 @@
  * License-Filename: LICENSE
  */
 
-import debounce from 'debounce';
 import { Filter, XCircle } from 'lucide-react';
-import { useEffect, useMemo, useState } from 'react';
+import { useEffect, useState } from 'react';
```
sschuberth, line 21: Commit message nit: "ie." -> "i.e."

sschuberth: Sorry, another nit: "behaviour" is British English, but we should use "behavior" (American English).

PR #3499, opened 2025-09-10
fix(api): Align calculation of issue count statistics

### LINKED ISSUE #3096: UI does not show all issues

wkl3nk:
Ensure the backend uses the same business logic for computing issue statistics as for retrieving detailed issue data. This removes inconsistencies in total and severity-based counts shown in the dashboard versus the issue list.

Fixes #3096.


wkl3nk:
The idea was to use the same code that also generates the issue items in the details view, just to make sure that the number of reported issues is consistent with the issue items in the details view.
However, while this **could** work for a **single** repository, the code is **very inefficient** when it would need to iterate over all repositories of a product, or even over all repositories of an organization.

To have an efficient solution, we need some architectural analysis and design how to solve this.

PR #3509, opened 2025-09-11
refactor(ui): Refactor plugins handling

Etsija:
This PR concludes the refactoring of the query client to use the hey-api instead of the old one.

After this PR and #3541 are merged, the old query client will be removed, and the error handling will be returned to have the more informative error messaged from back-end again.

This picture shows that the only import from the old query client is the `ApiError` which will be the topic of the final refactoring follow-up PR.
Screenshot from 2025-09-17 07-28-27

### DIFF
--- a/ui/src/routes/admin/plugins/$pluginType/$pluginId/create-template/index.tsx
+++ b/ui/src/routes/admin/plugins/$pluginType/$pluginId/create-template/index.tsx
@@ -19,6 +19,7 @@
 
 import { zodResolver } from '@hookform/resolvers/zod';
 import { CheckedState } from '@radix-ui/react-checkbox';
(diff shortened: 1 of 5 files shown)


mnonnenmacher (dismissed):
Changes look good to me, @lamppu do you want to have a look as well?

Etsija:
> 
> Actually an empty value is not allowed for secrets. Anyway, I did lots of testing and could not find any issue so I'll mark this as ready for review. I will do more testing for my other PR, so if I missed something I should find it when working on it.

So if an empty value is not allowed, the "optional" should be removed from the form?

mnonnenmacher:
@Etsija I think the issue you had with deleting templates is the one fixed in #3544.

mnonnenmacher (approved):

Issue #3314, opened 2025-08-08
Issues query sometimes makes the server unresponsive


lamppu:
We have a problem when navigating to see the issues of some runs, the server crashes. This seems to happen consistently for specific runs.

I've verified with Postman that making a single request doesn't overwhelm the server. For one example, the request takes about 40-50 seconds but eventually succeeds. (I gathered from the logs of one such request, that the database queries took about 8 seconds in total and the file list requests to S3 took about 46 seconds.)

When navigating to the issues page in the UI for the same run though, ~there is some default retry logic that TanStack Query does, and the same request is retried 3 times by default within a few seconds~ (I was wrong about this), and then the server crashes.

I'm seeing these kinds of warnings before the crash (similar warnings to these are logged tens of times):

```
time=2025-08-08T11:32:03.767 thread=eventLoopGroupProxy-4-1  level=WARN  logger=i.n.c.AbstractChannelHandlerContext message="Failed to submit an exceptionCaught() event."
java.util.concurrent.RejectedExecutionException: event executor terminated
	at io.netty.util.concurrent.SingleThreadEventExecutor.reject(SingleThreadEventExecutor.java:1005)
	at io.netty.util.concurrent.SingleThreadEventExecutor.offerTask(SingleThreadEventExecutor.java:388)
	at io.netty.util.concurrent.SingleThreadEventExecutor.addTask(SingleThreadEventExecutor.java:381)
	at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:907)
	at io.netty.util.concurrent.SingleThreadEventExecutor.execute0(SingleThreadEventExecutor.java:873)
	at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:863)
	at io.netty.channel.AbstractChannelHandlerContext.fireExceptionCaught(AbstractChannelHandlerContext.java:275)
	at io.ktor.server.netty.NettyApplicationCallHandler.exceptionCaught(NettyApplicationCallHandler.kt:66)
	at io.netty.channel.AbstractChannelHandlerContext.invokeExceptionCaught(AbstractChannelHandlerContext.java:290)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelUnregistered(AbstractChannelHandlerContext.java:196)
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:148)
[...]

PR #3537, opened 2025-09-16
refactor(services): Don't load file lists for result on IssueService

lamppu:
Loading the file lists is not necessary in order to access the needed functions in the `OrtResult` in this particular case. Add a `loadFileLists` flag for the `generateOrtResult` function to allow skipping the loading of the file lists, while still allowing to fetch other data from the database for the `ScannerRun` object.


sschuberth:
Should this already close #3314?

PR #3540, opened 2025-09-16
Return purl for organization and product vulnerabilities

lamppu:

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/OrganizationVulnerability.kt
+++ b/api/v1/model/src/commonMain/kotlin/OrganizationVulnerability.kt
@@ -32,6 +32,9 @@ data class OrganizationVulnerability(
     /** The package [Identifier]. */
     val identifier: Identifier,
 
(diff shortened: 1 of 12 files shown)


api/v1/model/src/commonMain/kotlin/OrganizationVulnerability.kt
```diff
@@ -45,5 +45,8 @@ data class OrganizationVulnerability(
      * Count of repositories that the [Vulnerability] is found in. Relative to the restrictions of the query that
      * provides the information.
      */
-    val repositoriesCount: Long
+    val repositoriesCount: Long,
+
+    /** The purl of the package that the [Vulnerability] is found in. */
+    val purl: String
```
sschuberth, line 51: @mnonnenmacher and @lamppu, is there a need / convention to add new properties for these API classes only at the end of the class? Otherwise putting `purl` next to `identifier` would probably be more readable.

sschuberth (approved):

PR #3549, opened 2025-09-17
refactor(ui): Query client refactor cleanup

Etsija:
This PR finally concludes (at least the main part of) the UI query client refactoring process. The old client is removed and the new client is installed back to `@/api`. Also the more informative error messages are back.

I I calculated correctly, the refactoring totalled to some 120 files.

I was actually able to retain the buildability of the UI after each commit, although it wasn't trivial and we discussed today with the team, whether an exception to the rule would be necessary.

Screenshot from 2025-09-17 14-04-19

### DIFF
--- a/ui/.eslintrc.cjs
+++ b/ui/.eslintrc.cjs
@@ -25,7 +25,7 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'plugin:react-hooks/recommended',
   ],
-  ignorePatterns: ['dist', '.eslintrc.cjs', 'src/api/', 'src/hey-api/'],
+  ignorePatterns: ['dist', '.eslintrc.cjs', 'src/api/'],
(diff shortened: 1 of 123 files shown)


ui/src/components/toast-error.tsx
```diff
@@ -17,39 +17,26 @@
...
+    cause = error.response?.data.cause;
   } else {
     message = 'An unknown error occurred.';
     cause = 'Please try again.';
   }
   return (
     <div className='grid gap-2'>
       <div>{message}</div>
-      <div className='break-all'>{cause}</div>
+      <div className='break-all'>{String(cause)}</div>
```
lamppu, line 39: Did this actually only need the surrounding String() when you were testing with something else?

lamppu (changes requested):

Issue #3177, opened 2025-07-16
Consider auto-generating zod schemas from Kotlin data classes


sschuberth:
See e.g. https://github.com/guimauvedigital/zodable.

lamppu:
> What would those generated schemas be used for?

For example, these kinds of [zod enum schemas](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/schemas/index.ts#L31-L38) could be auto-generated. I think there also might be cases where you could auto-generate some zod object schemas that could be used as form schemas (although sometimes with forms it's not one-to-one).

PR #3559, opened 2025-09-18
refactor(ui): Remove the "hand-written" validation schemas from the UI 

Etsija:

### DIFF
--- a/ui/openapi-ts.config.ts
+++ b/ui/openapi-ts.config.ts
@@ -24,5 +24,5 @@ export default defineConfig({
   output: {
     path: './src/api',
   },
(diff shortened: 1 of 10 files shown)


Etsija:
> This could probably close #3177.

Good catch, I'll mark it to resolve the issue.

sschuberth:
> to auto-generate Zod-enum themselves already.

Ok, double-checking this myself, that indeed seems to be the case, as this import comes from `zod.*gen*`

    import { zOrtRunStatus } from '@/api/zod.gen';

and the file is not there until running `pnpm -C ui generate:api` 👍🏻

PR #3564, opened 2025-09-19
ui: Show metadata and modified flags for packages

Etsija:
Screenshot from 2025-09-19 10-44-56

Screenshot from 2025-09-19 10-46-17

Please see the commit for details (and wording of the tooltips, which is subject to discussion).

After a lengthy discussion with the DoubleOpen team, it was decided _not_ so show these "exception" flags when they are `false`, after all, to not overwhelm the user with package data which is "normal" (i.e. most of the packages are probably non-modified and contain the source in some form).

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/packages/index.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/packages/index.tsx
@@ -99,24 +99,47 @@ const renderSubComponent = ({
 
   return (
     <div className='flex flex-col gap-4'>
-      {pkg.isModified && (
-        <div>
-          <Tooltip>
-            <TooltipTrigger>
-              <Badge
-                className={`border ${getIssueSeverityBackgroundColor('HINT')}`}
-              >
-                MODIFIED
-              </Badge>
-            </TooltipTrigger>
-            <TooltipContent>
-              The source code of the package has been modified compared to the
-              original source code, e.g., in case of a fork of an upstream Open
-              Source project.
-            </TooltipContent>
-          </Tooltip>
-        </div>
-      )}
+      {pkg.isModified ||
+        (pkg.isMetadataOnly && (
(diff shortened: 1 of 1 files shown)


sschuberth:
As in contrast to being "modified", there's nothing potentially "bad" about a "metadata-only" package, should we just use a blue color for that badge?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/packages/index.tsx
```diff
@@ -99,24 +99,46 @@ const renderSubComponent = ({
...
+                  <TooltipTrigger>
+                    <Badge
+                      className={`border ${getIssueSeverityBackgroundColor('WARNING')}`}
+                    >
+                      MODIFIED
+                    </Badge>
+                  </TooltipTrigger>
+                  <TooltipContent>
+                    The package has been modified compared to the original
+                    package, e.g., in case of a fork of an upstream Open Source
```
sschuberth, line 118: Nit: I believe there should be no comma after "e.g.".

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/packages/index.tsx
```diff
@@ -99,24 +99,46 @@ const renderSubComponent = ({
...
+                <Tooltip>
+                  <TooltipTrigger>
+                    <Badge
+                      className={`border ${getIssueSeverityBackgroundColor('WARNING')}`}
+                    >
+                      METADATA ONLY
+                    </Badge>
+                  </TooltipTrigger>
+                  <TooltipContent>
+                    This is a metadata only package.
```
sschuberth, line 135: Maybe a bit more verbose (and "metadata-only" with a hypthen): "This is a metadata-only package that has no source code associated to it.".

sschuberth (changes requested):

Issue #3370, opened 2025-08-18
Add search function for organization, product, and repository


hanna-modica:
Since our Server is growing, it is becoming harder to navigate to the desired organziation with only pagination. I would like to have a search bar in the UI, where I can search/filter for an organization. When inside an organization, I would like to search for a product and in the same way for a repository, when on that level.
Bonus points, if I could globally search for a repository for all organizations that are visible to me. If this should be a separate issue ticket, I can open it.

hanna-modica:
Hi @Etsija,
yes, if the label means, that the ui will receive such search bars. Thank you.

Etsija:
That's what I expected, as there were no filtering options in the `openapi.json`, either.

PR #3566, opened 2025-09-19
fix(core): Add missing filters to API docs

mnonnenmacher:
This is a fixup for 14c5f17.


Etsija:
> > Thanks for a speedy response! For reference to the reviewers, I need this PR to advance #3370.
> 
> What about reviewing it yourself then, @Etsija 😉 It seems to be trivial enough even for non-Kotlin-natives 😁

Well I already did, but wasn't aware that I have the rights to approve...

Etsija (approved):

sschuberth (approved):

PR #3569, opened 2025-09-19
docs(website): Add guide for configuring environment variables

wkl3nk:
Add a section to the User Guide explaining how to configure environment variables for package managers. Mention both options: referencing secrets from the Secrets Provider or using plain values.

### DIFF
--- a/website/docs/user-guide/concepts/environment-variables.md
+++ b/website/docs/user-guide/concepts/environment-variables.md
@@ -1,0 +1,71 @@
+# Environment Variables
+
+In the [Analyzer](https://oss-review-toolkit.org/ort/#analyzer)
+step of the [OSS Review Toolkit](https://oss-review-toolkit.org/ort/) pipeline,
+dependencies of your project are identified, and
+information about them is gathered, such as declared license, authors and source code location.
+
+In detail, package managers are used by ORT to analyze project dependencies.
+Since it is **not** appropriate to **hardcode** credentials or other sensitive information in files within your project’s
+source code repository, CI/CD pipelines typically provide such settings through environment variables instead.
+The following sections describe how to define these environment variables.
+
+## General usage
+
+To define these environment variables, use the `.ort.env.yml` file in your project repository. This also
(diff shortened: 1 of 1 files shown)


website/docs/user-guide/concepts/environment-variables.md
```diff
@@ -0,0 +1,68 @@
+# Environment Variables
+
+In the [Analyzer](https://oss-review-toolkit.org/ort/#analyzer)
+step of the [OSS Review Toolkit](https://oss-review-toolkit.org/ort/) pipeline,
+dependencies of your project are identified, and
+information about them is gathered, such as declared license, authors and source code location.
+
+In detail, package managers perform this task. As in most cases it is **not** appropriate to configure these package
+managers by files that are located in the source code repository of your project, typically environment variables
+are used by CI/CD pipelines and also by the OSS Review Toolkit pipeline.
```
mnonnenmacher, line 10: > As in most cases it is **not** appropriate to configure these package managers by files that are located in the source code repository...

That part is unclear to me, why is it not appropriate in "most cases"? It's also a bit weird because the solution below is to add a file to the repository.

website/docs/user-guide/concepts/environment-variables.md
```diff
@@ -0,0 +1,68 @@
...
+## General usage
+
+To define these environment variables, use the `.ort.env.yml` file in your project repository. This also
+ensures that the configuration is version controlled and provides full traceability for changes over time.
+
+There are two ways to define environment variables in the `.ort.env.yml` file:
+
+- For **sensitive** values, use a secret reference that references to a value in your _Secrets Provider_.
+  Of course, the reference must exist in your _Secrets Provider_ (e.g. HashiCorp Vault).
+  These references are resolved just before the scan runs.
```
mnonnenmacher, line 21: For users the secrets backend (like Vault) is not visible, all they see is the UI and API for manging secrets, so better mention that (and later it can link to the secrets docs, once created).

mnonnenmacher (changes requested):

PR #3582, opened 2025-09-22
chore(ort): Add path exclusion for test projects

MarcelBochtler:


.ort.yml
```diff
@@ -21,6 +21,14 @@ excludes:
...
     comment: >-
       The buildSrc directory contains build scripts for the Gradle build tool.
+  - pattern: "shared/ort-test-data/**"
+    reason: "TEST_OF"
+    comment: >-
+      Projects in this directory are used for testing only and are not distributed.
+  - pattern: "utils/test/**"
+    reason: "TEST_OF"
+    comment: >-
+      Projects in this directory are used for testing only and are not distributed.
```
mnonnenmacher, line 31: Currently both projects are published as artifacts to Maven Central, for example:
https://github.com/eclipse-apoapsis/ort-server/blob/f36d01a9c105499f94964dd7682da7afab0d10e7/shared/ort-test-data/build.gradle.kts#L23

I think we should go through the projects and decide which really need to be published (I think many do not have to be), but right now these statements are not correct.

mnonnenmacher (changes requested):

MarcelBochtler:
Closing.
@mnonnenmacher is correct and the modules are indeed published.

Issue #3587, opened 2025-09-23
GitHub-like layout for items tables


Etsija:
For brevity, in this issue, "items" currently refer to vulnerabilities, rule violations, issues, packages and projects, for which we have table views.

As discussed in #3403 (and probably in several other tickets), we need to consider switching from showing item data in the corresponding tables in columns to GitHub-like "data cards". The reason is, the more data we want to show in the table itself (instead of the expandable sub-row), the more crowded the table becomes horizontally, with ever increasing number of sortable and filterable columns.

When it comes to TanStack Table, which is the headless table library we use in the core of the tables, sorting and filtering are inherently built around columns, by manipulation of [ColumnSort](https://tanstack.com/table/latest/docs/guide/sorting#sorting-state) and [ColumnFilter](https://tanstack.com/table/latest/docs/guide/column-filtering#column-filter-state). With GitHub-like "card tables", of which an example is shown below,

<img width="1136" height="298" alt="Image" src="https://github.com/user-attachments/assets/ed9170ae-697e-48c9-bcf4-935e7fa98198" />

the whole concept of "columns" loses its meaning; all information is presented in only one column. We probably would have tables with only two columns:
- the leftmost column would be the "Details" column, which includes the down-caret to expand the sub-row for more information, and the chain icon for copying the item's URL to clipboard
- the rightmost column would show all data (text, indicators etc.) that needs to be seen on top-level

In the picture below, this would mean that "Details" would be one column, and all data (severity, status, package ID, rule etc.) would be another.

<img width="1039" height="294" alt="Image" src="https://github.com/user-attachments/assets/f2682847-61b3-488b-9ff5-66b2ebb3fa2b" />

Although, as I mentioned, TanStack Table "inherently" does sorting and filtering column-wise, it does also have a concept of [Sorting Functions](https://tanstack.com/table/latest/docs/guide/sorting#sorting-fns) and [Filtering Functions](https://tanstack.com/table/latest/docs/guide/column-filtering#filterfns). Both do take in as an argument the table row, which makes me believe this kind of "special" sorting and filtering of the data column would be technically possible.

I also do kind of like the way GitHub issues table works: the (one-column) table is filtered and sorted with the special elements rendered in the header. I believe it would be possible for us to achieve something similar.

PR #3599, opened 2025-09-24
fix(ui): Fix some problems with the issues table

Etsija:

Screenshot from 2025-09-24 12-45-59

NOTE: This kind of "dirty" UI fix very much emphasizes the need of #3587, as it becomes now very hard to work with columnized tables that are so crowded, and hacks are needed to resolve the overflow issues.

### DIFF
--- a/ui/src/helpers/get-issue-category.ts
+++ b/ui/src/helpers/get-issue-category.ts
@@ -35,7 +35,7 @@ const issueCategoryMap: Record<IssueCategory, RegExp[]> = {
     /Scan failed for job with ID '.*'/,
     /StreamResetException: stream was reset*/,
   ],
-  'Missing Data': [/IOException: Could not resolve provenance for .*/],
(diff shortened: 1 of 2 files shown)


sschuberth:
You should have two separate commits in this PR, one for each paragraph in the commit message:

> The issue message has changed into not having the "IOException" in the 
message anymore, so regex filter for "Missing Data" category didn't match
anymore, which lead to all "Could not resolve provenance for..." issue 
messages were categorized to "Other" category.

"so *the* regex"
"*does not* match anymore"
"which lead*s*"
"issue messages being categorized"

> Moreover, there was a bad horizontal overflow in the table for 
"problematic" package IDs, which even the component developed to solve
these cannot solve, so as a dirty fix, solve with a TailwindCSS class 
which crudely breaks the package ID from any point.

sschuberth (approved):

PR #3600, opened 2025-09-24
fix(dao): Add a missing ON DELETE CASCADE constraint

oheger-bosch:
This constraint was missing for the
`infrastructure_service_declarations_ort_runs` table preventing the deletion of runs with associated infrastructure service declarations.


dao/src/main/resources/db/migration/V116__deleteInfrastructureServiceDeclarationsCascade.sql
```diff
@@ -0,0 +1,7 @@
+-- This migration script adds a missing ON DELETE CASCADE to the infrastructure_service_declarations_ort_runs table.
+-- This is required to successfully delete ORT runs that have associated infrastructure service declarations.
+
+ALTER TABLE infrastructure_service_declarations_ort_runs
+  DROP CONSTRAINT infrastructure_service_declarations_ort_runs_ort_run_id_fkey,
+  ADD CONSTRAINT infrastructure_service_declarations_ort_runs
```
mnonnenmacher, line 6: For consistency, should this not use the same name as before?

PR #3604, opened 2025-09-25
ui: Tweak packages table

Etsija:
1. No repository information:

Screenshot from 2025-09-25 08-46-19

2. No source artifact:

Screenshot from 2025-09-25 08-47-27

3. Repository is of `UNKNOWN` type, the type information is rendered as a string:

Screenshot from 2025-09-25 08-49-48


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/projects/index.tsx
```diff
@@ -108,8 +108,7 @@ const renderSubComponent = ({ row }: { row: Row<Project> }) => {
 
       <div>
         <div className='font-semibold'>
-          {getRepositoryTypeLabel(project.vcsProcessed.type as RepositoryType)}{' '}
-          Repository
+          {getRepositoryTypeLabel(project.vcsProcessed.type)} Repository
```
sschuberth, line 111: I'm not sure whether it's correct to also omit the cast here, and thus using a string `type`. Have you verified that this would *not* cause the repository type to show up as "GIT" instead of "Git" for *projects* (your screenshots only refers to package details).

Etsija: No, they are shown as "Git". But I can omit the cast if that's required.

PR #3652, opened 2025-10-01
feat(cli): Make waiting for a run more robust

oheger-bosch:
Do not fail the waiting loop for the first error on a status check, but
only if there are multiple failures in sequence. This makes the polling
more stable against temporary network glitches.

### DIFF
--- a/cli/src/commonMain/kotlin/StartCommand.kt
+++ b/cli/src/commonMain/kotlin/StartCommand.kt
@@ -38,6 +38,7 @@ import com.github.ajalt.mordant.widgets.progress.progressBarContextLayout
 import com.github.ajalt.mordant.widgets.progress.spinner
 import com.github.ajalt.mordant.widgets.progress.text
 
+import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 
 import kotlinx.coroutines.delay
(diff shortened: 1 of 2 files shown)


cli/src/commonMain/kotlin/StartCommand.kt
```diff
@@ -62,9 +63,27 @@ import org.eclipse.apoapsis.ortserver.cli.utils.useJsonFormat
...
+ */
+private const val MAX_FAILURE_COUNT = 10
 
 class StartCommand : SuspendingCliktCommand(name = "start") {
+    companion object {
+        /**
+         * The interval to poll the server for updates when waiting for a run to finish. This is modifiable to allow
+         * tests to run faster.
+         */
+        private var pollInterval = 10.seconds
```
sschuberth, line 77: How about making this a command line option instead, and using that in the test?

oheger-bosch: We deliberately decided against this in the past, since we do not want to allow users to set the polling interval to a few nanoseconds.

cli/src/commonMain/kotlin/StartCommand.kt
```diff
@@ -62,9 +63,27 @@ import org.eclipse.apoapsis.ortserver.cli.utils.useJsonFormat
...
+
+        /**
+         * Set the polling interval to use when waiting for a run to finish. This is only intended to be used in tests.
+         */
+        internal fun setPollInterval(interval: Duration) {
+            pollInterval = interval
+        }
```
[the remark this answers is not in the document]
oheger-bosch: Encapsulation. Maybe I am old-fashioned, but when I see a property that is directly accessible, some alarm bells start ringing.

oheger-bosch:
> Should this be marked breaking as the `POLL_INTERVAL` environment variable is now not considered anymore?

It has never been documented and existed only for the reason to make the tests faster. Therefore, I don't think that this is actually a breaking change.

Ironically, that never worked; the tests did not use it, but tried to set a system property.

sschuberth:
> Ironically, that never worked; the tests did not use it, but tried to set a system property.

Yes, I was wondering about that as well 😅

PR #3672, opened 2025-10-06
Common base image for workers

oheger-bosch:
This PR introduces a common base container image to be used by all workers. Refer to the single commits for further details.


.github/workflows/docker-build.yml
```diff
@@ -47,13 +47,17 @@ jobs:
         #
         # Optional properties:
         # freeDiskSpace: Whether to free disk space before building the image.
+        # skipWorkerBaseImage: Whether to skip building the worker base image.
```
sschuberth, line 50: Maybe same this positively, e.g. "buildWorkerBaseImage"?

sschuberth: A general question: Why does this have to be modeled / optimized manually? Shouldn't the `:core:jibDockerBuild` task automatically skip building the base worker image if it the task is not depending on it?

Maybe you can detail this in the commit message more.

oheger-bosch:
Fixed a merge conflict. @sschuberth, @mnonnenmacher, is there anything left you want me to change?

workers/reporter/docker/Reporter.Dockerfile
```diff
@@ -37,40 +37,15 @@ RUN scancode-license-data --path /opt/scancode-license-data \
...
-    --shell /bin/bash \
-    --home-dir $HOMEDIR \
-    --create-home $USERNAME
-
-# Make sure the user executing the container has access rights in the home directory.
-RUN chgrp -R 0 /home/ort && chmod -R g+rwX /home/ort
-
-USER $USERNAME
-WORKDIR $HOMEDIR
+    && sudo rm -rf /var/lib/apt/lists/*
```
sschuberth, line 49: It it really intended that `sudo` is needed here now?

mnonnenmacher (dismissed):

build.gradle.kts
```diff
@@ -118,14 +118,21 @@ rootDir.walk().maxDepth(4).filter { it.isFile && it.extension == "Dockerfile" }.
             group = "Docker"
             description = "Builds the $name worker Docker image."
 
+            val imageName = if (name == "Base") {
+                "ort-server-base-image:latest"
+            } else {
+                dependsOn("buildBaseWorkerImage")
```
sschuberth: Nit: Probably also use an "absolute" task name here, with a leading `:`, like in line 157 below.

oheger-bosch:
> No, I'm still a bit puzzled about the requirement for the local registry but that's not a blocker to me. So the only question is if you have also verified that last iteration on a fork to make sure that the workflow doesn't brake? To me the two most important points are:
> 
>     * Nothing is pushed to the GitHub registry when the workflow runs for a PR.
> 
>     * Changes to main and release tags trigger a push to the GitHub registry.

What is new in this PR is that now worker base images built via Docker depend on a base image. While jib does not seem to have problems with resolving images built locally, this is not the case for Docker builds. They run in their own context and do not have access to images loaded elsewhere to a local Docker daemon. Therefore, the approach with the registry is necessary.

I have started a new run of the workflow on the fork, which was successful: https://github.com/boschglobal/ort-server/actions/runs/18865391520/job/53834976313

The logic when to push images is not touched by this PR. The new base image is never pushed (except for the local registry). So, this should be safe.

oheger-bosch:
> 
> But in the example here it works without a separate registry: https://docs.docker.com/build/ci/github-actions/named-contexts/#use-image-in-subsequent-steps

This example does not use a container builder as we do.

PR #3695, opened 2025-10-08
chore(compose): Make logging driver compatible

MarcelBochtler:
Not all container runtimes support `gelf` [1]. To support running the compose setup with e.g. Podman, make the logging driver configurable.

[1]: https://docs.podman.io/en/v5.0.1/markdown/podman-run.1.html#log-driver-driver


website/docs/admin-guide/getting-started/docker-compose.md
```diff
@@ -190,3 +190,11 @@ When upgrading the PostgreSQL version in the Docker Compose setup, the database
...
    docker compose up -d
    ```
+
+### Using Podman instead of Docker
+
+By default, the Compose setup uses a logging driver that is not supported by Podman. To disable central logging, set the `LOG_DRIVER` environment variable to `none`:
+
+   ```shell
+   LOG_DRIVER=none podman compose up
+   ```
```
sschuberth, line 200: I guess the indentation is problematic here.

PR #3777, opened 2025-10-22
Authorization component: authentication and authorization on route level

oheger-bosch:
This PR implements authorization support on route level for the new authorization component. Refer to the single commits for further details.


oheger-bosch:
@Etsija, yes, the plan is to have that fine-grained access control. To be honest, I was not aware that the current implementation does not support this.

Regarding filtering, the new component will also provide some functionality to determine the entities in the hierarchy for which a user has specific access rights. This is going to be used to retrieve overview lists (like the list of organizations, products, etc.) in a more efficient way. So far, all of these entities are fetched and then filtered in memory.

sschuberth:
> yes, the plan is to have that fine-grained access control.

Are there any ideas yet how to implemented that in practice? Like, would e.g. the role enums be extended with a `DENY` entry that can be used to restrict access on a lower hierarchy level again that was granted on a higher hierarchy level?

PR #3791, opened 2025-10-23
Authorization migration

oheger-bosch:
This PR implements a one-time migration to the new authorization component. The migration is triggered once on ORT Server startup. It creates entries in the database corresponding to the roles and permissions stored in Keycloak.

This should not be merged before all permission checks have been migrated to the new authorization component.


sschuberth:
Just curious, after the migration, will there be another follow-up PR that removes old code?

Issue #3815, opened 2025-10-27
Allow to specify environment configuration without `.ort.env.yml`


sschuberth:
For cases without write access to the repository, the approach of [committing a `.ort.env.yml` file](https://eclipse-apoapsis.github.io/ort-server/docs/user-guide/concepts/external-services#environment-definitions) to the repository to analyze does not work. While [the API basically offers a way](https://eclipse-apoapsis.github.io/ort-server/api/post-repository-run) to specify an `environmentConfig` object as part of the `jobConfigs` object for a particular job, this functionality is not exposed in the UI yet.

So we need to extend the form for analyzer options when creating a run to also allow for specifying an `environmentConfig`. This new section should probably come after the current "Environment variables" section.

sschuberth:
To clarify, an "environment configuration" may consist of
- environment definitions (e.g. package manager specific configuration files)
- environment variables (partly done already, see https://github.com/eclipse-apoapsis/ort-server/pull/1898)
- infrastructure services

PR #3829, opened 2025-10-28
ui: Inject infrastructure services into ORT runs

Etsija:
This PR takes one step towards solving #3815.

Screenshot from 2025-10-28 12-42-02

Screenshot from 2025-10-28 13-33-12

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/config/-components/analyzer-job-details.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/config/-components/analyzer-job-details.tsx
@@ -61,23 +61,24 @@ export const AnalyzerJobDetails = ({ run }: AnalyzerJobDetailsProps) => {
               value={jobConfigs.enabledPackageManagers}
               showIfEmpty={false}
             />
-            {jobConfigs.environmentConfig?.environmentVariables && (
-              <div>
-                <Label className='font-semibold'>Environment variables</Label>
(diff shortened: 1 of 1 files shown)


mnonnenmacher:
I checked how the environment config from the request is used. For cloning the repository, the infrastructure services from the hierarchy are combined with those from the request:

https://github.com/eclipse-apoapsis/ort-server/blob/6be938c74284e60017fe996f1095537eb88c80cf/workers/common/src/main/kotlin/common/env/EnvironmentService.kt#L77-L96

And for setting up the environment for the package managers both configurations are also combined:

https://github.com/eclipse-apoapsis/ort-server/blob/6be938c74284e60017fe996f1095537eb88c80cf/workers/common/src/main/kotlin/common/env/EnvironmentService.kt#L283-L312

So the documentation of `environmentConfig` is simply wrong and there is no need to add the infrastructure services that are already configured in the hierarchy to the run request again.

Etsija:
So, as a result, we should simply close this PR?

We have been _very_ confused with how this `EnvironmentConfig` works, that's why we have probably done some futile work to get what we need from the system.

oheger-bosch:
Initially, the environment configuration passed in the request had an override semantic. This has been changed later to merge the available data. Obviously, the documentation was not updated accordingly.

sschuberth:
> So, as a result, we should simply close this PR?

I believe so, yes. But let's discuss in the daily first.

> Obviously, the documentation was not updated accordingly.

I have updated it now (according to my understanding) as part of https://github.com/eclipse-apoapsis/ort-server/pull/3836.

Etsija:
I will close this PR now, and reuse some of its code in another PR.

PR #3846, opened 2025-10-30
Initial backend support for managing vulnerability resolutions on the server

lamppu:
This PR adds back-end support for making vulnerability resolutions on the server.

### DIFF
--- a/components/resolutions/api-model/src/commonMain/kotlin/PatchVulnerabilityResolution.kt
+++ b/components/resolutions/api-model/src/commonMain/kotlin/PatchVulnerabilityResolution.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.components.resolutions
+
+import kotlinx.serialization.Serializable
+
+import org.eclipse.apoapsis.ortserver.shared.apimodel.OptionalValue
+import org.eclipse.apoapsis.ortserver.shared.apimodel.VulnerabilityResolutionReason
+
+/**
+ * The request object for updating a vulnerability resolution.
+ */
+@Serializable
+data class PatchVulnerabilityResolution(
+    /**
+     * The list of vulnerability ID matchers (regular expressions) to match the ids of the vulnerabilities to resolve.
+     */
+    val idMatchers: OptionalValue<List<String>> = OptionalValue.Absent,
+
+    /** The reason why the vulnerability is resolved. */
+    val reason: OptionalValue<VulnerabilityResolutionReason> = OptionalValue.Absent,
+
+    /** A comment to further explain why the [reason] is applicable here. */
+    val comment: OptionalValue<String> = OptionalValue.Absent
+)
--- a/components/resolutions/backend/src/main/kotlin/VulnerabilityResolutionDefinitionService.kt
+++ b/components/resolutions/backend/src/main/kotlin/VulnerabilityResolutionDefinitionService.kt
@@ -31,6 +31,7 @@ import org.eclipse.apoapsis.ortserver.model.RepositoryId
 import org.eclipse.apoapsis.ortserver.model.UserDisplayName
 import org.eclipse.apoapsis.ortserver.model.VulnerabilityResolutionDefinition
 import org.eclipse.apoapsis.ortserver.model.VulnerabilityResolutionReason
+import org.eclipse.apoapsis.ortserver.model.util.OptionalValue
 import org.eclipse.apoapsis.ortserver.model.util.asPresent
 import org.eclipse.apoapsis.ortserver.services.ortrun.OrtRunService
 
@@ -94,6 +95,32 @@ class VulnerabilityResolutionDefinitionService(private val db: Database, private
         ortRunService.markAsOutdated(
             getAffectedRuns(id) + definition.contextRunId,
             "Vulnerability resolution definition restored."
+        )
+
+        definition
+    }
+
+    suspend fun update(
+        id: Long,
+        userDisplayName: UserDisplayName,
+        idMatchers: OptionalValue<List<String>> = OptionalValue.Absent,
+        reason: OptionalValue<VulnerabilityResolutionReason> = OptionalValue.Absent,
+        comment: OptionalValue<String> = OptionalValue.Absent
+    ): VulnerabilityResolutionDefinition = db.dbQuery {
+        VulnerabilityResolutionDefinitionsTable.updateDefinition(
+            id,
+            idMatchers,
+            reason,
+            comment
+        )
+
+        addChangeLogEvent(id, ChangeEventAction.UPDATE, userDisplayName)
+
+        val definition = VulnerabilityResolutionDefinitionsTable.get(id)
+
+        ortRunService.markAsOutdated(
+            getAffectedRuns(id) + definition.contextRunId,
+            "Vulnerability resolution definition updated."
         )
 
         definition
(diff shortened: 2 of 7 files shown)


components/resolutions/api-model/src/commonMain/kotlin/CreateVulnerabilityResolutionDefinition.kt
Etsija: As part of #3712 @mnonnenmacher asked me to also change the naming of all request objects to comply with the Get/Post/Patch/Delete paradigm. So I guess these data classes should be named `PostVulnerabilityResolutionDefinition` and `PatchVulnerabilityResolutionDefinition`, accordingly. I don't think this would lead to naming clash with the endpoint naming, as these request objects stay in the `api-model`, while the route definitions are on `backend`.

Etsija (changes requested):

api/v1/model/src/commonMain/kotlin/OrtRun.kt
```diff
@@ -138,7 +138,17 @@ data class OrtRun(
...
+
+    /**
+     * A flag to indicate if the results of the run are outdated, e.g. because of a new resolution.
+     */
+    val outdated: Boolean = false,
+
+    /**
+     * A message describing why the results of the run are outdated.
+     */
+    val outdatedMessage: String? = null
```
sschuberth: Do we need both `outdated` and `outdatedMessage`? Theoretically, we could define that a run is outdated if `outdatedMessage` is non-`null`, and never allow `outdatedMessage` to be `null` for and outdated run (i.e. outdated runs always need to provide a message, which is probably good practice anyway).

On the other hand, being explicit with `outdated` also has its advantages, but complicates the logic a bit, I guess.

model/src/commonMain/kotlin/ChangeEvent.kt
```diff
@@ -0,0 +1,53 @@
...
+}
+
+/**
+ * An enumeration of the actions that can be performed, resulting in a [ChangeEvent].
+ */
+enum class ChangeEventAction {
+    CREATE,
+    UPDATE,
+    ARCHIVE,
+    RESTORE
```
sschuberth: It would be great to have docs for these, esp. the latter two are a bit unclear to me.

model/src/commonMain/kotlin/VulnerabilityResolutionReason.kt
```diff
@@ -0,0 +1,67 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.model
+
+/**
+ * Possible reasons for resolving an [Vulnerability] using a [VulnerabilityResolution].
+ */
+enum class VulnerabilityResolutionReason {
```
sschuberth: Also see the [PR discussion in ORT](https://github.com/oss-review-toolkit/ort/pull/10886) to "Rework VulnerabilityResolutionReason for CRA / DORA requirements". So these enums are likely to change. I was briefly wondering whether we should already use that we believe to be "the right solution" here, abstracting away what ORT currently uses until the PR is merged. But it's probably easier (for the developers) to keep these in sync with ORT.

components/resolutions/backend/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
...
+
+    testImplementation(ktorLibs.serialization.kotlinx.json)
+    testImplementation(ktorLibs.server.auth)
+    testImplementation(ktorLibs.server.contentNegotiation)
+    testImplementation(ktorLibs.server.statusPages)
+    testImplementation(ktorLibs.server.testHost)
+    testImplementation(libs.kotestAssertionsKtor)
+    testImplementation(libs.kotestRunnerJunit5)
+    testImplementation(libs.kotlinxSerializationJson)
+    testImplementation(libs.mockk)
```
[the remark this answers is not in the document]
Etsija: That was exactly the task I was also missing, thanks! But what's the "project" here? So if I want to do `projectHealth` for, say, components/secrets, how would I use it?

components/resolutions/backend/src/routes/kotlin/routes/vulnerabilities/RestoreVulnerabilityResolution.kt
```diff
@@ -0,0 +1,128 @@
...
+        HttpStatusCode.NoContent to {
+            description = "The vulnerability resolution is not archived."
+        }
+    }
+}) {
+    val id = call.requireIdParameter("id")
+
+    val definition = vulnerabilityResolutionDefinitionService.getById(id)
+
+    if (definition == null) throw AuthorizationException()
```
sschuberth: Similar `?: throw` shortcut could be made.

api/v1/model/src/commonMain/kotlin/OrtRun.kt
lamppu: I'm not sure, this was something I was wondering about too.

model/src/commonMain/kotlin/VulnerabilityResolutionReason.kt
lamppu: I'm afraid that that would cause some errors when the vulnerability resolutions [get mapped to the ORT model's VulnerabilityResolution](https://github.com/eclipse-apoapsis/ort-server/blob/main/services/ort-run/src/main/kotlin/OrtServerMappings.kt#L733-L736).

shared/api-model/src/commonMain/kotlin/ChangeEvent.kt
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * An enumeration of the actions that can be performed, resulting in a [ChangeEvent].
+ */
+@Serializable
+enum class ChangeEventAction {
+    CREATE,
+    UPDATE,
+    ARCHIVE,
+    RESTORE
```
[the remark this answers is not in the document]
lamppu: I would like @mnonnenmacher's opinion on this.

api/v1/model/src/commonMain/kotlin/OrtRun.kt
mnonnenmacher: I assume there is a limited list of reasons why a run can get outdated so I would suggest to create an enum for that, then it could be `outdated: List<OutdatedReason>` with, for example, `OutdatedReason.ISSUE_RESOLUTIONS_CHANGED`.

model/src/commonMain/kotlin/VulnerabilityResolutionReason.kt
mnonnenmacher: I would not add the new reasons already but leave out the redundant `_VULNERABILITY` suffix.

dao/src/main/kotlin/repositories/repositoryconfiguration/DaoRepositoryConfigurationRepository.kt
```diff
@@ -55,7 +60,38 @@ class DaoRepositoryConfigurationRepository(private val db: Database) : Repositor
         licenseChoices: LicenseChoices,
         provenanceSnippetChoices: List<ProvenanceSnippetChoices>
     ): RepositoryConfiguration = db.blockingQuery {
-        RepositoryConfigurationDao.new {
+        val ortRun = OrtRunDao[ortRunId].mapToModel()
```
mnonnenmacher: The repository configuration is the representation of the `.ort.yml` file, adding the resolutions here can create confusion. A cleaner approach would be to make a custom `ResolutionProvider` implementation for the server that takes resolutions from the database into account.

api/v1/model/src/commonMain/kotlin/OrtRun.kt
sschuberth: What your opinion @mnonnenmacher about having both these properties?

mnonnenmacher: No strong opinion but I have a tendency to only have one property with the list of reasons and if it is empty or null it means the run is not outdated.

components/resolutions/backend/build.gradle.kts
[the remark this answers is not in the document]
Etsija: Yeah, me too, seems as if the task really isn't as intelligent as it should. It suggests removing almost all of the dependencies for my component, but totally breaks even if I remove a couple.

dao/src/main/kotlin/tables/ChangeLogTable.kt
```diff
@@ -0,0 +1,74 @@
...
+import org.jetbrains.exposed.sql.SortOrder
+import org.jetbrains.exposed.sql.Table
+import org.jetbrains.exposed.sql.and
+import org.jetbrains.exposed.sql.insert
+import org.jetbrains.exposed.sql.kotlin.datetime.timestamp
+
+/*
+ * A table to store change log entries representing user-performed change events.
+ */
+object ChangeLogTable : Table("change_log") {
```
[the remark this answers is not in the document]
lamppu: So do you think it could be something like `ResolutionsEventStore`, or do you think there should be a separate one for each type of resolution? The changes can be stored as well if so wanted, I didn't add that yet, as the answer I got for that question in the issue was that it's not necessarily needed yet.

components/resolutions/backend/build.gradle.kts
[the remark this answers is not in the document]
sschuberth: The short answer here is: [The DAGP plugin does not (fully) support Kotlin/Multiplatform projects yet](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/228), only Kotlin/JVM projects.

PR #3878, opened 2025-11-04
Search component

Etsija:
This PR introduces a new component "search", which is implemented with the sliced architecture. Initial implementation for the service and endpoint returns a list of all ORT run ids which include a given ORT ID. The service can later on be extended to other search functionality, for example to provide all products/repositories that are affected by a given vulnerability.

Screenshot from 2025-12-04 17-07-03

Screenshot from 2025-12-08 10-26-32

Details:
- so far, only ORT IDs are used for searching. PURL implementation (taking into account curated PURLs) will later on extend this
- the search can be scoped to a specific organization, product, or repository
- the search is `regEx` based, so for example substring of a package ID can be used
- the search is filtered such that it doesn't return any runs that the user doesn't have access to

NOTE: the filtering part is retained as a separate (last) commit in the PR, for reviewers to more easily review it.
NOTE2: it has manually been tested that the endpoint returns runs properly as an admin. User-filtered results will be verified later with UI code in a real-life situation (but they are tested with service and integration tests).

### DIFF
--- a/components/search/backend/build.gradle.kts
+++ b/components/search/backend/build.gradle.kts
@@ -38,11 +38,12 @@ repositories {
 dependencies {
     api(libs.exposedCore)
 
+    implementation(projects.components.authorization.backend)
     implementation(projects.components.search.apiModel)
     implementation(projects.dao)
     implementation(projects.model)
 
-    routesImplementation(projects.components.authorizationKeycloak.backend)
+    routesImplementation(projects.components.authorization.backend)
     routesImplementation(projects.shared.apiModel)
     routesImplementation(projects.shared.ktorUtils)
 
@@ -50,7 +51,6 @@ dependencies {
     routesImplementation(ktorLibs.server.core)
     routesImplementation(libs.ktorOpenApi)
 
-    testImplementation(testFixtures(projects.clients.keycloak))
     testImplementation(testFixtures(projects.dao))
     testImplementation(testFixtures(projects.shared.ktorUtils))
 
--- a/components/search/backend/src/main/kotlin/SearchService.kt
+++ b/components/search/backend/src/main/kotlin/SearchService.kt
@@ -19,6 +19,8 @@
 
 package org.eclipse.apoapsis.ortserver.components.search.backend
 
+import org.eclipse.apoapsis.ortserver.components.authorization.rights.RepositoryPermission
+import org.eclipse.apoapsis.ortserver.components.authorization.service.AuthorizationService
 import org.eclipse.apoapsis.ortserver.components.search.apimodel.RunWithPackage
 import org.eclipse.apoapsis.ortserver.dao.blockingQuery
 import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerjob.AnalyzerJobsTable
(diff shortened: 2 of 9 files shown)


Etsija:
NOTE: as decided in today's ORT Server Community Meeting, this PR will stay on hold until after the new authorization concept is fully merged (it needs then some changes to permissions handling and filtering the search results). Meanwhile, I am still open to general review comments regarding my PR.

components/search/api-model/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+plugins {
+    id("ort-server-kotlin-multiplatform-conventions")
+    id("ort-server-publication-conventions")
+    alias(libs.plugins.kotlinSerialization)
```
sschuberth: Nit: This should use conventional comments / group like e.g. [here](https://github.com/eclipse-apoapsis/ort-server/blob/73ea99f04792ea01b78b48aaf7f2701cb716c67a/components/admin-config/api-model/build.gradle.kts#L21-L25).

components/search/api-model/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+group = "org.eclipse.apoapsis.ortserver.components.search"
+
+kotlin {
+    jvm()
```
sschuberth: I guess `jvm()` is not required. We what it nowhere else.

components/search/backend/src/routes/kotlin/routes/GetRunsWithPackage.kt
```diff
@@ -0,0 +1,128 @@
...
+import org.eclipse.apoapsis.ortserver.components.authorization.keycloak.permissions.ProductPermission
+import org.eclipse.apoapsis.ortserver.components.authorization.keycloak.permissions.RepositoryPermission
+import org.eclipse.apoapsis.ortserver.components.authorization.keycloak.requirePermission
+import org.eclipse.apoapsis.ortserver.components.authorization.keycloak.requireSuperuser
+import org.eclipse.apoapsis.ortserver.components.search.apimodel.RunWithPackage
+import org.eclipse.apoapsis.ortserver.components.search.backend.SearchService
+import org.eclipse.apoapsis.ortserver.shared.ktorutils.jsonBody
+import org.eclipse.apoapsis.ortserver.shared.ktorutils.requireParameter
+
+@Suppress("LongMethod")
```
sschuberth: Is this really necessary? The function does not seem to be that long.

components/search/backend/src/main/kotlin/SearchService.kt
```diff
@@ -0,0 +1,119 @@
...
+import org.jetbrains.exposed.sql.and
+import org.jetbrains.exposed.sql.innerJoin
+import org.jetbrains.exposed.sql.stringLiteral
+
+class SearchService(private val db: Database) {
+    /**
+     * Search for Analyzer runs containing the given package identifier, with optional scoping.
+     * Throws IllegalArgumentException for invalid scoping hierarchy.
+     */
+    @Suppress("LongMethod")
```
sschuberth: Is this really necessary? The function does not seem to be that long.

components/search/backend/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
...
+repositories {
+    exclusiveContent {
+        forRepository {
+            maven("https://repo.gradle.org/gradle/libs-releases/")
+        }
+        filter {
+            includeGroup("org.gradle")
+        }
+    }
+}
```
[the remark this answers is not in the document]
sschuberth: This block is only needed if either this project or projects it depends on need to retrieve artifacts from non-standard locations (in this case the Gradle-specific Maven repository).

To find out whether it's really needed we first need to double-check that the list of dependencies is as minimal as it can be, and then simply try whether everything builds also without this block or not.

components/search/backend/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
...
+
+    testImplementation(ktorLibs.serialization.kotlinx.json)
+    testImplementation(ktorLibs.server.auth)
+    testImplementation(ktorLibs.server.contentNegotiation)
+    testImplementation(ktorLibs.server.statusPages)
+    testImplementation(ktorLibs.server.testHost)
+    testImplementation(libs.kotestAssertionsKtor)
+    testImplementation(libs.kotestRunnerJunit5)
+    testImplementation(libs.kotlinxSerializationJson)
+    testImplementation(libs.mockk)
```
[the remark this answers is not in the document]
Etsija: It was checked with Claude Code, which inspected the dependencies and concluded that all deps are needed. So how should I double-check it, as you have said earlier that the Gradle task for this is really not working atm.

components/search/backend/src/main/kotlin/SearchService.kt
Etsija: See above.

components/search/backend/build.gradle.kts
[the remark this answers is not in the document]
Etsija: As @lamppu has said earlier, that list certainly cannot be trusted: she tried, and the build broke from a slightest removal of the deps suggested by projectHealth. That's what I meant by saying the task is not functioning as it should.

Etsija:
@oheger-bosch could I kindly ask from you a "sanity check" review for this PR, especially regarding the usage of the new db authorization system? As I am not nearly enough familiar with it, I might have omitted some useful methods and workflows from it. I kept all relevant implementation (especially query filtering) in the last commit 312d97ab5ba78c9a41a1d54d59830361c32c46a2 for easier review.

I am especially sceptic about the need of `requireSearchPermission()` in the `GetRunsWithPackage.kt` file; would there be simpler and more elegant way of checking the permissions dynamically, depending of the query params?

If you think it's still needed, I believe I could at least refactor it to an own file, as similar search endpoints (in future PRs) will need this kind of dynamic permissions checking, because they will also use query parameters dynamically.

components/search/backend/src/main/kotlin/SearchService.kt
```diff
@@ -30,90 +32,124 @@ import org.eclipse.apoapsis.ortserver.dao.repositories.ortrun.OrtRunsTable
...
+                stringLiteral(":"),
+                IdentifiersTable.namespace,
+                stringLiteral(":"),
+                IdentifiersTable.name,
+                stringLiteral(":"),
+                IdentifiersTable.version
+            )
 
-        val whereClause = conditions.reduce { acc, expression -> acc and expression }
+            val scopeCondition = when (scope) {
```
oheger-bosch: I think, this condition should not be required. The condition generated by the `HierarchyFilter` should already include exactly the hierarchy elements that are visible for the user.

components/search/backend/src/routes/kotlin/Validation.kt
```diff
@@ -0,0 +1,122 @@
...
+ * Use [parseScope] to extract and validate the scope from request parameters.
+ */
+fun requireScopedReadPermission(): AuthorizationChecker =
+    object : AuthorizationChecker {
+        override suspend fun loadEffectiveRole(
+            service: AuthorizationService,
+            userId: String,
+            call: ApplicationCall
+        ): EffectiveRole? {
+            val scope = call.parseScope().getOrNull()
```
oheger-bosch: IIUC, if `parseScope` returns a failed result, then superuser rights are required. Wouldn't it be better to rethrow the exeption (or some other exception), so that the endpoint can return a 400 response in this case?

components/search/backend/src/routes/kotlin/routes/GetRunsWithPackage.kt
```diff
@@ -89,40 +87,24 @@ internal fun Route.getRunsWithPackage(searchService: SearchService) =
                 }
             }
         }
-    }) {
+    }, requireScopedReadPermission()) {
         val identifierParam = call.requireParameter("identifier")
-        val organizationIdParam = call.request.queryParameters["organizationId"]?.toLongOrNull()
-        val productIdParam = call.request.queryParameters["productId"]?.toLongOrNull()
-        val repositoryIdParam = call.request.queryParameters["repositoryId"]?.toLongOrNull()
+        val scopeResult = call.parseScope()
```
oheger-bosch: Related to my previous comment. I would propose to handle invalid scopes in the checker returned by `requireScopedReadPermission()` and use some exception mapping logic in [StatusPages](https://github.com/eclipse-apoapsis/ort-server/blob/main/core/src/main/kotlin/plugins/StatusPages.kt) to yield a 400 BAD RESPONSE code.

components/search/backend/src/routes/kotlin/routes/GetRunsWithPackage.kt
```diff
@@ -19,24 +19,91 @@
...
+            service: AuthorizationService,
+            userId: String,
+            call: ApplicationCall
+        ): EffectiveRole? {
+            val organizationIdParam = call.request.queryParameters["organizationId"]?.toLongOrNull()
+            val productIdParam = call.request.queryParameters["productId"]?.toLongOrNull()
+            val repositoryIdParam = call.request.queryParameters["repositoryId"]?.toLongOrNull()
+
+            return when {
+                repositoryIdParam != null && productIdParam != null && organizationIdParam != null -> {
```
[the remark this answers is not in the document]
oheger-bosch: In #4096 I have added a corresponding warning to the documentation of the authorization service.

PR #3884, opened 2025-11-04
Authorization: Improvements of the new component

oheger-bosch:
This PR adds some improvements and missing features to the new Authorization component that are needed to actually switch the routes to the new functionality.

### DIFF
--- a/components/authorization/backend/src/main/kotlin/routes/AuthorizedRoutes.kt
+++ b/components/authorization/backend/src/main/kotlin/routes/AuthorizedRoutes.kt
@@ -17,6 +17,8 @@
  * License-Filename: LICENSE
  */
 
+@file:Suppress("TooManyFunctions")
+
 package org.eclipse.apoapsis.ortserver.components.authorization.routes
 
 import com.auth0.jwt.interfaces.Payload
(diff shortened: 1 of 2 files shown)


components/authorization/backend/src/main/kotlin/service/AuthorizationService.kt
```diff
@@ -93,11 +93,11 @@ interface AuthorizationService {
...
 
     /**
-     * Return a [Map] with the IDs of all users and their assigned roles on the hierarchy element identified by the
+     * Return a [Map] with the IDs of all users and their assigned role on the hierarchy element identified by the
      * given [compoundHierarchyId]. The result includes users who inherit access rights on this hierarchy element from
-     * higher levels, such as organization admins.
+     * higher levels, such as organization admins, but no superusers.
      */
-    suspend fun listUsers(compoundHierarchyId: CompoundHierarchyId): Map<String, Set<Role>>
+    suspend fun listUsers(compoundHierarchyId: CompoundHierarchyId): Map<String, Role>
```
mnonnenmacher: Do I understand this correctly that this function solves the issue that when you list the users for a repository, you also want to see the users who inherited the role from a parent entity? If so, I think it would be a great addition for later to include in the result if the role was inherited or not, this will be very useful for the UI.

components/authorization/backend/src/main/kotlin/rights/OrganizationPermission.kt
```diff
@@ -49,3 +49,12 @@ enum class OrganizationPermission {
     /** Permission to delete the [Organization]. */
     DELETE
 }
+
+/**
+ * The set of permissions required by the role to read an organization. (This is defined here to avoid circular
+ * dependencies, as it is referenced by multiple role classes.)
+ */
```
mnonnenmacher: Are you sure that is required? I think the compiler can handle this kind of circular dependencies, I must made a small test that compiles:

```kotlin
enum class OrgRoles(val orgPermissions: List<String>, val prodPermissions: List<String>) {
    READER(
        orgPermissions = listOf("READ"),
        prodPermissions = ProdRoles.READER.prodPermissions,
    )
}

enum class ProdRoles(val orgPermissions: List<String>, val prodPermissions: List<String>) {
    READER(
        orgPermissions = OrgRoles.READER.orgPermissions,
        prodPermissions = listOf("READ"),
    )
}
```

mnonnenmacher (changes requested):

components/authorization/backend/src/main/kotlin/service/AuthorizationService.kt
oheger-bosch: Yes, the returned map includes users who inherit their access rights from other levels. Adding this information to the result is possible, however, it would probably require some changes in the data structures used by the underlying `HierarchyPermission` class. I would therefore like to defer this change and first focus on the basics to get the new authorization component life.

PR #3895, opened 2025-11-05
fix(ui): Fix CycloneDX logo variant in system mode

Etsija:
Screenshot from 2025-11-05 15-03-35

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/sbom/index.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/sbom/index.tsx
@@ -120,6 +120,10 @@ const SBOMComponent = () => {
     (filename) => filename.toLowerCase().includes('cyclonedx')
   );
 
+  const systemMode = window.matchMedia('(prefers-color-scheme: dark)').matches
+    ? 'dark'
+    : 'light';
+
   return (
     <div className='flex flex-col gap-4'>
       <div className='grid grid-cols-2 gap-4'>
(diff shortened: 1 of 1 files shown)


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/sbom/index.tsx
```diff
@@ -134,7 +138,12 @@ const SBOMComponent = () => {
                 <TooltipTrigger>
                   <img
                     alt='CycloneDX'
-                    src={mode === 'dark' ? CycloneDXLight : CycloneDXDark}
+                    src={
+                      mode === 'dark' ||
+                      (mode === 'system' && systemMode === 'dark')
```
sschuberth: Does it maybe make sense to add a commit which extracts this to an `isDarkMode` helper function that does no care whether it's set explicitly or via the system, and use it to also simplify the code at

https://github.com/eclipse-apoapsis/ort-server/blob/800de3b0307210bc3d485e73f9cea18fad0d7ffb/ui/src/components/theme-provider.tsx#L59-L62

Etsija: I just KNEW you'd ask :rofl:

sschuberth: You could have just said "No, it does not make sense", but of course I'd then ask "why not"? 😁

Etsija: The whole theming concept is very much tied to React, with hooks, states and providers, and it would be an interesting rehearsal to extract the information outside React (ie. in a plain TypeScript function)...but maybe not this time.

PR #3930, opened 2025-11-11
ui: Start creating e2e tests for the UI

Etsija:
The purpose is not to start creating elaborate e2e tests, but as has been seen many times, it is possible to create a PR which totally breaks the UI, so best to implement the e2e test framework and run at least some trivial UI tests.


.github/workflows/build-and-test.yml
```diff
@@ -214,3 +214,124 @@ jobs:
         name: integration-test-results
         path: "**/build/reports/tests"
         retention-days: 7
+
```
mnonnenmacher: Commit message nit: "waits for core/the UI to..."

PR #3957, opened 2025-11-14
chore(compose): Set a name for the admin user

mnonnenmacher:
With the upcoming changes in [1] it will be required that the name is not empty. Before, the admin user did not have any name set which lead to the `name` claim missing from the JWT.

[1]: https://github.com/eclipse-apoapsis/ort-server/pull/3908

### DIFF
--- a/scripts/docker/keycloak/master-realm.json
+++ b/scripts/docker/keycloak/master-realm.json
@@ -411,6 +411,7 @@
   "users" : [ {
     "id" : "b4192f3c-d52a-43a8-8e5d-4ea429f08ca4",
     "username" : "admin",
+    "firstName" : "admin",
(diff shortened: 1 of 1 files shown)


scripts/docker/keycloak/master-realm.json
```diff
@@ -411,6 +411,7 @@
   "users" : [ {
     "id" : "b4192f3c-d52a-43a8-8e5d-4ea429f08ca4",
     "username" : "admin",
+    "firstName" : "admin",
```
sschuberth: Is there also a "lastName" that we should set for consistency, too?

mnonnenmacher:
Obsolete, instead the name will be made optional in #3908.

PR #3967, opened 2025-11-16
docs(dao): Avoid duplicate "provenance" in an error message

sschuberth:
Instead of "for provenance 'RepositoryProvenance(..." shorten to just "for 'RepositoryProvenance(...".


dao/src/main/kotlin/repositories/scannerrun/DaoScannerRunRepository.kt
```diff
@@ -143,7 +143,7 @@ class DaoScannerRunRepository(private val db: Database) : ScannerRunRepository {
                         nestedProvenanceResolutionIssue = Issue(
                             timestamp = Clock.System.now(),
                             source = "Scanner",
-                            message = "Could not resolve nested provenance for provenance '$packageProvenance'.",
+                            message = "Could not resolve nested provenance for '$packageProvenance'.",
```
sschuberth: @mnonnenmacher, at this point, would we be able to get any more details about *why* nested provenance resolution failed?

PR #3975, opened 2025-11-17
fix(config): Clear HTTP basic auth cache to prevent credential conflicts

MarcelBochtler:
JGit uses `HttpURLConnection` for HTTP(S) connections [1], which caches
credentials when basic authentication is used [2] [3]. This causes issues
when accessing multiple repositories on the same host with different
credentials.

E.g. for GitHub repositories accessed via HTTPS, HttpURLConnection sets
the cache key as `s:BASIC:https:github.com:443:GitHub`. When both the
config file provider and a user repository use the same Git hosting
service but require different credentials, the cached credentials from
the first operation are incorrectly reused for the second, causing
authentication failures.

Clear the authentication cache after updating the working tree of the
`GitConfigFileProvider` to ensure that subsequent Git operations for
each repositories use the correct credentials.

Note:
This was only tested with OpenJDK Temurin-21.0.9+10. It is possible,
that this fix won't work for other versions.

[1]: https://github.com/eclipse-jgit/jgit/blob/v7.5.0.202510290248-m2/org.eclipse.jgit/src/org/eclipse/jgit/transport/http/JDKHttpConnection.java
[2]: https://bugs.openjdk.org/browse/JDK-6626700
[3]: https://github.com/openjdk/jdk/blob/676e6fd8d5152f4e0d14ae59ddd7aa0a7127ea58/src/java.base/share/classes/sun/net/www/protocol/http/AuthenticationInfo.java#L306-L308

### DIFF
--- a/config/git/src/main/kotlin/GitConfigFileProvider.kt
+++ b/config/git/src/main/kotlin/GitConfigFileProvider.kt
@@ -23,6 +23,7 @@ import com.typesafe.config.Config
 
 import java.io.File
 import java.io.InputStream
+import java.util.HashMap
 
 import kotlin.time.measureTime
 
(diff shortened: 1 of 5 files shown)


config/git/src/main/kotlin/GitConfigFileProvider.kt
```diff
@@ -133,7 +134,42 @@ class GitConfigFileProvider internal constructor(
...
+
+        val setMapMethod = authCacheImplClass.getDeclaredMethod("setMap", HashMap::class.java)
+        // Replace the cache map with an empty one.
+        setMapMethod.invoke(defaultCache, HashMap<Any, Any>())
+
+        logger.debug("Successfully cleared JGit HTTP authentication cache.")
+    }.onFailure { e ->
+        logger.warn(
+            "Failed to clear JGit HTTP authentication cache. This may lead to Git authentication issues. Consider " +
+                    "setting '--add-opens java.base/sun.net.www.protocol.http=ALL-UNNAMED' javaOpts.",
```
mnonnenmacher: Why don't you add this to the default opts for the Docker images that need it?

mnonnenmacher (approved):

sschuberth:
Would switching to `org.eclipse.jgit.transport.http.apache.HttpClientConnection` be another, cleaner option?

PR #3985, opened 2025-11-18
deps: Remove unused ORT version declaration

MarcelBochtler:
This was made obsolete by using the ORT version catalog.

### DIFF
--- a/buildSrc/src/main/kotlin/ort-server-kotlin-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/ort-server-kotlin-conventions.gradle.kts
@@ -37,7 +37,7 @@ plugins {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


MarcelBochtler:
Turns out this outdated version is still being used. I'll look into it.

sschuberth:
> > Turns out this outdated version is still being used. I'll look into it.
> 
> I guess we'd need something like https://github.com/radoslaw-panuszewski/typesafe-conventions-gradle-plugin to fix this due to Gradle limitations. @MarcelBochtler, I can look into this if you lack time.

Took me a while to find out, but it works without the plugin with this block after `private val Project.libs: ...`:

```kotlin
private val catalogs = extensions.getByType<VersionCatalogsExtension>()
private val detektRulesVersion = catalogs.named("ortLibs").findLibrary("detektRules").get().get().version
```

And then using `$detektRulesVersion` instead of `${rootProject.libs.versions.ort.get()}`.

Issue #1149, opened 2024-10-03
UI issues on mobile


mnonnenmacher:
I just tested the UI on my phone and found a few issues:

- The title bar does not fit on the screen. 
- The breadcrumbs appear above other content. 
- The menu button should not appear if the menu is empty anyway. 

I only did a quick test, there are probably more issues on the other pages.

![Screenshot_20241003-185613.png](https://github.com/user-attachments/assets/7612c193-f92b-4b32-b977-5b7750729204)

![Screenshot_20241003-185751.png](https://github.com/user-attachments/assets/1392868b-ef76-4ef7-9d2b-701582e3750e)

Etsija:
Good that you created an issue about this; the UI needs a complete layout review, as it's currently not tested properly on different sizes/platforms.  Some layout issues can be easily fixed with by using `TailwindCSS` media queries properly, while some others (for example the breadcrumb) need a bit of rethinking.

PR #4007, opened 2025-11-21
Fix the horizontal overflow of repository runs table

Etsija:
Screenshot from 2025-11-21 09-10-01


sschuberth:
> That would be great, and has been requested numerous times, but alas, it's not possible. Only numbers are accepted.

Ok, then let's merge this for now until we have a better solution (that also allows us to address https://github.com/eclipse-apoapsis/ort-server/issues/1149).

PR #4028, opened 2025-11-24
Improve secret resolution for provider plugin configs

mnonnenmacher:
Implement resolving user secrets for provider plugins configs in the same way as it was done in #3831 for plugin configs. See the commit messages for details.


oheger-bosch (commented):
In general, I find this duplication of resolvable configs and plain configs rather difficult to understand. Is there some hope that this will be replaced in the future?

mnonnenmacher:
> In general, I find this duplication of resolvable configs and plain configs rather difficult to understand. Is there some hope that this will be replaced in the future?

Right now I would not know how.

oheger-bosch (approved):

PR #4043, opened 2025-11-26
fix(compose): Use a Docker volume for the secrets provider

### LINKED ISSUE #4037: No secrets at all can be created in compose environment
I tested all levels. Also tried to delete the `scripts/compose/secrets-provider/secrets.store` in hopes that a new valid one will be created, but it isn't happening. @lamppu has confirmed this also happens for her.

Image

mnonnenmacher:
Since bfaaa05 the core image uses UID 1001 instead of 1000. This breaks the mount for the secrets provider directory in the Compose setup if the directory on the host system belongs to a different user. As it is difficult to find a generic solution with a bind mount, use a Docker volume for the secrets provider instead and use an init container to configure the permissions correctly.

While at it, also add the secrets provider volume to services where it was missing.

Fixes #4037.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -143,9 +143,19 @@ services:
         gelf-address: "udp://127.0.0.1:12201"
         tag: "graphite"
 
+  # Give UID 1001 ownership of the secrets-provider volume as that UID is used by default by the ORT Server images.
+  init-secrets:
+    image: busybox
+    command: [ "sh", "-c", "chown -R 1001:1001 /mnt/secrets-provider" ]
+    volumes:
+    - secrets:/mnt/secrets-provider
+    restart: "no"
+
   core:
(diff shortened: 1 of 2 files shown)


Etsija (changes requested):
I checked, and secrets can be created again. But you also removed `.gitignore` from the .../secrets-provider directory, was this intentional? Because now, git wants to commit my `secrets.store` which I'm sure is wrong.

mnonnenmacher:
> I checked, and secrets can be created again. But you also removed `.gitignore` from the .../secrets-provider directory, was this intentional? Because now, git wants to commit my `secrets.store` which I'm sure is wrong.

Yes, because the directory is not used anymore, you have to delete that file or manually move it to the Docker volume if you want to keep the content.

Etsija (approved):

PR #4055, opened 2025-11-28
deps(website): Run `pnpm update`, mainly to update `node-forge` to >=1.3.2

sschuberth:
See [1].

[1]: https://github.com/eclipse-apoapsis/ort-server/security/dependabot/123

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -43,24 +43,24 @@
     "@docusaurus/utils": "3.9.2",
     "@docusaurus/utils-common": "3.9.2",
     "@docusaurus/utils-validation": "3.9.2",
-    "@easyops-cn/docusaurus-search-local": "^0.52.0",
-    "@iconify/react": "^6.0.0",
-    "@mdx-js/react": "^3.0.0",
-    "clsx": "^2.0.0",
+    "@easyops-cn/docusaurus-search-local": "^0.52.1",
+    "@iconify/react": "^6.0.2",
+    "@mdx-js/react": "^3.1.1",
+    "clsx": "^2.1.1",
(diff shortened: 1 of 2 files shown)


mnonnenmacher:
> Could we do this as a follow-up PR?

Sure.

> Anyway, I'm a little bit confused about Renovate's docs here, e.g. when it says "convert ranges to exact versions". This only makes sense to me when referring to the `package.json`, whereas usually Renovate only touches the `pnpm-lock.yaml`.

It also updates the package.json if required, but if version ranges are used it is rarely required. The default "auto" strategy is explained on the page I linked.

mnonnenmacher (approved):

sschuberth:
> Or maybe even better, set the `rangeStrategy` to `pin` or `bump`: https://docs.renovatebot.com/configuration-options/#rangestrategy

I guess this requires a general decision about how we want to manage (P)NPM dependencies:
1. "Maven-style", by always pinning exact versions already in `package.json`, leaving it to Renovate to update both `package.json` and `pnpm-lock.yaml`.
2. Use ranges in `package.json`, allow Renovate to update `pnpm-lock.yaml` only without these manually maintained ranges.
3. Use ranges in `package.json`, but allow Renovate also to bump the ranges, and update `pnpm-lock.yaml`.

1. seems unusual in the NPM world, and 3. seems a bit "arbitrary", because why to have ranges if Renovate can bump them to its liking anyway? SO I personally like 2. somewhat as a compromise of having some manual control over the ranges, but let Renovate to the boring update work of `pnpm-lock.yaml`. The downside is that missing to manually bump a range can lead to missing security-relevant updates.

What do @eclipse-apoapsis/technology-apoapsis-security think?

PR #4058, opened 2025-11-28
Curation metadata

sschuberth:


Etsija (commented):
Verified that it works in the UI. Is there a need for an integration test?

Etsija (approved):

sschuberth:
> We’ve discussed this internally and are confident that this part of the API is currently not being used by our users.

That's good to hear, and is also what I was hoping for to be the case. Anyway, I'd now like to get your explicit approvals on this PR, from both @MarcelBochtler and @mnonnenmacher, please.

MarcelBochtler:
> Do you really mean "approved" or rather "merged" here, @MarcelBochtler?

My concern is that a breaking API change was communicated only through a commit message. It could have been mentioned in the community meeting we had half an hour before this PR was marked as ready for review.

> Because I don't really see a problem with a PR that includes both API and UI changes being approved for the UI part, as happened in this case, but then not being merged yet due to awaiting further approvals.

I agree that the approval status is not the real issue. It just caused that I was rushing for the "request changes" button, because I thought you might merge it before anybody else had a chance to intervene.
But let's discuss that as part of the community meeting where I want to find an agreement for a process for changes like this (in ORT Server 0.x).

MarcelBochtler:
@sschuberth add a note about the breaking API change in the next ORT Server version release notes as we did for 0.44.0:
https://github.com/eclipse-apoapsis/ort-server/releases/tag/0.44.0

sschuberth:
> @sschuberth add a note

Yes, sir! 🫡

sschuberth:
> > @sschuberth add a note
> 
> Yes, sir! 🫡

Done: https://github.com/eclipse-apoapsis/ort-server/releases/tag/0.45.0.

PR #4071, opened 2025-12-01
ui: Show secrets from all levels when creating an infrastructure service

Etsija:

Note for @sschuberth: no fancy pictures this time, as my Ubuntu won't let me take a screenshot with a dropdown open, so the changed UI listing the inherited secrets cannot be shown - but it's looking exactly the same as the one for the environment secrets in ORT run creation form.

### DIFF
--- a/ui/src/routes/organizations/$orgId/infrastructure-services/create/index.tsx
+++ b/ui/src/routes/organizations/$orgId/infrastructure-services/create/index.tsx
@@ -18,16 +18,13 @@
  */
 
 import { zodResolver } from '@hookform/resolvers/zod';
-import { useMutation, useQuery } from '@tanstack/react-query';
+import { useMutation } from '@tanstack/react-query';
 import { createFileRoute, useNavigate } from '@tanstack/react-router';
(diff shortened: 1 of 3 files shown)


ui/src/routes/organizations/$orgId/infrastructure-services/create/index.tsx
```diff
@@ -218,18 +215,27 @@ const CreateInfrastructureServicePage = () => {
...
                     <SelectContent>
-                      {secrets?.data.map((secret) => (
-                        <SelectItem key={secret.name} value={secret.name}>
-                          {secret.name}
-                        </SelectItem>
-                      ))}
+                      {secrets.map((secret) => {
+                        const hierarchyLabel =
+                          secret.hierarchy.charAt(0).toUpperCase() +
+                          secret.hierarchy.slice(1);
```
sschuberth: How about extracting this to a `capitalize()` (or similar) helper function for reuse, covered with a few tests? As part of testing, we should also define how to behave for mixed-case strings like "fOO"... I guess we show lowercase the tail, so the final string would be "Foo".

PR #4078, opened 2025-12-02
deps(reporter): Force JRuby 9.x for the Reporter worker

oheger-bosch:
It has been observed that there is a slight performance decrease when generating PDF reports with JRuby 10.x. Also, AsciidoctorJ officially still depends on JRuby 9.x. Therefore, override the JRuby 10.x dependency from ORT to use JRuby 9.x in the Reporter image.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -41,6 +41,7 @@ commonsText = "1.14.0"
 exposed = "0.61.0"
 flyway = "11.18.0"
 hikari = "7.0.2"
+jrubyAscciiDoc = "9.4.14.0"
 jsonSchemaValidator = "2.0.0"
 kaml = "0.104.0"
 keycloakTestcontainerVersion = "4.0.0"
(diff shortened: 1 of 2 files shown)


sschuberth:
So, as we have been discussing this a bit, why was this approach chosen over downgrading JRuby in ORT?

oheger-bosch:
The requirements of ORT for the JRuby version in the Analyzer are not fully clear to me. In ORT Server, it is possible to have different JRuby versions for the Analyzer and the Reporter image, so we could stick in the Reporter with the 9.x version until version 10.x is further optimized, while the Analyzer could already take advantage of  features from the new version.

oheger-bosch:
We noticed the increase of execution time when our acceptance tests started to run into timeouts; the delay seems to be larger when the system is under load, and this was sufficient to reach the threshold in this case. And, although we are talking about seconds only, multiplied with the number of runs, this can pile up.

But from your points, it is not clear to me what is your preferred solution. I see the following options:
* Apply this PR, at least as a temporary solution
* Instead downgrade the JRuby version in ORT to 9.x
* Do nothing and hope that the issue will be fixed in future 10.x versions of JRuby

PR #4109, opened 2025-12-08
Use the "build-health" settings plugin instead of the "dependency-analysis" plugin

sschuberth:


oheger-bosch:
@mnonnenmacher, @sschuberth, I don't know about the root cause. From my local tests, I could only see that the errors start to happen with commit 2218c74aac11c8b6a7487af9fe4ad48ccf08fadb. Unless there is no better idea, I would opt for reverting this.

PR #4115, opened 2025-12-09
build(gradle): Simplify accessing the version catalog from `buildSrc`

sschuberth:
Make use of [1].

[1]: https://github.com/radoslaw-panuszewski/typesafe-conventions-gradle-plugin


sschuberth:
> > Unfortunately not. Still the same error:
> 
> Indeed unfortunate, but I believe this PR is useful in any case.

@lamppu, do you feel comfortable reviewing / approving this?

PR #4123, opened 2025-12-10
Gradle chores

mnonnenmacher:


api/v1/model/build.gradle.kts
```diff
@@ -37,7 +37,7 @@ kotlin {
     sourceSets {
         commonMain {
             dependencies {
-                api(projects.components.infrastructureServices.apiModel)
+                api(projects.components.infrastructureServices.infrastructureServicesApiModel)
```
sschuberth: Same here, isn't this a breaking change for consumers of the published artifacts?

PR #4137, opened 2025-12-12
Package search with PURL

Etsija:
In order to add vulnerability search to the back-end, the search service needs to be extended to handle curated PURL
based search. UI implementation for this search will be done in a follow-up PR.

Screenshot from 2025-12-12 09-27-44

Screenshot from 2025-12-12 09-28-02

Screenshot from 2025-12-12 09-28-20

### DIFF
--- a/components/search/backend/src/test/kotlin/routes/GetRunsWithPackageIntegrationTest.kt
+++ b/components/search/backend/src/test/kotlin/routes/GetRunsWithPackageIntegrationTest.kt
@@ -21,8 +21,11 @@ package ort.eclipse.apoapsis.ortserver.components.search.routes
 
 import io.kotest.assertions.ktor.client.shouldHaveStatus
 import io.kotest.matchers.collections.shouldBeEmpty
+import io.kotest.matchers.collections.shouldBeSingleton
 import io.kotest.matchers.collections.shouldContainExactly
 import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
+import io.kotest.matchers.nulls.shouldBeNull
+import io.kotest.matchers.shouldBe
 
 import io.ktor.client.call.body
 import io.ktor.client.request.get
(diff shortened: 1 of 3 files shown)


components/search/api-model/src/commonMain/kotlin/RunWithPackage.kt
```diff
@@ -31,5 +31,6 @@ data class RunWithPackage(
     val ortRunIndex: Long,
     val revision: String?,
     val createdAt: Instant,
-    val packageId: String,
+    val packageId: String?,
```
sschuberth: Commit message nits:
- Double space in "pattern  used"
- "Note1" -> "Note 1" (same for 2)

components/search/backend/src/main/kotlin/SearchService.kt
```diff
@@ -70,50 +81,133 @@ class SearchService(
...
+     */
+    private fun findByPurl(
+        query: org.jetbrains.exposed.sql.Join,
+        purl: String,
+        hierarchyFilter: HierarchyFilter
+    ): List<RunWithPackage> {
+        val effectivePurl = createEffectivePurlExpression()
+
+        @Suppress("UNCHECKED_CAST")
+        val purlCondition = (effectivePurl as Expression<String>).applyRegex(purl)
```
sschuberth: Unchecked casts are always suspicious. I haven't checked, but please verify that this is really needed.

components/search/backend/src/test/kotlin/routes/GetRunsWithPackageIntegrationTest.kt
```diff
@@ -348,6 +353,117 @@ class GetRunsWithPackageIntegrationTest : SearchIntegrationTest({
...
+                val response = client.get(SEARCH_ROUTE) {
+                    parameter("purl", packageIdentifier.toPurl())
+                }
+
+                response shouldHaveStatus HttpStatusCode.OK
+
+                val body = response.body<List<RunWithPackage>>()
+                body.size shouldBe 1
+                body.first().packageId.shouldBeNull()
+                body.first().purl shouldBe packageIdentifier.toPurl()
```
sschuberth: This can be simplified to

                body.shouldBeSingleton {
                    it.packageId.shouldBeNull()
                    it.purl shouldBe packageIdentifier.toPurl()
                }

Similar for 3 other places that use the `size shouldBe 1` approahc.

PR #4157, opened 2025-12-15
fix(compose): Adjust the path of the Postgres volume

nnobelis:
The Postgres 18 Docker image requires the filesystem path where the DB is stored to contain the database version. See [1].

[1]: https://github.com/docker-library/postgres/pull/1259


sschuberth:
Did you see 240ce8215641b191a6ecd263c90d1838c7144299 in this regard?

sschuberth:
> What do you think ? Should I close current PR ?

Ideally, you'd be able to redo the upgrade with `pg_upgrade --link` to confirm the existing fix, then we could close this with confidence.

Otherwise, let's wait for some more feedback from others.

PR #4176, opened 2025-12-18
Vulnerability search

Etsija:
This PR extends the search component with an endpoint for searching ORT runs containing packages with a given vulnerability. Scoping and regular expressions supported. The returned package identifiers can be ORT IDs or purls.

Screenshot from 2025-12-22 11-44-29

Screenshot from 2025-12-22 08-37-37

Screenshot from 2025-12-18 09-02-19

### DIFF
--- a/components/search/backend/src/test/kotlin/SearchServiceTest.kt
+++ b/components/search/backend/src/test/kotlin/SearchServiceTest.kt
@@ -265,7 +265,6 @@ class SearchServiceTest : WordSpec({
             val pkgId = Identifier("maven", "org.example", "library", "1.0.0")
             val curatedPurl = "pkg:maven/org.example/library-corrected@1.0.0"
             val run = createRunWithCuratedPurl(
-                db = db,
                 fixtures = fixtures,
                 repoId = repositoryId,
                 pkgId = pkgId,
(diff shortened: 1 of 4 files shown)


components/search/api-model/src/commonMain/kotlin/RunWithVulnerability.kt
```diff
@@ -0,0 +1,37 @@
...
+data class RunWithVulnerability(
+    val organizationId: Long,
+    val productId: Long,
+    val repositoryId: Long,
+    val ortRunId: Long,
+    val ortRunIndex: Long,
+    val revision: String?,
+    val createdAt: Instant,
+    val externalId: String,
+    val packageId: String?,
```
lamppu: I noticed now that this is a `String` for the package search too, but was there some particular reason why it's not an `Identifier`?

lamppu: I noticed now that `packageId` is a `String` for the package search too, but was there some particular reason why it's not an `Identifier`? It's just a bit inconsistent, as I think other endpoints return an `Identifier`. Also if there's ever changes to the display representation of an identifier in the UI, this would need to be changed / need a workaround.

PR #4181, opened 2025-12-18
Update ORT to version 74.0.0

MarcelBochtler:
This PR is mostly based on https://github.com/eclipse-apoapsis/ort-server/pull/4126.
With two important changes:
1. I've removed the breaking REST API change
2. I've removed the breaking change to use an Enum instead of a String for the `licenseSource`.
3. The two changes above removed the requirement to adjust the UI code.

The reason for this approach is, that we can continue to update ORT without the need to have a breaking change in the API.

However, this means that the feature to support multiple license sources, won't work fully with the ORT Server.
If multiple license sources are detected, they will be stored correctly in the database, but neither the API nor the UI will show more than one license source. They will both fall back to the first one in the set.


api/v1/mapping/src/commonMain/kotlin/ApiMappings.kt
```diff
@@ -619,7 +619,7 @@ fun RuleViolation.mapToApi() = ApiRuleViolation(
     rule = rule,
     id = id?.mapToApi(),
     license = license,
-    licenseSource = licenseSource,
+    licenseSource = licenseSources.firstOrNull(),
```
sschuberth: I believe this should be marked with a `TODO` in the code *and* we should create a follow-up issue from these lines once merged to not forget about adding support for multiple license sources going forward.

nnobelis (dismissed):

api/v1/mapping/src/commonMain/kotlin/ApiMappings.kt
```diff
@@ -619,7 +619,8 @@ fun RuleViolation.mapToApi() = ApiRuleViolation(
     rule = rule,
     id = id?.mapToApi(),
     license = license,
-    licenseSource = licenseSource,
+    // TODO: Add support for multiple license sources, see https://github.com/eclipse-apoapsis/ort-server/issues/4185.
```
nnobelis: Better see issue #4185.

nnobelis (approved):

mnonnenmacher:
>     2. I've removed the breaking change to use an Enum instead of a String for the `licenseSource`.

@MarcelBochtler Was that really a breaking change for API users or only for consumers of the OpenAPI spec? And why did you also remove the code that introduced the enum in backend code?

PR #4188, opened 2025-12-19
feat(logs): Add custom header for UI and CLI clients

MarcelBochtler:
Add the value of the `X-Client-Type` header for tracking UI and CLI clients to the MDC logs. This can be used to distinguish all API requests originating from the known clients from unknown ones.

### DIFF
--- a/core/src/main/kotlin/plugins/Monitoring.kt
+++ b/core/src/main/kotlin/plugins/Monitoring.kt
@@ -19,12 +19,16 @@
 
 package org.eclipse.apoapsis.ortserver.core.plugins
 
+import io.ktor.http.HttpHeaders
 import io.ktor.server.application.Application
 import io.ktor.server.application.install
 import io.ktor.server.plugins.calllogging.CallLogging
+import io.ktor.server.request.httpMethod
 import io.ktor.server.request.path
 
 import java.util.UUID
+
+import org.eclipse.apoapsis.ortserver.utils.system.CustomHttpHeaders
 
 import org.slf4j.event.Level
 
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/plugins/Monitoring.kt
```diff
@@ -42,5 +45,10 @@ fun Application.configureMonitoring() {
         mdc("organizationId") { it.parameters["organizationId"] }
         mdc("productId") { it.parameters["productId"] }
         mdc("repositoryId") { it.parameters["repositoryId"] }
+        mdc("clientType") { call ->
+            call.request.headers[CustomHttpHeaders.ClientType]
+                ?: call.request.headers[HttpHeaders.UserAgent]
```
nnobelis: User agents from browser can be pretty long, for instance:

`Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36`

Is this no going to swell the logs ?

nnobelis: Ok, we can assume not so many users access ort-server outside of the UI and the CLI.

MarcelBochtler: > Ok, we can assume not so many users access ort-server outside of the UI and the CLI.

And if they do, we would be interested in the actual User Agent (e.g. to see if the request was made from Kubernetes Liveness or from `curl` or from `Postman`, ...

nnobelis (approved):

core/src/main/kotlin/plugins/Monitoring.kt
```diff
@@ -42,5 +45,10 @@ fun Application.configureMonitoring() {
...
+                ?: "unknown"
+        }
```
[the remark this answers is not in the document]
MarcelBochtler: Yes, you're right. When I'm adding a log line into a route, it also contains the `clientType`, and I agree that it is only required for the request log.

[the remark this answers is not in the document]
MarcelBochtler: Changed to print the client type as part of the message and not as MDC.

PR #4238, opened 2026-01-07
ui: Fix and extend table column sizing options

Etsija:
Column sizing with TanStack Table has been problematic, and has needed lots of hard-coded fine-tunings to make the tables look right. What's especially been missing is a "width as a percentage", which is often part of modern table libraries. There is a [GitHub discussion](https://github.com/TanStack/table/discussions/4179) about the matter, which gave an insight to solve this sizing problem. Most importantly, the new `isGrow` and `widthPercentage` column meta properties extend the layout possibilities of the data tables in the UI.

The PR is non-destructive by nature, and all existing tables should (the author has visually inspected them) look as they are for now. The PR enables a few extensions for column sizing, which will probably be used in some upcoming PRs, to make the column sizing work better and be more responsive:

### Column Definition Examples

**Fixed width (existing behavior)**
```typescript
columnHelper.accessor('actions', {
  header: 'Actions',
  size: 70,
})
```

**Percentage of container width**
```typescript
columnHelper.accessor('name', {
  header: 'Name',
  minSize: 100,
  maxSize: 400,
  meta: { widthPercentage: 30 },
})
```

**Grow to fill remaining space**
```typescript
columnHelper.accessor('description', {
  header: 'Description',
  minSize: 200,
  meta: { isGrow: true },
})
```

**No sizing specified (equal distribution)**
```typescript
columnHelper.accessor('status', {
  header: 'Status',
  // No size, widthPercentage, or isGrow
  // Will share remaining space equally with other unsized columns
})
```

### DIFF
--- a/ui/src/components/data-table/data-table-body.tsx
+++ b/ui/src/components/data-table/data-table-body.tsx
@@ -48,8 +48,8 @@ export function DataTableBody<TData>({
                 <TableCell
                   key={cell.id}
                   style={{
-                    minWidth: cell.column.columnDef.size,
-                    maxWidth: cell.column.columnDef.size,
+                    width: cell.column.getSize(),
+                    minWidth: cell.column.columnDef.minSize,
                   }}
                 >
                   {cell.getIsPlaceholder()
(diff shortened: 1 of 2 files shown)


ui/src/hooks/use-table-sizing.ts
```diff
@@ -0,0 +1,235 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
```
lamppu: Nit: copyright year

Issue #4239, opened 2026-01-07
QOL: Display the analyzed/scanned branch when listing runs


nnobelis:
When listing all the runs of a repository, it is often required to find the run for a specific revision (git hash or rolling revision).

Currently, one way is to view (enter) each run till the desired one is found.

IMO, it should be possible to quickly find out the run(s) for a specific revision. Either with a tooltip or a UI filter:

<img width="1181" height="340" alt="Image" src="https://github.com/user-attachments/assets/b7200f40-f393-48f4-a54a-93064b95cbf4" />

Etsija:
Listing the revision is already part of my table refactoring PR #4245.

Etsija:
@nnobelis as I commented, filtering the table with the branch is currently not possible (no support from the endpoint), and #4245 refactored the table into a card layout, so the branch is now readily available in the main table view. Can this ticket be closed, so as not to leave it hanging?

nnobelis:
@Etsija I think you can close the ticket. Thanks !

PR #4245, opened 2026-01-08
ui: Refactor repository runs table into a card layout

Etsija:
The refactored repository runs table:
Screenshot from 2026-01-09 11-29-26

Abbreviated SHA1 shown elsewhere in the UI with the new component:
Screenshot from 2026-01-09 09-41-31

NOTE: More information will be added to this table later, back-end permitting. There's some room in the middle section of each cell, which will be showing rule violations, issues etc. as badges in the future.

NOTE 2: The table uses the old `DataTable` component despite the data column having the "card" layout, to give it some meaningful header titles ("Index", "Run Details", "Actions"). Once some data manipulation (sorting, filtering) to this table is actually possible (see [my comment](https://github.com/eclipse-apoapsis/ort-server/issues/4239#issuecomment-3725610286)), it will be changed into a `DataTableCards` component.


Etsija:
An idea to make the view less clogged would be to use GitHub's 7-character abbreviation of the SHA in most places in the UI (also in the run overview bar). I believe 7 characters from the beginning tells enough to the user, he doesn't need to see all of it. Maybe showing the full SHA in a tooltip.

Screenshot from 2026-01-08 14-25-15

PR #4250, opened 2026-01-09
fix(vault): Fix a failing request after token renewal

oheger-bosch:
When the API token expired and was renewed, the retried request fails for unknown reasons. Adding a retry configuration to the HTTP client fixes the problem. Retrying failed requests for a number of times is beneficial anyway.


secrets/vault/src/main/kotlin/VaultSecretsProvider.kt
```diff
@@ -140,6 +141,10 @@ class VaultSecretsProvider(
                 )
             }
 
+            install(HttpRequestRetry) {
+                maxRetries = 3
+            }
```
MarcelBochtler: Other places also configure `retryIf` to only conditionally retry, e.g. only when server errors occur.
Your implementation would also retry on `400` or `401` responses.

Do you know anything about the failure reasons to add a `retryIf`?

https://ktor.io/docs/client-request-retry.html#conditions

PR #4279, opened 2026-01-14
fix(workers): Always consider all secrets when resolving services

oheger-bosch:
There was an optimization not to load the secrets of the hierarchy if the
environment configuration does not reference any secrets. This does not
work any more after infrastructure services reference their secrets by
name. Fix this by loading all secrets of the current hierarchy
unconditionally in `EnvironmentConfigLoader`.

This is a follow-up of c1101909.


nnobelis (approved):
LGTM

PR #4305, opened 2026-01-19
fix(ui): Abbreviate long usernames and emails

Etsija:
Too long usernames and email addresses break the admin/users table, so use ellipsis to hide them.

Screenshot from 2026-01-19 12-39-57

@sschuberth sadly, there is no TailwindCSS class to put the ellipsis in the middle of the e-mail address, and I'd rather not start solving this programmatically (with a function to cut in the middle and replace by ellipsis), since this solution is so quick and simple.


Etsija:
The long-standing solution to this would be once again to move on to using a card table.

PR #4306, opened 2026-01-19
Fix vault token renewal (2nd attempt)

oheger-bosch:
This is a follow-up to #4250. Unfortunately, the referenced PR did not fix the problem. Therefore, this PR has another idea about the root cause and provides a corresponding fix.


oheger-bosch:
> > Unfortunately, the referenced PR did not fix the problem.
> 
> Should the changes from that PR be fully reverted then?

No, I don't think this is needed. The changes were useful on their own and are further improved in this PR.

PR #4309, opened 2026-01-20
deps(docker): Upgrade to ScanCode 32.5.0

sschuberth:
See [1].

[1]: https://github.com/aboutcode-org/scancode-toolkit/releases/tag/v32.5.0


oheger-bosch:
From reading the referenced release notes, do you think this new command line option `--package-in-compiled` would need to be enabled to get comparable results?

MarcelBochtler (approved):

Issue #1727, opened 2025-01-03
Show stats for each icon in product and organization lists


mnonnenmacher:
The product and organization pages show the total counts of issues, vulnerabilities, and so on:

![image](https://github.com/user-attachments/assets/c525e097-a24c-471f-9ac2-4e21a446d391)

To quickly recognize which items (repos or products) need attention, it would be nice to also show the stats for each item in the list, for example, by showing the icons with colored numbers next to them.

sschuberth:
> by showing the icons with colored numbers next to them.

Could you create a mock-up for this? I'm not sure I understand what you mean.

mnonnenmacher:
The basic idea is like this, but in beautiful:
![grafik](https://github.com/user-attachments/assets/e14026ff-fd45-4a0e-9468-dac3fc22d955)
The details need to be figured out, like if the icons should be colored or the numbers only be shown on hover, and of course where exactly that should be shown.

Etsija:
I like the idea, and I think that's doable.

sschuberth:
> To not clutter the ui too much and get a rough overview in one glance, I'd just use the icons and total number of corr. items, without coloring.

@Etsija could you please share a mockup as a basis for discussion.

PR #4374, opened 2026-01-27
ui: Show vulnerabilities, rule violations, and issues counts for ORT runs

Etsija:
Screenshot from 2026-01-29 10-13-34

NOTE: This is now changed according to the DoubleOpen team review, and aligned such that upcoming PRs, which will introduce similar counts for organization and product overview pages (issue #1727) will have the run status and item count UI elements co-located together, which is better for the UX.


oheger-bosch:
> I don't know what this new maintenance-tasks Docker image is, but it's totally failing the PR check now.

This image is there for a long time already. There seems to be a (probably temporary) issue when building it. Maybe try to just restart the failing check.

PR #4409, opened 2026-01-31
refactor!: Remove the defects model

mnonnenmacher:
There are currently no advisor plugins available in ORT that support reporting defects and there are no known plans to implement such plugins. Therefore remove the defects model to simplify the model classes and the database schema.


mnonnenmacher:
I'd like a review from both @eclipse-apoapsis/doubleopen and @eclipse-apoapsis/bosch for this change. Also see https://github.com/oss-review-toolkit/ort/issues/11389.

PR #4442, opened 2026-02-06
ui: Refactor run configuration page

Etsija:
Screenshot from 2026-02-09 06-51-54


mnonnenmacher:
@Etsija What about always showing the labels and parameters even if the are empty? And would it be easy to add an "expand/collapse all" button?

lamppu (approved):

PR #4460, opened 2026-02-09
Revert "fix(docker): Copy only cargo directory instead of full Rust installation

sschuberth:
This reverts commit ff635bf as it breaks Cargo analysis with:

```
error: rustup could not choose a version of cargo to run, because one
wasn't specified explicitly, and no default is configured.

help: run 'rustup default stable' to download the latest stable release of
Rust and set it as your default toolchain.
```

### DIFF
--- a/workers/analyzer/docker/Analyzer.Dockerfile
+++ b/workers/analyzer/docker/Analyzer.Dockerfile
@@ -501,8 +501,9 @@ RUN chmod -R g+rwX $NVM_DIR
 # Rust
 ENV RUST_HOME=/opt/rust
 ENV CARGO_HOME=$RUST_HOME/cargo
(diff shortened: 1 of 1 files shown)


sschuberth:
[The same](https://github.com/oss-review-toolkit/ort/commit/d7f1669e4ae0c22ea4e376c1d681ec014f7c06cd#commitcomment-176757212) happens for the ORT Docker image, BTW.

sschuberth:
> I did the Analyzer update PR #4422 with the assumption that the ORT Docker image is actually tested and working 😼

No one is blaming you. The ORT Docker image actually _should_ have been tested by running the functional tests against it, and it is currently unclear why this bug was not caught.

PR #4475, opened 2026-02-11
Apply resolutions after each job

Etsija:


workers/analyzer/src/main/kotlin/AnalyzerWorker.kt
```diff
@@ -119,14 +121,27 @@ internal class AnalyzerWorker(
...
+                vulnerabilities = emptyList(),
+                resolutionProvider = resolutionProvider
+            )
+
+            // Calculate unresolved issues for logging.
             val unresolvedIssues = allIssues.filter { issue ->
-                repositoryConfigIssueResolutions.none { it.matches(issue) } &&
-                        globalIssueResolutions.none { it.matches(issue) }
+                resolvedItems.issues.keys.none { resolvedIssue ->
+                    resolvedIssue == issue.mapToModel()
```
oheger-bosch: The `mapToModel()` call should be moved out of the inner loop, so that it is not done repeatedly.

PR #4558, opened 2026-02-25
Refactor the `/packages` endpoint for performance

Etsija:


services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -56,144 +65,124 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
...
+
+            val packageIds = query.map { it[PackagesTable.id] }
+
+            if (packageIds.isEmpty()) {
+                return@blockingQuery ListQueryResult(emptyList(), parameters, totalCount)
             }
-        }
 
-        val sortedResult = filteredResult.sortedWith(comparator)
+            val packages = packageIds.map { PackageDao[it] }
```
oheger-bosch: This looks suspicious as well. Please check that this does not create a SELECT per package.

PR #4565, opened 2026-02-26
fix: Persist Analyzer/Advisor issues before resolution mapping

Etsija:
A recent PR #4475 for applying resolutions after each job has a bug, where the Analyzer and Advisor are storing resolution mappings before storing the runs. This causes the resolved-item links for these workers to be missing in `resolved_issues`, which might skew resolved/unresolved reporting.

Fix by correcting the order so resolution mappings are created against existing issue rows.

NOTE: The author hasn't found any test cases for this from local dev environment, but the fix is probably still best to be done, to be on the safe side and align the order of storing and resolving items in Analyzer and Advisor to the order used in Scanner and Evaluator.

### DIFF
--- a/workers/advisor/src/main/kotlin/AdvisorWorker.kt
+++ b/workers/advisor/src/main/kotlin/AdvisorWorker.kt
@@ -95,8 +95,8 @@ internal class AdvisorWorker(
 
             db.dbQuery {
                 getValidAdvisorJob(jobId)
-                ortRunService.storeResolvedItems(job.ortRunId, resolvedItems)
                 ortRunService.storeAdvisorRun(advisorRun.mapToModel(jobId))
+                ortRunService.storeResolvedItems(job.ortRunId, resolvedItems)
             }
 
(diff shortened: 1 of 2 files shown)


workers/advisor/src/main/kotlin/AdvisorWorker.kt
```diff
@@ -95,8 +95,8 @@ internal class AdvisorWorker(
 
             db.dbQuery {
                 getValidAdvisorJob(jobId)
-                ortRunService.storeResolvedItems(job.ortRunId, resolvedItems)
                 ortRunService.storeAdvisorRun(advisorRun.mapToModel(jobId))
+                ortRunService.storeResolvedItems(job.ortRunId, resolvedItems)
```
mnonnenmacher: Commit message: Please reference the commit you refer to instead of saying "a recent PR".

> which might skew resolved/unresolved reporting

Why so vague, isn't it clearly a bug if a resolved issue is stored as unresolved?

PR #4570, opened 2026-02-26
Refactor the `/issues` endpoint for performance

Etsija:
Similar to PR #4558, eliminate totally the usage of `OrtResult` and expensive in-memory calculations based on it, and replace it with direct SQL queries, to improve the performance of all issues queries.

No numerical data available, but manual verification using local development environment shows the issues query executes almost lightning-fast now (earlier, it took ~1 sec for 50 issues). The real benefit can probably better be assessed on production environment.

Note to @oheger-bosch: I exercised special care to alleviate N+1 query issues this time.

One could also argue that the conventional commit could be `perf(service)` - but I retained `feat(service)` because no-one had anything to complain about it in #4558.

### DIFF
--- a/services/ort-run/src/main/kotlin/IssueService.kt
+++ b/services/ort-run/src/main/kotlin/IssueService.kt
@@ -20,91 +20,248 @@
 package org.eclipse.apoapsis.ortserver.services.ortrun
 
 import org.eclipse.apoapsis.ortserver.dao.QueryParametersException
+import org.eclipse.apoapsis.ortserver.dao.blockingQuery
 import org.eclipse.apoapsis.ortserver.dao.dbQuery
-import org.eclipse.apoapsis.ortserver.dao.repositories.advisorrun.AdvisorRunDao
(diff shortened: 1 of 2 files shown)


services/ort-run/src/main/kotlin/IssueService.kt
```diff
@@ -20,91 +20,248 @@
...
+                .innerJoin(IssuesTable, { issueId }, { id })
+                .join(IdentifiersTable, JoinType.LEFT, OrtRunsIssuesTable.identifierId, IdentifiersTable.id)
+
+            // Step 1: Build the ID query with filtering.
+            val query = baseJoin
+                .select(OrtRunsIssuesTable.id)
+                .where { OrtRunsIssuesTable.ortRunId eq ortRunId }
+
+            // Step 2: Apply resolved/unresolved filter.
+            val resolvedIssueIdsSubquery = ResolvedIssuesTable
```
oheger-bosch: Minor nit: This variable is not used in all cases. Its unnecessary instantiation could be prevented, e.g. by using `by lazy` or extracting the creation into a separate function.

Issue #4471, opened 2026-02-10
Use stored resolution result for `/runs/{runId}/vulnerabilities`


mnonnenmacher:
The vulnerabilities endpoint currently loads the whole ORT result from the database to get the applied resolutions. With the introduction of the `resolved_vulnerabilities` table in #4304 this can now be implemented much more efficiently without loading the whole ORT result.

PR #4573, opened 2026-02-26
Add initial support for managing vulnerability resolutions via the API

mnonnenmacher:
Add a new `resolutions` component which will be used for managing resolutions on the server. For the start, support for vulnerability resolutions is implemented.

This PR replaces #3846 and takes several of the review comments into account. It does not yet add information about new resolutions to the `/runs/{runId}/vulnerabilities` endpoint as #3846 did, because that feature requires more design work for proper UI support and is probably easier to implement once #4471 is done. The reason is that the UI needs to know for each vulnerability resolution if it is managed by the server or not, and because currently resolutions are taken from the `OrtResult` class that information is not available.

### DIFF
--- a/components/resolutions/backend/build.gradle.kts
+++ b/components/resolutions/backend/build.gradle.kts
@@ -31,6 +31,7 @@ dependencies {
     api(projects.components.resolutions.resolutionsApiModel)
 
     api(libs.exposedJdbc)
+    api(libs.kotlinResult)
 
     implementation(projects.dao)
     implementation(projects.model)
@@ -39,7 +40,6 @@ dependencies {
 
     implementation(libs.exposedCore)
     implementation(libs.exposedKotlinDatetime)
-    implementation(libs.kotlinResult)
     implementation(libs.kotlinxSerializationJson)
 
     routesImplementation(projects.components.authorization.authorizationBackend)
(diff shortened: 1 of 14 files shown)


Etsija:
There's a PR in for resolving #4471 now.

components/resolutions/backend/src/main/kotlin/vulnerabilities/VulnerabilityResolutionEvent.kt
```diff
@@ -0,0 +1,59 @@
...
+import kotlin.time.Clock
+import kotlin.time.Instant
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+import org.eclipse.apoapsis.ortserver.model.RepositoryId
+import org.eclipse.apoapsis.ortserver.model.runs.repository.VulnerabilityResolutionReason
+
+internal data class VulnerabilityResolutionEvent(
```
oheger-bosch: Did we ever agree to use event sourcing as a general approach? As far as I recall, it was agreed that
1. you do an experimental implementation on one or two components
2. you present your implementation
3. the team decides about the approach

The steps 2 and 3 never happened as far as I am aware of.

I am rather concerned about the current state of the implementation. It adds significant complexity, while I am pretty sure that many corner cases are not handled.

lamppu (approved):

PR #4581, opened 2026-02-27
ui: Fix search params for links in item count components

Etsija:


ui/src/components/item-counts.tsx
```diff
@@ -0,0 +1,261 @@
...
+  );
+};
+
+type CountBadgeLinkProps = {
+  params: ItemCountsLinkParams;
+  to:
+    | typeof issuesRoute
+    | typeof vulnerabilitiesRoute
+    | typeof ruleViolationsRoute;
+  search?: IssuesSearch | VulnerabilitiesSearch | RuleViolationsSearch;
```
lamppu: I would check if some of these would not need separate type definitions but maybe could use `LinkProps` from TanStack Router, e.g. `LinkProps['params']` (also in the `ItemsCountsLink` type).

lamppu (approved):

PR #4595, opened 2026-03-02
Beautify test assertions

sschuberth:


Etsija:
Regarding these beautifications: they are somewhat of "implicit knowledge", probably known well to seasoned back-end test writers, but not so well to others. We have the [Contributing Guide](https://github.com/eclipse-apoapsis/ort-server/?tab=contributing-ov-file), but could we also have an additional guide for code conventions somewhere? For example, I always have trouble asserting the size of objects in the way the team wants, and the AI agents won't know the conventions, either, unless I can point them to a document which gives the guidelines.

sschuberth:
@eclipse-apoapsis/backend-devs please review to avoid me running into further conflicts.

PR #4597, opened 2026-03-03
Expose hierarchical permissions to OpenAPI, UI

Etsija:
Note for the UI developers: this is a breaking change, and requires regeneration of the OpenAPI specification (`./gradlew :core:generateOpenApiSpec`) and UI API (`pnpm generate:api`).

### DIFF
--- a/components/authorization/api-model/src/commonMain/kotlin/UserInfo.kt
+++ b/components/authorization/api-model/src/commonMain/kotlin/UserInfo.kt
@@ -34,10 +34,14 @@ data class UserInfo(
     /** The full name of the current user. */
     val fullName: String,
 
-    /**
-     * A set with the names of permissions the current user has on the current level of the hierarchy.
-     */
-    val permissions: Set<String>,
+    /** A set with the organization permissions the current user has. */
+    val organizationPermissions: Set<OrganizationPermission> = emptySet(),
+
+    /** A set with the product permissions the current user has. */
+    val productPermissions: Set<ProductPermission> = emptySet(),
+
+    /** A set with the repository permissions the current user has. */
+    val repositoryPermissions: Set<RepositoryPermission> = emptySet(),
 
     /** A flag whether the current user is a superuser. */
     val isSuperuser: Boolean
--- a/components/authorization/backend/src/main/kotlin/routes/Mappings.kt
+++ b/components/authorization/backend/src/main/kotlin/routes/Mappings.kt
@@ -19,11 +19,17 @@
 
 package org.eclipse.apoapsis.ortserver.components.authorization.routes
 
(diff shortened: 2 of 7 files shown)


components/authorization/backend/src/routes/kotlin/routes/userinfo/GetUserInfo.kt
```diff
@@ -88,43 +98,57 @@ internal fun Route.getUserInfo() = get("/authorization/userinfo", {
...
+        repositoryPermissions = permissions.repositoryPermissions,
         isSuperuser = principal.effectiveRole.isSuperuser
     )
     call.respond(HttpStatusCode.OK, userInfo)
 }
 
 /**
- * Return a [Set] with the names of all permissions contained in the given [effectiveRole] for the current hierarchy
- * level. If the current user is a superuser, the set contains all permissions of the ADMIN role on this level.
+ * Return typed permissions contained in the given [effectiveRole] for the current hierarchy level. For wildcard
```
oheger-bosch: Similar to my comment on the commit message: Since the behavior wrt the wildcard level does not change, why does the comment has to be altered? I am fine with adding the sentence about empty sets for this level, but the statement about superusers is still valid.

components/authorization/backend/src/main/kotlin/routes/Mappings.kt
```diff
@@ -46,6 +52,37 @@ fun RepositoryRole.mapToModel() = when (this) {
     RepositoryRole.ADMIN -> ModelRepositoryRole.ADMIN
 }
 
+fun ModelOrganizationPermission.mapToApi() = when (this) {
```
oheger-bosch: Just thinking: As has been discussed, there will be a cleanup on permissions, which will then be another breaking change. Afterward, the permissions model should be changed only in a backward-compatible way. So, it should then be fine to add new permissions, but not to change or delete existing ones. Is this correct?

components/authorization/api-model/src/commonMain/kotlin/UserInfo.kt
```diff
@@ -34,10 +34,14 @@ data class UserInfo(
     /** The full name of the current user. */
     val fullName: String,
 
-    /**
```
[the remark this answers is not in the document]
mnonnenmacher: > Because this changes the API contract, it is considered a breaking change,
however it has been discussed that the CLI consumers are not using user
info endpoint of the API currently.

I would just scratch this sentence, the first part is redundant and the second part is a bit weird (why "CLI consumers", not "API consumers"?) and it references a community discussion where it was said that the users of one specific instance do not use that endpoint which cannot be generalized in this way.

PR #4643, opened 2026-03-09
Store the sources of resolutions

mnonnenmacher:
Store from which source resolutions are originating (global configuration file, repository configuration file, or the server). This is mainly required for the UI later on to decide which resolutions are editable. See the commit messages for details.

### DIFF
--- a/workers/common/src/test/kotlin/resolutions/OrtServerResolutionProviderTest.kt
+++ b/workers/common/src/test/kotlin/resolutions/OrtServerResolutionProviderTest.kt
@@ -466,15 +466,15 @@ class OrtServerResolutionProviderTest : WordSpec({
             val issue2 = Issue(source = "source", message = "match-2")
 
             val expectedResolutionsForIssue1 = listOf(
-                globalResolutions.issues[0].mapToModel(ResolutionSource.REPOSITORY_FILE),
+                globalResolutions.issues[0].mapToModel(ResolutionSource.GLOBAL_FILE),
                 repositoryConfigurationResolutions.issues[0].mapToModel(ResolutionSource.REPOSITORY_FILE),
-                managedResolutions.issues[0].mapToModel(ResolutionSource.REPOSITORY_FILE)
+                managedResolutions.issues[0].mapToModel(ResolutionSource.SERVER)
             )
 
             val expectedResolutionsForIssue2 = listOf(
-                globalResolutions.issues[1].mapToModel(ResolutionSource.REPOSITORY_FILE),
+                globalResolutions.issues[1].mapToModel(ResolutionSource.GLOBAL_FILE),
                 repositoryConfigurationResolutions.issues[1].mapToModel(ResolutionSource.REPOSITORY_FILE),
-                managedResolutions.issues[1].mapToModel(ResolutionSource.REPOSITORY_FILE)
+                managedResolutions.issues[1].mapToModel(ResolutionSource.SERVER)
             )
 
             val matchResult = provider.matchResolutions(listOf(issue1, issue2), emptyList(), emptyList())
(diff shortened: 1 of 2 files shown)


workers/common/src/test/kotlin/resolutions/OrtServerResolutionProviderTest.kt
```diff
@@ -0,0 +1,406 @@
...
+                    comment = "matching global rule violation resolution"
+                ),
+                RuleViolationResolution(
+                    message = "match",
+                    reason = RuleViolationResolutionReason.CANT_FIX_EXCEPTION,
+                    comment = "matching repository configuration rule violation resolution"
+                )
+            )
+
+            provider.getResolutionsFor(ruleViolation) should containExactlyInAnyOrder(expectedRuleViolationResolutions)
```
oheger-bosch: Nit: Actually, these are multiple test cases.

workers/advisor/src/main/kotlin/AdvisorWorker.kt
```diff
@@ -86,10 +86,11 @@ internal class AdvisorWorker(
             val allIssues = advisorRun.results.values.flatten().flatMap { it.summary.issues }
             val allVulnerabilities = advisorRun.results.values.flatten().flatMap { it.vulnerabilities }
 
-            val resolutionProvider = workerContext.createResolutionProvider(
-                RepositoryId(ortRun.repositoryId),
-                ortResult,
+            val resolutionProvider = OrtServerResolutionProvider.create(
```
oheger-bosch: Just wondering why this does not require adjustments in tests?

shared/api-model/src/commonMain/kotlin/ResolutionSource.kt
```diff
@@ -0,0 +1,32 @@
...
+/** An enum that defines the different sources from which a resolution can originate. */
+enum class ResolutionSource {
+    /** The resolution is defined in the global configuration file. */
+    GLOBAL,
+
+    /** The resolution is defined in the repository configuration file. */
+    REPOSITORY,
+
+    /** The resolution is managed by the server. */
+    SERVER
```
oheger-bosch: I assume, this refers to resolutions that have been created via the UI, right? The name `SERVER` does not seem to fit, also the comment "...managed by the server".  I have, however, no better idea either.

Also, `REPOSITORY` might be a bit misleading, since it refers only to the configuration file for this repository and not the structure in the hierarchy.

sschuberth: Maybe we can clarify all of these enum names a bit, like: `GLOBAL_FILE`, `REPOSITORY_FILE`, `SERVER_UI`. Instead of the "_FILE" suffix maybe also "_CONFIG" could be used.

And BTW, I do not think that a "_UI" suffix would be misleading here, just because resolutions _could_ also be sent via `curl` to the API directly, because this should more indicate how that type of resolution is supposed to be managed, so I believe saying "it's supposed to be managed via the UI" is fine.

Etsija: I agree, _REPOSITORY would be misleading, but let's not invent any monster names, either... :smirk:

Why can't we just use SERVER?

dao/src/main/resources/db/migration/V132__addResolutionSource.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE issue_resolutions
```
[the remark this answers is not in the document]
mnonnenmacher: I forgot to push the changes.

oheger-bosch:
From my PoV, the changes are fine.

shared/api-model/src/commonMain/kotlin/ResolutionSource.kt
mnonnenmacher: @sschuberth @Etsija None of you has made a change request, so do you insist on changing the names or not?

Etsija: I have no objections to the original ones, but if @sschuberth wants to, I'm also OK with GLOBAL_FILE, REPOSITORY_FILE, and SERVER.

PR #4661, opened 2026-03-12
Deal with issues from the dependency graph

oheger-bosch:
Issues associated with the dependency graph (for instance, if package metadata could not be resolved) were not returned by the issues service, since they were not stored in the `issues` database table - only as part of the dependency graph in JSON format. Because of that, they were also not shown by the UI.

This PR addresses this problem by writing such issues explicitly to the `issues` table, so that they are handled in the same way as any other issues. When constructing an ORT result from the database, it then has to be ensured that these issues are removed again from the Analyzer result, as they would be duplicates otherwise.


oheger-bosch:
> As a side question: could this PR also help in reporting a bit more helpful Issue than just this for an error in dependency graph creation (for which I have just reported [this ticket](https://github.com/oss-review-toolkit/ort/issues/11597) to ORT core)? Currently, the Issue doesn't even record the Analyzer as the source.

I fear this is a different issue which will not be addressed by this fix.

PR #4692, opened 2026-03-18
Fix UI image on OpenShift

oheger-bosch:
This is a fixup for #4671 which broke the execution of the UI container image in OpenShift.

@mnonnenmacher, please check whether these changes are compatible with your target environment. We may need to come iteratively to a solution that works for both of us.

### DIFF
--- a/ui/docker/entrypoint.sh
+++ b/ui/docker/entrypoint.sh
@@ -25,7 +25,12 @@
 : "${UI_URL:=http://localhost:8082/}"
 
 # Copy template files to the serving directory. This allows the serving directory to be a writable volume mount.
-cp -r /usr/share/nginx/html-template/. /usr/share/nginx/html/
+# Clear the directory first if it exists and has content. This is needed to gracefully deal with restarts of the
+# container.
+if [ -d /usr/share/nginx/html ] && [ "$(ls -A /usr/share/nginx/html)" ]; then
+  rm -rf /usr/share/nginx/html/*
+fi
+cp -r /usr/share/nginx/html-template/* /usr/share/nginx/html/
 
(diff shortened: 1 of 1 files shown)


oheger-bosch:
> Just for my understanding, are these changes required when not mounting emptydir volumes to the target directories?

I don't know about mounting empty volumes. What we saw on OpenShift were permission issues. Here is the error log:

```
cp: can't create directory '/usr/share/nginx/html/./assets': Permission denied
cp: can't create '/usr/share/nginx/html/./favicon.svg': Permission denied
cp: can't create '/usr/share/nginx/html/./index.html': File exists
find: /usr/share/nginx/html/assets: No such file or directory
find: /usr/share/nginx/html/assets: No such file or directory
find: /usr/share/nginx/html/assets: No such file or directory
find: /usr/share/nginx/html/assets: No such file or directory
find: /usr/share/nginx/html/assets: No such file or directory
/entrypoint.sh: line 39: can't create /etc/nginx/conf.d/default.conf: Permission denied
sed: can't create temp file '/usr/share/nginx/html/index.htmlXXXXXX': Permission denied
2026/03/17 15:24:28 [warn] 1#1: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
2026/03/17 15:24:28 [emerg] 1#1: bind() to 0.0.0.0:80 failed (13: Permission denied)
nginx: [emerg] bind() to 0.0.0.0:80 failed (13: Permission denied)
```

PR #4776, opened 2026-03-31
Resolve default package managers in config worker

mnonnenmacher:
Move the `getDefaultPackageManagers` logic from the analyzer worker to
the config worker so that the default package managers are resolved
before the validation script runs. This makes the resolution transparent
to the script, which can now see and potentially override the defaults.

The resolved defaults are stored in `resolvedJobConfigs`, keeping all
job config resolution in one place. The analyzer worker no longer needs
to inject default package managers at runtime.

This is also a preparation for applying plugin templates in the config
worker which requires that the actually used package managers are
already known.

### DIFF
--- a/components/plugin-manager/backend/src/main/kotlin/PluginTemplateService.kt
+++ b/components/plugin-manager/backend/src/main/kotlin/PluginTemplateService.kt
@@ -263,7 +263,7 @@ class PluginTemplateService(
      *  [organizationId]. This is either a template that is assigned to the organization, a global template if no
      *  template is assigned to the organization, or `null` if no template exists for the organization.
      */
-    internal fun getTemplateForOrganization(
+    fun getTemplateForOrganization(
         pluginType: PluginType,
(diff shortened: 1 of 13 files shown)


components/plugin-manager/backend/src/main/kotlin/PluginTemplateService.kt
```diff
@@ -263,7 +263,7 @@ class PluginTemplateService(
      *  [organizationId]. This is either a template that is assigned to the organization, a global template if no
      *  template is assigned to the organization, or `null` if no template exists for the organization.
      */
-    internal fun getTemplateForOrganization(
+    fun getTemplateForOrganization(
```
sschuberth, line 266: Commit message nit: "plugin templated" -> "plugin template"

sschuberth (changes requested):

sschuberth (approved):

PR #4802, opened 2026-04-07
Add optional repository name

### LINKED ISSUE #1368: Relax uniqueness constraint of Repository URLs inside a Product and add optional name field for repositories
I'd like to be able to add multiple instances of same repository URL inside of a Product to organize analysis pipelines better. The URL is however currently enforced to be unique inside a Product. I'd like this constraint to be removed.

As having multiple Repositories with same URL inside one Product makes distinguishing them impossible, it would be great to be able to define a _name_ field for each Repository, that would be shown in the Repository list and the breadcrumb. This would help distinguishing between multiple repositories also without duplicate URLs inside a Product, as it's likely that the repositoriy URLs of a Product share a lot of elements ("https://provider.com/organization").

I'd prefer the name field to be optional to keep the UX straight-forward in cases on single Repository Products, and in those cases display the URL instead of name.

### LINKED ISSUE #1599: Allow custom names / descriptions for repositories
Currently a repository is only identified via its URL, which is quite "builky" in the UI. Also, some way to attach free text information to repositories in needed, e.g. to better describe the "thing" that is hosted in that repository.

To solve both issues at once, I propose to create a free text "Name" property for a repository. That name should then be used as a product-wide unique identifier for the repository that is also used in the breadcrumb bar and elsewhere.

Less preferable, the URL could stay the identifier, but an additional "Description" property could be added for a repository.

Etsija:
### 1. Adding a new repository: an optional name can be given
Screenshot from 2026-04-08 11-23-46

### 2. Editing repository details: you can add/modify/delete the name
Screenshot from 2026-04-08 11-25-02

### 3. The name (if given) is seen in the repository details page and breadcrumb
Screenshot from 2026-04-08 11-25-33

### 4. In product repositories table, name is also shown in case it exists
Screenshot from 2026-04-15 13-08-55

NOTE1: @sschuberth I didn't yet remove the repository ID - maybe @hanna-modica could tell if it's still needed for the CLI usage? We could provide with something handier for copying the URL needed for the CLI.

NOTE2: Although the API commit is marked as breaking (as it changes the API), it is not breaking as such, as the PR only adds an optional `name` to the GET/CREATE/UPDATE repository endpoints, and no back-filling to the database is needed.

NOTE3: This PR also resolves #1368 by relaxing the uniqueness requirement of the repositories: now triplet `(product_id, url, name)` must be unique, so there can be multiple repositories with the same URL inside a product if their names differ. It was also tested manually in the UI that creating a repository with the same URL and omitting the name gives is prevented:

Screenshot from 2026-04-14 06-25-06

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/Repository.kt
+++ b/api/v1/model/src/commonMain/kotlin/Repository.kt
@@ -20,6 +20,7 @@
 package org.eclipse.apoapsis.ortserver.api.v1.model
 
 import io.konform.validation.Validation
+import io.konform.validation.constraints.pattern
 
 import io.ktor.http.parseUrl
 
@@ -48,12 +49,17 @@ data class Repository(
     /** The url to the repository. */
     val url: String,
 
+    /** The name of the repository. */
+    val name: String? = null,
+
     /** The description of the repository. */
     val description: String? = null
 ) {
     companion object {
         const val INVALID_URL_MESSAGE = "The repository URL is malformed."
         const val USER_INFO_MESSAGE = "The repository URL must not contain userinfo."
+        val NAME_PATTERN_REGEX = Product.NAME_PATTERN_REGEX
+        const val NAME_PATTERN_MESSAGE = Product.NAME_PATTERN_MESSAGE
 
         fun isValidUrl(url: String): Boolean = parseUrl(url)?.host?.isValidHost() ?: false
 
@@ -73,6 +79,7 @@ private fun String.isValidHost() = all { it.isLetterOrDigit() || it == '.' || it
 data class PostRepository(
     val type: RepositoryType,
     val url: String,
+    val name: String? = null,
     val description: String? = null
 ) {
     companion object {
@@ -87,6 +94,10 @@ data class PostRepository(
                         !Repository.hasUserInfo(it)
                     } hint Repository.USER_INFO_MESSAGE
                 }
+
+                PostRepository::name ifPresent {
+                    pattern(Repository.NAME_PATTERN_REGEX) hint Repository.NAME_PATTERN_MESSAGE
+                }
             }.invoke(obj)
         }
     }
@@ -99,6 +110,7 @@ data class PostRepository(
 data class PatchRepository(
     val type: OptionalValue<RepositoryType> = OptionalValue.Absent,
     val url: OptionalValue<String> = OptionalValue.Absent,
+    val name: OptionalValue<String?> = OptionalValue.Absent,
     val description: OptionalValue<String?> = OptionalValue.Absent,
 
     /**
@@ -124,6 +136,15 @@ data class PatchRepository(
                             is OptionalValue.Absent -> true
                         }
                     } hint Repository.USER_INFO_MESSAGE
+                }
+
+                PatchRepository::name {
+                    constrain("must match the expected pattern '${Repository.NAME_PATTERN_REGEX.pattern}'") {
+                        when (it) {
+                            is OptionalValue.Present -> it.value?.matches(Repository.NAME_PATTERN_REGEX) ?: true
+                            is OptionalValue.Absent -> true
+                        }
+                    } hint Repository.NAME_PATTERN_MESSAGE
                 }
             }.invoke(obj)
         }
--- a/core/src/main/kotlin/apiDocs/ProductsDocs.kt
+++ b/core/src/main/kotlin/apiDocs/ProductsDocs.kt
@@ -163,14 +163,16 @@ val getProductRepositories: RouteConfig.() -> Unit = {
                                 organizationId = 2,
                                 productId = 3,
                                 type = RepositoryType.GIT,
-                                url = "https://example.com/first/repo.git"
+                                url = "https://example.com/first/repo.git",
+                                name = "first-repo"
                             ),
                             Repository(
                                 id = 2,
                                 organizationId = 3,
                                 productId = 4,
                                 type = RepositoryType.SUBVERSION,
-                                url = "https://example.com/second/repo"
+                                url = "https://example.com/second/repo",
+                                name = "second-repo"
                             )
                         ),
                         PagingData(
@@ -199,7 +201,8 @@ val postRepository: RouteConfig.() -> Unit = {
             example("Create repository") {
                 value = PostRepository(
                     type = RepositoryType.GIT,
-                    url = "https://example.com/namspace/repo.git"
+                    url = "https://example.com/namspace/repo.git",
+                    name = "repo"
                 )
             }
         }
@@ -215,7 +218,8 @@ val postRepository: RouteConfig.() -> Unit = {
                         organizationId = 2,
                         productId = 3,
                         type = RepositoryType.GIT,
-                        url = "https://example.com/namspace/repo.git"
+                        url = "https://example.com/namspace/repo.git",
+                        name = "repo"
                     )
                 }
             }
--- a/core/src/main/kotlin/apiDocs/ProductsDocs.kt
+++ b/core/src/main/kotlin/apiDocs/ProductsDocs.kt
@@ -147,7 +147,7 @@ val getProductRepositories: RouteConfig.() -> Unit = {
         standardListQueryParameters()
 
         queryParameter<String>("filter") {
-            description = "A regular expression to filter repositories by URL."
+            description = "A regular expression to filter repositories by name or URL."
         }
     }
 
--- a/core/src/test/kotlin/api/ProductsRouteIntegrationTest.kt
+++ b/core/src/test/kotlin/api/ProductsRouteIntegrationTest.kt
@@ -279,18 +279,22 @@ class ProductsRouteIntegrationTest : AbstractIntegrationTest({
                 val type = RepositoryType.GIT
                 val url1 = "https://example.com/repo1.git"
                 val url2 = "https://example.com/repo2.git"
+                val name1 = "repo1"
+                val name2 = "repo2"
                 val description = "description"
 
                 val createdRepository1 = productService.createRepository(
                     type = type,
                     url = url1,
                     productId = createdProduct.id,
+                    name = name1,
                     description = description
                 )
                 val createdRepository2 = productService.createRepository(
                     type = type,
                     url = url2,
                     productId = createdProduct.id,
+                    name = name2,
                     description = description
                 )
 
@@ -299,8 +303,24 @@ class ProductsRouteIntegrationTest : AbstractIntegrationTest({
                 response shouldHaveStatus HttpStatusCode.OK
                 response shouldHaveBody PagedResponse(
                     listOf(
-                        Repository(createdRepository1.id, orgId, createdProduct.id, type.mapToApi(), url1, description),
-                        Repository(createdRepository2.id, orgId, createdProduct.id, type.mapToApi(), url2, description)
+                        Repository(
+                            id = createdRepository1.id,
+                            organizationId = orgId,
+                            productId = createdProduct.id,
+                            type = type.mapToApi(),
+                            url = url1,
+                            name = name1,
+                            description = description
+                        ),
+                        Repository(
+                            id = createdRepository2.id,
+                            organizationId = orgId,
+                            productId = createdProduct.id,
+                            type = type.mapToApi(),
+                            url = url2,
+                            name = name2,
+                            description = description
+                        )
                     ),
                     PagingData(
                         limit = DEFAULT_LIMIT,
@@ -319,24 +339,29 @@ class ProductsRouteIntegrationTest : AbstractIntegrationTest({
                 val type = RepositoryType.GIT
                 val url1 = "https://example.com/repo1.git"
                 val url2 = "https://example.com/repo2.git"
+                val name1 = "repo1"
+                val name2 = "repo2"
                 val description = "description"
 
                 val createdRepository1 = productService.createRepository(
                     type = type,
                     url = url1,
                     productId = createdProduct.id,
+                    name = name1,
                     description = description
                 )
                 val createdRepository2 = productService.createRepository(
                     type = type,
                     url = url2,
                     productId = createdProduct.id,
+                    name = name2,
                     description = description
                 )
                 productService.createRepository(
                     type = type,
                     url = "https://example.com/hidden-repo.git",
                     productId = createdProduct.id,
+                    name = "hidden-repo",
                     description = "You cannot see me"
                 )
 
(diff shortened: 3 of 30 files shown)


lamppu (commented):
LGTM

Etsija:
I realised the repositories can still only be filtered by their URL, so I need to add support for filtering by name as well now.

sschuberth:
I haven't checked, but what about the delete confirmation? Should deletion still be confirmed by URL if a name is present, or should then (only) the name be used? Let's discuss with the team.

Etsija:
> I haven't checked, but what about the delete confirmation? Should deletion still be confirmed by URL if a name is present, or should then (only) the name be used? Let's discuss with the team.

I can do either; it's URL confirmation only for now. We didn't actually discuss this with the team today, so which way do you want it, @sschuberth?

sschuberth:
> We didn't actually discuss this with the team today

I also forgot to ask: While this PR addresses #1599, it does not address #1368 yet (for the reason mentioned [here](https://github.com/eclipse-apoapsis/ort-server/issues/1599#issuecomment-3996366665)) yet, correct? If so, the primary identifier of a repository stays to be the URL, and while that's the case, IMO also the URL should be used in the delete confirmation. Does that make sense to you as well, @Etsija?

Etsija:
> > We didn't actually discuss this with the team today
> 
> I also forgot to ask: While this PR addresses #1599, it does not address #1368 yet (for the reason mentioned [here](https://github.com/eclipse-apoapsis/ort-server/issues/1599#issuecomment-3996366665)) yet, correct? If so, the primary identifier of a repository stays to be the URL, and until that's the case, IMO also the URL should be used in the delete confirmation. Does that make sense to you as well, @Etsija?

Yes it does.

Etsija:
@sschuberth what's the status of this PR, could it be approved now?

sschuberth:
> @sschuberth what's the status of this PR, could it be approved now?

It changes areas of the backend code that I'd prefer others to review. I don't see a need to rush this until @mnonnenmacher is back on Monday, but of course @eclipse-apoapsis/bosch is welcome to review, too.

dao/src/main/resources/db/migration/V142__addRepositoryName.sql
```diff
@@ -0,0 +1,3 @@
+-- Add a name column to the repositories table.
+
+ALTER TABLE repositories ADD COLUMN name TEXT;
```
mnonnenmacher: I'm missing a constraint here, I don't think it should be allowed to use the same name multiple names within a single product. I think it would be best to replace the existing `repositories_url_product_id_key` constraint with one that requires the triple of URL, product ID and name to be unique.

dao/src/test/kotlin/repositories/repository/DaoRepositoryRepositoryTest.kt
```diff
@@ -70,25 +70,58 @@ class DaoRepositoryRepositoryTest : StringSpec({
...
         val dbEntry = repositoryRepository.get(createdRepository.id)
 
-        dbEntry shouldBe Repository(createdRepository.id, orgId, productId, type, url, description)
+        dbEntry shouldBe Repository(
+            id = createdRepository.id,
+            organizationId = orgId,
+            productId = productId,
+            type = type,
+            url = url,
+            description = description
```
mnonnenmacher, line 85: For all test cases that set the optional description like this one it would be consistent to also set the optional name now.

model/src/commonMain/kotlin/repositories/RepositoryRepository.kt
```diff
@@ -35,7 +35,13 @@ interface RepositoryRepository {
     /**
      * Create a repository.
      */
-    fun create(type: RepositoryType, url: String, productId: Long, description: String?): Repository
+    fun create(
+        type: RepositoryType,
+        url: String,
+        productId: Long,
+        name: String? = null,
```
mnonnenmacher: If `description` does not have a default value, `name` should not have one either. Or a default value for `description` should be introduced in a preceding commit.

services/hierarchy/src/main/kotlin/ProductService.kt
```diff
@@ -63,9 +63,16 @@ class ProductService(
         type: RepositoryType,
         url: String,
         productId: Long,
-        description: String?
+        description: String?,
+        name: String? = null
```
mnonnenmacher: The parameter order and default values in the repository and service should be consistent.

services/hierarchy/src/test/kotlin/ProductServiceTest.kt
```diff
@@ -87,6 +88,26 @@ class ProductServiceTest : WordSpec({
         }
     }
 
+    "createRepository" should {
+        "create a repository with a name" {
```
mnonnenmacher: The "with a name" is redundant as the function tests all parameters.

services/hierarchy/src/test/kotlin/RepositoryServiceTest.kt
```diff
@@ -294,6 +294,23 @@ class RepositoryServiceTest : WordSpec({
...
+            )
+
+            val updatedRepository = service.updateRepository(
+                repositoryId = repository.id,
+                name = "Updated repository name".asPresent()
+            )
+
+            updatedRepository.name shouldBe "Updated repository name"
+            service.getRepository(repository.id)?.name shouldBe "Updated repository name"
+        }
```
mnonnenmacher: This should be included in the test above instead.

core/src/test/kotlin/api/ProductsRouteIntegrationTest.kt
```diff
@@ -437,22 +465,37 @@ class ProductsRouteIntegrationTest : AbstractIntegrationTest({
...
         }
 
         "respond with 'Bad Request' if the repository's URL is malformed" {
             integrationTestApplication {
                 val createdProduct = createProduct()
 
-                val repository = PostRepository(ApiRepositoryType.GIT, "https://git hub.com/org/repo.git")
+                val repository = PostRepository(
+                    type = ApiRepositoryType.GIT,
+                    url = "https://git hub.com/org/repo.git"
```
mnonnenmacher: Are this change and the two below required?

api/v1/model/src/commonMain/kotlin/Repository.kt
```diff
@@ -73,6 +76,7 @@ private fun String.isValidHost() = all { it.isLetterOrDigit() || it == '.' || it
 data class PostRepository(
     val type: RepositoryType,
     val url: String,
+    val name: String? = null,
```
mnonnenmacher, line 82: There should be some validation for the name, I would suggest to use the same pattern as for product and organization names.

ui/src/routes/organizations/$orgId/products/$productId/-components/product-repository-table.tsx
```diff
@@ -97,7 +97,7 @@ export const ProductRepositoryTable = () => {
                 repoId: row.original.id.toString(),
               }}
             >
-              {row.original.url}
+              {row.original.name || row.original.url}
```
mnonnenmacher: IMO we should not hide the URL in this table if a name is set. This will be confusing later on if a filter matches the URL but only the name is shown.

core/src/main/kotlin/apiDocs/ProductsDocs.kt
```diff
@@ -147,7 +147,7 @@ val getProductRepositories: RouteConfig.() -> Unit = {
         standardListQueryParameters()
 
         queryParameter<String>("filter") {
-            description = "A regular expression to filter repositories by URL."
+            description = "A regular expression to filter repositories by name or URL."
```
mnonnenmacher, line 150: Commit message: IMO that's a feat not a fix.
Nit: "apidocs" -> "API docs"

dao/src/main/kotlin/repositories/repository/DaoRepositoryRepository.kt
```diff
@@ -120,6 +128,9 @@ class DaoRepositoryRepository(private val db: Database) : RepositoryRepository {
     override fun deleteByProduct(productId: Long): Int = db.blockingQuery {
         RepositoriesTable.deleteWhere { RepositoriesTable.productId eq productId }
     }
+
+    private fun repositoryNameExpression() =
+        CustomFunction<String>("coalesce", TextColumnType(), RepositoriesTable.name, stringLiteral(""))
```
mnonnenmacher: Why is this needed? For example, the name filter for products just uses `ProductsTable.name.applyIRegex(it.value)`.

model/src/commonMain/kotlin/repositories/RepositoryRepository.kt
```diff
@@ -55,8 +55,8 @@ interface RepositoryRepository {
     fun getHierarchy(id: Long): Hierarchy
 
     /**
-     * List all repositories according to the given [parameters]. Optionally, a [urlFilter] on the repository URL and a
-     * [hierarchyFilter] can be provided.
+     * List all repositories according to the given [parameters]. Optionally, a [urlFilter] on the repository name or
+     * URL and a [hierarchyFilter] can be provided.
```
mnonnenmacher: `urlFilter` should be renamed if it filters by name as well.

services/hierarchy/src/main/kotlin/ProductService.kt
```diff
@@ -105,7 +105,7 @@ class ProductService(
 
     /**
      * List all repositories for a [product][productId] that are visible to a specific [user][userId] according to the
-     * given [parameters] and [urlFilter].
+     * given [parameters] and [urlFilter], which is applied to the repository name or URL.
```
mnonnenmacher: `urlFilter` should be renamed if it filters by name as well.

mnonnenmacher (changes requested):

dao/src/main/resources/db/migration/V142__addRepositoryName.sql
Etsija: Wouldn't replacing the existing `(url, product_id)` uniqueness constraint with the triplet `(url, product_id, name)` actually weaken the condition, as it would allow the same URL to appear multiple times in the same product, as long as the names differ?

If the intended uniqueness rules are:
- URL stays the primary identifier within a product
- optional non-null names must also be unique within a product

then I think it could be achieved with a partial unique index such as
```
CREATE UNIQUE INDEX repositories_product_id_name_key
    ON repositories (product_id, name)
    WHERE name IS NOT NULL;
```

That would enforce that
- repository names are optional
- if a name is set, it must be unique within the product

mnonnenmacher: I thought if the index needs to be changed anyway it could be done so that it already solves #1368 in addition to #1599.

services/hierarchy/src/main/kotlin/ProductService.kt
Etsija: This and the previous comment required to touch also API integration tests outside the scope of this PR, to make them compile (name needed to be added).

core/src/test/kotlin/api/ProductsRouteIntegrationTest.kt
Etsija: For this one, yes, as I added `name` now to it, to test all params. For the next two, the layout change wasn't necessary.

ui/src/routes/organizations/$orgId/products/$productId/-components/product-repository-table.tsx
Etsija: Let's think about how to address this with a follow-up; @sschuberth had an opinion that the current way is acceptable, even if your comment contains a valid point.

sschuberth: > This will be confusing later on if a filter matches the URL but only the name is shown.

I don't have a too strong opinion here. For me it's fine for a simple feature like filtering to apply some "fuzzyness" that happens to silently match a URL.

dao/src/test/kotlin/repositories/repository/DaoRepositoryRepositoryTest.kt
Etsija: Solved by replacing the constraint with a new one.

dao/src/main/kotlin/repositories/repository/DaoRepositoryRepository.kt
```diff
@@ -46,11 +46,18 @@ import org.jetbrains.exposed.v1.jdbc.deleteWhere
 import org.jetbrains.exposed.v1.jdbc.select
 
 class DaoRepositoryRepository(private val db: Database) : RepositoryRepository {
-    override fun create(type: RepositoryType, url: String, productId: Long, description: String?) = db.blockingQuery {
+    override fun create(
```
mnonnenmacher, line 53: Please mark this commit as breaking and add a message like this to the commit message body:

```
BREAKING CHANGE: The minimum required PostgreSQL version is now 15 because this commit uses `UNIQUE NULLS NOT DISTINCT` which is only available since this version.
```

ui/src/routes/organizations/$orgId/products/$productId/-components/product-repository-table.tsx
mnonnenmacher: > For me it's fine for a simple feature like filtering to apply some "fuzzyness" that happens to silently match a URL.

To me that's bad UX, it could look like the filter logic is broken.

But filter is just one aspect, I would generally want to always see the URL, just like we always show the description.

dao/src/main/kotlin/repositories/repository/DaoRepositoryRepository.kt
```diff
@@ -72,13 +76,17 @@ class DaoRepositoryRepository(private val db: Database) : RepositoryRepository {
...
-    override fun list(parameters: ListQueryParameters, urlFilter: FilterParameter?, hierarchyFilter: HierarchyFilter) =
+    override fun list(parameters: ListQueryParameters, filter: FilterParameter?, hierarchyFilter: HierarchyFilter) =
         db.blockingQuery {
-            val urlCondition = urlFilter?.let {
-                RepositoriesTable.url.applyIRegex(it.value)
+            val filterCondition = filter?.let {
+                RepositoriesTable.url.applyIRegex(it.value) or
+                    (
+                        RepositoriesTable.name.isNotNull() and
+                            RepositoriesTable.name.castTo(TextColumnType()).applyIRegex(it.value)
```
mnonnenmacher, line 85: Is `castTo(TextColumnType())` required and why?

mnonnenmacher (changes requested):

dao/src/main/kotlin/repositories/repository/DaoRepositoryRepository.kt
Etsija: Yes, because `name` is `Column<String?>` which is nullable, while `applyIRegex` requires `Expression<String>` which is non-nullable. The `isNotNull()` handles null safety at the SQL level, and `castTo` handles it at the Kotlin type level.

ui/src/routes/organizations/$orgId/products/$productId/-components/product-repository-table.tsx
Etsija: Name added, now URL always shown, and name (if exists) as well. One picture about it added to the PR description.

dao/src/main/kotlin/repositories/repository/DaoRepositoryRepository.kt
mnonnenmacher, line 53: Commit message: The `BREAKING CHANGE` section needs to be a footer, so the last section before the `Signed-off-by`. See: https://www.conventionalcommits.org/en/v1.0.0/#specification

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #4815, opened 2026-04-10
Query and show detected licenses

### LINKED ISSUE #2074: Need an endpoint to get license findings
Either a dedicated endpoint for this is needed, or then [this endpoint](https://eclipse-apoapsis.github.io/ort-server/api/get-packages-by-run-id) needs to be extended to provide with the detected licenses (which are available after the Scanner job is completed).

Etsija:
This PR adds a view for all license detections from the Scanner to the ORT run details. It is implemented as three endpoints and correspondingly, the UI consists of three tables within each other:

#### 1. Top-level view: all detected licenses with the corresponding package counts
Screenshot from 2026-04-13 12-01-18

#### 2. All projects/packages that have a detected license
Screenshot from 2026-04-13 12-01-53

#### 3. License detection data for a license and project/package
Screenshot from 2026-04-13 12-02-31

NOTE: This is the first, sub-optimal implementation of the feature. It is recognized that the database queries are far from optimal, and even for a relatively small project, the top-level query takes between 1-2 seconds. There are several points for optimization, most notably:
- possibly decomposing the SPDX license expressions into their respective atomic license components, which will change the views considerably
- package curations are _not_ taken into account at this stage. This can lead to false data shown, and needs to be corrected
- the queries for fetching even relatively simple data are _very_ long join chains on our normalized database. Most notable performance gains would probably be available with the following:

1. (easy) Add index on `scanner_jobs.ort_run_id`.
2. (schema change) Add a direct FK from `scan_results` to `package_provenances`. The queries need a cross-product between `scan_results` and `package_provenances` via `scanner_runs`, filtered by the `provenanceCondition()` multi-column string comparison (matching URL/hash fields across two tables). There's no direct FK between the two — they're linked by matching denormalized text fields. Adding a `package_provenance_id` FK column to `scan_results` would replace this expensive string-comparison filter with an indexed join.


sschuberth:
One thing that crossed my mind while evaluating https://github.com/mstykow/provenant is: Where / how should we display (potentially different) results from different scanners that were all configured to be run?

I'm not saying that we should address that question in this PR already, but it would be important to me that we don't go in a wrong direction here, which then makes adding this kind of information later on unnecessary hard.

Probably again something we should try to discuss briefly in a meeting to start with.

Etsija:
If it is possible to run several scanners in an ORT run, then I would think the relation would (or should, if it isn't already) be retained in `scanner_runs.ort_run_id`. That could maybe already provide a starting point for implementation.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings/-components/detected-license-findings-table.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  if (isError) {
+    toastError('Unable to load data', error);
+    return <></>;
+  }
+
+  return (
+    <div className='space-y-2 p-2'>
+      <div className='text-muted-foreground text-sm'>
+        License findings ({findings.pagination.totalCount} in total)
```
sschuberth: As discussed, this should say "Detected licenses".

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings/-components/license-findings-view.tsx
```diff
@@ -0,0 +1,298 @@
...
+  const filtersInUse =
+    totalDetectedLicenses.pagination.totalCount !==
+    detectedLicenseFindings.pagination.totalCount;
+  const matching = `, ${detectedLicenseFindings.pagination.totalCount} matching filters`;
+
+  return (
+    <Card className='h-fit'>
+      <CardHeader>
+        <CardTitle>
+          License Findings ({totalDetectedLicenses.pagination.totalCount} in
```
sschuberth: As discussed, this should say "Detected licenses".

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/route.tsx
```diff
@@ -64,14 +65,15 @@ const Layout = () => {
...
-        // This menu option is temporarily disabled until the license findings page is implemented.
-        // The code is still left here as a bookmark/reminder for future implementation.
-        //{
-        //  title: 'License Findings',
-        //  to: '/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings',
-        //  params,
-        //  icon: () => <FileText className='h-4 w-4' />,
-        //},
+        {
+          title: 'License Findings',
```
sschuberth: As discussed, this should say "Detected licenses".

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings/-components/license-findings-view.tsx
```diff
@@ -0,0 +1,298 @@
...
+          total
+          {filtersInUse && matching})
+        </CardTitle>
+        <CardDescription>
+          Detected licenses across packages in this run.
```
sschuberth: I'd propose something like "Licenses as detected in projects and packages by the configured scanners".

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings/-components/detected-license-findings-table.tsx
```diff
@@ -0,0 +1,162 @@
...
+    }),
+    findingColumnHelper.accessor('endLine', {
+      id: 'endLine',
+      header: 'End Line',
+      meta: {
+        widthPercentage: 10,
+      },
+    }),
+    findingColumnHelper.accessor('score', {
+      id: 'score',
```
sschuberth, line 98: As discussed, esp. the score is very scanner specific, so we should try to get the scanner name and display it here, probably as a new last column.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings/-components/detected-license-findings-table.tsx
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
```
sschuberth: Is the year correct here?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings/-components/detected-license-packages-table.tsx
```diff
@@ -0,0 +1,262 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
```
sschuberth: Is there year correct here?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings/-components/license-findings-view.tsx
```diff
@@ -0,0 +1,298 @@
+/*
+ * Copyright (C) 2024 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
```
sschuberth: Is the year correct here?

sschuberth (dismissed):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings/-components/detected-license-findings-table.tsx
Etsija: Do you want to also show the scanner_version? Both the name and version are available in the db.

sschuberth: If it's already there, then yes. I don't think we need a separate column for the version though, but just showing it in the "Scanner" column as e.g. "ScanCode 32.0.0" should suffice.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings/-components/detected-license-findings-table.tsx
Etsija: That would be wrong: here, we rather need "license findings", otherwise, it would read "there were 95 detected licenses as total", when it should read as "there were 95 license findings with this license in this package".

Etsija: To clarify: I believe this is how it should be.

Screenshot from 2026-04-13 11-39-37

sschuberth: I was probably commenting on the wrong occurrence (due to the "in total" I though it was the page title). Anyway, I agree with your screenshot.

components/license-findings/backend/src/main/kotlin/LicenseFindingService.kt
```diff
@@ -0,0 +1,350 @@
...
+import org.jetbrains.exposed.v1.jdbc.select
+
+class LicenseFindingService(private val db: Database) {
+    fun getRunDetectedLicenses(
+        ortRunId: Long,
+        parameters: ListQueryParameters,
+        licenseFilter: String?
+    ): ListQueryResult<DetectedLicense> = db.blockingQuery {
+        // includeAnalyzerData = false: skips the 4 LEFT JOINs to analyzer tables that are only
+        // needed for purl retrieval. Including them causes row multiplication via packages_analyzer_runs.
```
mnonnenmacher: This only needs to be documented for the `buildQueryContext` function, not on the call side.

components/license-findings/backend/src/main/kotlin/LicenseFindingService.kt
```diff
@@ -0,0 +1,350 @@
...
+            when (orderField.name) {
+                "license" -> query.orderBy(LicenseFindingsTable.license to sortOrder)
+                "packageCount" -> query.orderBy(packageCount to sortOrder)
+                else -> throw QueryParametersException("Unsupported sort field: '${orderField.name}'.")
+            }
+        }
+
+        val limit = parameters.limit ?: ListQueryParameters.DEFAULT_LIMIT
+        val offset = parameters.offset ?: 0L
+        query.limit(limit).offset(offset)
```
mnonnenmacher: Why not just:
```kotlin
query.limit(parameters.limit ?: ListQueryParameters.DEFAULT_LIMIT).offset(parameters.offset ?: 0L)
```

components/license-findings/backend/src/main/kotlin/LicenseFindingService.kt
```diff
@@ -0,0 +1,350 @@
...
+import org.jetbrains.exposed.v1.core.isNotNull
+import org.jetbrains.exposed.v1.core.min
+import org.jetbrains.exposed.v1.core.or
+import org.jetbrains.exposed.v1.core.stringLiteral
+import org.jetbrains.exposed.v1.jdbc.Database
+import org.jetbrains.exposed.v1.jdbc.andWhere
+import org.jetbrains.exposed.v1.jdbc.select
+
+class LicenseFindingService(private val db: Database) {
+    fun getRunDetectedLicenses(
```
mnonnenmacher: I think we usually would use a more expressive name like `getDetectedLicensesForRun`.
And the public functions in this service should be documented.

components/license-findings/backend/src/main/kotlin/LicenseFindingService.kt
```diff
@@ -0,0 +1,350 @@
...
+                    license = row[LicenseFindingsTable.license],
+                    packageCount = row[packageCount]
+                )
+            },
+            params = parameters,
+            totalCount = rows.firstOrNull()?.get(totalCount) ?: 0L
+        )
+    }
+
+    fun getRunPackagesWithDetectedLicense(
```
mnonnenmacher: `getPackagesWithDetectedLicenseForRun`

components/license-findings/backend/src/main/kotlin/LicenseFindingService.kt
```diff
@@ -0,0 +1,350 @@
...
+
+                "purl" -> query.orderBy(purl to sortOrder)
+
+                else -> throw QueryParametersException("Unsupported sort field: '${orderField.name}'.")
+            }
+        }
+
+        val limit = parameters.limit ?: ListQueryParameters.DEFAULT_LIMIT
+        val offset = parameters.offset ?: 0L
+        query.limit(limit).offset(offset)
```
mnonnenmacher: Why not just:
```kotlin
query.limit(parameters.limit ?: ListQueryParameters.DEFAULT_LIMIT).offset(parameters.offset ?: 0L)
```

components/license-findings/backend/src/main/kotlin/LicenseFindingService.kt
```diff
@@ -0,0 +1,350 @@
...
+                    ortIdentifier = buildOrtIdentifier(row),
+                    purl = row.getOrNull(purl)
+                )
+            },
+            params = parameters,
+            totalCount = rows.firstOrNull()?.get(totalCount) ?: 0L
+        )
+    }
+
+    fun getRunDetectedLicenseFindings(
```
mnonnenmacher: `getLicenseFindingsForRun`

components/license-findings/backend/src/main/kotlin/LicenseFindingService.kt
```diff
@@ -0,0 +1,350 @@
...
+import org.jetbrains.exposed.v1.core.eq
+import org.jetbrains.exposed.v1.core.isNotNull
+import org.jetbrains.exposed.v1.core.min
+import org.jetbrains.exposed.v1.core.or
+import org.jetbrains.exposed.v1.core.stringLiteral
+import org.jetbrains.exposed.v1.jdbc.Database
+import org.jetbrains.exposed.v1.jdbc.andWhere
+import org.jetbrains.exposed.v1.jdbc.select
+
+class LicenseFindingService(private val db: Database) {
```
mnonnenmacher, line 67: Add docs for the service.

components/license-findings/backend/src/main/kotlin/LicenseFindingService.kt
```diff
@@ -0,0 +1,350 @@
...
+        append(row[IdentifiersTable.name])
+        append(':')
+        append(row[IdentifiersTable.version])
+    }
+
+private fun OrderDirection.toSortOrder(): SortOrder =
+    when (this) {
+        OrderDirection.ASCENDING -> SortOrder.ASC
+        OrderDirection.DESCENDING -> SortOrder.DESC
+    }
```
mnonnenmacher: There already is a helper function for this in the `dao` module.

components/license-findings/backend/src/test/kotlin/LicenseFindingServiceTest.kt
```diff
@@ -0,0 +1,556 @@
...
+            endLine = finding.endLine
+            score = finding.score
+            this.scanSummary = scanSummary
+        }
+    }
+
+    return scanSummary
+}
+
+private fun Identifier.toCoordinates(): String = "$type:$namespace:$name:$version"
```
mnonnenmacher: There already is the same function in the search component, instead of repeating it here it should be made available globally by moving it to the `Identifier` class in a separate commit.

components/license-findings/backend/src/test/kotlin/LicenseFindingServiceTest.kt
```diff
@@ -0,0 +1,556 @@
...
+    init {
+        beforeEach {
+            db = dbExtension.db
+            fixtures = dbExtension.fixtures
+            service = LicenseFindingService(db)
+        }
+
+        "getRunDetectedLicenses" should {
+            "return all licenses with package counts sorted by license ascending" {
+                val seed = seedData(fixtures, db)
```
mnonnenmacher: If every test is calling this anyway it should be done in `beforeEach` instead. There is just one test which uses `duplicatePackageEntries = true`, this test should adapt the default scenario accordingly instead.

core/src/main/kotlin/di/Module.kt
```diff
@@ -188,6 +189,7 @@ fun ortServerModule(config: ApplicationConfig, db: Database?, authorizationServi
     singleOf(::ContentManagementService)
     singleOf(::InfrastructureServiceService)
     singleOf(::IssueService)
+    singleOf(::LicenseFindingService)
```
mnonnenmacher, line 192: This should be part of the commit where `core` starts using the new service.

core/build.gradle.kts
```diff
@@ -75,6 +75,12 @@ dependencies {
             requireCapability("$group:routes")
         }
     }
+    implementation(projects.components.licenseFindings.licenseFindingsBackend)
+    implementation(projects.components.licenseFindings.licenseFindingsBackend) {
+        capabilities {
+            requireCapability("$group:routes:$version")
+        }
+    }
```
mnonnenmacher, line 83: This should be part of the commit where `core` starts using the new service.

components/license-findings/backend/src/routes/kotlin/routes/GetRunDetectedLicenseFindings.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.eclipse.apoapsis.ortserver.shared.apimodel.SortProperty
+import org.eclipse.apoapsis.ortserver.shared.ktorutils.jsonBody
+import org.eclipse.apoapsis.ortserver.shared.ktorutils.pagingOptions
+import org.eclipse.apoapsis.ortserver.shared.ktorutils.requireIdParameter
+import org.eclipse.apoapsis.ortserver.shared.ktorutils.standardListQueryParameters
+
+internal fun Route.getRunDetectedLicenseFindings(
+    service: LicenseFindingService,
+    checker: AuthorizationChecker
+) = get("licenses/{license}/packages/{identifier}/findings", {
```
mnonnenmacher: Routes should be self-contained like for other components, so this should be the full path `runs/{runId}/licenses/...`. Otherwise it depends on the context where it is called if the `runId` parameter is actually available.

components/license-findings/backend/src/routes/kotlin/routes/GetRunDetectedLicenses.kt
```diff
@@ -0,0 +1,86 @@
...
+                                offset = 0,
+                                totalCount = 2,
+                                sortProperties = listOf(SortProperty("license", SortDirection.ASCENDING))
+                            )
+                        )
+                    }
+                }
+            }
+        }
+    }, checker) {
```
mnonnenmacher: The required permission should not be passed as a `checker` variable and instead be hard-coded here like for other components.

core/src/main/kotlin/api/RunsRoute.kt
```diff
@@ -162,6 +165,8 @@ fun Route.runs() = route("runs") {
     }
 
     route("{runId}") {
+        licenseFindingRoutes(licenseFindingService, requireRunPermission())
```
mnonnenmacher: The routes from the new component should be included in `Routes.kt` like for other components.

mnonnenmacher (changes requested):

components/license-findings/backend/src/main/kotlin/LicenseFindingService.kt
Etsija: I fixed also limiting for the third service function, and as one function was missing the limit and offset test, I also added it.

core/src/main/kotlin/api/RunsRoute.kt
Etsija: I cannot find `Routes.kt` from the codebase. Did you mean `core/src/main/kotlin/plugins/Routing.kt`?

sschuberth: Maybe `components/authorization/backend/src/main/kotlin/routes/AuthorizedRoutes.kt`?

components/license-findings/api-model/src/commonMain/kotlin/DetectedLicensePackage.kt
```diff
@@ -0,0 +1,28 @@
...
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.components.licensefindings
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class DetectedLicensePackage(
+    val ortIdentifier: String,
```
mnonnenmacher: We call this just `identifier` everywhere else and it should be of type `Identifier` instead of `String`.

components/license-findings/api-model/src/commonMain/kotlin/LicenseFindingDetail.kt
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.components.licensefindings
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class LicenseFindingDetail(
```
mnonnenmacher: Why not just `LicenseFinding`?

components/license-findings/api-model/src/commonMain/kotlin/DetectedLicensePackage.kt
```diff
@@ -0,0 +1,28 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.eclipse.apoapsis.ortserver.components.licensefindings
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class DetectedLicensePackage(
```
mnonnenmacher: Why does this class have `DetectedLicense` in the name, why not just call it `PackageIdentifier(s)` or so?.

components/search/backend/src/test/kotlin/routes/GetRunsWithPackageIntegrationTest.kt
```diff
@@ -41,14 +41,14 @@ import org.eclipse.apoapsis.ortserver.model.OrganizationId
 import org.eclipse.apoapsis.ortserver.model.ProductId
 import org.eclipse.apoapsis.ortserver.model.RepositoryId
 import org.eclipse.apoapsis.ortserver.model.runs.Identifier
+import org.eclipse.apoapsis.ortserver.model.runs.toCoordinates
```
mnonnenmacher: > - make `toSortOrder()` in DAO public to be able to reuse it
> - move `toCoordinates()` to `Identifier` class in model for reuse, and 
>  update its call sites in search component

Separate refactorings should go to separate commits.

model/src/commonMain/kotlin/runs/Identifier.kt
```diff
@@ -28,3 +28,8 @@ data class Identifier(
     val name: String,
     val version: String
 )
+
+/**
+ * Return this identifier in ORT coordinates format.
+ */
+fun Identifier.toCoordinates(): String = "$type:$namespace:$name:$version"
```
mnonnenmacher: This does not need to be an extension function anymore, then it does not have to be imported.

components/license-findings/backend/src/main/kotlin/LicenseFindingService.kt
```diff
@@ -0,0 +1,358 @@
...
+                (ScanResultsTable.vcsType eq vcsAlias[VcsInfoTable.type]) and
+                (ScanResultsTable.vcsUrl eq vcsAlias[VcsInfoTable.url]) and
+                (ScanResultsTable.vcsRevision eq vcsAlias[VcsInfoTable.revision]) and
+                (vcsAlias[VcsInfoTable.path] eq stringLiteral(""))
+            )
+}
+
+// includeAnalyzerData controls whether the 4 LEFT JOINs for purl lookup are included. Set to false
+// when purl is not needed, as packages_analyzer_runs is a junction table that multiplies rows by the
+// number of analyzer runs a package has been part of, forcing expensive COUNT(DISTINCT) to compensate.
```
mnonnenmacher: This should be proper KDoc for the function, not a code comment.

components/license-findings/backend/src/routes/kotlin/Validation.kt
```diff
@@ -0,0 +1,50 @@
...
+
+fun requireRunReadPermission(ortRunRepository: OrtRunRepository): AuthorizationChecker =
+    object : AuthorizationChecker {
+        override suspend fun loadEffectiveRole(
+            service: AuthorizationService,
+            userId: String,
+            call: ApplicationCall
+        ): EffectiveRole? {
+            val runId = call.requireIdParameter("runId")
+            val ortRun = ortRunRepository.get(runId) ?: throw InvalidHierarchyIdException(RepositoryId(runId))
```
mnonnenmacher: `runId` is not a `RepositoryId`, use an `IllegalArgumentException` instead.

components/license-findings/backend/src/test/kotlin/routes/GetRunDetectedLicensesIntegrationTest.kt
```diff
@@ -0,0 +1,95 @@
...
+                val response = client.get("/runs/$ortRunId/detected-licenses?license=apache")
+
+                response.status shouldBe HttpStatusCode.OK
+                val body = response.body<PagedResponse<DetectedLicense>>()
+                body.pagination.totalCount shouldBe 1
+                body.data shouldContainExactly listOf(DetectedLicense("Apache-2.0", 2))
+            }
+        }
+
+        "return 200 with empty data for an unknown run" {
```
mnonnenmacher: For an unknown run it should return 404.

core/src/main/kotlin/plugins/Routing.kt
```diff
@@ -52,10 +55,14 @@ fun Application.configureRouting() {
...
             downloads()
             authenticate(AuthenticationProviders.TOKEN_PROVIDER) {
+                val licenseFindingService = get<LicenseFindingService>()
+                val ortRunRepository = get<OrtRunRepository>()
+
                 admin()
                 adminConfigRoutes(get())
                 authorizationRoutes()
                 infrastructureServicesRoutes(get())
+                licenseFindingRoutes(licenseFindingService, ortRunRepository)
```
mnonnenmacher: Just use `get()` here for the parameters like for the other functions.

mnonnenmacher (changes requested):

components/license-findings/backend/src/routes/kotlin/Validation.kt
Etsija: Effectively, I moved the checking and 404 reporting to API layer (your next comment) to align with other components, so I believe this now doesn't need to return an exception anymore.

mnonnenmacher (approved):

PR #4855, opened 2026-04-15
feat(workers): Generate package manager config files only in analyzer

mnonnenmacher:
If any infrastructure services are defined, all implementations of `EnvironmentConfigGenerator` were invoked by the analyzer, scanner, and reporter workers. As a result, package manager specific files like `.npmrc` were generated in all of those workers, even though they are only required by the analyzer.

To avoid generating unneeded files, make the generation of package manager specific files optional and enable it only in the analyzer worker.

### DIFF
--- a/workers/common/src/main/kotlin/env/EnvironmentModule.kt
+++ b/workers/common/src/main/kotlin/env/EnvironmentModule.kt
@@ -34,8 +34,11 @@ import org.koin.dsl.module
 /**
  * Return a [Module] with bean definitions that provide an [EnvironmentService] instance and its dependencies. This
  * module can be used by worker implementations that need to set up a build environment.
+ *
+ * If [includePackageManagerGenerators] is `true`, generators for package manager-specific configuration files are
+ * included.
  */
-fun buildEnvironmentModule(): Module = module {
+fun buildEnvironmentModule(includePackageManagerGenerators: Boolean = false): Module = module {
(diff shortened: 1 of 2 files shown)


workers/common/src/main/kotlin/env/EnvironmentModule.kt
```diff
@@ -34,8 +34,11 @@ import org.koin.dsl.module
 /**
  * Return a [Module] with bean definitions that provide an [EnvironmentService] instance and its dependencies. This
  * module can be used by worker implementations that need to set up a build environment.
+ *
+ * If [includePackageManagerGenerators] is `true`, generators for package manager-specific configuration files are
+ * included.
  */
-fun buildEnvironmentModule(): Module = module {
+fun buildEnvironmentModule(includePackageManagerGenerators: Boolean = false): Module = module {
```
sschuberth, line 41: Should there maybe be no default value? This seems to be called only once, so having a default value does not save code, and making it explicit could be more readable.

sschuberth (approved):
Approving as my comment is optional.

workers/common/src/main/kotlin/env/EnvironmentModule.kt
sschuberth: > This seems to be called only once

Scratch that, it's called more than once without arguments.

mnonnenmacher:
Updated because I noticed a typo in the commit message title.

PR #3251, opened 2025-07-29
Add support for advisor and reporter plugin options in the UI

mnonnenmacher:
Add support for configuration plugin options for advisor and reporter plugins in the UI.
The forms are generated from the plugin descriptors and take default values and plugin config templates into account.

Example for advisor plugins:

image

Example for reporter plugins:

image

This is a first version that will help to identify required improvements and missing functionality.

@lamppu @Etsija This is still a draft but a review would already be welcome, just ignore the first few Kotlin commits which I will also commit separately.

PR #4858, opened 2026-04-15
Add support for configuring advisor plugin options in the UI

mnonnenmacher:
Add support for configuring plugin options for advisor plugins in the UI.
The forms are generated from the plugin descriptors and take default values and plugin config templates into account.

This is based on #3251, support for other plugin types will be added in follow-up commits.

Example for advisor plugins:

image

This is a first version that will help to identify required improvements and missing functionality.

### DIFF
--- a/ui/src/components/form/plugin-multi-select-field.tsx
+++ b/ui/src/components/form/plugin-multi-select-field.tsx
@@ -167,6 +167,7 @@ export const PluginMultiSelectField = <
                               <Checkbox
                                 checked={field.value as CheckedState}
                                 onCheckedChange={field.onChange}
+                                disabled={option.isFixed}
                               />
                             ) : option.isRequired ? (
                               <Input
@@ -178,6 +179,7 @@ export const PluginMultiSelectField = <
                                     : 'text'
                                 }
                                 value={field.value}
+                                disabled={option.isFixed}
                               />
                             ) : (
                               <OptionalInput
@@ -189,9 +191,16 @@ export const PluginMultiSelectField = <
                                     : 'text'
                                 }
                                 value={field.value}
+                                disabled={option.isFixed}
                               />
                             )}
                           </FormControl>
+                          {option.isFixed && (
+                            <FormDescription className='text-muted-foreground font-semibold text-yellow-700'>
+                              This option is set by an administrator and cannot
+                              be changed.
+                            </FormDescription>
+                          )}
                         </FormItem>
                       )}
                     />
--- a/ui/src/components/form/plugin-multi-select-field.tsx
+++ b/ui/src/components/form/plugin-multi-select-field.tsx
@@ -1,0 +1,207 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import { CheckedState } from '@radix-ui/react-checkbox';
+import React from 'react';
+import {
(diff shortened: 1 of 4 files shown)


sschuberth:
A nit, from just looking at the screenshot: Could we apply some "code"-style formatting to the option names, similar to like backticks in Markdown would do? Currently, then seems to be a bit "pale" next to the colored type badges.

ui/src/components/form/plugin-multi-select-field.tsx
```diff
@@ -0,0 +1,207 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
```
sschuberth, line 2: Is 2025 correct here?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
```diff
@@ -62,7 +65,62 @@ const environmentVariableSchema = z.object({
...
 
-export const createRunFormSchema = () => {
+function optionTypeToZodType(type: PluginOptionType): ZodType {
+  switch (type) {
+    case 'BOOLEAN':
+      return z.boolean();
+    case 'INTEGER':
+      return z.coerce.string();
+    case 'LONG':
+      return z.coerce.string();
```
sschuberth, line 75: Just checking: No integral types here? If that's correct, does this maybe deserve a code comment?

ui/src/components/form/plugin-multi-select-field.tsx
```diff
@@ -27,7 +27,7 @@ import {
   UseFormReturn,
 } from 'react-hook-form';
 
-import { PreconfiguredPluginDescriptor } from '@/api';
+import { PreconfiguredPluginDescriptor, Secret } from '@/api';
```
sschuberth, line 30: Commit message:

> secrets available in the context of the repository.

This sounds a bit as if secrets inherited from the organization level would now be shown, but really only those defined on exactly the repository level. Maybe just say "... available in the the current context."?

Etsija (changes requested):
`create-run-utils.ts` is now already over 800 code lines, and very hard to read. Please consider refactoring it, as it would easily grow to over 1000 lines when supporting other plugins.

Ideally, create a directory `.../_repo-layout/create-run/-components` and do the refactoring there. The `-components` directory excludes the directory from TanStack Router's route context and is the usual way of co-locating helper files and directories along the UI routes.

Docs: https://tanstack.com/router/latest/docs/api/file-based-routing#routefileignoreprefix

ui/src/components/form/plugin-multi-select-field.tsx
mnonnenmacher: Yes, the commit is old.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
mnonnenmacher: What do you mean with "integral types"?

ui/src/components/form/plugin-multi-select-field.tsx
mnonnenmacher: I have tried to clarify this now in the message.

mnonnenmacher:
> A nit, from just looking at the screenshot: Could we apply some "code"-style formatting to the option names, similar to like backticks in Markdown would do? Currently, then seems to be a bit "pale" next to the colored type badges.

I want to make a follow-up PR with several UI improvements, I will try that out there.

mnonnenmacher:
> `create-run-utils.ts` is now already over 800 code lines, and very hard to read. Please consider refactoring it, as it would easily grow to over 1000 lines when supporting other plugins.
> 
> Ideally, create a directory `.../_repo-layout/create-run/-components` and do the refactoring there. The `-components` directory excludes the directory from TanStack Router's route context and is the usual way of co-locating helper files and directories along the UI routes.
> 
> Docs: https://tanstack.com/router/latest/docs/api/file-based-routing#routefileignoreprefix

Before doing that I would like to make a follow-up PR with two more commits from #3251 that generalize some functions to make them usable for all plugin types, otherwise there would be lots of conflicts.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-create-run-utils.ts
sschuberth: I mean that I (naively) would have expected INTEGER to be mapped to Integer, and LONG to Long, not both to String.

Etsija (approved):

PR #4864, opened 2026-04-16
Filter packages table by dependency type

### LINKED ISSUE #4761: The "Packages" table should have a filter for direct and / or transitive dependencies
Esp. when trying to estimate which dependencies can be easily upgraded / swapped out in the context of vulnerability management / resolution, it would be great to be able to filter the table by direct and / or transitive dependencies. The filter should look similar to the Status" filter for (un-)resolved vulnerabilities in the "Vulnerabilities" table.

I was also considering to get more generic and filter by dependency depth, but I do not think that necessary in the vast majority of cases, and probably is to technical. Other products also seem to only distinguish between direct / transitive, see e.g. [this](https://central.sonatype.com/artifact/org.scala-sbt/launcher/dependencies).

Etsija:
Screenshot from 2026-04-17 06-10-02

Screenshot from 2026-04-17 06-11-06

The main rule: if a package is listed as a direct dependency of the project _for at least one of the scopes_, then it is considered a direct dependency of the project.

The author has verified from the UI that this logic is applied correctly (for some multi-scope projects).

### DIFF
--- a/core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
+++ b/core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
@@ -28,6 +28,7 @@ import io.kotest.engine.spec.tempdir
 import io.kotest.inspectors.forAll
 import io.kotest.matchers.collections.beEmpty
 import io.kotest.matchers.collections.shouldBeSingleton
+import io.kotest.matchers.collections.shouldContainExactly
 import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
 import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.file.aFile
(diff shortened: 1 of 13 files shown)


sschuberth:
Isn't it a bit of an unusual UX to allow none to be selected of a (mutually exclusive) radio button group? So instead of the "Clear selection" menu entry, I'd probably follow the UX of the example linked in the issue, and offer a third "All" radio button (however, I'd put that first, and select it by default).

Etsija:
> Isn't it a bit of an unusual UX to allow none to be selected of a (mutually exclusive) radio button group? So instead of the "Clear selection" menu entry, I'd probably follow the UX of the example linked in the issue, and offer a third "All" radio button (however, I'd put that first, and select it by default).

Got too much hung up with the idea of making the single-select filter look as much as possible like the multi-select one. But yeah, the typical UX flow of a radio button group is that one is always selected, just like those old radios, if someone else still remembers such old things :rofl: 

Pictures updated in the description.

sschuberth:
> Pictures updated in the description.

Thanks. I'm not too sure about that separator line though. It makes "All" stand out so much that it does not seem to be part of the "mutually exclusive" group. But this is one of the things we could easily tweak any time later.

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -118,6 +121,28 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
                 }
             }
 
+            filters.isDirectDependency?.let { filter ->
+                /**
+                 * The subquery keeps the existing query shape unchanged and applies the new filter as an
+                 * existence check, rather than joining a one-to-many-per-package table into the main query,
+                 * because extending the main join tree would duplicate package rows and make counting,
+                 * pagination, and sorting more fragile.
+                 */
```
sschuberth: For inline code comments (i.e. not class or function documentation), do not use KDoc syntax (starting with `/**`) but precede each line with `//`.

core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
```diff
@@ -1353,6 +1353,116 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
...
+                        )
+                    )
+                )
+
+                val response = superuserClient.get("/api/v1/runs/${ortRun.id}/packages?isDirectDependency=false")
+
+                response shouldHaveStatus HttpStatusCode.OK
+                val packages = response.body<PagedSearchResponse<ApiPackage, PackageFilters>>()
+
+                packages.data.map { it.identifier } shouldBe listOf(identifier2.mapToApi())
```
sschuberth: I guess the `shouldContainExactly` matcher would be more suitable.

core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
```diff
@@ -1353,6 +1353,116 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
...
+                        )
+                    )
+                )
+
+                val response = superuserClient.get("/api/v1/runs/${ortRun.id}/packages?isDirectDependency=true")
+
+                response shouldHaveStatus HttpStatusCode.OK
+                val packages = response.body<PagedSearchResponse<ApiPackage, PackageFilters>>()
+
+                packages.data.map { it.identifier } shouldBe listOf(identifier1.mapToApi())
```
sschuberth: I guess the `shouldContainExactly` matcher would be more suitable.

Etsija:
> Thanks. I'm not too sure about that separator line though. It makes "All" stand out so much that it does not seem to be part of the "mutually exclusive" group. But this is one of the things we could easily tweak any time later.

I kind of like the separator there.

mnonnenmacher (approved):

PR #4869, opened 2026-04-17
UI: Prepare supporting more plugin types and fix an issue with required fields

### LINKED ISSUE #4868: Consider disabling advisors with mandatory API keys in Compose by default
I haven't selected BlackDuck or OSSIndex Advisor for the run.

Image

If this is a user error, then some instructions to make it work in Docker Compose would be welcome.

mnonnenmacher:


Etsija:
Indeed, TypeScript type validation can be extremely frustrating at times.

Etsija (approved):

PR #4872, opened 2026-04-17
Store the ORT server version for each run

mnonnenmacher:

The version is shown in the run details bar:

image


sschuberth:
I totally agree with the first commit, but is this information really important enough to the majority of user to justify it being shown prominently in the overview bar? Or would it be sufficient to be shown in the configuration menu of a run?

Etsija:
To me, it's most important as a basis for informing users of a feature which might skew old run results (in context of #4870, for example) - but we need to probably think about how to implement that in the UI in a clever way.

mnonnenmacher:
I have just dropped the second commit for now, I can't spend more time on this today.

sschuberth (approved):

PR #4875, opened 2026-04-17
fix(orchestrator): Keep recipient addresses after job completion

### LINKED ISSUE #4865: Notifier job configuration is ignored when clicking "rerun"
I trigger a run with a Notifier configuration:

Image

The run is successful, I receive a mail.

If I click "rerun", the notifier configuration is empty:

Image

When rerunning, the step configuration should be preserved.

Joelp03:
Previously, the `cleanupJobConfigs` function intentionally cleared the recipient addresses from the notifier configuration when a run completed. This caused the recipient addresses to be lost when a user tried to rerun a previous run, as the UI would fetch the cleaned configuration from the database.

This change removes the line that clears `recipientAddresses`, allowing them
to be preserved for rerun scenarios.

Resolves #4865.

### DIFF
--- a/orchestrator/src/main/kotlin/Orchestrator.kt
+++ b/orchestrator/src/main/kotlin/Orchestrator.kt
@@ -24,7 +24,6 @@ import java.sql.Connection
 import kotlin.time.Clock
 
 import org.eclipse.apoapsis.ortserver.dao.blockingQueryCatching
-import org.eclipse.apoapsis.ortserver.model.JobConfigurations
 import org.eclipse.apoapsis.ortserver.model.JobStatus
 import org.eclipse.apoapsis.ortserver.model.OrtRun
 import org.eclipse.apoapsis.ortserver.model.OrtRunStatus
@@ -53,7 +52,6 @@ import org.eclipse.apoapsis.ortserver.model.orchestrator.WorkerMessage
 import org.eclipse.apoapsis.ortserver.model.repositories.OrtRunRepository
 import org.eclipse.apoapsis.ortserver.model.repositories.RepositoryRepository
 import org.eclipse.apoapsis.ortserver.model.runs.Issue
-import org.eclipse.apoapsis.ortserver.model.util.OptionalValue
 import org.eclipse.apoapsis.ortserver.model.util.asPresent
 import org.eclipse.apoapsis.ortserver.transport.AdvisorEndpoint
 import org.eclipse.apoapsis.ortserver.transport.AnalyzerEndpoint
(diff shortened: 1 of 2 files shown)


sschuberth:
> intentionally cleared the recipient addresses

I wonder whether this was a requirement from @eclipse-apoapsis/bosch to not store email addresses, because they could been regarded as personal data? But on the other hand, email addresses are in Keycloak already anyway.

Joelp03:
I wasn't sure if this issue was exclusive to the frontend. I will investigate a solution that focuses solely on the `UI` layer.

oheger-bosch:
Yes, the functionality to clear email addresses from the configuration has been added originally to avoid dealing with personal data. I think we may be able to find some argumentation in the line that this storage is part of the regular processing, as re-triggering a run with the same notifications should be a normal use case.

However, if I understand the legal requirements correctly, every user can at any time demand that their data is removed from the system. This would then probably require a mechanism that scans the configurations of all existing runs and deletes just specific mail addresses. Would you agree to this interpretation?

In any case, if we follow this path, then the whole `cleanupJobConfigs` function is obsolete and should probably be dropped.

sschuberth:
> I wasn't sure if this issue was exclusive to the frontend. I will investigate a solution that focuses solely on the `UI` layer.

Let's first wait for an answer from the people of Bosch. I've put this one the [agenda](https://github.com/eclipse-apoapsis/ort-server/discussions/3987#discussioncomment-16638247) of today's ORT Server community meeting (although I'll not be able to participate).

mnonnenmacher:
> However, if I understand the legal requirements correctly, every user can at any time demand that their data is removed from the system. This would then probably require a mechanism that scans the configurations of all existing runs and deletes just specific mail addresses. Would you agree to this interpretation?

To me it makes sense, it would certainly simplify things if we would not have to clean up the job configs. If it is unlikely that someone requests deletion of their data this could also be kept a manual step for now.

sschuberth:
> In any case, if we follow this path, then the whole `cleanupJobConfigs` function is obsolete and should probably be dropped.

Sounds good to me!

oheger-bosch:
For some use cases, we would also like to keep the recipient email addresses, so from Bosch side, we are in principle fine with the change. However, we have sent a query to our data security officer to clarify whether there are some aspects that require special care. I will comment again when we have the response.

sschuberth:
> I will comment again when we have the response.

Does this mean that we have to wait with the change until the response, or that we could merge now and eventually improve later?

oheger-bosch:
We got the response (a typical "Yes, but..."). So, we should be fine to go ahead with this approach. But the recommendation is to apply some encryption for the personal data at rest in the future.

sschuberth:
Great, thanks. So @Joelp03, would mind amending your commit to remove the function completely, providing the rational from the above discussion?

Joelp03:
Just to confirm. Is the idea to remove the `cleanupJobConfigs` function and create a new one to encrypt the email? 
Or will the email encryption be part of future changes?

mnonnenmacher:
> Just to confirm. Is the idea to remove the `cleanupJobConfigs` function and create a new one to encrypt the email? Or will the email encryption be part of future changes?

I understand @oheger-bosch's response so that we can remove the function now and add encryption later.

orchestrator/src/main/kotlin/Orchestrator.kt
```diff
@@ -449,24 +448,13 @@ class Orchestrator(
...
         ortRunRepository.get(ortRunId)?.let { ortRun ->
-            ortRunRepository.update(
-                id = ortRunId,
-                jobConfigs = OptionalValue.Present(cleanupJobConfigs(ortRun.jobConfigs)),
-                resolvedJobConfigs = ortRun.resolvedJobConfigs?.let {
-                    OptionalValue.Present(cleanupJobConfigs(it))
-                } ?: OptionalValue.Absent
-            )
+            ortRunRepository.update(id = ortRunId, jobConfigs = OptionalValue.Present(ortRun.jobConfigs))
         }
```
sschuberth: This seems to be a no-op now, so I believe the whole block should be removed.

Joelp03:
I have removed the `ortRunRepository.update` method and resolved the test issues.

sschuberth (approved):

Issue #4317, opened 2026-01-20
Apply plugin templates in config worker


mnonnenmacher:
Apply the configured plugin templates in the config worker before invoking the parameter validation script. This ensures that plugin templates are taken into account for all runs and still allows the validation script to override them which is useful for testing.

PR #4893, opened 2026-04-20
ui: Add support for scanner plugin options

mnonnenmacher:

image

### DIFF
--- a/components/plugin-manager/backend/src/main/kotlin/PluginTemplateService.kt
+++ b/components/plugin-manager/backend/src/main/kotlin/PluginTemplateService.kt
@@ -36,6 +36,7 @@ import org.eclipse.apoapsis.ortserver.dao.blockingQuery
 import org.eclipse.apoapsis.ortserver.model.OrganizationId
 import org.eclipse.apoapsis.ortserver.model.RepositoryId
 import org.eclipse.apoapsis.ortserver.model.ResolvablePluginConfig
+import org.eclipse.apoapsis.ortserver.model.ResolvableSecret
 import org.eclipse.apoapsis.ortserver.model.repositories.OrganizationRepository
 import org.eclipse.apoapsis.ortserver.model.repositories.RepositoryRepository
 
@@ -392,7 +393,7 @@ class PluginTemplateService(
                 }
 
                 // Check if option is set which is fixed by a template.
-                (config.options.keys + config.secrets.keys).forEach { option ->
+                (config.options + config.secrets).forEach { (option, value) ->
                     val pluginOption = descriptor.options.find { it.name == option }
 
                     if (pluginOption == null) {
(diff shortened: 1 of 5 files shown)


Etsija (changes requested):
When opening up the scanner tab for DOS (this probably applies to other scanners as well), this picture shows scanner scope "Both" is chosen:

Image

However, if the user doesn't touch the button, for example by clicking "Packages only", then back to "Both", the option isn't actually set to the form payload:

Image

mnonnenmacher:
> However, if the user doesn't touch the button, for example by clicking "Packages only", then back to "Both", the option isn't actually set to the form payload:

I have tested several scenarios now and it should be fixed. ~There is one issue left which is that if a user selects only a project scanner the backend ignores that and falls back to the defaults. I will create a separate issue for that.~ (this was also a frontend issue)

Etsija:
Still some weird errors from back-end when testing this on UI - they might be "user errors", however.

What I did:
- I have created a global template for DOS scanner plugin, where I set `url`, `token`, `frontendUrl`, and `writeToStorage` and declare them as `Final` so they cannot be changed for runs
- I try to create a new run, and the UI correctly tells me what options are locked by an admin and cannot be changed in the form (only partial screenshot shown so as not to expose the endpoint addresses publicly):

Screenshot from 2026-04-22 08-42-29

- but when I try to create a run, I get this error message:

Screenshot from 2026-04-22 08-41-36

But I did not try to change those options at all, just trying to trigger a run with the fixed options there! Is this a bug, or do I completely misunderstand the meaning of `Final` in a plugin template?

When I configure the scanners this way, with global templates which have `Final` options set, the options are still sent in the request payload. Maybe there's something in the back-end erroring then, and the admin-locked options should _not_ be included in the payload?

mnonnenmacher:
> But I did not try to change those options at all, just trying to trigger a run with the fixed options there! Is this a bug, or do I completely misunderstand the meaning of `Final` in a plugin template?
> 
> When I configure the scanners this way, with global templates which have `Final` options set, the options are still sent in the request payload. Maybe there's something in the back-end erroring then, and the admin-locked options should _not_ be included in the payload?

I have added a commit to not fail the validation if a plugin option is set to the final value. I think I would prefer to not even include those options in the payload, but that would require #4317.

Etsija:
> I have added a commit to not fail the validation if a plugin option is set to the final value. I think I would prefer to not even include those options in the payload, but that would require #4317.

Sorry, now not failing for the three other options I have locked in plugin management, but still failing on `token`:

Screenshot from 2026-04-22 13-11-56

Same happens for both DOS and SCANOSS scanner tokens.

mnonnenmacher:
> Same happens for both DOS and SCANOSS scanner tokens.

Fixed this now for secrets as well.

Etsija (approved):

PR #4901, opened 2026-04-21
Add logaccess implementation for Elasticsearch

lamppu:
Please see the commit message and the module's README.md for more information on implementation details such as `search_after`, the need for a `sortId` and how the fields should be indexed.

The implementation has been tested in the local docker compose environment. Both authentication methods have been tested, and also downloading the logs for a longer run has been tested to ensure correct behavior for `search_after`.


logaccess/elasticsearch/src/main/kotlin/ElasticsearchLogFileProvider.kt
```diff
@@ -0,0 +1,248 @@
...
+    ): JsonObject =
+        buildJsonObject {
+            put("size", config.pageSize)
+            putJsonObject("query") {
+                putJsonObject("bool") {
+                    putJsonArray("filter") {
+                        add(termQuery("namespace", JsonPrimitive(config.namespace)))
+                        add(termQuery("component", JsonPrimitive(source.component)))
+                        add(termQuery("ortRunId", JsonPrimitive(ortRunId.toString())))
+                        add(termsQuery("level", levels.map { it.name }))
```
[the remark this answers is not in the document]
lamppu: The levels set is formed in `LogFileService` using the `levelOrHigher` function from [LogLevel](https://github.com/eclipse-apoapsis/ort-server/blob/40bcb2e5cff7063ae33bae3d755c6669e69d91da/model/src/commonMain/kotlin/LogLevel.kt#L49). I would argue that that implementation only changes very rarely if at all, and to my understanding the set it forms is deterministic, so I feel this change would be unnecessary, (unless other reviewers think otherwise).

logaccess/elasticsearch/README.md
```diff
@@ -0,0 +1,144 @@
...
+| Field | Expected Elasticsearch type | Purpose |
+|-------|-----------------------------|---------|
+| `namespace` | `keyword` | Exact-match deployment namespace filter; must match `elasticsearchNamespace`. |
+| `component` | `keyword` | Exact-match ORT component filter using ORT Server component names. |
+| `ortRunId` | `keyword` with optional `long` subfield, for example `ortRunId.numeric` | Exact-match ORT run ID filter. |
+| `level` | `keyword` | Exact-match log level filter using ORT Server log level names. |
+| `time` | `date` | ORT log event timestamp used for range queries and primary sorting. |
+| `sortId` | `keyword` | Secondary sort key for stable `search_after` pagination. |
+| `message` | present in `_source`; `text` recommended for Kibana searches | Rendered log line written to the downloaded log file. |
+
```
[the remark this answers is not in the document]
lamppu: There's just one `|`, not making any changes.

logaccess/elasticsearch/README.md
```diff
@@ -0,0 +1,144 @@
...
+| Property | Variable | Description | Default | Secret |
+|----------|----------|-------------|---------|--------|
+| `elasticsearchServerUrl` | `ELASTICSEARCH_SERVER_URL` | Base URL of the Elasticsearch instance. | mandatory | no |
+| `elasticsearchIndex` | `ELASTICSEARCH_INDEX` | Index or index pattern to query. | mandatory | no |
+| `elasticsearchNamespace` | `ELASTICSEARCH_NAMESPACE` | Namespace label used to restrict queries. | mandatory | no |
+| `elasticsearchPageSize` | `ELASTICSEARCH_PAGE_SIZE` | Number of hits to fetch per search request. | `1000` | no |
+| `elasticsearchUsername` | `ELASTICSEARCH_USERNAME` | Optional username for Basic Auth. Ignored when an API key is configured. | undefined | no |
+| `elasticsearchPassword` | `ELASTICSEARCH_PASSWORD` | Optional password for Basic Auth. | undefined | yes |
+| `elasticsearchApiKey` | `ELASTICSEARCH_API_KEY` | Optional Elasticsearch API key. Takes precedence over Basic Auth. | undefined | yes |
+| `elasticsearchTimeoutSec` | `ELASTICSEARCH_TIMEOUT_SEC` | Optional request timeout in seconds. | `30` | no |
```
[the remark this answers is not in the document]
lamppu: There's just one `|`, not making any changes.

logaccess/elasticsearch/src/main/kotlin/ElasticsearchConfig.kt
```diff
@@ -0,0 +1,121 @@
...
+
+        /** The default number of hits requested per page when no explicit page size is configured. */
+        private const val DEFAULT_PAGE_SIZE = 1000
+
+        /**
+         * Return a new instance of [ElasticsearchConfig] initialized from the passed in [configManager].
+         */
+        fun create(configManager: ConfigManager): ElasticsearchConfig {
+            val username = configManager.getStringOrNull(USERNAME_PROPERTY)?.ifBlank { null }
+            val apiKey = configManager.getStringOrNull(API_KEY_PROPERTY)?.ifBlank { null }?.let {
```
mnonnenmacher: For the API key only `getSecret` should be used, like for the password. But this function throws an exception and there is no `getSecretOrNull`, so calling it unconditionally without catching the exception will prevent that the optional password can be read below.

logaccess/elasticsearch/src/main/kotlin/ElasticsearchLogFileProvider.kt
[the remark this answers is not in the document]
mnonnenmacher: I also wonder if it matters for the response how the levels are sorted here. If not, I would not care about sorting.

logaccess/elasticsearch/src/main/kotlin/ElasticsearchLogFileProvider.kt
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * Create an [HttpClient] with a configuration to communicate with Elasticsearch.
+     */
+    private fun createClient(): HttpClient =
+        HttpClient(OkHttp) {
+            defaultRequest {
+                if (config.apiKey != null) {
+                    header(HttpHeaders.Authorization, "ApiKey ${config.apiKey}")
+                }
+            }
```
mnonnenmacher: The `if` could be moved out of the `defaultRequest` block, then it could be combined with setting up basic auth to make the either/or more visible:

```kotlin
if (config.apiKey != null) {
  defaultRequest { ... }
} else if (config.username != null && config.password != null) {
  install(Auth) { ... }
}
```

logaccess/elasticsearch/src/main/kotlin/ElasticsearchLogFileProvider.kt
```diff
@@ -0,0 +1,251 @@
...
+                        credentials {
+                            BasicAuthCredentials(config.username, config.password)
+                        }
+                        sendWithoutRequest { true }
+                    }
+                }
+            }
+
+            install(HttpTimeout) {
+                val httpClientTimeoutMillis = (config.timeoutSec ?: HTTP_CLIENT_DEFAULT_TIMEOUT_SEC) * 1000L
```
mnonnenmacher: Consider already applying the default value when creating the config object, then this code would not have to know about it.

mnonnenmacher (changes requested):

logaccess/elasticsearch/src/main/kotlin/ElasticsearchLogFileProvider.kt
lamppu: Should I make a similar change for [Loki](https://github.com/eclipse-apoapsis/ort-server/blob/main/logaccess/loki/src/main/kotlin/LokiLogFileProvider.kt#L203), and do you think the default value should be provider specific or a shared default?

mnonnenmacher: Feel free to make the same change for Loki if you want, to me it makes more sense to handle default values on that level than when consuming the config. But the value should be provider specific IMO.

logaccess/elasticsearch/src/main/kotlin/ElasticsearchLogFileProvider.kt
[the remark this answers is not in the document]
lamppu: It seems it might have an effect on whether a cache is hit or not, but shouldn't have an effect on the actual response.

logaccess/elasticsearch/src/main/kotlin/ElasticsearchLogFileProvider.kt
lamppu: Yes it does look better to me too.

mnonnenmacher (approved):

PR #4905, opened 2026-04-22
build(gradle): Disable Detekt's HTML and Markdown reports

sschuberth:
Detekt's tasks with type-resolution seem to depend on the report output of the tasks without type-resolution. As some of the reports contain timestamps, their content changes even if there are no new findings. Avoid this case of unnecessary Detekt task reruns by disabling these unneeded reports.


Etsija:
Would this PR be OK to merge @eclipse-apoapsis/backend-devs? It might make running `detekt` tasks on local dev. environment faster, but are any teams actually using those reports, or are they not needed?

sschuberth:
> It might make running `detekt` tasks on local dev. environment faster,

This is only a step into the right direction. I don't expect you seeing any performance benefits just yet.

> or are they not needed?

They are not used by anyone to the best of my knowledge.

Etsija (approved):

PR #4907, opened 2026-04-22
feat(secrets): Add configurable HTTP timeout for Vault client

Juli0q:
Add a `vaultHttpTimeoutSec` config property to set the HTTP timeout for requests to the Vault service, defaulting to 10 seconds.

### DIFF
--- a/secrets/vault/src/main/kotlin/VaultConfiguration.kt
+++ b/secrets/vault/src/main/kotlin/VaultConfiguration.kt
@@ -21,9 +21,13 @@ package org.eclipse.apoapsis.ortserver.secrets.vault
 
 import com.typesafe.config.Config
 
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
 import org.eclipse.apoapsis.ortserver.config.ConfigManager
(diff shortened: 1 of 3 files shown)


oheger-bosch (approved):
Maybe as a follow-up, you can also update the documentation to mention the new configuration property.

secrets/vault/src/main/kotlin/VaultConfiguration.kt
```diff
@@ -75,9 +82,15 @@ data class VaultConfiguration(
         /** Name of the configuration property defining the namespace to be passed to the vault service. */
         private const val NAMESPACE_PROPERTY = "vaultNamespace"
 
+        /** Name of the configuration property defining the timeout for HTTP requests in seconds. */
+        private const val TIMEOUT_PROPERTY = "vaultHttpTimeoutSec"
```
oheger-bosch, line 86: You also have to update the `application.conf` to reference the new property.

oheger-bosch (changes requested):

nnobelis (dismissed):

oheger-bosch (approved):

PR #4913, opened 2026-04-23
Dependency graphs for an ORT run

### LINKED ISSUE #944: Add an endpoint to the get the dependency graphs for projects
In the front-end, we'd like to show the dependency graphs immediately after an analyzer run, e.g. as per-project expandable / collapsible trees.

We need to think of a good representation of the data. Maybe multiple calls to different endpoints would be needed to assemble the necessary data.

### LINKED ISSUE #2242: Show the dependency graphs for projects
The distinction between "Dependencies" and "Packages" on the menu-level seems to be too artificial. As "Dependencies" is still empty anyway, the idea is to remove it completely. Afterwards, "Packages" should be renamed to "Dependencies" (also taking the latter's icon), as "Dependencies" seems to be the more common term for users.

Later, once https://github.com/eclipse-apoapsis/ort-server/issues/944 is implemented, the "Dependencies" page should get a graph view in addition to the table view, probably on a separate tab (similar to like in the WebApp report).

Etsija:
Screenshot from 2026-04-27 13-58-29

First implementation somewhat mimics the WebApp report's Tree view - that's not to say there aren't points of improvement from it, so this was just the first goal in this PR.

Graph sorting is not as straightforward as for other, plainly list-formatted data, so sorting is implemented in the last two commits for easier review. The changes in the commits can be absorbed to the earlier ones if needed.

#### Possible points of improvement in follow-ups
1. The graphs could be made queryable on an ecosystem basis, which would mean that for example Gradle trees would only be queried, when the user chooses the corresponding tab from the UI. The UI implementation has however been tested locally for a repository with over 1600 packages and it has shown no signs of performance problems yet, so this optimization will be done as a follow-up if needed.
2. The graphs could be filtered with `scopes`.


sschuberth:
Just some quick remarks regarding the visuals / naming: I think "Graphs" is fine, but should we be more verbose and say "Dependency Graphs"? Or does that take up too much horizontal space in the menu then?

In any case, I believe we have somewhat of a convention to name the title of the right-hand section after the menu item. So that should also say "(Dependency) Graphs" instead of "Dependencies".

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/dependencies/index.tsx
```diff
@@ -17,27 +17,446 @@
...
+  const defaultManager = managerEntries[0]?.[0];
+
   return (
     <Card className='h-fit'>
       <CardHeader>
         <CardTitle>Dependencies</CardTitle>
-        <CardDescription>Dependency graphs of the project.</CardDescription>
+        <CardDescription>
+          Dependency graphs of the ORT run for each ecosystem. The number in
+          each tab is the number of unique packages across the ecosystem graph.
```
sschuberth: At least in output visible to the regular user, I'd like to avoid the mention of "ORT" generally. That's more or less an implementation detail that's just "noise" to the regular user in description like this.

So maybe just "The dependency graphs per ecosystem as discovered during the run. The number of unique packages per ecosystem is shown in each tab.".

sschuberth:
> I think "Graphs" is fine, but should we be more verbose and say "Dependency Graphs"? Or does that take up too much horizontal space in the menu then?

Thanks for updating the screenshots as well. Now seeing it, I think it's on the borderline of being too long. What do @mnonnenmacher and @lamppu think? Should we revert to just "Graphs" for both menu and title? But then again, this could be confused with graphical graphs, like in charts.

Etsija:
> > I think "Graphs" is fine, but should we be more verbose and say "Dependency Graphs"? Or does that take up too much horizontal space in the menu then?
> 
> Thanks for updating the screenshots as well. Now seeing it, I think it's on the borderline of being too long. What do @mnonnenmacher and @lamppu think? Should we revert to just "Graphs" for both menu and title? But then again, this could be confused with graphical graphs, like in charts.

I'd say keep it as is for now, improve later if it starts to annoy us.

mnonnenmacher:
> > I think "Graphs" is fine, but should we be more verbose and say "Dependency Graphs"? Or does that take up too much horizontal space in the menu then?
> 
> Thanks for updating the screenshots as well. Now seeing it, I think it's on the borderline of being too long. What do @mnonnenmacher and @lamppu think? Should we revert to just "Graphs" for both menu and title? But then again, this could be confused with graphical graphs, like in charts.

I prefer "Dependency Graphs" because it's clearer to the user (and they maybe know the term from GitHub already).

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/dependencies/-components/dependency-graph-utils.ts
```diff
@@ -0,0 +1,206 @@
...
+  pkg: Identifier | undefined
+): string => {
+  if (!pkg) return '';
+
+  if (pkg.type) {
+    return identifierToString(pkg);
+  }
+
+  return [pkg.namespace, pkg.name, pkg.version].filter(Boolean).join(':');
+};
```
lamppu: Can you explain what kind of a case this helper function is needed for, why isn't `identifierToString` sufficient on its own?

components/dependency-graph/backend/src/main/kotlin/DependencyGraphService.kt
```diff
@@ -0,0 +1,285 @@
...
+import org.jetbrains.exposed.v1.core.eq
+import org.jetbrains.exposed.v1.core.inList
+import org.jetbrains.exposed.v1.core.inSubQuery
+import org.jetbrains.exposed.v1.core.innerJoin
+import org.jetbrains.exposed.v1.core.isNotNull
+import org.jetbrains.exposed.v1.jdbc.Database
+import org.jetbrains.exposed.v1.jdbc.select
+
+class DependencyGraphService(private val db: Database) {
+    fun getDependencyGraphs(ortRunId: Long): DependencyGraphs = db.blockingQuery {
```
mnonnenmacher: Please add KDoc for the class and function with a short description what they do.

components/dependency-graph/backend/src/routes/kotlin/routes/Routing.kt
```diff
@@ -0,0 +1,32 @@
...
+package org.eclipse.apoapsis.ortserver.components.dependencygraph.routes
+
+import io.ktor.server.routing.Route
+import io.ktor.server.routing.route
+
+import org.eclipse.apoapsis.ortserver.components.authorization.routes.AuthorizationChecker
+import org.eclipse.apoapsis.ortserver.components.dependencygraph.backend.DependencyGraphService
+
+fun Route.dependencyGraphRoutes(service: DependencyGraphService, checker: AuthorizationChecker) {
+    route("dependency-graph") {
```
mnonnenmacher: As already mentioned in an earlier PR, routes in components should contain the full path and be imported in `Routing.kt` of `core`, otherwise there is no guarantee that required path parameters are available and it will be hard to keep the overview where routes are coming from.

components/dependency-graph/backend/src/test/kotlin/routes/GetDependencyGraphIntegrationTest.kt
```diff
@@ -0,0 +1,190 @@
...
+                response shouldHaveStatus HttpStatusCode.OK
+                response.body<DependencyGraphs>() shouldBe DependencyGraphs(emptyMap())
+            }
+        }
+
+        "return an empty graphs map when no analyzer run exists" {
+            dependencyGraphTestApplication { client ->
+                val response = client.get("/runs/${dbExtension.fixtures.ortRun.id}/dependency-graph")
+
+                response shouldHaveStatus HttpStatusCode.OK
```
mnonnenmacher: I think in this case `NotFound` should be returned instead.

shared/ktor-utils/src/main/kotlin/OpenApiHelpers.kt
```diff
@@ -37,6 +37,17 @@ inline fun <reified T> ResponseConfig.jsonBody(noinline block: SimpleBodyConfig.
         block()
     }
 
+/**
```
mnonnenmacher, line 40: Commit message nit: Add the scope `refactor(ktor-utils): ...`.

mnonnenmacher (changes requested):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/dependencies/-components/dependency-graph-utils.ts
Etsija: Unnecessary function, removed.

components/dependency-graph/backend/src/test/kotlin/routes/GetDependencyGraphIntegrationTest.kt
Etsija: That would make the API much more complicated for this nuance, as the service returns empty map in both cases: when analyzer actually has no dependency graphs / the analyzer hasn't been run. Do you really want this?

mnonnenmacher: AFAIK other `/runs/{runId}` endpoints return 404 if a run does not exist, it should be consistent. Why would that be much more complicated?

components/dependency-graph/backend/src/main/kotlin/DependencyGraphService.kt
mnonnenmacher: Docs are still missing in the second commit.

components/dependency-graph/backend/src/routes/kotlin/routes/Routing.kt
mnonnenmacher: Also not fixed yet.

Etsija: Can you be more specific: what's not fixed here? I don't quite understand your review comment, but thought I had fixed it already.

mnonnenmacher: It's basically the same request as here:
https://github.com/eclipse-apoapsis/ort-server/pull/4815#discussion_r3072390878

Routes in components should set the full path (`/runs/{runId}/...`) and be imported in the `Routing.kt` of `core`.

Etsija: Ah, they're not imported yet in the core Routing.kt.

Etsija: Now hopefully fixed.

mnonnenmacher: The fix ended up in the wrong commit (it should have been in the third commit that introduces the route, not the last one). If it is a lot of effort to move the fix to the third commit I would be ok with keeping it as is.

Etsija: Yeah it kind of is due to git creating hunks that are tied to two commits, and then I'd need to do some git magic to assign the hunks all to their specific commits. I'd be grateful if this could be kept as-is.

Even GitButler cannot fight against the sometimes-stupid hunk creation that happens in Git.

mnonnenmacher (approved):

PR #4918, opened 2026-04-24
fix(secrets): Scope in-use check to the secret's hierarchy level

Juli0q:
Only check infrastructure services at the same hierarchy level when deleting a secret, as same-named secrets at different levels are independent.


sschuberth:
@Juli0q, for future reference, mind adding a rationale _why_ this was closed unmerged?

PR #4922, opened 2026-04-24
fix(ui): Adjust padding and margins in plugin and advisor fields

### LINKED ISSUE #4894: Inconsistent title / text spacing in advisor options
See

Image

All spacing should be like between "Enabled advisors" and its below text. I.e. spacing below "Skip excluded" is too small, and spacing after property names with a badge are too big. The latter is probably due to the height of the badge. So either decrease its height / size, or align all spacing to the one needed if there is a badge present, and use that spacing also if no badge is present.

Joelp03:
This PR addresses the spacing inconsistencies reported in the issue, using the spacing between "Enabled advisors" and the text below as the standard reference.

- Vertical Spacing: Unified vertical distance between elements to maintain consistency across the view.
- Properties with Badges: Fixed excessive spacing after property names caused by the badge height.
- Subitems: Added margin-left to subitems to improve visual hierarchy and indentation.

image

Resolves #4894.

### DIFF
--- a/ui/src/components/form/plugin-multi-select-field.tsx
+++ b/ui/src/components/form/plugin-multi-select-field.tsx
@@ -123,7 +123,7 @@ export const PluginMultiSelectField = <
               Enable/disable all
             </Label>
           </div>
-          <Separator />
+          <Separator className='my-2' />
           {plugins.map((plugin) => (
             <FormItem
               key={plugin.id}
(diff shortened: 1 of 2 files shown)


Joelp03:
I added margin-left to the "subitems" because there was no visual distinction from their parents. However, I only applied it to the Advisor component for now; we will need to add it to other components that require it later.

sschuberth:
Thanks! The visuals look good to me, but the Prettier checks fails. Assigning to @Etsija for the actual code review.

Etsija (approved):

PR #4923, opened 2026-04-24
Store the excluded status for projects and packages

lamppu:


dao/src/main/kotlin/repositories/analyzerrun/DaoAnalyzerRunRepository.kt
```diff
@@ -145,7 +147,11 @@ private fun createAnalyzerConfiguration(
     return analyzerConfigurationDao
 }
 
-private fun createProject(analyzerRun: AnalyzerRunDao, project: Project): ProjectDao {
+private fun createProject(
+    analyzerRun: AnalyzerRunDao,
+    project: Project,
+    excludedProjectIds: Set<Identifier>
```
mnonnenmacher: I would prefer to just pass `isExcluded: Boolean` here and do the check on the call side, same for the other function.

mnonnenmacher (changes requested):

sschuberth:
@lamppu, there's also a conflict now that you need to resolve.

mnonnenmacher (approved):

PR #4924, opened 2026-04-24
build: Fix the logic to opt-in to the experimental serialization API

mnonnenmacher:
The previous approach checked if the serialization plugin is applied to a module, but this fails the compilation for source sets which do not have the serialization library on the classpath. Therefore, the compiler options were only applied if the task name started with "compileKotlin", but as a result tasks like `compileRoutesKotlin` were not configured correctly.

Instead of changing the filter logic which might break again with later changes, apply the opt-in to all compilations that have `kotlinx-serialization-core` on the classpath.

### DIFF
--- a/buildSrc/src/main/kotlin/ort-server-kotlin-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/ort-server-kotlin-conventions.gradle.kts
@@ -54,18 +54,18 @@ detekt {
 }
 
 tasks.withType<KotlinCompile>().configureEach {
-    if (!name.startsWith("compileKotlin")) return@configureEach
-
-    val hasSerializationPlugin = plugins.hasPlugin(libs.plugins.kotlinSerialization.get().pluginId)
-
-    val optInRequirements = listOfNotNull(
-        "kotlinx.serialization.ExperimentalSerializationApi".takeIf { hasSerializationPlugin }
-    )
+    val serializationOptIn = libraries.elements.map { files ->
+        buildList {
+            if (files.any { it.asFile.name.startsWith("kotlinx-serialization-core") }) {
+                add("kotlinx.serialization.ExperimentalSerializationApi")
+            }
+        }
+    }
 
     compilerOptions {
         allWarningsAsErrors = true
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/ort-server-kotlin-conventions.gradle.kts
```diff
@@ -54,18 +54,18 @@ detekt {
...
 
 tasks.withType<KotlinCompile>().configureEach {
-    if (!name.startsWith("compileKotlin")) return@configureEach
-
-    val hasSerializationPlugin = plugins.hasPlugin(libs.plugins.kotlinSerialization.get().pluginId)
-
-    val optInRequirements = listOfNotNull(
-        "kotlinx.serialization.ExperimentalSerializationApi".takeIf { hasSerializationPlugin }
-    )
+    val serializationOptIn = libraries.elements.map { files ->
```
sschuberth, line 57: How about keeping most of the original code, and just replace the `hasSerializationPlugin` boolean with `val hasKxS = libraries.files.any { it.name.startsWith("kotlinx-serialization-core") }`?

mnonnenmacher: This doesn't work, the classpath is not known at configuration time, I have to use `libraries.elements` to create a provider which is evaluated at build time.

sschuberth: Does moving the logic to inside `doFirst` help to avoid that?

mnonnenmacher: I couldn't make it work with `doFirst` but maybe I tried something else than you had in mind. But what's the problem with the current implementation?

sschuberth: > But what's the problem with the current implementation?

It's just ugly 😸 Quite a bit of nesting, and usage of `buildList` for a list that contains at most one element seems overkill.

But if there's no low-effort to make it nicer, let's go with it.

sschuberth (approved):

PR #4925, opened 2026-04-24
Plugin template improvements

mnonnenmacher:

### DIFF
--- a/components/plugin-manager/backend/src/main/kotlin/PluginTemplateState.kt
+++ b/components/plugin-manager/backend/src/main/kotlin/PluginTemplateState.kt
@@ -44,7 +44,12 @@ internal class PluginTemplateState(
 
     fun apply(event: PluginTemplateEvent) = apply {
         when (event.payload) {
-            is Deleted -> isDeleted = true
+            is Deleted -> {
+                options = emptyList()
+                isGlobal = false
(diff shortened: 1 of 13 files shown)


components/plugin-manager/backend/src/main/kotlin/PluginTemplateState.kt
```diff
@@ -44,7 +44,12 @@ internal class PluginTemplateState(
 
     fun apply(event: PluginTemplateEvent) = apply {
         when (event.payload) {
-            is Deleted -> isDeleted = true
+            is Deleted -> {
+                options = emptyList()
+                isGlobal = false
```
sschuberth, line 49: Commit message: Typo in "isGloabl"

sschuberth: Does `version` also need to be reset here?

sschuberth (changes requested):

components/plugin-manager/backend/src/main/kotlin/PluginTemplateState.kt
mnonnenmacher: No, the `version` must always increase.

sschuberth (approved):

Issue #4381, opened 2026-01-27
Verify that compression for HTTP responses is enabled


sschuberth:
It seems that Ktor is not currently configured to do compression for HTTP responses. We should enable that by installing the [compression Ktor plugin](https://ktor.io/docs/server-compression.html).

sschuberth:
Just for my understanding @oheger-bosch, was https://github.com/eclipse-apoapsis/ort-server/pull/4992 a preparation to be able to address this issue in a central place, and will you take care of this issue?

oheger-bosch:
#4992 was not strictly a preparation for this issue, it was just a consequent next step after the introduction of the standard HTTP client. But having now all HTTP-related configuration centrally, makes this task easier of course. However, before further extending this configuration, #4975 should be completed first.

I can try to pick this up between my other tasks, but cannot promise when this will be. From having a short glance at the compression plugin, it seems to have quite a bunch of options. Do you already have an idea how these should be reflected in the standard configuration and which defaults should be set?

sschuberth:
> Do you already have an idea how these should be reflected in the standard configuration and which defaults should be set?

I'd simply [stick to the defaults](https://ktor.io/docs/server-compression.html#install_plugin) by specifying `install(Compression)`, or maybe enable `zstd()` on [top of it](https://ktor.io/docs/server-compression.html#add_specific_encoders). That's it.

sschuberth:
@oheger-bosch, any update here? I had a brief discussion with @MarcelBochtler, and we believe this should not be blocked by / wait for #4975.

oheger-bosch:
No update.

PR #4955, opened 2026-04-29
Introduce a ktor-client-utils module

oheger-bosch:
This new project contains functionality to create an HTTP client with a robust configuration to handle timeouts, server failures, and rate limiting. For now, this new client is integrated with the Vault secrets provider (which currently causes us some trouble). In follow-up PRs, other modules can be converted to use this functionality, too.


sschuberth:
@oheger-bosch, this sounds as if it would make https://github.com/eclipse-apoapsis/ort-server/issues/4381 easier to implement (in a single place). If that's so, would you mind enabling compression as part of this PR?

shared/ktor-client-utils/src/main/kotlin/HttpClient.kt
```diff
@@ -0,0 +1,70 @@
...
+            connectTimeoutMillis = config.connectTimeout.inWholeMilliseconds
+            socketTimeoutMillis = config.socketTimeout.inWholeMilliseconds
+            requestTimeoutMillis = config.requestTimeout.inWholeMilliseconds
+        }
+
+        install(HttpRequestRetry) {
+            retryOnServerErrors(maxRetries = config.maxRetries)
+
+            if (config.retryExponentialBackoff) {
+                exponentialDelay(base = config.retryDelay.inWholeMilliseconds.toDouble())
```
MarcelBochtler: If I understand the docs in `HttpClientConfig.retryDelay` correct, you have to pass `config.retryDelay.inWholeMilliseconds.toDouble()` as `baseDelayMs` not as `base`.

oheger-bosch:
The new central HTTP client configuration is certainly the right place to enable compression for all HTTP clients. However, I would rather handle compression in a follow-up PR. We have currently some trouble with Vault and therefore would need these changes soon. The [compression plugin](https://ktor.io/docs/server-compression.html) defines a number of settings which means that the corresponding configuration options need some careful design. Also, to make use of the standard HTTP client configuration, the consuming modules need to be updated.

shared/ktor-client-utils/src/main/kotlin/RateLimitHandler.kt
```diff
@@ -0,0 +1,215 @@
...
+            retryCount++
+
+            val delayMs = resolveDelay(
+                retryAfterHeader = call.response.headers["Retry-After"],
+                retryCount = retryCount,
+                config = config
+            )
+
+            logger.warn(
+                "Rate limit exceeded (HTTP 429). Waiting ${delayMs}ms before retry " +
```
MarcelBochtler: `delayMs` is of type `Duration`, so the unit of measurement is baked into the type. It is not always in ms.

Additionally, the log message will have douplicated units, because the `.toString()` of `Duration` will already print the unit.

shared/ktor-client-utils/src/main/kotlin/RateLimitHandler.kt
```diff
@@ -0,0 +1,215 @@
...
+import kotlin.time.Clock
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
+
+import kotlinx.coroutines.delay
+
+import org.slf4j.LoggerFactory
+
+private val logger = LoggerFactory.getLogger(RateLimitConfig::class.java)
```
MarcelBochtler, line 39: Wrong class for the logger.

secrets/vault/src/main/resources/application.conf
```diff
@@ -22,5 +22,20 @@ secretsProvider {
   vaultRootPath = ${?VAULT_ROOT_PATH}
   vaultPrefix = ${?VAULT_PREFIX}
   vaultNamespace = ${?VAULT_NAMESPACE}
-  vaultHttpTimeoutSec = ${?VAULT_HTTP_TIMEOUT_SEC}
```
MarcelBochtler, line 25: Removing this value is a breaking change and should be documented as such.

MarcelBochtler (changes requested):

shared/ktor-client-utils/src/main/kotlin/RateLimitHandler.kt
oheger-bosch: Since the rate limiting functionality is implemented as an extension function, there is no central class which could be used here. Which one would you propose?

shared/ktor-client-utils/src/main/kotlin/HttpClient.kt
oheger-bosch: Good catch! Changed.

sschuberth:
> I would rather handle compression in a follow-up PR.

Fine with me! But does that mean you'd be willing to take that task, @oheger-bosch?

shared/ktor-client-utils/src/main/kotlin/RateLimitHandler.kt
MarcelBochtler: I'd suggest:
```kt
private val logger = LoggerFactory.getLogger("RateLimitHandler")
```

oheger-bosch: Actually, I had this first. The name without a package is a bit problematic, because in some places we have logging configurations that adapt the log level for the "org.eclipse.apoapsis" package. This would then no longer match.

While this is probably not that relevant for the logging of rate limits - which only uses the levels WARN and ERROR that should always be enabled -, the logger in the HttpClient file has the same issue. Here the DEBUG logs are discarded if no logger in the org.eclipse.apoapsis is used.

shared/ktor-client-utils/src/main/kotlin/RateLimitHandler.kt
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * A strategy implementing an exponential backoff algorithm. Starting with [RateLimitConfig.defaultDelay], for
+     * each retry, the delay is doubled.
+     */
+    BACKOFF {
+        override fun computeDelay(
+            retryCount: Int,
+            config: RateLimitConfig
+        ): Duration = (config.defaultDelay.inWholeMilliseconds * (1L shl (retryCount - 1))).milliseconds
```
MarcelBochtler: If I understand this correctly this is just doing:
```kt
config.defaultDelay * 2.0.pow(retryCount - 1)
```
Which is way simpler to read for me. Do you mind changing it?

secrets/vault/src/main/resources/application.conf
oheger-bosch: Changed the commit message and added a "!".

shared/ktor-client-utils/src/main/kotlin/RateLimitHandler.kt
oheger-bosch: If you insist, I will change it. But the current approach has the advantage that it works with integer arithmetic and does not require double conversions.
I see there are not so many developers around any more that have worked with Assembler.

MarcelBochtler: > I see there are not so many developers around any more that have worked with Assembler.

Correct.

> If you insist, I will change it.

I'd appreciate that.

sschuberth: Just to add to that, there'll most likely also be no performance gain, as this is one of the low-hanging fruit optimizations that almost all compilers will do automatically. So general readability should be preferred here.

sschuberth: Maybe it makes sense to instead use the `HttpRequestRetry` plugin which supports exponential backoff out of the box?

sschuberth: Oh, sorry, it's already in use.

oheger-bosch: Changed.

oheger-bosch:
Note: Since this PR contains a breaking change and introduces a new rather global feature, we decided to not include it in today's ORT Server release to give all parties more time for the review.

shared/ktor-client-utils/build.gradle.kts
mnonnenmacher, line 1: Commit message nit: "KTOR" -> "Ktor"

shared/ktor-client-utils/build.gradle.kts
```diff
@@ -17,12 +17,23 @@
...
-export * from './evaluator-fields';
-export * from './move-repository';
-export * from './notifier-fields';
-export * from './package-manager-field';
-export * from './reporter-fields';
-export * from './repository-runs-table';
-export * from './scanner-fields';
+plugins {
+    // Apply precompiled plugins.
+    id("ort-server-kotlin-jvm-conventions")
```
mnonnenmacher, line 22: Would it make sense to make this a multiplatform module so that it could potentially be used by the CLI as it should be able to handle rate limits configured for the API (also a question to @MarcelBochtler)?

shared/ktor-client-utils/src/main/resources/application.conf
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+rateLimit {
```
mnonnenmacher: Could this be renamed to make it clearer that this is client configuration? Maybe `httpclient.rateLimit`? Just to avoid any confusion with the server side rate limits at `ktor.rateLimit`.
And how does this config relate to the vault specific config options introduced in a later commit? Is this file even needed?

oheger-bosch: It is moved under a `httpClient` section in a later commit.

In this file, the global default values for the several configuration options are set. In the `application.conf` files of the consuming modules, only environment variables are defined that can be used to override all or a subset of these values.

shared/ktor-client-utils/build.gradle.kts
oheger-bosch: Maybe, but this can certainly be done later? From my experience, this multiplatform stuff tends to make everything more complex.

MarcelBochtler: Yes, rate limiting handling will have to be implemented as well for the CLI HTTP client.

> but this can certainly be done later?

Not sure about that. When this isn't done from the beginning migrating will be way harder, as developers start to rely on JVM-only libraries (like it is already done with `typesafeConfig` and `slf4j` in this PR).

> this multiplatform stuff tends to make everything more complex.

The alternative is to have another similar implementation for rate limit handling.

However, I think that this should not block this change.

MarcelBochtler (approved):

shared/ktor-client-utils/build.gradle.kts
oheger-bosch: Actually, the functionality related to configuration (based on typesafe config) is a major motivation for this PR; so that a concrete installation of ORT Server can override these settings for all relevant modules. I assume, for the CLI, some settings would need to be hard-coded, or the configuration would be done differently (via command line options?).

mnonnenmacher: It would probably be sufficient to have the config file support only in the JVM module, so only the logging might be a problem if this is migrated to multiplatform later.

PR #4964, opened 2026-04-29
Fix detected licenses showing scan results from wrong run

### LINKED ISSUE #4957: UI shows scan results from wrong scanner
The "detected licenses" UI currently shows all stored scan results for packages, not only those for the scanners that were selected for the run. For example, in this run only ScanCode was selected but results for Askalono and Licensee are shown as well:

Image

The root cause is probably the join table introduced in #4870.

To reproduce the issue, start a run with one scanner and the rerun it with another scanner.

Etsija:
This was tested with run 13 using only ScanCode, and run 12 only DOS.

Screenshot from 2026-04-29 13-50-18

Screenshot from 2026-04-29 13-50-46

Please see the commits for details. The latter one fixes one annoying missing feature from switching the run views, which was helpful for producing exactly same views from both runs.

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/-components/run-details-bar.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/-components/run-details-bar.tsx
@@ -111,10 +111,11 @@ export const RunDetailsBar = ({ className }: RunDetailsBarProps) => {
   // E.g., if on .../runs/5/config, navigating to run 4 should go to
   // .../runs/4/config instead of /runs/4.
   const buildRunPath = (runIndex: number) => {
-    return location.pathname.replace(
+    const pathname = location.pathname.replace(
       `/runs/${params.runIndex}`,
       `/runs/${runIndex}`
     );
+    return location.searchStr ? `${pathname}${location.searchStr}` : pathname;
   };
 
   return (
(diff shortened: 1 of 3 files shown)


mnonnenmacher (approved):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/-components/run-details-bar.tsx
```diff
@@ -111,10 +111,11 @@ export const RunDetailsBar = ({ className }: RunDetailsBarProps) => {
   // E.g., if on .../runs/5/config, navigating to run 4 should go to
   // .../runs/4/config instead of /runs/4.
   const buildRunPath = (runIndex: number) => {
-    return location.pathname.replace(
+    const pathname = location.pathname.replace(
       `/runs/${params.runIndex}`,
       `/runs/${runIndex}`
     );
+    return location.searchStr ? `${pathname}${location.searchStr}` : pathname;
```
lamppu, line 118: I'm wondering what happens if pagination states are preserved and the tables in different runs have a different amount of pages, and the user gets navigated to a page that doesn't exist in the table..

Etsija: Well I only verified that the UI survives from a run where there are _some_ results -> to a run where no results exist (because scanner didn't run). But yes, did not think of pagination. It could be safest to leave pagination out of the search params?

Etsija: I actually tested this out: TanStack Router / Table do prevent the UI from crashing in such cases - the search params related to pages "outside the data" just return to 1. Kudos to TanStack!

PR #4965, opened 2026-04-29
Distinguish empty states for detected licenses

### LINKED ISSUE #4892: Detected licenses view should show special message if no scanner was run
Instead of an empty table like

Image

it would be more user-friendly to explain that no detected licenses are available because the scanner job was not run. Maybe even with a "rerun with the scanner job enabled" link that takes the user straight to the rerun form, to an enabled / expanded scanner section.

Etsija:
#### Empty state due to no findings / no filtered findings / scanner failure

Screenshot from 2026-04-29 15-04-50

#### Empty state due to scanner not being run as part of the ORT run

Screenshot from 2026-04-29 15-04-22

@sschuberth I decided not to implement your proposal for linking to run creation form with the scanner enabled - at this stage - due to it being a bit complicated and leaving open questions - specifically, what scanners to enable and with what options. However, the PR passes the optional "empty state" message as a `ReactNode` so in future, the message can be tailored for different purposes - including links, buttons, special formatting etc. So the feature is future-ready, and the tailored empty-state message can be used for any table in our UI.

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings/-components/license-findings-view.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings/-components/license-findings-view.tsx
@@ -251,6 +251,13 @@ export const LicenseFindingsView = () => {
     totalDetectedLicenses.pagination.totalCount !==
     detectedLicenseFindings.pagination.totalCount;
   const matching = `, ${detectedLicenseFindings.pagination.totalCount} matching filters`;
+  const scannerWasIncludedInRun = ortRun.jobs.scanner != null;
+  const noResultsContent = !scannerWasIncludedInRun ? (
+    <div className='text-muted-foreground text-sm'>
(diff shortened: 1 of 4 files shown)


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/license-findings/-components/license-findings-view.tsx
```diff
@@ -251,6 +251,13 @@ export const LicenseFindingsView = () => {
     totalDetectedLicenses.pagination.totalCount !==
     detectedLicenseFindings.pagination.totalCount;
   const matching = `, ${detectedLicenseFindings.pagination.totalCount} matching filters`;
+  const scannerWasIncludedInRun = ortRun.jobs.scanner != null;
+  const noResultsContent = !scannerWasIncludedInRun ? (
+    <div className='text-muted-foreground text-sm'>
+      No detected licenses are available because the scanner job was not
+      included in this run.
```
sschuberth: Nit: Maybe "enabled for this run"?

sschuberth (approved):

PR #4968, opened 2026-04-30
ui: Some UX tunings for the plugins page

### LINKED ISSUE #4959: Plugin template descriptions use enum names instead of display names
See e.g.

Image

Instead of "PACKAGE_CONFIGURATION_PROVIDER", we should say "Package Configuration Provider" (like on the overview page of installed plugins).

Maybe relate, also the URL path contains "admin/plugins/PACKAGE_CONFIGURATION_PROVIDER/DOS". This does not _need_ to change from my point of view, but I also would not mind if it _does_ change as a side-effect of fixing the above issue.

Etsija:
#### All plugin types shown in a single compact accordion

Screenshot from 2026-04-30 07-12-52

#### One plugin type can be opened at a time

Screenshot from 2026-04-30 07-13-12

#### Render plugin descriptions as markdown to eg. have the links functional

Screenshot from 2026-04-30 07-13-40

#### Use descriptive plugin types for templates

Screenshot from 2026-04-30 07-41-41

### DIFF
--- a/ui/src/components/markdown-renderer.tsx
+++ b/ui/src/components/markdown-renderer.tsx
@@ -19,13 +19,19 @@
 
 import Markdown from 'react-markdown';
 
+import { cn } from '@/lib/utils';
+
 type MarkdownRendererProps = {
   markdown?: string;
+  className?: string;
 };
 
-export const MarkdownRenderer = ({ markdown }: MarkdownRendererProps) => {
+export const MarkdownRenderer = ({
+  markdown,
+  className,
+}: MarkdownRendererProps) => {
   return (
-    <div className='prose dark:prose-invert text-sm'>
+    <div className={cn('prose dark:prose-invert text-sm', className)}>
       <Markdown
         components={{
           h1: ({ children }) => (
--- a/ui/src/lib/types.ts
+++ b/ui/src/lib/types.ts
@@ -19,7 +19,7 @@
 
 import { z } from 'zod';
 
-import { RepositoryType } from '@/api';
+import { PluginType, RepositoryType } from '@/api';
 import { zEnvironmentConfig } from '@/api/zod.gen';
 
 /**
@@ -237,6 +237,22 @@ export function getRepositoryTypeLabel(type: RepositoryType | string): string {
     return repositoryTypeLabels[type as RepositoryType];
   }
   return type ? `"${type}"` : 'Unset';
+}
+
+const pluginTypeLabels: Record<PluginType, string> = {
+  ADVISOR: 'Advisor',
+  PACKAGE_CONFIGURATION_PROVIDER: 'Package Configuration Provider',
+  PACKAGE_CURATION_PROVIDER: 'Package Curation Provider',
+  PACKAGE_MANAGER: 'Package Manager',
+  REPORTER: 'Reporter',
+  SCANNER: 'Scanner',
+};
+
+export function getPluginTypeLabel(type: PluginType | string): string {
+  if (type in pluginTypeLabels) {
+    return pluginTypeLabels[type as PluginType];
+  }
+  return type ? `"${type}"` : 'Unknown';
 }
 
 // Type and constant for color themes. New themes can be added to the project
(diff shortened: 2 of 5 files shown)


ui/src/routes/admin/plugins/index.tsx
```diff
@@ -29,6 +29,12 @@ import {
   restrictPluginMutation,
 } from '@/api/@tanstack/react-query.gen';
 import { PluginAvailabilityToggle } from '@/components/plugin-availability-toggle';
+import {
```
sschuberth, line 33: So are Accordions collapsed by default?

To solve the problem of loosing overview, I was also wondering whether we should have a TOC at the top, with anchors linking to the sections for the different plugin types.

ui/src/components/markdown-renderer.tsx
```diff
@@ -19,13 +19,19 @@
 
 import Markdown from 'react-markdown';
 
+import { cn } from '@/lib/utils';
```
sschuberth, line 22: Commit message nit: I think it's "Markdown".

ui/src/lib/types.ts
```diff
@@ -239,6 +239,22 @@ export function getRepositoryTypeLabel(type: RepositoryType | string): string {
   return type ? `"${type}"` : 'Unset';
 }
 
+const pluginTypeLabels: Record<PluginType, string> = {
+  ADVISOR: 'Advisor',
```
sschuberth, line 243: Hmm, don't we have this mapping somewhere else already? Or can we somehow get it from the API endpoint for plugins, @mnonnenmacher? It would be good to not have to maintain this here.

Etsija: I knew you'd ask: sadly, we don't get the list from back-end to OpenAPI, so currently have to maintain the list in the UI. That's true for many other things, currently.

If we get a new plugin type in the future (I couldn't think of what it would be), the UI is prepared: if a new type is added to back-end, and the UI doesn't know of it yet, the UI will render `Unknown` in its place. This is exactly the same behaviour as for example the repository types rendering.

ui/src/routes/admin/plugins/index.tsx
Etsija: They are collapsed by default. I don't see how a TOC would help, as the view in the first screenshot is what the user sees when visiting the page, and it's a self-explanatory TOC by itself.

sschuberth: > the view in the first screenshot is what the user sees when visiting the page, and it's a self-explanatory TOC by itself.

Yeah, it probably would have been _either_ a TOC _or_ Accordions.

Etsija: IMHO TOCs are old school. Accordions used in this way act as a functional TOC, and the UI is kept compact as-is.

ui/src/lib/types.ts
mnonnenmacher: I don't think we have this mapping anywhere. But instead of a hardcoding the mapping, would it be an option to just replace `_` with ` ` and title case the words?

Etsija: I'd really not reinvent the method we already are using (in `/lib`) for these cases. If not for anything else, the code file is actually useful, as it tells us what's still missing from OpenAPI. Kind of  "do-to list".

sschuberth: Thinking about it: Coming up with a display name for the plugin type enum is a visual thing after all. So I don't believe anymore it should come from the backend / API, but the frontend should decide how to name things in user-facing places. In that sense, having the "enum name" -> "display name" mapping here in the frontend code is correct.

sschuberth (approved):

ui/src/lib/types.ts
mnonnenmacher: I just don't like that it will only be visible at runtime if a mapping is missing, a compile time error like with Kotlin's exhaustive `when` would be much better.

PR #4969, opened 2026-04-30
ui: Add link to the repository to the breadcrumb

### LINKED ISSUE #4939: Add a clickable repository link to the overview bar
I often miss a clickable link to the repository when looking at a run. To ensure the link is visible on any page of a run, I propose to add it to the overview bar. Esp. with the new feature to name repositories, there's otherwise no link visible:

Image

More concretely, the proposal is to have a "Repository: https://..." link between the current "Run ID" and "Revision" entries. The UI should be prepared for handling long links here, probably by cropping them with ellipses (`...`).

Etsija:
Screenshot from 2026-04-30 11-40-18

### DIFF
--- a/ui/src/components/siblings.tsx
+++ b/ui/src/components/siblings.tsx
@@ -19,7 +19,7 @@
 
 import { useQuery } from '@tanstack/react-query';
 import { Link, useParams, useRouter } from '@tanstack/react-router';
-import { Check, ChevronDown } from 'lucide-react';
+import { Check, ChevronDown, ExternalLink } from 'lucide-react';
 
 import {
   getOrganizationProductsOptions,
@@ -28,6 +28,7 @@ import {
   getRepositoryRunsOptions,
 } from '@/api/@tanstack/react-query.gen';
 import { BreadcrumbItem, BreadcrumbLink } from '@/components/ui/breadcrumb';
+import { Button } from '@/components/ui/button';
 import {
   DropdownMenu,
   DropdownMenuContent,
@@ -110,6 +111,11 @@ export const Siblings = ({ entity, pathName }: SiblingsProps) => {
     enabled: entity === 'run' || !!params.repoId,
   });
 
+  const orgs = organizations?.data;
+  const prods = products?.data;
+  const repos = repositories?.data;
+  const runIndexes = runs?.data;
+
   const name =
     entity === 'organization'
       ? breadcrumbs.organization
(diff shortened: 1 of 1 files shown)


mnonnenmacher:
It still feels wrong to me to have this button in this place, just two more ideas:
* How about having the button with just the icon directly in the breadcrumbs? Is that something you could quickly try out?
* Having the revision link directly to the correct revision in the repository would be even better, but this would be more complex to implement.

sschuberth:
> Having the revision link directly to the correct revision in the repository would be even better, but this would be more complex to implement.

I was now thinking about that as well. We do have the code to create "deep"- / "perma"-links to GitHub and other hosting platforms [in ORT](https://github.com/oss-review-toolkit/ort/blob/60316b797569f05fc7835f6e2f3b2bb72340f71b/downloader/src/main/kotlin/VcsHost.kt), but that's not available to the front-end (yet).

Maybe https://github.com/npm/hosted-git-info could be used?

_Edit:_ AI says like so:

```JavaScript
import hostedGitInfo from 'hosted-git-info';

function commitPermalink(repoUrl, sha) {
  const info = hostedGitInfo.fromUrl(repoUrl);
  if (!info) throw new Error('Unsupported git host');
  return info.browse(`commit/${sha}`);
}

console.log(
  commitPermalink(
    'git@github.com:user/repo.git',
    'abc123def456'
  )
);

// https://github.com/user/repo/commit/abc123def456
```

Etsija:
I made this minimal by having the link to the repository in the breadcrumb. There are some problems with using `hosted-git-info`, so I did not use it here:
- it's a pure Javascript library with no type definitions in `DefinitelyTyped`, so we would have to create our own `.d.ts` type declaration file for the library to pass TypeScript
- as I was suspecting from one of the long-open issue tickets, it does not generate a `/commit/sha` structure

If we want a link to the specific commit somewhere in our UI, we will probably need to write a DIY utility function for it.

ui/src/components/siblings.tsx
```diff
@@ -277,6 +283,13 @@ export const Siblings = ({ entity, pathName }: SiblingsProps) => {
       <BreadcrumbLink asChild>
         <Link to={pathName}>{name}</Link>
       </BreadcrumbLink>
+      {currentRepoUrl && (
+        <Button variant='ghost' size='icon' className='h-4 w-4' asChild>
+          <Link to={currentRepoUrl} target='_blank'>
```
sschuberth: Does this also need the `rel='noopener noreferrer'` stuff?

sschuberth (approved):

PR #4974, opened 2026-04-30
feat: Add support for advisor's provider-level issues

### LINKED ISSUE #4647: Take an advisor's provider-level issues into account
> @MarcelBochtler, this needs adjustments due to the snippet-related class renamings.

We also need to adjust for the changes in https://github.com/oss-review-toolkit/ort/pull/11536 because the advisor now creates issues for failed providers instead of throwing exceptions.

_Originally posted by @mnonnenmacher in https://github.com/eclipse-apoapsis/ort-server/issues/4640#issuecomment-4024691654_

lamppu:
Persist the issues and ensure that they are correctly retrieved, and that they are handled in model conversions. Also take them into account when determining run status and when storing resolutions.

Resolves https://github.com/eclipse-apoapsis/ort-server/issues/4647.

### DIFF
--- a/dao/src/main/kotlin/repositories/advisorrun/AdvisorRunsTable.kt
+++ b/dao/src/main/kotlin/repositories/advisorrun/AdvisorRunsTable.kt
@@ -72,6 +72,7 @@ class AdvisorRunDao(id: EntityID<Long>) : LongEntity(id) {
             endTime = endTime,
             environment = environment.mapToModel(),
             config = advisorConfiguration.mapToModel(),
+            providerIssues = runIssues.filter { it.identifier == null }.toSet(),
             results = results.associate { it.mapToModel(runIssues) }
         )
     }
--- a/dao/src/main/kotlin/repositories/advisorrun/AdvisorRunsTable.kt
+++ b/dao/src/main/kotlin/repositories/advisorrun/AdvisorRunsTable.kt
@@ -72,6 +72,7 @@ class AdvisorRunDao(id: EntityID<Long>) : LongEntity(id) {
             endTime = endTime,
             environment = environment.mapToModel(),
             config = advisorConfiguration.mapToModel(),
+            providerIssues = runIssues.filterTo(mutableSetOf()) { it.identifier == null },
             results = results.associate { it.mapToModel(runIssues) }
         )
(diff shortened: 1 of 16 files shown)


dao/src/main/kotlin/repositories/advisorrun/AdvisorRunsTable.kt
```diff
@@ -72,6 +72,7 @@ class AdvisorRunDao(id: EntityID<Long>) : LongEntity(id) {
             endTime = endTime,
             environment = environment.mapToModel(),
             config = advisorConfiguration.mapToModel(),
+            providerIssues = runIssues.filter { it.identifier == null }.toSet(),
```
sschuberth: Prefer: `.filterTo(mutableSetOf())`

model/src/commonMain/kotlin/repositories/AdvisorRunRepository.kt
```diff
@@ -40,6 +41,7 @@ interface AdvisorRunRepository {
         endTime: Instant,
         environment: Environment,
         config: AdvisorConfiguration,
+        providerIssues: Set<Issue> = emptySet(),
```
sschuberth: Why does this get a default value, but all other properties don't?

sschuberth (changes requested):
@mnonnenmacher, mind having a look, too?

mnonnenmacher:
> @mnonnenmacher, mind having a look, too?

I have nothing to add on top of your comments.

model/src/commonMain/kotlin/repositories/AdvisorRunRepository.kt
lamppu: I can remove it, I just added it so there wouldn't be a need to add `providerIssues = emptySet()` to some tests that call the create function.

sschuberth: I believe we shouldn't tweak production code for testability only in that way. So I'd prefer to remove the default argument here, esp. as it's not the last argument, which means setting `results` would always need a named argument with default `providerIssues`.

sschuberth (approved):

PR #4989, opened 2026-05-04
feat(ui): Add the environment config path to the ORT run

### LINKED ISSUE #4051: Add environmentConfigPath to create run in UI
In the API (https://eclipse-apoapsis.github.io/ort-server/api/post-repository-run) there is a parameter called 'environmentConfigPath', however, in the UI, when creating a run, I cannot find that option.

nnobelis:
This parameter is the optional path to an environment configuration file. If this is not defined, the environment configuration is read from the default location `.ort.env.yml`.

Fixes #4051.

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
@@ -174,6 +174,7 @@ export function defaultValues(
     },
     labels: undefined,
     jobConfigContext: '',
+    environmentConfigPath: '',
   };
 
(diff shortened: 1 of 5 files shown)


Etsija:
I usually add some pictures of the UI to UI PRs, for the reviewers, as not everyone has time to take the PR to their local environment and try it out.

nnobelis:
Run without the parameter:
image

Run with the parameter:
image

Rerun without the parameter:
image

Rerun with the parameter:
image

New run:
image

Etsija (changes requested):
I would not unnecessarily use the abbreviated "config" in the UI. After all, we do have also "Configuration context", not "Config context".

For the variables and objects in the code, abbreviating is OK, but we do have enough horizontal real estate in the UI to spell "configuration" in all the places you changed.

Etsija (approved):

PR #4990, opened 2026-05-04
fix(ui): Omit blank plugin secret default from run payload

Etsija:
Screenshot from 2026-05-05 07-27-50

### DIFF
--- a/ui/src/components/form/plugin-multi-select-field.tsx
+++ b/ui/src/components/form/plugin-multi-select-field.tsx
@@ -54,6 +54,23 @@ import { cn } from '@/lib/utils';
 
 export type ScannerScope = 'both' | 'packages' | 'projects';
 
+const UNDEFINED_SECRET_VALUE = '__undefined_secret__';
+
+function mapSecretSelectValue(value: string | undefined): string | undefined {
+  return value === UNDEFINED_SECRET_VALUE ? undefined : value;
+}
+
+function getSecretSelectDisplayValue(
+  value: string | undefined,
+  isRequired: boolean
+): string | undefined {
+  if (value === undefined) {
+    return isRequired ? undefined : UNDEFINED_SECRET_VALUE;
+  }
+
+  return value;
+}
+
 type PluginMultiSelectFieldProps<
   TFieldValues extends FieldValues,
   TName extends FieldPathByValue<TFieldValues, Array<string>>,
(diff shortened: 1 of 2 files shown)


sschuberth:
Actually, I think this rather is a bug in ORT. If a secret is given, even if it's empty, it should be passed to the plugin as-is. What do you think @mnonnenmacher?

mnonnenmacher (commented):
If you are working on this anyway, could you also make it possible to explicitly set no secret if the value is not required? Currently, once a secret was selected there is no way to go back. I think the dropdown should have an explicit option for that.

mnonnenmacher:
> Actually, I thin this rather is a bug in ORT. If a secret is given. even if it's empty, it should be passed to the plugin as-is. What do you think @mnonnenmacher?

The difference between ORT and the Server is that with the Server the user provides the name of a secret instead of the value directly. So if the option is set to an empty string, the server tries to find a secret with that name to resolve the actual secret value from the secret storage. So if a secret option is optional and the user does not want to provide a secret, the UI should not send an empty string.

Etsija:
> If you are working on this anyway, could you also make it possible to explicitly set no secret if the value is not required? Currently, once a secret was selected there is no way to go back. I think the dropdown should have an explicit option for that.

Set it where, and where should the dropdown be located? In the run creation form, or admin/plugin management templates?

Etsija:
> > If you are working on this anyway, could you also make it possible to explicitly set no secret if the value is not required? Currently, once a secret was selected there is no way to go back. I think the dropdown should have an explicit option for that.
> 
> Set it where, and where should the dropdown be located? In the run creation form, or admin/plugin management templates?

I implemented this with a "Not defined" option, see screenshot in the PR description.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/plugin-utils.ts
```diff
@@ -237,7 +241,9 @@ export function getPluginDefaultValues(
       plugin.options?.forEach((option) => {
         if (option.defaultValue !== undefined) {
           if (option.type === 'SECRET') {
-            secrets[option.name] = String(option.defaultValue);
+            if (isNonBlankString(option.defaultValue)) {
+              secrets[option.name] = option.defaultValue;
```
mnonnenmacher: Actually, I think the UI should never use the default value for secret options, because they are values not references to secrets defined in the server. Instead, the UI should just not set the secret if `option.isRequired == false` and the user did not select a secret (see also my comment in the next commit).

ui/src/components/form/plugin-multi-select-field.tsx
```diff
@@ -294,15 +306,22 @@ export const PluginMultiSelectField = <
...
+                                  )}
                                   disabled={option.isFixed}
                                 >
                                   <SelectTrigger>
                                     <SelectValue placeholder='Select a secret' />
                                   </SelectTrigger>
                                   <SelectContent>
+                                    <SelectItem value={UNDEFINED_SECRET_VALUE}>
+                                      Not defined
+                                    </SelectItem>
```
mnonnenmacher: Should this option only be added if `option.isRequired == false`?

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #4991, opened 2026-05-04
View license detections with Software Heritage service

Etsija:
Screenshot from 2026-05-05 07-57-44

Screenshot from 2026-05-04 15-28-19

This feature will be improved in follow-up PRs, so it's an initial version only.

@sschuberth the tooltip dimming in the first picture is just a screenshot artefact.

### DIFF
--- a/ui/src/lib/software-heritage.ts
+++ b/ui/src/lib/software-heritage.ts
@@ -1,0 +1,190 @@
+/*
+ * Copyright (C) 2026 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import { PackageURL } from 'packageurl-js';
+
+const SWH_BROWSE_BASE = 'https://archive.softwareheritage.org/browse/content/';
+
+/**
+ * Map a parsed purl to the canonical origin URL used by Software Heritage.
+ * Return null for unsupported package types or missing required fields.
+ */
+function purlToSwhOriginUrl(pkg: PackageURL): string | null {
+  switch (pkg.type) {
+    case 'github':
+      // pkg:github/owner/repo — namespace is owner, name is repo.
+      return pkg.namespace && pkg.name
+        ? `https://github.com/${pkg.namespace}/${pkg.name}`
+        : null;
+    case 'npm': {
+      // pkg:npm/%40scope/name or pkg:npm/name.
+      const name = pkg.namespace ? `${pkg.namespace}/${pkg.name}` : pkg.name;
+      return `https://www.npmjs.com/package/${name}`;
+    }
+    case 'pypi':
+      return `https://pypi.org/project/${pkg.name}`;
+    case 'maven':
+      // SWH archives Maven Central; namespace is groupId with dots.
+      if (pkg.namespace) {
+        const groupPath = pkg.namespace.replace(/\./g, '/');
+        return `https://repo1.maven.org/maven2/${groupPath}/${pkg.name}`;
+      }
+      return null;
+    default:
+      return null;
+  }
+}
+
+/**
+ * Return the SWH browse URL version parameter for a parsed purl.
+ *
+ * Package archives (npm, PyPI, Maven) are indexed as "releases" in SWH.
+ * GitHub refs that look like a 40-char hex SHA map to "revision"; everything
+ * else (tags, branch names) maps to "branch".
+ */
+function swhVersionParam(pkg: PackageURL): Record<string, string> {
+  if (!pkg.version) return {};
+
+  if (pkg.type === 'github') {
+    const isSha = /^[0-9a-f]{40}$/i.test(pkg.version);
+    return isSha ? { revision: pkg.version } : { branch: pkg.version };
+  }
+
+  return { release: pkg.version };
+}
+
+/**
+ * Build a Software Heritage browse URL for a specific file and line range.
+ * Return null if the purl type is not supported.
+ */
+export function buildSwhBrowseUrl(
+  purl: string,
+  path: string,
+  startLine: number,
+  endLine: number
+): string | null {
+  let pkg: PackageURL;
+  try {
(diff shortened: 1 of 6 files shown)


ui/src/lib/software-heritage.ts
```diff
@@ -0,0 +1,192 @@
...
+ * License-Filename: LICENSE
+ */
+
+import { PackageURL } from 'packageurl-js';
+
+const SWH_BROWSE_BASE = 'https://archive.softwareheritage.org/browse/content/';
+
+/**
+ * Maps a parsed purl to the canonical origin URL used by Software Heritage.
+ * Returns null for unsupported package types or missing required fields.
```
sschuberth: Just a general remark: Purls have implicit default package registry URLs, like [Maven Central for the Maven ecosystem](https://github.com/package-url/purl-spec/blob/1f8c6f2461cfe8dd3eda85eb6afa30d99b3cc6d0/types/maven-definition.json#L9). This code currently only supports mapping to this implicit default. However, the default may be [overridden by a custom `repository_url` query parameter](https://github.com/package-url/purl-spec/blob/1f8c6f2461cfe8dd3eda85eb6afa30d99b3cc6d0/types/maven-definition.json#L51). I'm ok with this limitation as long as it's documented. On the other hand, for the few mapping we have, maybe we should just make them complete by supporting custom registries.

ui/src/lib/software-heritage.ts
```diff
@@ -0,0 +1,192 @@
...
+      const name = pkg.namespace ? `${pkg.namespace}/${pkg.name}` : pkg.name;
+      return `https://www.npmjs.com/package/${name}`;
+    }
+    case 'pypi':
+      return `https://pypi.org/project/${pkg.name}`;
+    case 'maven':
+      // SWH archives Maven Central; namespace is groupId with dots
+      if (pkg.namespace) {
+        const groupPath = pkg.namespace.replace(/\./g, '/');
+        return `https://repo1.maven.org/maven2/${groupPath}/${pkg.name}`;
```
sschuberth, line 46: I guess we should strictly follow the Purl standard and use https://repo.maven.apache.org/maven2/ here, even if https://repo1.maven.org/maven2/ is a mirror of it. Also see [this comment](https://github.com/package-url/purl-spec/blob/1f8c6f2461cfe8dd3eda85eb6afa30d99b3cc6d0/types/maven-definition.json#L10).

ui/src/lib/software-heritage.ts
```diff
@@ -0,0 +1,192 @@
...
+const SWH_BROWSE_BASE = 'https://archive.softwareheritage.org/browse/content/';
+
+/**
+ * Maps a parsed purl to the canonical origin URL used by Software Heritage.
+ * Returns null for unsupported package types or missing required fields.
+ */
+function purlToSwhOriginUrl(pkg: PackageURL): string | null {
+  switch (pkg.type) {
+    case 'github':
+      // pkg:github/owner/repo — namespace is owner, name is repo
```
sschuberth: Nit: Missing periods comments after comment sentences here and below.

ui/src/lib/software-heritage.ts
```diff
@@ -0,0 +1,192 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import { PackageURL } from 'packageurl-js';
+
+const SWH_BROWSE_BASE = 'https://archive.softwareheritage.org/browse/content/';
+
+/**
+ * Maps a parsed purl to the canonical origin URL used by Software Heritage.
```
sschuberth: I don't remember about TS code, but at least in Kotlin code we have the convention of using imperative mood not only in commit messages. but also for function documentation. So say "Map a ...". Same several times below.

ui/src/lib/software-heritage.ts
```diff
@@ -0,0 +1,192 @@
...
+  const params = new URLSearchParams({
+    origin_url: originUrl,
+    path,
+    ...swhVersionParam(pkg),
+  });
+  return `${SWH_BROWSE_BASE}?${params.toString()}#L${startLine}-L${endLine}`;
+}
+
+// Unit tests
+
```
sschuberth: I believe we can omit this.

sschuberth:
> @sschuberth the tooltip dimming in the first picture is just a screenshot artefact.

Ack. But what about the idea to put the icon directly after the file path, did you revert that idea?

ui/src/lib/software-heritage.ts
Etsija: For an unknown reason, the repo1.maven.org URL is the only one that works with Software Heritage, the "standard" version leads to "Not found".

ui/src/lib/software-heritage.ts
Etsija: The convention in the UI code is mixed, but OK.

Etsija:
> > @sschuberth the tooltip dimming in the first picture is just a screenshot artefact.
> 
> Ack. But what about the idea to put the icon directly after the file path, did you revert that idea?

Yes, "reverted" as in "forgot". Screenshot updated.

ui/src/lib/software-heritage.ts
Etsija: Maybe discuss this today?

sschuberth: I forgot to mention this today. But the more I think about it, the more I believe that the current solution does not scale well. It would be much more elegant if we had the `sha1_git` as the _content_ hash, independently of the actual provenance. Then we could create links like https://archive.softwareheritage.org/browse/content/sha1_git:94a9ed024d3859793618152ea559a168bbcbb5e2/#L10-L11.

However, this would imply quite some changes, including DOS to use `sha1_git` instead of (or in addition to) `sha256`. So I'm good with this implementation as a proof-of-concept to link to Software Heritage.

ui/src/lib/software-heritage.ts
sschuberth: Ok, personally I like the non-Apache host better anyway, as it sounds more neutral.

But out of curiosity, so you have examples for full generated links that work for `repo1.maven.org` and not for `repo.maven.apache.org`?

sschuberth (approved):

ui/src/lib/software-heritage.ts
Etsija: Here's one. I did click around, and all Maven packages return 404 with the `repo.maven.apache.org`.

Screenshot from 2026-05-05 15-11-13

Screenshot from 2026-05-05 15-11-48

ui/src/lib/software-heritage.ts
sschuberth: FYI, we're using similar code in ORT's static HTML report to link to "normalized" license text only, see [this](https://github.com/oss-review-toolkit/ort/blob/013bee02074edf08bf5f466b127ad3c6122ec017/plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt#L648-L663).

ui/src/lib/software-heritage.ts
sschuberth: Ah, so https://repo.maven.apache.org/maven2/com/fasterxml/classmate/ works just fine, but Software Heritage does not recognize it as a unique source, probably because it de-duplicates the artifact link to the `repo1.maven.org` host.

PR #4992, opened 2026-05-04
Use the standard http client

oheger-bosch:
This PR reworks the remaining modules that are using an HTTP client to create this client via the new standard functionality to make sure that proper configuration is set (analogously as was done for the Vault module before).

Note that for some modules, this causes a breaking change in the configuration related to the timeouts: While timeouts had been configured via a dedicated property for these modules, this is now done via the configuration of the HTTP client.


logaccess/loki/src/main/kotlin/LokiConfig.kt
```diff
@@ -86,15 +83,12 @@ data class LokiConfig(
         /** The configuration property that defines the tenant ID if in multi-tenant mode. */
         private const val TENANT_PROPERTY = "lokiTenantId"
 
-        /** The configuration property that defines a timeout for queries against the Loki REST API in seconds. */
-        private const val TIMEOUT_SEC_PROPERTY = "lokiTimeoutSec"
+        /** The path in the application configuration that contains Loki-specific HTTP client overrides. */
```
mnonnenmacher, line 86: Commit message title in this and subsequent commits: "Use to the" -> "Use the" (or did you mean "Switch to the"?)

mnonnenmacher (changes requested):

logaccess/loki/src/main/kotlin/LokiConfig.kt
oheger-bosch: It was obviously a mixture of both. Changed to "Use the...".

mnonnenmacher (approved):

PR #4996, opened 2026-05-05
deps(analyzer): Update the default Node.js to the latest LTS version

MarcelBochtler:


sschuberth:
Should the commit be of type "deps" instead?

oheger-bosch (approved):

PR #4997, opened 2026-05-05
fix(secrets): Scope in-use check to the secret's hierarchy level

Juli0q:
Only check infrastructure services at the same hierarchy level when deleting a secret, as same-named secrets at different levels are independent.

### DIFF
--- a/components/infrastructure-services/backend/src/main/kotlin/InfrastructureServiceService.kt
+++ b/components/infrastructure-services/backend/src/main/kotlin/InfrastructureServiceService.kt
@@ -25,6 +25,9 @@ import org.eclipse.apoapsis.ortserver.dao.UniqueConstraintException
 import org.eclipse.apoapsis.ortserver.dao.blockingQuery
 import org.eclipse.apoapsis.ortserver.dao.dbQuery
 import org.eclipse.apoapsis.ortserver.dao.findSingle
+import org.eclipse.apoapsis.ortserver.dao.repositories.product.ProductsTable
+import org.eclipse.apoapsis.ortserver.dao.repositories.repository.RepositoriesTable
+import org.eclipse.apoapsis.ortserver.dao.repositories.secret.SecretsTable
 import org.eclipse.apoapsis.ortserver.dao.utils.apply
 import org.eclipse.apoapsis.ortserver.dao.utils.listQuery
 import org.eclipse.apoapsis.ortserver.model.CredentialsType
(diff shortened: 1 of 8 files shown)


components/infrastructure-services/backend/src/main/kotlin/InfrastructureServiceService.kt
```diff
@@ -214,13 +219,80 @@ class InfrastructureServiceService(
...
      */
-    suspend fun listForSecret(secretName: String): List<InfrastructureService> = db.dbQuery {
+    suspend fun listForSecret(secretName: String, id: HierarchyId): List<InfrastructureService> = db.dbQuery {
+        val secretFilter = (InfrastructureServicesTable.usernameSecret eq secretName) or
+                (InfrastructureServicesTable.passwordSecret eq secretName)
+
         list(ListQueryParameters.DEFAULT) {
-            InfrastructureServicesTable.usernameSecret eq secretName or
-                    (InfrastructureServicesTable.passwordSecret eq secretName)
+            secretFilter and when (id) {
```
oheger-bosch, line 230: I have to admit that I was not aware that the deletion of secrets can be so tricky. And what is worse: I am still not convinced that all cases are handled:

For instance, if a secret on repository level is deleted, you return the services referencing a secret with this name for this repository. It may, however, be the case that there is another secret on product or organization level with the same name, which would then be used by these services. So, they do not necessarily block the deletion of the secret.

Could the following approach work:
* Based on the passed in ID, you resolve the organization.
* You define queries for all products and all repositories in this organization, as you already do in the case of an organization ID.
* You load all secrets with this name in this organization and all infrastructure services referencing a secret with this name in this organization.
* You group the secrets and the infrastructure services by their hierarchy level.
* You find the secret on the highest level that remains after the selected secret is deleted (if any).
* Now the problematic services are those that are on a higher level than the highest secret. (Or all services if no secret remains.)

Please verify if this is correct; I am not sure about it.

components/infrastructure-services/backend/src/main/kotlin/InfrastructureServiceService.kt
```diff
@@ -214,13 +219,119 @@ class InfrastructureServiceService(
...
```
oheger-bosch, line 230: Sorry, I do not understand this code. My gut feeling is that this approach is unnecessarily complex with products and repositories containing the secret and an orgFallback (?). There should be a more straight-forward way to check whether a secret is in use.

In any case, this code needs to be made more readable. First, it does not help that the whole filter condition is constructed inline in a `when` construct going over multiple pages. Could you please extract the logic in multiple functions (probably grouped by the different hierarchy levels) where each function has a comment describing the intention and the approach? This would probably help to grasp what is going on here.

Juli0q: The simplified algorithm unfortunately doesn't hold up when there are multiple products/repos with different secret configurations. For example if product 1 has its own product level secret but product 2 doesn't, deleting an org secret would still affect repos under product 2, even though a product level secret exists elsewhere in the org. The per-entity checks are needed to handle this correctly.

I've extracted the logic into separate functions per hierarchy level with comments describing the approach. Hopefully that makes it clearer.

oheger-bosch: Thanks for the rework. I hope, we never need to touch this again. Maybe it would have been easier to load all the services and secrets of the organization and do the logic in memory, not as queries on the DB. But let's leave this for now.

oheger-bosch (approved):

PR #4999, opened 2026-05-05
fix(ui): Preserve scanner plugin config on rerun

Etsija:

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
@@ -262,7 +262,8 @@ export function defaultValues(
               ortRun.jobConfigs.scanner?.scanners ?? null,
               ortRun.jobConfigs.scanner?.projectScanners ?? null,
               scannerDefaults,
-              scannerPluginDefaultValues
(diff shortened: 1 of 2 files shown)


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
```diff
@@ -262,7 +262,8 @@ export function defaultValues(
               ortRun.jobConfigs.scanner?.scanners ?? null,
               ortRun.jobConfigs.scanner?.projectScanners ?? null,
               scannerDefaults,
-              scannerPluginDefaultValues
+              scannerPluginDefaultValues,
+              ortRun.jobConfigs.scanner?.config
```
mnonnenmacher, line 266: @Etsija Could you apply the same fix for the advisor config above?

mnonnenmacher (approved):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
Etsija: I considered it, but advisor seems to already have taken care of it. But I'll need to check with a test run.

PR #5000, opened 2026-05-06
Get and show snippet findings

### LINKED ISSUE #4706: Need an endpoint to get snippet findings
Analog to #2074.

### LINKED ISSUE #5023: Need a UI for the snippet findings
Very similar to [license findings](https://github.com/eclipse-apoapsis/ort-server/tree/main/ui/src/routes/organizations/%24orgId/products/%24productId/repositories/%24repoId/runs/%24runIndex/license-findings): 
- top-level table with snippets found from project, along with the count of matching upstream snippets
- sub-table for each snippet, showing upstream matching data

### LINKED ISSUE #5250: Evaluator and reporter fail with "The resolved revision must not be blank." when SCANOSS snippet findings are present
### Summary

When a run uses SCANOSS as the scanner and produces snippet findings, both the evaluator and reporter jobs fail with: "The resolved revision must not be blank."

### Steps to reproduce

1. Create or use a repository that contains files with partial code matches (e.g. the [test-snippet-findings
  repository](https://github.com/Etsija/test-snippet-findings)), so it produces snippet findings with SCANOSS.
2. Configure the run to use SCANOSS as the scanner.
3. Trigger a run.
4. Observe that the analyzer and scanner jobs finish successfully, but the evaluator and reporter jobs both fail with
  the error above.

### Root cause

`SnippetsTable` in ORT Server has no `resolved_revision` column. When the scanner worker stores a snippet with `RepositoryProvenance`, `SnippetDao.put()` (`SnippetsTable.kt`) persists only the `vcsInfo` foreign key and silently drops the `resolvedRevision`.

ORT's SCANOSS result parser has no real commit SHA available for snippet matches, so it uses "." as a placeholder `resolvedRevision` (see `ScanOssResultParser.kt`). The `vcsInfo` parsed from the snippet URL has an empty revision field.

When the evaluator (or reporter) later reconstructs the ORT result, `SnippetDao.mapToModel()` rebuilds `RepositoryProvenance` using `vcsInfo.revision` as the `resolvedRevision`. Since that field is empty, `RepositoryProvenance.mapToOrt()` passes "" to ORT's `RepositoryProvenance` constructor, whose init block throws:

```kotlin
require(resolvedRevision.isNotBlank()) { "The resolved revision must not be blank." }
```

### Expected behavior

The evaluator and reporter should complete successfully when snippet findings are present.

### Possible fix

Add a `resolved_revision` column to `SnippetsTable`, populate it in `SnippetDao.put()`, and use it in `SnippetDao.mapToModel()`. A database migration is probably also required.

### Additional Notes: FOSSID vs SCANOSS Provenance Handling

FOSSID does not trigger this issue because it uses `ArtifactProvenance` for snippet findings ([`FossIdScanResults.kt`](https://github.com/oss-review-toolkit/ort/blob/main/plugins/scanners/fossid/src/main/kotlin/
  FossIdScanResults.kt)):

```kotlin
val snippetProvenance = ArtifactProvenance(RemoteArtifact(url, Hash.NONE))
```

`ArtifactProvenance` has no `resolvedRevision` field, so it is unaffected.

SCANOSS instead uses `RepositoryProvenance`, which requires a non-blank `resolvedRevision`, despite no actual commit SHA being available from the SCANOSS API. The ORT SCANOSS result parser already acknowledges this with a TODO comment (ScanOssResultParser.kt):

```kotlin
// TODO: No resolved revision is available. Should a ArtifactProvenance be created instead?
val provenance = RepositoryProvenance(vcsInfo, ".")
```

This suggests an alternative fix at the ORT core level: switch SCANOSS snippet findings to use `ArtifactProvenance` (as FOSSID does), which would avoid the need for a placeholder `resolvedRevision` entirely and make the two scanners consistent in their provenance handling.

Etsija:
This PR introduces three new endpoints:
- one for listing all provenances for a snippet scan in an ORT run
- one for listing all snippet findings for an ORT run and a provenance
- one for listing the snippets for each snippet finding

It also implements the UI around the snippets, which follows closely the three-level table layout that was earlier implemented for detected licenses.

### Provenances which include snippets (top level table)

Screenshot from 2026-06-15 12-23-00

### Snippet findings for a provenance (second level table)

Screenshot from 2026-06-15 12-40-08

### Upstream snippet sources for a snippet finding (third level table)

Screenshot from 2026-06-15 12-39-25

NOTE 1: To (artificially) produce snippet findings, this repository has been used, which contains some abbreviated files (and a submodule) from known OSS repositories: https://github.com/Etsija/test-snippet-findings/tree/main

NOTE 2: As the author doesn't have enough information as to what data is relevant, the UI in this PR is built to expose almost all data the endpoints retrieve. The importance of each field is to be discussed and UI fine-tuned, accordingly.

NOTE 3: The data exposed for the snippets probably shows some promise for using Software Heritage or similar for showing the actual code. This PR is regarded as laying the ground for such feature.


nnobelis:
I will review this when I have the time, today or Friday.
Please do not merge before.

Etsija:
> I will review this when I have the time, today or Friday. Please do not merge before.

Agreed. As discussed in #5023 which will be the corresponding UI implementation, this probably needs to handle provenances properly. I've just had difficulties providing/producing a repository for testing, which would produce snippet findings at all, and especially findings that belong to different provenances, so haven't been able to advance this PR.

nnobelis:
As I mentioned, the multiple provenance thing should happen with git sub modules.
To get snippet findings at all, take an open source project like Semver4j. You should be several hits with the existing forks.

So my suggestion: A test repository containing Semver4j source code with a git sub module containing the code of another open source project.

Etsija:
> As I mentioned, the multiple provenance thing should happen with git sub modules. To get snippet findings at all, take an open source project like Semver4j. You should be several hits with the existing forks.

Already tried: no snippet findings at all with SCANOSS. I have had incredible difficulties to produce any snippet findings with that scanner, in general.

> So my suggestion: A test repository containing Semver4j source code with a git sub module containing the code of another open source project.

Already tried with [ort-test-data-git-submodules](https://github.com/oss-review-toolkit/ort-test-data-git-submodules) but the whole analysis fails on Analyzer already due to duplicate definitions of `isArray`, so that didn't help, either.

I will probably need to employ an agent to produce me an artificial repository which would contain submodules, trigger snippet findings - and in general be analyzable with ORT / ORT Server.

Etsija:
@nnobelis the problem with SCANOSS seems to be [this part](https://github.com/oss-review-toolkit/ort/blob/808223333307501e28f5ef231aec8a4e825b3d12/plugins/scanners/scanoss/src/main/kotlin/ScanOssResultParser.kt#L57-L64) of `ScanOssResultParser` in ORT core: a "complete fork" that copies files from upstream repository file-by-file doesn't produce snippets due to `MatchType.file`. It produces license findings instead.

The only way I have ever been able to get any snippet findings from SCANOSS has been to fork an OSS repository, then "mutilate" some of the files in my fork, to not produce whole-file findings. Seems like with at least SCANOSS that's just the way it is - and we do not have access to FOSSID on our organization, so haven't been able to test that.

nnobelis:
> doesn't produce snippets due to MatchType.file. It produces license findings instead

This is a difference between the implementation of FossId and ScanOSS clients in ORT: when there is a full file match, the FossId client stills creates a snippet finding for it. To have it as a license finding only, you need to identify the snippet in the FossId UI or write a snippet choice for it.

components/snippet-findings/backend/src/main/kotlin/SnippetFindingService.kt
```diff
@@ -0,0 +1,252 @@
...
+}
+
+private class QueryContext(val join: Join)
+
+/**
+ * Build the common query context for snippet findings queries. Joins through the
+ * [ScanResultPackageProvenancesTable] junction table to guarantee correct provenance matching at the
+ * DB level.
+ */
+private fun buildQueryContext(): QueryContext {
```
nnobelis: I don't think the logic here is right. You don't use the provenance in the returned snippets findings (i.e. the content of the table `package_provenances`).

So what you return are all the snippet finding for all provenance, flattened. You are effectively losing information which snippet finding belong to which provenance.
In `scanner_runs_scan_results`, there can be several rows with the same `scanner_run_id`.

I think it must be done as explained in my comment https://github.com/eclipse-apoapsis/ort-server/issues/5023#issuecomment-4406094408:

Provide an endpoint to list all the provenances of a scanner run for an ort run, meaning all `ScannerRun.ScanResult[].provenance`, id included.

Provide a snippet finding endpoint that takes a ort run id and a provenance id and return the corresponding snippets findings.
This corresponds to `ScannerRun.ScanResult[provenanceId].ScanSummary.SnippetFindings`.

Then in the UI, I imagine you first see the provenances. When you click on one, it opens and you see the snippet findings for this provenance.

Of course you can also just return `<Provenance, <List<SnippetFindings>>` from the backend and construct the UI at once, but in this case you have no paging which could be problematic because the amount of snippet findings can be huge.

nnobelis: If you have problem creating a test repository, maybe you can just manipulate the DB data, i.e. create multiple entries in `scanner_runs_scan_results` with the same `scanner_run_id` and different provenance in the scan results.

Etsija: I'm in the process of creating an artificial repository for testing right now.

Etsija: I have now finally produced an artificial repository which has another artificial repository as a submodule. I was able to analyse it and get snippets with different provenances. I'll put this to draft and get to work.

Etsija: Screenshot from 2026-05-15 13-57-08

@nnobelis I'm not sure the provenances endpoint should be this tightly coupled to under snippet findings. What's your opinion?

nnobelis: How about `/runs/{runId}/provenances` ?

Etsija: I was thinking of that, but I didn't know whether any scanner returns the same list of provenances? If yes, then that would probably be good. But what would happen if I run several scanners, like SCANOSS and ScanCode, in the same run? If there's a chance that the list of provenances is different for different scanners in an ORT run, then I guess I'd really need to keep the endpoint `/snippet-findings/provenances`.

nnobelis: This is correct, the UI cannot bluntly return all the provenances from all the `ScanResults`, because there is nothing to present for non-snippet scanners.

But we are talking about the backend here. So you could have a API endpoint that returns you all the provenances and the scanner that processed them. This endpoint could be in `/runs/` and the UI could then filter the results to keep only the snippet-scanned provenances.

But since this is ineficient to fetch stuff you don't need,  maybe `/snippet-findings/provenances` is indeed better.

Etsija:
The PR is finally ready for review.

components/snippet-findings/backend/src/main/kotlin/SnippetFindingService.kt
```diff
@@ -0,0 +1,443 @@
...
+        parameters: ListQueryParameters
+    ): ListQueryResult<SnippetFindingProvenance> = db.blockingQuery {
+        // Two separate queries are needed because sub-repository scan results are not linked via
+        // ScanResultPackageProvenancesTable — they can only be reached through NestedProvenancesTable.
+        val directProvenances = queryProvenances(buildDirectProvenancesJoin(), ortRunId)
+        val nestedProvenances = queryProvenances(buildNestedProvenancesJoin(), ortRunId)
+
+        // Merge, giving direct matches priority (a scan result should not appear on both paths,
+        // but deduplicate by id just to be safe).
+        val merged = LinkedHashMap<Long, SnippetFindingProvenance>()
```
mnonnenmacher: Use `mutableMapOf<Long, SnippetFindingProvenance>()` instead. And give the variable a more speaking name like `mergedProvenances` or `allProvenances`.

components/snippet-findings/backend/src/main/kotlin/SnippetFindingService.kt
```diff
@@ -0,0 +1,443 @@
...
+    ScanResultsTable.vcsUrl,
+    ScanResultsTable.vcsRevision
+)
+
+private fun queryProvenances(join: Join, ortRunId: Long): List<SnippetFindingProvenance> =
+    join.select(provenanceColumns)
+        .where { ScannerJobsTable.ortRunId eq ortRunId }
+        .map { row ->
+            val artifactUrl = row.getOrNull(ScanResultsTable.artifactUrl)
+            val vcsUrl = row.getOrNull(ScanResultsTable.vcsUrl)
```
mnonnenmacher, line 337: These two should be inlined below just like for all other properties.

components/snippet-findings/backend/src/main/kotlin/SnippetFindingService.kt
```diff
@@ -0,0 +1,443 @@
...
+        // Merge, giving direct matches priority (a scan result should not appear on both paths,
+        // but deduplicate by id just to be safe).
+        val merged = LinkedHashMap<Long, SnippetFindingProvenance>()
+        (directProvenances + nestedProvenances).forEach { merged.putIfAbsent(it.id, it) }
+
+        // Attach snippet finding counts so callers can see how many findings each provenance has.
+        val counts = countSnippetFindingsByProvenanceId(merged.keys.toList())
+        var result = merged.values.map { it.copy(snippetFindingCount = counts[it.id] ?: 0L) }
+
+        // Sort in Kotlin because the two query paths cannot share a single ORDER BY clause.
```
mnonnenmacher, line 84: Please add a TODO comment that this needs to be improved, always loading all rows to sort and limit them in memory is very inefficient.

sschuberth:
Just a quick remark on how the provenance is shown based on the screenshots: Can we please get the "open link" icon next to the repository URL like we have in the breadcrumb bar for repositories, and use an abbreviated commit SHA1 in the UI (while still copying the full commit SHA1 to the clipboard, of course)?

components/snippet-findings/backend/src/main/kotlin/SnippetFindingService.kt
```diff
@@ -0,0 +1,443 @@
...
+}
+
+private class QueryContext(val join: Join)
+
+/**
+ * Build the common join for snippet findings queries.
+ *
+ * Goes directly from the snippet finding's scan result to the scanner run via [ScannerRunsScanResultsTable]. This
+ * works for both root-package scan results and nested sub-repository scan results, unlike the old path through
+ * [ScanResultPackageProvenancesTable] which excluded nested scan results.
```
mnonnenmacher: What "old path" is this referring to? If this refers to an earlier iteration of the implementation it should be dropped.

components/snippet-findings/backend/src/main/kotlin/SnippetFindingService.kt
```diff
@@ -0,0 +1,443 @@
...
+    return SnippetFindingsTable
+        .innerJoin(ScanSummariesTable)
+        .join(ScanResultsTable, JoinType.INNER, ScanSummariesTable.id, ScanResultsTable.scanSummaryId)
+        .select(ScanResultsTable.id, count)
+        .where { scanResultIds.map { ScanResultsTable.id eq it }.reduce(Op<Boolean>::or) }
+        .groupBy(ScanResultsTable.id)
+        .associate { it[ScanResultsTable.id].value to it[count] }
+}
+
+private class QueryContext(val join: Join)
```
mnonnenmacher: This class seems redundant, the `Join` could just be passed around directly. It also lacks any documentation to clarify the purpose.

mnonnenmacher (changes requested):

components/snippet-findings/backend/src/main/kotlin/SnippetFindingService.kt
Etsija: They are actually used in two places below (see calculation of `provenanceType`), so are you sure they should be both inlined in both places, as then they need to be computed twice?

Etsija:
> Just a quick remark on how the provenance is shown based on the screenshots: Can we please get the "open link" icon next to the repository URL like we have in the breadcrumb bar for repositories, and use an abbreviated commit SHA1 in the UI (while still copying the full commit SHA1 to the clipboard, of course)?

As discussed, I also changed the upstream snippet provenance URLs into links.

mnonnenmacher (approved):

PR #5001, opened 2026-05-06
Add Git and Bazel credential helpers to the workers' base image

nnobelis:


workers/common/src/main/kotlin/env/BazelGenerator.kt
```diff
@@ -47,7 +47,7 @@ class BazelGenerator : EnvironmentConfigGenerator<BazelDefinition> {
         }
 
         builder.buildInUserHome(BAZEL_RC_FILE_NAME) {
-            println("common --credential_helper=/opt/bazel/bazel_cred_wrapper.sh")
+            println("common --credential_helper=/opt/bazel/bin/bazel-credential-helper")
```
oheger-bosch, line 50: The test has to be adapted accordingly. Also, is such a change necessary for the Git credentials helper, too? I think so. For the Git implementation, also the `useHttpPath` flag has to be set in the configuration. See https://github.com/eclipse-apoapsis/ort-server-credential-helper#git-credential-helper .

oheger-bosch (approved):

PR #5003, opened 2026-05-06
chore(docker): Update Bazelisk to version 1.29

nnobelis:
See [1].

[1]: https://github.com/bazelbuild/bazelisk/releases/tag/v1.29.0


nnobelis:
@mnonnenmacher It would be nice to have it automatically updated by Renovate.

I found https://docs.renovatebot.com/modules/manager/bazelisk/#default-config, but it seems you can only configure the file name.
Can Renovate update any content, even when there is extra tokens `ARG BAZELISK_VERSION=`  like in our case ?

sschuberth:
> chore(docker): Update Bazelisk to version 1.29

I believe the commit type should be `deps` instead of `chore`.

sschuberth:
> Can Renovate update any content, even when there is extra tokens `ARG BAZELISK_VERSION=` like in our case ?

Yes, you can have [regex matchers](https://docs.renovatebot.com/modules/manager/regex/), similar to like [here](https://github.com/eclipse-apoapsis/helm-charts/blob/d6580b05aa595fcf05fa1fe1a75a0b666837b350/renovate.json#L14-L23).

sschuberth (approved):

PR #5013, opened 2026-05-07
ui: Change accordion open/close logic for items with resolutions

### LINKED ISSUE #5010: Rename resolutions accordions and collapse them by default
Especially when viewing ORT run issues, having the resolutions accordion expanded by default  makes the UX a bit slow, as the user has to always scroll down before seeing the issue description.

It is better to collapse the resolutions sections in vulnerability, rule violations, and issue resolutions by default, and also rename them to "Resolve vulnerability", "Resolve rule violation" and "Resolve issue", respectively.

Etsija:
Screenshot from 2026-05-07 11-58-59

Screenshot from 2026-05-07 11-59-33

### DIFF
--- a/ui/src/helpers/resolutions.ts
+++ b/ui/src/helpers/resolutions.ts
@@ -101,6 +101,42 @@ export function hasRuleViolationResolutionActivity(item: RuleViolation) {
   );
 }
 
+function hasResolutionActivity(item: ItemWithResolutions) {
+  if (isVulnerabilityItem(item)) {
+    return hasVulnerabilityResolutionActivity(item);
+  }
+
+  if (isRuleViolationItem(item)) {
+    return hasRuleViolationResolutionActivity(item);
+  }
+
+  if (isIssueItem(item)) {
+    return hasIssueResolutionActivity(item);
+  }
+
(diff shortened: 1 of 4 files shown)


ui/src/helpers/resolutions.ts
```diff
@@ -101,6 +101,42 @@ export function hasRuleViolationResolutionActivity(item: RuleViolation) {
   );
 }
 
+function hasResolutionActivity(item: ItemWithResolutions) {
```
sschuberth, line 104: Commit message:
- It should say "to concentrate _on_ the item description".
- Maybe instead of "to:" say "so that" (no colon at the end), which reads nicer with then following "when"s. But actually, the "when"s should be "if"s 😸 .
- The bullet item sentences lacks periods or commas. Either all of them should have periods, or only the last should have a period and the other ones commas.

sschuberth: As the last two cases result in the same behavior, maybe combine them as "if a resolution is pending or exists, ...".

sschuberth (dismissed):

lamppu (approved):

PR #5020, opened 2026-05-07
ui: Allow to specify environment configuration for each supported package manager

### LINKED ISSUE #3815: Allow to specify environment configuration without `.ort.env.yml`

Etsija:
#### Keep the form tidier by wrapping the whole section into a collapsible

Screenshot from 2026-05-07 18-31-33

#### Opening the collapsible shows each package manager section in its own accordion

Screenshot from 2026-05-07 18-31-55

#### Example of a configuration section for a package manager

Screenshot from 2026-05-07 18-32-28

NOTE1: Field descriptions are taken from back-end code comments in [here](https://github.com/eclipse-apoapsis/ort-server/tree/main/workers/common/src/main/kotlin/env/definition) and certainly need some polishing.

NOTE2: I don't want to turn this description into a slideshow, so any customer of this feature should take the PR and try it out.

NOTE3: @mnonnenmacher already pointed out that the initial assumption of having only one environment configuration per package manager is wrong. The team decided that this PR should really be merged first, anyway, to gain experience (and possible bug reports) from using it, then improve.

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/analyzer-fields.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/analyzer-fields.tsx
@@ -19,7 +19,7 @@
 
 import { useParams } from '@tanstack/react-router';
 import { PlusIcon, TrashIcon } from 'lucide-react';
-import { useEffect, useRef } from 'react';
+import { useEffect } from 'react';
 import { useFieldArray, UseFormReturn } from 'react-hook-form';
 
 import { InfrastructureService } from '@/api';
@@ -30,7 +30,6 @@ import {
   AccordionTrigger,
 } from '@/components/ui/accordion';
 import { Button } from '@/components/ui/button';
-import { Card, CardContent, CardTitle } from '@/components/ui/card';
 import {
   FormControl,
   FormDescription,
@@ -56,13 +55,8 @@ import {
   ProductPermissions,
   RepositoryPermissions,
 } from '@/lib/permissions.ts';
-import {
-  EnvironmentDefinitions,
-  NpmAuthMode,
-  npmAuthModes,
-  npmEnvironmentDefinitions,
-} from '@/lib/types';
 import { CreateRunFormValues } from '@/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components';
+import { EnvironmentDefinitionsFields } from './environment-definitions';
 import { PackageManagerField } from './package-manager-field';
 
 type AnalyzerFieldsProps = {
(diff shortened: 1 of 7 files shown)


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/analyzer-fields.tsx
```diff
@@ -19,7 +19,7 @@
 
 import { useParams } from '@tanstack/react-router';
 import { PlusIcon, TrashIcon } from 'lucide-react';
-import { useEffect, useRef } from 'react';
+import { useEffect } from 'react';
```
mnonnenmacher, line 22: Commit message nit: "concentrates **on**", "**the** frontend"

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/environment-definitions.tsx
```diff
@@ -129,6 +130,127 @@ export const EnvironmentDefinitionsFields = ({
...
+                          field.onChange(checked ? 'true' : 'false')
+                        }
+                      />
+                    </FormControl>
+                  </FormItem>
+                )}
+              />
+            </div>
+          </AccordionContent>
+        </AccordionPrimitive.Item>
```
mnonnenmacher: That is still a lot of repetitive code for each package manager. As there is a limited amount of field types like string, bool, or enum, have you considered defining the missing model for the package manager config in the frontend and use it to generate the form, similar to how it is done for the plugin options? This would reduce the amount of code and make it easier to update if the backend model changes.

For example, this would be the information needed to generate the form for Conan:

```json
"conan": {
  "name": {
    "type": "string",
    "title": "Remote name",
    "description": "Name of the Conan remote, used in commands like <code>conan list</code>"
  },
  "URL": {
    "verifySsl": {
      "type": "string",
      "title": "Verify SSL",
      "description": "Verify the SSL certificate of the remote URL."
    }
  }
}
```

mnonnenmacher (changes requested):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/environment-definitions.tsx
Etsija: I was worried about the amount of repetition, yes, and was playing with the same idea as you have. Let me see how this would be best implemented.

Etsija: Would it be acceptable to have this as a separate "second refactoring commit" on top of the current commit stack, or do you want the changes be amended to the earlier ones (which might be difficult)?

mnonnenmacher: Yes, would be fine with me.

mnonnenmacher (approved):

PR #5024, opened 2026-05-08
ui: Improve plugin templates handling

### LINKED ISSUE #5014: Improve the admin / plugin templates handling
- make it possible to also edit a created plugin template (there's already an endpoint for it). Infrastructure services and secrets on different hierarchy levels give good references for implementation
- remove the "Delete Template" label from the remove icon (for consistency in the UI), and once editing is possible, add an editing icon to the same place
- check if it's possible to also deassign an organization from the template once it is chosen
- creating a template is cumbersome, as initially, all template options are checkmarked as "undefined", which locks the field from editing, and to enable editing, the user has to first uncheck "undefined", which is not obvious. An "undefined" field should still be editable, and once the user starts editing, "undefined" checkmark should simply go to off state. If the user cancels editing by emptying the field, "undefined" checkmark should go back to on state
- boolean template values show weirdly an indeterminate state when undefined. A shadcn-ui Toggle (defaulting to false) would be a better option for boolean values

Image

- both the create template / and the upcoming edit template are missing Cancel button, which should lead back to $pluginType/$pluginId route

Etsija:
This PR includes some visual and usability improvements for the plugin templates handling, most notably new template editing feature for which there already existed an endpoint but no UI yet.
Screenshot from 2026-05-08 10-47-45

Screenshot from 2026-05-08 10-48-31

Screenshot from 2026-05-08 10-48-46

### DIFF
--- a/ui/src/routes/admin/plugins/$pluginType/$pluginId/create-template/index.tsx
+++ b/ui/src/routes/admin/plugins/$pluginType/$pluginId/create-template/index.tsx
@@ -55,6 +55,7 @@ import {
   FormMessage,
 } from '@/components/ui/form';
 import { Input } from '@/components/ui/input';
+import { Switch } from '@/components/ui/switch';
 import { ApiError } from '@/lib/api-error';
 import { toast, toastError } from '@/lib/toast';
 import { getPluginTypeLabel } from '@/lib/types';
@@ -63,7 +64,7 @@ import { Route as LayoutRoute } from '../../../route.tsx';
 function optionTypeToZodType(type: PluginOptionType): ZodType {
   switch (type) {
     case 'BOOLEAN':
-      return z.boolean();
+      return z.boolean().default(false);
     case 'INTEGER':
       return z.coerce.number();
     case 'LONG':
@@ -124,7 +125,8 @@ const CreateTemplate = () => {
       [templateName]: '',
       ...(plugin?.options?.reduce(
         (acc, option) => {
-          acc[option.name] = option.defaultValue ?? '';
+          acc[option.name] =
+            option.type === 'BOOLEAN' ? false : (option.defaultValue ?? '');
           acc[`${option.name}_isFinal`] = false;
           acc[`${option.name}_isNotSet`] = true;
           return acc;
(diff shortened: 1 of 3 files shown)


sschuberth:
Visuals LGTM!

ui/src/routes/admin/plugins/$pluginType/$pluginId/create-template/index.tsx
```diff
@@ -333,7 +348,12 @@ const CreateTemplate = () => {
                         >
                           <Checkbox
                             checked={field.value as CheckedState}
-                            onCheckedChange={field.onChange}
+                            onCheckedChange={(checked) => {
+                              field.onChange(checked);
+                              if (checked) {
+                                form.setValue(option.name, '');
```
mnonnenmacher: I think this should revert to the default value instead because fields are initialized with the default values of the options.

ui/src/routes/admin/plugins/$pluginType/$pluginId/edit-template/$templateName/index.tsx
```diff
@@ -0,0 +1,443 @@
...
+              disabled={isUpdateTemplatePending}
+            >
+              Cancel
+            </Button>
+          </CardFooter>
+        </form>
+      </Form>
+    </Card>
+  );
+};
```
mnonnenmacher, line 431: I think this mostly duplicates the creation form so any future improvements would have to be done in two places. Would it be worth it to extract the form to a shared place or are the differences between create and edit too large?

mnonnenmacher (changes requested):

ui/src/routes/admin/plugins/$pluginType/$pluginId/edit-template/$templateName/index.tsx
Etsija: There are differencies in editable/read-only inputs, form initialization, mutation functions at least - and we are using the same pattern everywhere; separate create/edit forms throughout the UI.

mnonnenmacher (approved):

PR #5037, opened 2026-05-09
ci: Run pnpm from the project dir to fix corepack version detection

mnonnenmacher:
Set the working directory when running pnpm as corepack picks up the version from the `package.json` in the current directory. Otherwise, it might install the wrong version which then causes the version check of pnpm to fail.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -123,13 +123,16 @@ jobs:
       run: corepack enable
 
     - name: Install dependencies
-      run: pnpm -C ui install
+      working-directory: ui
+      run: pnpm install
 
     - name: Build UI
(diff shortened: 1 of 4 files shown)


mnonnenmacher:
We could also consider switching to https://github.com/pnpm/action-setup instead of using corepack, that would still require to define the package.json location to pick up the correct pnpm version, but only once. But if this PR fixes the current issues I would merge it first to unblock other PRs.

sschuberth (approved):

sschuberth:
> We could also consider switching to https://github.com/pnpm/action-setup instead of using corepack

Note that 6e8f3046270a1ef242c338127ef67012bb4b015f did the exact opposite.

PR #5064, opened 2026-05-12
Switch to back-end data manipulation for run vulnerabilities

### LINKED ISSUE #5053: Add filtering to run vulnerabilities endpoint
Add filtering support for the run vulnerabilities endpoint (`rating`, `identifier`, `purl`, `externalId`) `runs/:runId/vulnerabilities`, to align with `organizations/:organizationId/vulnerabilities` and `products/:productId/vulnerabilities`.

### LINKED ISSUE #5054: Change run vulnerabilities table to use server-side data manipulation
Once the run vulnerabilities endpoint supports filtering.

Etsija:
This is to be able to switch the corresponding UI table finally to use server-side data manipulation, as is the target for most of UI tables. The PR prepares to add filtering by advisors to all three vulnerabilities endpoints in the upcoming PRs.

### DIFF
--- a/api/v1/mapping/src/commonMain/kotlin/ApiMappings.kt
+++ b/api/v1/mapping/src/commonMain/kotlin/ApiMappings.kt
@@ -851,7 +851,15 @@ fun ApiPackageFilters.mapToModel(): PackageFilters =
 
 fun ApiRuleViolationFilters.mapToModel(): RuleViolationFilters = RuleViolationFilters(resolved = resolved)
 
-fun ApiVulnerabilityFilters.mapToModel(): VulnerabilityFilters = VulnerabilityFilters(resolved = resolved)
+fun ApiVulnerabilityFilters.mapToModel(): VulnerabilityFilters = VulnerabilityFilters(
+    resolved = resolved,
+    rating = rating?.mapToModel { ratingSet ->
+        ratingSet.mapTo(mutableSetOf()) { it.mapToModel() }
+    },
+    identifier = identifier?.mapToModel { it },
+    purl = purl?.mapToModel { it },
+    externalId = externalId?.mapToModel { it }
+)
 
 fun Project.mapToApi() = ApiProject(
     identifier = identifier.mapToApi(),
--- a/core/src/main/kotlin/api/RunsRoute.kt
+++ b/core/src/main/kotlin/api/RunsRoute.kt
@@ -225,7 +225,7 @@ fun Route.runs() = route("runs") {
 
         route("vulnerabilities") {
             get(getRunVulnerabilities, requireRunPermission()) {
-                val pagingOptions = call.pagingOptions(SortProperty("externalId", SortDirection.ASCENDING))
+                val pagingOptions = call.pagingOptions(SortProperty("rating", SortDirection.DESCENDING))
                 val filters = call.vulnerabilityFilters()
 
                 val vulnerabilitiesForOrtRun =
--- a/core/src/main/kotlin/api/RunsRoute.kt
+++ b/core/src/main/kotlin/api/RunsRoute.kt
@@ -48,6 +48,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.OrtRunStatus
 import org.eclipse.apoapsis.ortserver.api.v1.model.PackageFilters
 import org.eclipse.apoapsis.ortserver.api.v1.model.RuleViolationFilters
 import org.eclipse.apoapsis.ortserver.api.v1.model.VulnerabilityFilters
+import org.eclipse.apoapsis.ortserver.api.v1.model.VulnerabilityRating
 import org.eclipse.apoapsis.ortserver.components.authorization.rights.EffectiveRole
 import org.eclipse.apoapsis.ortserver.components.authorization.rights.HierarchyPermissions
 import org.eclipse.apoapsis.ortserver.components.authorization.rights.RepositoryPermission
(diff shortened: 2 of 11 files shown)


api/v1/mapping/src/commonMain/kotlin/ApiMappings.kt
```diff
@@ -851,7 +851,15 @@ fun ApiPackageFilters.mapToModel(): PackageFilters =
 
 fun ApiRuleViolationFilters.mapToModel(): RuleViolationFilters = RuleViolationFilters(resolved = resolved)
 
-fun ApiVulnerabilityFilters.mapToModel(): VulnerabilityFilters = VulnerabilityFilters(resolved = resolved)
+fun ApiVulnerabilityFilters.mapToModel(): VulnerabilityFilters = VulnerabilityFilters(
+    resolved = resolved,
+    rating = rating?.mapToModel { ratingSet ->
+        ratingSet.map { it.mapToModel() }.toSet()
```
sschuberth: This should directly map to a set for efficiency:

    ratingSet.mapTo(mutableSetOf()) { it.mapToModel() }

services/ort-run/src/main/kotlin/VulnerabilityService.kt
```diff
@@ -380,7 +338,7 @@ class VulnerabilityService(
                 ].orEmpty(),
                 unappliedResolutions = unappliedResolutions,
                 advisor = AdvisorDetails(name = pagedRow.advisorName),
-                purl = purlByIdentifierId[pagedRow.identifierId] ?: identifier.toFallbackPurl()
+                purl = pagedRow.purl.ifBlank { identifier.toFallbackPurl() }
```
sschuberth: Nit: Maybe `ifEmpty` is enough? That expresses better that we never expect a PURL to be blank but non-empty.

core/src/main/kotlin/api/RunsRoute.kt
```diff
@@ -549,7 +550,18 @@ private fun ApplicationCall.issueFilters() =
  */
 private fun ApplicationCall.vulnerabilityFilters() =
     VulnerabilityFilters(
-        resolved = parameters["resolved"]?.lowercase()?.toBooleanStrictOrNull()
+        resolved = parameters["resolved"]?.lowercase()?.toBooleanStrictOrNull(),
+        rating = parameters["rating"]?.let {
+            val parts = it.split(',')
```
sschuberth: I just filed https://github.com/eclipse-apoapsis/ort-server/pull/5065, and I believe we should do it similarly here:

```kotlin
rating = parameters["rating"]?.let { rating ->
    val (operators, ratings) = rating.split(',').partition { it == "-" }
    val enums = ratings.mapTo(mutableSetOf()) { findByName<VulnerabilityRating>(it) }
    FilterOperatorAndValue(if (operators.isEmpty()) ComparisonOperator.IN else ComparisonOperator.NOT_IN, enums)
},
```

core/src/main/kotlin/apiDocs/RunsDocs.kt
```diff
@@ -296,7 +298,25 @@ val getRunVulnerabilities: RouteConfig.() -> Unit = {
                 """.trimIndent()
         }
 
-        standardListQueryParameters()
+        queryParameter<String>("rating") {
+            description = "Defines the ratings to filter the results by. This is a comma-separated string with the" +
```
sschuberth: Nit: We usually put trailing spaces at the end of the previous line instead of the start of the next line for multi-line strings.

core/src/main/kotlin/api/RunsRoute.kt
```diff
@@ -224,7 +225,7 @@ fun Route.runs() = route("runs") {
 
         route("vulnerabilities") {
             get(getRunVulnerabilities, requireRunPermission()) {
-                val pagingOptions = call.pagingOptions(SortProperty("externalId", SortDirection.ASCENDING))
+                val pagingOptions = call.pagingOptions(SortProperty("rating", SortDirection.DESCENDING))
```
sschuberth, line 228: I believe this change of default sorting should be factored out to a separate commit, because it makes sense without the other changes in this commit, and justifies a separate rationale (to explain why it makes more sense to sort by `rating` than by `externalId`).

sschuberth (dismissed):

sschuberth:
@lamppu, any remarks on the code?

services/ort-run/src/main/kotlin/OrtRunVulnerabilityQueryBuilder.kt
```diff
@@ -0,0 +1,334 @@
...
+                    )
+                    .Else(stringLiteral(":"))
+                val concatenatedIdentifier = concat(
+                    IdentifiersTable.type,
+                    stringLiteral(":"),
+                    namespaceWithSlash,
+                    IdentifiersTable.name,
+                    stringLiteral("@"),
+                    IdentifiersTable.version
+                )
```
lamppu: This seems to still use the old formatting, I think it should now just be like this (to match an identifier search string like "Maven:junit:junit:4.12" for instance):
```
                val concatenatedIdentifier = concat(
                    IdentifiersTable.type,
                    stringLiteral(":"),
                    IdentifiersTable.namespace,
                    stringLiteral(":"),
                    IdentifiersTable.name,
                    stringLiteral(":"),
                    IdentifiersTable.version
                )
```

Etsija: Thanks, this was a really good catch! I made a separate fix commit as the last one in the stack, which fixes all vulnerability endpoints with the same fix. Tested it - now the search works as intended on all three endpoints.

lamppu (approved):

PR #5066, opened 2026-05-12
fix(scanner): Return issues from scan summaries

oheger-bosch:
When scan results were loaded from `OrtServerScanResultStorage` issues
from the scan summary were dropped. Fix this by loading these issues and
adding them to the internal `issuesMap` which is then queried by the
scanner runner. That way, issues from newly executed scans and those
loaded from the database can be treated in the same way.


sschuberth:
@lamppu, could you please have a look at this as well?

lamppu:
> @lamppu, could you please have a look at this as well?

LGTM

lamppu (approved):

PR #5074, opened 2026-05-13
Advisor endpoints

### LINKED ISSUE #5055: Add endpoints to return all used advisors
Add three separate advisors endpoints, so each endpoint already knows about the scope: `organizations/{orgId}/vulnerabilities/advisors`, `products/{productId}/vulnerabilities/advisors`, `runs/{runId}/vulnerabilities/advisors`.

Hints for implementation:
- `runs/{runId}/vulnerabilities/advisors` just has to check which advisors were activated in the run request, which should be super fast lookup
- `products/{productId}/vulnerabilities/advisors` would have to do the same lookup but for the last run of each repository in the product (last run has the highest `runIndex`)
- `organizations/{orgId}/vulnerabilities/advisors`: same lookup, but as it's the highest hierarchy, it will need to loop through products and their repositories to search for the last runs

Etsija:
This PR adds three new endpoints:
- get a list of advisors used for vulnerabilities in an ORT run
- get a list of advisors used for product vulnerabilities (from the latest runs of all product repositories)
- get a list of advisors used for organization vulnerabilities (from the latest runs of all repositories belonging to products in the organization)

This collection of endpoints enables possible tunings to this "latest run" logic later, if needed.

### DIFF
--- a/core/src/main/kotlin/api/OrganizationsRoute.kt
+++ b/core/src/main/kotlin/api/OrganizationsRoute.kt
@@ -58,6 +58,7 @@ import org.eclipse.apoapsis.ortserver.core.apiDocs.getOrganizationProducts
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getOrganizationRunStatistics
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getOrganizationUsers
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getOrganizationVulnerabilities
+import org.eclipse.apoapsis.ortserver.core.apiDocs.getOrganizationVulnerabilityAdvisors
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getOrganizations
 import org.eclipse.apoapsis.ortserver.core.apiDocs.patchOrganization
 import org.eclipse.apoapsis.ortserver.core.apiDocs.postOrganization
@@ -267,6 +268,18 @@ fun Route.organizations() = route("organizations") {
                     .toSearchResponse(filters)
 
                 call.respond(HttpStatusCode.OK, pagedSearchResponse)
+            }
+
+            route("advisors") {
+                get(getOrganizationVulnerabilityAdvisors, requirePermission(OrganizationPermission.READ)) {
+                    val organizationId = call.requireIdParameter("organizationId")
+                    val repositoryIds = organizationService.getRepositoryIdsForOrganization(organizationId)
+                    val ortRunIds = repositoryIds.mapNotNull { repositoryId ->
+                        repositoryService.getLatestOrtRunIdWithSuccessfulAdvisorJob(repositoryId)
+                    }
+                    val advisors = vulnerabilityService.getAdvisorsForOrtRunIds(ortRunIds)
+                    call.respond(HttpStatusCode.OK, advisors.toList())
+                }
             }
         }
 
--- a/core/src/main/kotlin/api/ProductsRoute.kt
+++ b/core/src/main/kotlin/api/ProductsRoute.kt
@@ -56,6 +56,7 @@ import org.eclipse.apoapsis.ortserver.core.apiDocs.getProductRepositories
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getProductRunStatistics
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getProductUsers
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getProductVulnerabilities
+import org.eclipse.apoapsis.ortserver.core.apiDocs.getProductVulnerabilityAdvisors
 import org.eclipse.apoapsis.ortserver.core.apiDocs.patchProduct
 import org.eclipse.apoapsis.ortserver.core.apiDocs.postProductRuns
 import org.eclipse.apoapsis.ortserver.core.apiDocs.postRepository
@@ -231,6 +232,18 @@ fun Route.products() = route("products/{productId}") {
                 .toSearchResponse(filters)
 
             call.respond(HttpStatusCode.OK, pagedSearchResponse)
+        }
+
+        route("advisors") {
+            get(getProductVulnerabilityAdvisors, requirePermission(ProductPermission.READ)) {
+                val productId = call.requireIdParameter("productId")
+                val repositoryIds = productService.getRepositoryIdsForProduct(productId)
+                val ortRunIds = repositoryIds.mapNotNull { repositoryId ->
+                    repositoryService.getLatestOrtRunIdWithSuccessfulAdvisorJob(repositoryId)
+                }
+                val advisors = vulnerabilityService.getAdvisorsForOrtRunIds(ortRunIds)
+                call.respond(HttpStatusCode.OK, advisors.toList())
+            }
         }
     }
 
--- a/core/src/main/kotlin/api/RunsRoute.kt
+++ b/core/src/main/kotlin/api/RunsRoute.kt
@@ -69,6 +69,7 @@ import org.eclipse.apoapsis.ortserver.core.apiDocs.getRunReport
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getRunRuleViolations
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getRunStatistics
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getRunVulnerabilities
+import org.eclipse.apoapsis.ortserver.core.apiDocs.getRunVulnerabilityAdvisors
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getRuns
 import org.eclipse.apoapsis.ortserver.core.utils.findByName
 import org.eclipse.apoapsis.ortserver.logaccess.LogFileService
@@ -238,6 +239,13 @@ fun Route.runs() = route("runs") {
                 val pagedResponse = vulnerabilitiesForOrtRun.mapToApi { it }
 
                 call.respond(HttpStatusCode.OK, pagedResponse)
+            }
+
+            route("advisors") {
+                get(getRunVulnerabilityAdvisors, requireRunPermission()) {
+                    val advisors = vulnerabilityService.getAdvisorsForOrtRunId(call.ortRun.id)
+                    call.respond(HttpStatusCode.OK, advisors.toList())
+                }
             }
         }
 
--- a/core/src/main/kotlin/apiDocs/ProductsDocs.kt
+++ b/core/src/main/kotlin/apiDocs/ProductsDocs.kt
@@ -369,6 +369,31 @@ val getProductVulnerabilities: RouteConfig.() -> Unit = {
     }
 }
 
+val getProductVulnerabilityAdvisors: RouteConfig.() -> Unit = {
+    operationId = "getProductVulnerabilityAdvisors"
+    summary = "Get the advisors used for vulnerability checks in a product"
+    description = "Get the distinct advisor names from the latest successful advisor runs across the repositories " +
+            "in a product."
+    tags = listOf("Products")
+
+    request {
+        pathParameter<Long>("productId") {
+            description = "The product's ID."
+        }
+    }
+
+    response {
+        HttpStatusCode.OK to {
+            description = "Success. The advisors are sorted by name."
+            jsonBody<List<String>> {
+                example("Get vulnerability advisors for a product") {
+                    value = listOf("NexusIQ", "OSV", "VulnerableCode")
+                }
+            }
+        }
+    }
+}
+
 val getProductRunStatistics: RouteConfig.() -> Unit = {
     operationId = "getProductRunStatistics"
     summary = "Get statistics about ORT runs across the repositories of a product"
--- a/core/src/test/kotlin/api/ProductsRouteIntegrationTest.kt
+++ b/core/src/test/kotlin/api/ProductsRouteIntegrationTest.kt
@@ -24,6 +24,7 @@ import io.kotest.data.forAll
 import io.kotest.data.row
 import io.kotest.inspectors.forAll
 import io.kotest.matchers.collections.beEmpty
+import io.kotest.matchers.collections.containExactly
 import io.kotest.matchers.collections.shouldContain
 import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
 import io.kotest.matchers.collections.shouldHaveSize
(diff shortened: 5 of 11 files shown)


services/ort-run/src/main/kotlin/VulnerabilityService.kt
```diff
@@ -544,6 +544,22 @@ class VulnerabilityService(
 
             CountByCategory(ratingToCountMap)
         }
+
+    suspend fun getAdvisorsForOrtRunId(runId: Long): List<String> =
+        getAdvisorsForOrtRunIds(listOf(runId))
```
sschuberth: This overload seems to be exclusively used for testing, and only in 3 places. So I'd not make it part of the production code, but let test code call the below function.

services/ort-run/src/main/kotlin/VulnerabilityService.kt
```diff
@@ -544,6 +544,22 @@ class VulnerabilityService(
...
+    suspend fun getAdvisorsForOrtRunIds(runIds: List<Long>): List<String> = db.dbQuery {
+        if (runIds.isEmpty()) return@dbQuery emptyList()
+        AdvisorJobsTable
+            .select(AdvisorJobsTable.configuration)
+            .where { AdvisorJobsTable.ortRunId inList runIds }
+            .flatMap { row ->
+                val cfg = row[AdvisorJobsTable.configuration]
+                cfg.advisors + cfg.config?.keys.orEmpty()
+            }
+            .distinct()
```
sschuberth: `distinct()` converts to a set and back to a list. We could avoid that (and also the separate sorting below) by flat-mapping straight to a sorted set, and also changing the function's return value to `Set<String>`:

```kotlin
    suspend fun getAdvisorsForOrtRunIds(runIds: List<Long>): Set<String> = db.dbQuery {
        if (runIds.isEmpty()) return@dbQuery emptySet()
        AdvisorJobsTable
            .select(AdvisorJobsTable.configuration)
            .where { AdvisorJobsTable.ortRunId inList runIds }
            .flatMapTo(sortedSetOf()) { row ->
                val cfg = row[AdvisorJobsTable.configuration]
                cfg.advisors + cfg.config?.keys.orEmpty()
            }
    }
```

core/src/main/kotlin/apiDocs/ProductsDocs.kt
```diff
@@ -369,6 +369,31 @@ val getProductVulnerabilities: RouteConfig.() -> Unit = {
...
+            description = "The product's ID."
+        }
+    }
+
+    response {
+        HttpStatusCode.OK to {
+            description = "Success."
+            jsonBody<List<String>> {
+                example("Get vulnerability advisors for a product") {
+                    value = listOf("NexusIQ", "OSV", "VulnerableCode")
```
sschuberth, line 390: BTW, NexusIQ was removed a long time ago as a vulnerability provider. I'll probably make a clean-up PR for the whole code base.

services/ort-run/src/main/kotlin/VulnerabilityService.kt
Etsija: No it is definitely used in RunsRoute:
```kotlin
val advisors = vulnerabilityService.getAdvisorsForOrtRunId(call.ortRun.id)
```

sschuberth: Ok, I missed that one. Still not sure whether it's worth having that overload. But now that it's there, let's keep it, instead of spending the time to remove it.

services/ort-run/src/main/kotlin/VulnerabilityService.kt
Etsija: I agree about avoiding `distinct()` and the extra `sorted()` pass, but I'd like to keep the public return type as `List<String>`, though, because the endpoints expose an ordered JSON array and the API docs are already modeled as `List<String>`. If you specifically want to have `Set<String>` as the return type, then this will probably have to be changed in the following commits as well. Not saying it's hard to do, just making a note here.

Etsija: So it would be something like
```kotlin
  suspend fun getAdvisorsForOrtRunIds(runIds: List<Long>): List<String> = db.dbQuery {
      if (runIds.isEmpty()) return@dbQuery emptyList()

      AdvisorJobsTable
          .select(AdvisorJobsTable.configuration)
          .where { AdvisorJobsTable.ortRunId inList runIds }
          .flatMapTo(sortedSetOf()) { row ->
              val cfg = row[AdvisorJobsTable.configuration]
              cfg.advisors + cfg.config?.keys.orEmpty()
          }
          .toList()
  }

```

sschuberth: > If you specifically want to have `Set<String>` as the return type

I like that better as it semantically expresses that you're not supposed to expect any duplicates. But I'd be curious about @mnonnenmacher's opinion here, as he's the one with whom you've discussed these endpoints, I believe.

Etsija: Or, I can just use `Set` and change the endpoints, accordingly, to speed this up. We did not discuss in this much detail with @mnonnenmacher.

sschuberth: As you prefer.

Etsija: Done.

core/src/main/kotlin/apiDocs/ProductsDocs.kt
Etsija: Was this a thing you wanted me to change now? Is it blocking the PR?

sschuberth: No and no.

services/ort-run/src/test/kotlin/VulnerabilityServiceTest.kt
```diff
@@ -2571,6 +2571,85 @@ class VulnerabilityServiceTest : WordSpec() {
...
+        "getAdvisorsForOrtRunId" should {
+            "return advisor names from the advisors list" {
+                val repo = fixtures.createRepository()
+                val ortRun = fixtures.createOrtRun(repositoryId = repo.id)
+                fixtures.createAdvisorJob(
+                    ortRunId = ortRun.id,
+                    configuration = AdvisorJobConfiguration(advisors = listOf("OSV"))
+                )
+
+                service.getAdvisorsForOrtRunId(ortRun.id) shouldBe setOf("OSV")
```
mnonnenmacher: Having more than one advisor would make this test more meaningful. And use `should containExactlyInOrder` instead (here and in all tests below).

services/ort-run/src/test/kotlin/VulnerabilityServiceTest.kt
```diff
@@ -2571,6 +2571,85 @@ class VulnerabilityServiceTest : WordSpec() {
...
+                        config = mapOf(
+                            "VulnerableCode" to ResolvablePluginConfig(
+                                options = emptyMap(),
+                                secrets = emptyMap()
+                            )
+                        )
+                    )
+                )
+
+                service.getAdvisorsForOrtRunId(ortRun.id) shouldBe setOf("VulnerableCode")
```
mnonnenmacher: This is actually wrong, the `AdvisorRunner` only uses the list of advisors to enable the plugins, config for plugins that are not in the list is ignored.

services/ort-run/src/main/kotlin/VulnerabilityService.kt
```diff
@@ -544,6 +544,20 @@ class VulnerabilityService(
 
             CountByCategory(ratingToCountMap)
         }
+
```
mnonnenmacher, line 549: Commit message is missing a rationale.

core/src/main/kotlin/api/RunsRoute.kt
```diff
@@ -69,6 +69,7 @@ import org.eclipse.apoapsis.ortserver.core.apiDocs.getRunReport
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getRunRuleViolations
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getRunStatistics
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getRunVulnerabilities
+import org.eclipse.apoapsis.ortserver.core.apiDocs.getRunVulnerabilityAdvisors
```
mnonnenmacher, line 72: Commit message is missing a rationale. It should at least shortly explain what the change does, like "Add an endpoint to [do something]", and ideally also explain why, like "which will be used to [do something]".

core/src/main/kotlin/api/ProductsRoute.kt
```diff
@@ -56,6 +56,7 @@ import org.eclipse.apoapsis.ortserver.core.apiDocs.getProductRepositories
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getProductRunStatistics
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getProductUsers
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getProductVulnerabilities
+import org.eclipse.apoapsis.ortserver.core.apiDocs.getProductVulnerabilityAdvisors
```
mnonnenmacher, line 59: Commit message only talks about the tests but is not giving a rationale for the change itself. Also, the rationale for using `AdvisorJobConfiguration` is not needed at all IMO as this is how the fixtures API is supposed to be used.

core/src/test/kotlin/api/ProductsRouteIntegrationTest.kt
```diff
@@ -91,6 +91,7 @@ import org.eclipse.apoapsis.ortserver.components.pluginmanager.PluginService
 import org.eclipse.apoapsis.ortserver.components.pluginmanager.PluginType
 import org.eclipse.apoapsis.ortserver.core.SUPERUSER
 import org.eclipse.apoapsis.ortserver.core.TEST_USER
+import org.eclipse.apoapsis.ortserver.model.AdvisorJobConfiguration as DomainAdvisorJobConfiguration
```
mnonnenmacher: We usually use `Model` instead of `Domain` as prefix in such cases.

core/src/main/kotlin/api/OrganizationsRoute.kt
```diff
@@ -58,6 +58,7 @@ import org.eclipse.apoapsis.ortserver.core.apiDocs.getOrganizationProducts
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getOrganizationRunStatistics
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getOrganizationUsers
 import org.eclipse.apoapsis.ortserver.core.apiDocs.getOrganizationVulnerabilities
+import org.eclipse.apoapsis.ortserver.core.apiDocs.getOrganizationVulnerabilityAdvisors
```
mnonnenmacher, line 61: Commit message is missing a rationale.

mnonnenmacher (changes requested):

services/ort-run/src/main/kotlin/VulnerabilityService.kt
mnonnenmacher: IMO if sorting doesn't matter `Set` should be used, otherwise `SortedSet` (if feasible).

services/ort-run/src/test/kotlin/VulnerabilityServiceTest.kt
Etsija: There is no `containExactlyInOrder` in the codebase. I searched for it using the IntelliJ's double-shift and found no usages. Also IntelliJ's autocomplete doesn't suggest it. There is `containExactlyInAnyOrder` but that would just weaken the test, wouldn't it?

sschuberth: > There is `containExactlyInAnyOrder` but that would just weaken the test, wouldn't it?

That's exactly the point: Esp. if we decided to use a regular / unsorted set, we'd not care about the ordering, and the test should not break if the order does not match the expected one.

Etsija: I'm sorry but I don't understand your rationale: if I change the test to use `should containExactlyInAnyOrder`, then a result of `"VulnerableCode,OSV"` will pass the test, although the result is not in alphabetical order, no? What's the point, then?

Etsija: More specifically: is the current `shouldBe setOf` a worse assertion than `should containExactlyInAnyOrder`?

sschuberth: > although the result is not in alphabetical order, no? What's the point, then?

That's why I wrote "f we decided to use a regular / unsorted set". Of course, if we [decide](https://github.com/eclipse-apoapsis/ort-server/pull/5074#discussion_r3241003285) to stick with `SortedSet` instead of `Set`, ordering _should_ be asserted. But as least to me it's still an open question whether the API should already do the sorting, or whether the front-end code should do that. Do we have other API endpoints that already return data strictly sorted?

> More specifically: is the current `shouldBe setOf` a worse assertion than `should containExactlyInAnyOrder`?

Yes, it's a code smell, as `containExactlyInAnyOrder` more clearly expresses the intention.

Etsija: I don't see a reason for not to sort it alphabetically already in back-end. Why put the burden to the UI, if it's really easy to do in the service? We're talking about a list of a few strings only. I think this is becoming a bit academic now; the sorting is working, it is tested (test was extended for that), the sorted lists are proven to work in the UI (with my upcoming PR which depends on this being merged first).

@mnonnenmacher said to use `SortedSet` "if feasible". Well, should I once again change the service and all corresponding endpoints to use `SortedSet` instead of `Set`? I can, if we want to.

sschuberth: > Why put the burden to the UI, if it's really easy to do in the service?

Because you could argue that sorting is a purely visual thing, and no other API consumers than a UI would want it.

We've made the mistake in the pat in ORT to sort data for in-memory representations, which resulted in massive performance hints, instead of sorting the data only once on serialization to a results file (the "visual" representation of the data). That's why I'm being cautious here.

> Well, should I once again change the service and all corresponding endpoints to use `SortedSet` instead of `Set`?

Not from my point of view, but again I'd like @mnonnenmacher to comment. In any case, we should take the time to have these "academic" discussions instead of rushing things in.

mnonnenmacher: @Etsija As you already use `flatMapTo(sortedSetOf())` I would say yes, the functions should return a `SortedSet` (and the function docs could be updated as well).

mnonnenmacher: Also, the tests should then validate the order as well by using `containsExactly` instead of `containsExactlyInAnyOrder`.

> There is no containExactlyInOrder in the codebase.

That was a typo, I meant `containExactlyInAnyOrder`.

> is the current shouldBe setOf a worse assertion than should containExactlyInAnyOrder?

The `contain...` matchers also provide better error messages than `shouldBe` if they fail.

Etsija: I tried to remove the `.toList()` functions from the API, but all corresponding tests failed after that with 500, so they're in.

mnonnenmacher (approved):

services/ort-run/src/test/kotlin/VulnerabilityServiceTest.kt
sschuberth: I'm proposing to address that by just generalizing the function return types from `SortedSet` to `Set`, see https://github.com/eclipse-apoapsis/ort-server/pull/5109.

PR #5076, opened 2026-05-13
Re-enable publishing of the config worker artifact

nnobelis:

### DIFF
--- a/buildSrc/src/main/kotlin/ort-server-kotlin-jvm-application-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/ort-server-kotlin-jvm-application-conventions.gradle.kts
@@ -28,7 +28,11 @@ plugins {
 }
 
 dependencies {
-    implementation(enforcedPlatform(libsCatalog.findLibrary("kotlinBom").get()))
+    implementation(platform(libsCatalog.findLibrary("kotlinBom").get())) {
+        version {
+            require(libsCatalog.findVersion("kotlinPlugin").get().requiredVersion)
+        }
+    }
 }
 
 tinyJib {
--- a/workers/config/build.gradle.kts
+++ b/workers/config/build.gradle.kts
@@ -28,6 +28,8 @@ plugins {
 
     // Apply precompiled plugins.
     id("ort-server-kotlin-jvm-application-conventions")
+
+    id("ort-server-publication-conventions")
 }
 
 group = "org.eclipse.apoapsis.ortserver.workers"


buildSrc/src/main/kotlin/ort-server-kotlin-jvm-application-conventions.gradle.kts
```diff
@@ -26,7 +26,7 @@ plugins {
 }
 
 dependencies {
-    implementation(enforcedPlatform(libs.kotlinBom))
+    implementation(platform(libs.kotlinBom))
```
sschuberth: Commit message: I was confused by the reference to the change in ORT. How does it related to this?

sschuberth: Also in general, I'm not sure this is what we want: Because we _do_ want to enforce (not just recommend) Kotlin versions in _applications_, or? What is the actual error you're seeing? Could you please add a link to the commit message?

sschuberth: Ah, it becomes clearer with the second commit. I propose to squash them.

sschuberth: So why not do it like in ORT and keep `enforcedPlatform` but suppress the bogus error?

nnobelis: @mnonnenmacher recommended it as it prevents the Gradle warning.

sschuberth: Let's wait for him to be back from sick leave to comment.

mnonnenmacher: My motivation for checking for alternatives to suppressing the `enforcedPlatform` error to be able to publish application modules is this section from the [Gradle docs](https://docs.gradle.org/current/userguide/platforms.html#sec:enforced-platform):

> When using enforcedPlatform, exercise caution if your software component is intended for consumption by others. This declaration is transitive and affects the dependency graph of your consumers. If they disagree with any enforced versions, they’ll need to use exclude. Instead, if your reusable component strongly favors specific third-party dependency versions, consider using a [rich version](https://docs.gradle.org/current/userguide/dependency_versions.html#sec:strict-version) declaration with strictly.

So in ORT the only reason I know of to publish the CLI module to Maven Central again was it's use in a rather uncommon test setup in the [package manager plugin example project](https://github.com/oss-review-toolkit/ort-package-manager-plugin/). But here the situation is a bit different because you want to use the `ConfigValidator` from the config worker application module as a dependency in a test project.

`enforcedPlatform` was introduced because Jib copied the wrong version of the Kotlin stdlib to the libs folder in the generated Docker images, as this happened randomly, we probably cannot be sure that this cannot happen again when using `platform`, even if it works now. Based on the docs snippet I think that using the suggested rich version with `require` to set the minimum Kotlin version might be a better solution than switching to `platform`.

However, I also wouldn't be against suppressing the error if the alternatives are too complicated. In any case I still think the long-term goal should be #2211 which would require to decouple the `ConfigValidator` from the worker context god object anyway which I believe is the main problem with moving it to a library module.

nnobelis: So IIUC @mnonnenmacher, we should try something like this:

```
dependencies {
    implementation(libs.kotlinBom) {
        version {
            require( libs.versions.kotlinPlugin.get() )
        }
    }
}
```

nnobelis: ping @mnonnenmacher  (my last comment was a question :smile: )

mnonnenmacher: Sorry, I missed that. Yes, that was the suggestion because it seems like the cleaner approach. IIRC the problem was caused by okhttp depending on an old version of the kotlin-stdlib. If you replace `enforcePlatform` with your code above you could check with `./gradlew :core:dependencies` if the okhttp->kotlin-stdlib is upgraded to the correct version.

But like I said:
> I also wouldn't be against suppressing the error if the alternatives are too complicated.

nnobelis: So I tested the Gradle command and `okhttp` has direct dependency to `kotlin-stdlib` at 2.3.21 (the correct version).
In some cases, I see this:
```
|    +--- io.ktor:ktor-client-okhttp:3.4.3 (*)
|    +--- io.ktor:ktor-serialization-kotlinx-json:3.4.3
|    |    \--- io.ktor:ktor-serialization-kotlinx-json-jvm:3.4.3
|    |         +--- org.jetbrains.kotlinx:kotlinx-serialization-json-io:1.9.0 -> 1.11.0
|    |         |    \--- org.jetbrains.kotlinx:kotlinx-serialization-json-io-jvm:1.11.0
|    |         |         +--- org.jetbrains.kotlinx:kotlinx-serialization-bom:1.11.0 (*)
|    |         |         +--- org.jetbrains.kotlin:kotlin-stdlib:2.3.20 -> 2.3.21 (*)
|    |         \--- org.jetbrains.kotlin:kotlin-stdlib:2.3.0 -> 2.3.21 (*)
```

So it seems working as intended ?

mnonnenmacher: Looks good to me, to be on the safe side you could build the Docker images for this branch and make a test deployment.

nnobelis: I did a local deployment with the compose setup and I triggered and completed a run without issue.

mnonnenmacher (approved):

PR #5087, opened 2026-05-14
test(keycloak): Share a single Keycloak test container to speed up tests

mnonnenmacher:
Creating a new instance of the Keycloak test container for every test class that uses the `KeycloakTestExtension` takes a lot of time and slows down running the tests on CI significantly. As the extension is managing the creation and deletion of the test realm and tests are not executed in parallel, reuse the same container across all tests.

### DIFF
--- a/clients/keycloak/src/testFixtures/kotlin/KeycloakTestExtension.kt
+++ b/clients/keycloak/src/testFixtures/kotlin/KeycloakTestExtension.kt
@@ -25,19 +25,18 @@ import io.kotest.core.extensions.MountableExtension
 import io.kotest.core.listeners.AfterEachListener
 import io.kotest.core.listeners.AfterSpecListener
 import io.kotest.core.listeners.BeforeEachListener
+import io.kotest.core.listeners.BeforeSpecListener
 import io.kotest.core.spec.Spec
 import io.kotest.core.test.TestCase
 import io.kotest.engine.test.TestResult
-
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.withContext
 
 import org.keycloak.admin.client.Keycloak
 import org.keycloak.representations.idm.RealmRepresentation
 
 /**
  * A test extension for integration tests that need access to Keycloak. The extension sets up a
- * [Keycloak test container][KeycloakContainer] with the provided [realm] which defaults to [testRealm].
+ * [shared Keycloak test container][SharedKeycloakTestContainer] with the provided [realm] which defaults to
+ * [testRealm].
  *
  * By default, the extension creates the realm once per spec. If [createRealmPerTest] is set to `true` the realm is
  * created once per test. This provides better isolation of tests but also extends the duration of each test case by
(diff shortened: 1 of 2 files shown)


mnonnenmacher:
Compared to previous runs this change reduces the duration of test jobs:
* authorization-test: 15m -> 11m
* integration-test: 24m -> 18m

sschuberth:
> tests are not executed in parallel

Just to clarify what this means: AFAIK, tests within a single spec (in Kotest-speak) are _not_ executed in parallel. But specs _are_ executed in parallel, which means that tests from different specs also are executed in parallel. Do we agree here?

mnonnenmacher:
> > tests are not executed in parallel
> 
> Just to clarify what this means: AFAIK, tests within a single spec (in Kotest-speak) are _not_ executed in parallel. But specs _are_ executed in parallel, which means that tests from different specs also are executed in parallel. Do we agree here?

No, we don't set the `SpecExecutionMode` anywhere so the default `Sequential` is used: https://kotest.io/docs/framework/concurrency6.html#spec-concurrency-mode

sschuberth:
> so the default `Sequential` is used

Oh, that's somewhat unexpected on my end. I was pretty sure to have seen parallel test output on the console when running tests locally. Well 🤷🏻‍♂️

sschuberth (approved):

sschuberth:
> I was pretty sure to have seen parallel test output on the console

I think that's because: We do set

https://github.com/eclipse-apoapsis/ort-server/blob/b5379b86d22a6de845df1ce06281d5115d8a3011/gradle.properties#L23

to let Gradle run independent tasks in parallel. And executing tests is done via running tasks. So even if the test framework (here: Kotest) itself has parallel test execution disabled, parallel test execution happens at a higher level via Gradle, I believe.

So for example,

    ./gradlew :projectA:test :projectB:test

would run tests of `projectA` and `projectB` in parallel.

*Edit:* This can also be observed by interleaving `STARTED` / `PASSED` test output [in the logs](https://github.com/eclipse-apoapsis/ort-server/actions/runs/25790466425/job/75754750608?pr=5077#step:4:640), see

```
AzureServicebusConfigTest > createConfig should > create an instance from a ConfigManager object STARTED
ArtemisConfigTest > createConfig should > create an instance from a Config object STARTED
AzureServicebusConfigTest > createConfig should > create an instance from a ConfigManager object PASSED
KubernetesMessageReceiverFactoryTest > Messages can be received via the Kubernetes transport STARTED
ArtemisConfigTest > createConfig should > create an instance from a Config object PASSED
```

sschuberth:
So to clarify once more @mnonnenmacher: As I believe we _do_ have tests running parallel as described above, you still believe your solution is safe, correct?

mnonnenmacher:
> > I was pretty sure to have seen parallel test output on the console
> 
> I think that's because: We do set
> 
> https://github.com/eclipse-apoapsis/ort-server/blob/b5379b86d22a6de845df1ce06281d5115d8a3011/gradle.properties#L23
> 
> to let Gradle run independent tasks in parallel. And executing tests is done via running tasks. So even if the test framework (here: Kotest) itself has parallel test execution disabled, parallel test execution happens at a higher level via Gradle, I believe.
> 
> So for example,
> 
> ```
> ./gradlew :projectA:test :projectB:test
> ```
> 
> would run tests of `projectA` and `projectB` in parallel.
> 
> _Edit:_ This can also be observed by interleaving `STARTED` / `PASSED` test output [in the logs](https://github.com/eclipse-apoapsis/ort-server/actions/runs/25790466425/job/75754750608?pr=5077#step:4:640), see
> 
> ```
> AzureServicebusConfigTest > createConfig should > create an instance from a ConfigManager object STARTED
> ArtemisConfigTest > createConfig should > create an instance from a Config object STARTED
> AzureServicebusConfigTest > createConfig should > create an instance from a ConfigManager object PASSED
> KubernetesMessageReceiverFactoryTest > Messages can be received via the Kubernetes transport STARTED
> ArtemisConfigTest > createConfig should > create an instance from a Config object PASSED
> ```

I wonder how Gradle determines how many tasks can run in parallel, I'm only aware of the `maxParallelForks` setting for the test task but nothing similar for `org.gradle.parallel`. With the limited resources on GitHub hosted runners it might actually slow things down if too many test containers are launched in parallel.

> So to clarify once more @mnonnenmacher: As I believe we do have tests running parallel as described above, you still believe your solution is safe, correct?

Yes, because AFAIK Gradle spawns separate JVMs when executing tasks in parallel.

sschuberth:
> I wonder how Gradle determines how many tasks can run in parallel

It defaults to the number of processors, see `--max-workers` [here](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:command_line_performance).

> Yes, because AFAIK Gradle spawns separate JVMs when executing tasks in parallel.

No, it actually does not. Gradle executes tasks in parallel using threads within the same Gradle Daemon JVM, not by spawning multiple JVMs.

mnonnenmacher:
> > I wonder how Gradle determines how many tasks can run in parallel
> 
> It defaults to the number of processors, see `--max-workers` [here](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:command_line_performance).
> 
> > Yes, because AFAIK Gradle spawns separate JVMs when executing tasks in parallel.
> 
> No, it actually does not. Gradle executes tasks in parallel using threads within the same Gradle Daemon JVM, not by spawning multiple JVMs.

It seems that this depends on the specific task implementation: https://docs.gradle.org/current/userguide/worker_api.html#isolation-modes
However, the `Test` task always starts [at least one separate JVM](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html) for executing tests:
> Test are always run in (one or more) separate JVMs.

So that explains why the singleton container does not cause conflicts when running test tasks in parallel. On the GitHub runners with just two cores it might have been luck, but I also haven't seen any issues when running the tests locally with 16 cores.

PR #4121, opened 2025-12-10
build(gradle): Revert using the "build-health" settings plugin

oheger-bosch:
This reverts commits 7491d9c9402d89afe8677666b0d313adf9848d97 and 2218c74aac11c8b6a7487af9fe4ad48ccf08fadb. The changes were causing strange build errors for some developers locally and in the release workflow. The root cause still has to be investigated.

PR #5102, opened 2026-05-16
Enable project health plugin again

mnonnenmacher:
Enable the project health plugin again which was disabled in #4121. ~The errors are currently not reproducible, after merge I will create a release to check if there are still issues with the release workflow if the plugin is applied.~

Also enable most rules and address the findings, and add a CI job to verify project health. Findings of the `onRuntimeOnly` rule are still ignored, because investigating them takes more time and will be done in a follow-up.

Update: The error occurred again when build scans were enabled. I fixed this by not using the `libs` extension in convention plugins which to me is an acceptable inconvenience if it allows to run `projectHealth` in CI now.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -39,7 +39,7 @@ dependencyAnalysis {
                 exclude(projects.utils.test)
             }
             onUsedTransitiveDependencies { severity("ignore") }
-            onIncorrectConfiguration { severity("ignore") }
+            onIncorrectConfiguration { severity("fail") }
             onCompileOnly { severity("ignore") }
             onRuntimeOnly { severity("ignore") }
             onUnusedAnnotationProcessors { severity("ignore") }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,6 +29,31 @@ plugins {
 }
 
 dependencyAnalysis {
+    issues {
+        all {
+            onUnusedDependencies {
+                severity("fail")
+
+                // Exclude modules which are automatically added by the "ort-server-kotlin-jvm-conventions".
+                exclude(projects.utils.logging)
+                exclude(projects.utils.test)
+            }
+            onUsedTransitiveDependencies { severity("ignore") }
+            onIncorrectConfiguration { severity("ignore") }
+            onCompileOnly { severity("ignore") }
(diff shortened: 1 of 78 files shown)


build.gradle.kts
```diff
@@ -33,11 +33,13 @@ dependencyAnalysis {
         all {
             onUnusedDependencies {
                 severity("fail")
+
+                // Exclude modules which are automatically added by the "ort-server-kotlin-jvm-conventions".
```
sschuberth, line 37: This should go to the previous commit.

sschuberth (changes requested):

buildSrc/src/main/kotlin/ort-server-kotlin-jvm-conventions.gradle.kts
```diff
@@ -29,7 +29,6 @@ plugins {
     id("ort-server-kotlin-conventions")
 
     // Apply third-party plugins.
-    id("com.autonomousapps.dependency-analysis")
```
sschuberth, line 32: The old error still [seems to be there](https://github.com/eclipse-apoapsis/ort-server/actions/runs/25965719497/job/76328983666?pr=5102#step:4:67).

mnonnenmacher: See update PR description and the last commit.

buildSrc/src/main/kotlin/ort-server-kotlin-conventions.gradle.kts
```diff
@@ -22,6 +22,7 @@ import dev.detekt.gradle.Detekt
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 private val catalogs = extensions.getByType<VersionCatalogsExtension>()
+private val detektVersion = catalogs.named("libs").findVersion("detektPlugin").get().requiredVersion
```
sschuberth, line 25: Why use `requiredVersion` here? What's the difference to `get().version` below?

mnonnenmacher: When using `findVersion` there is no `version` property.

sschuberth (approved):

PR #5109, opened 2026-05-18
chore(core): Avoid unnecessary `toList()` conversions

sschuberth:
Generalize the `SortedSet` return types for advisor endpoints to `Set`. This way no custom serializer is required for the response. The documentation still makes clear that entries are sorted.


Etsija:
`List -> Set -> SortedSet -> Set` :rofl:

sschuberth:
> `List -> Set -> SortedSet -> Set` 🤣

Note that `Set` is just an interface which `SortedSet` implements. So every `SortedSet` is also a `Set`, and no conversion happens. The consumers just "see" the `SortedSet` differently.

Anyway, @mnonnenmacher already expressed to me privately that he prefers to have more specialized signatures even at the cost of additional conversions, so I'm closing this again.

PR #5125, opened 2026-05-20
fix(dao): Correctly match issues when storing resolutions

oheger-bosch:
When storing resolutions for issues make sure that also the timestamp and the identifier are matched correctly. Otherwise, resolutions may be missed.

### DIFF
--- a/dao/src/main/kotlin/repositories/resolvedconfiguration/DaoResolvedConfigurationRepository.kt
+++ b/dao/src/main/kotlin/repositories/resolvedconfiguration/DaoResolvedConfigurationRepository.kt
@@ -49,6 +49,7 @@ import org.eclipse.apoapsis.ortserver.dao.tables.shared.IssuesTable
 import org.eclipse.apoapsis.ortserver.dao.tables.shared.OrtRunsIssuesTable
 import org.eclipse.apoapsis.ortserver.dao.tables.shared.ResolvedIssuesTable
 import org.eclipse.apoapsis.ortserver.dao.utils.DigestFunction
+import org.eclipse.apoapsis.ortserver.dao.utils.toDatabasePrecision
 import org.eclipse.apoapsis.ortserver.model.repositories.ResolvedConfigurationRepository
 import org.eclipse.apoapsis.ortserver.model.resolvedconfiguration.AppliedPackageCurationRef
 import org.eclipse.apoapsis.ortserver.model.resolvedconfiguration.ResolvedConfiguration
@@ -60,7 +61,6 @@ import org.eclipse.apoapsis.ortserver.model.runs.RuleViolation
 import org.eclipse.apoapsis.ortserver.model.runs.advisor.Vulnerability
 import org.eclipse.apoapsis.ortserver.model.runs.repository.PackageConfiguration
 
-import org.jetbrains.exposed.v1.core.Op
 import org.jetbrains.exposed.v1.core.QueryParameter
 import org.jetbrains.exposed.v1.core.TextColumnType
 import org.jetbrains.exposed.v1.core.and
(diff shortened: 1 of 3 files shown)


dao/src/main/kotlin/repositories/resolvedconfiguration/DaoResolvedConfigurationRepository.kt
```diff
@@ -303,7 +303,8 @@ class DaoResolvedConfigurationRepository(private val db: Database) : ResolvedCon
                         ) and
                     (IssuesTable.severity eq issue.severity) and
                     (IssuesTable.affectedPath eq issue.affectedPath) and
-                        (identifierId?.let { OrtRunsIssuesTable.identifierId eq it } ?: Op.TRUE)
+                    (OrtRunsIssuesTable.timestamp eq issue.timestamp.toDatabasePrecision()) and
```
sschuberth, line 306: Interestingly, the function docs where already saying to match by timestamp, but the code was actually missing...

However, I still have a general / naive question: It seems to me that the matching now gets stricter with taking the timestamp into account. But the commit message says "Otherwise, resolutions may be missed". How can we now miss less by a stricter check?

dao/src/test/kotlin/repositories/resolvedconfiguration/DaoResolvedConfigurationRepositoryTest.kt
```diff
@@ -772,6 +773,110 @@ class DaoResolvedConfigurationRepositoryTest : WordSpec({
             storedResolvedIssues shouldHaveSize 2
         }
 
+        "handle issues that differ only in the timestamp correctly" {
```
sschuberth: Again a general remark: I believe test titles like this are not very useful, because they don't tell the occasional reviewer _what_ the correct behavior is. Something like "match resolutions for issues that only differ in timestamp" would probably be better.

dao/src/main/kotlin/repositories/resolvedconfiguration/DaoResolvedConfigurationRepository.kt
oheger-bosch: This is indeed counter-intuitive, and it took me a while to understand what is going on here. I will try to explain:

The modified function is supposed to find an issue associated with the current run. Only then, a resolution is recorded for this issue. If the query yields multiple results, it returns the first one. This is the problematic part. Because of the less strict conditions, multiple issues were matched, and only the first one (maybe the one with the lowest ID?) was returned. This had the effect that only a subset of issues for the run was matched, and only for those resolutions were recorded. With the stricter condition, only a single issue is selected (if any). This makes sure that all issues with resolutions are found.

dao/src/test/kotlin/repositories/resolvedconfiguration/DaoResolvedConfigurationRepositoryTest.kt
oheger-bosch: Changed the test titles.

dao/src/main/kotlin/repositories/resolvedconfiguration/DaoResolvedConfigurationRepository.kt
sschuberth: Thanks for the explanation! Could you please add it to the commit message?

nnobelis (approved):
LGTM. Feel free to wait for @sschuberth  approval

sschuberth:
> LGTM. Feel free to wait for @sschuberth approval

I'd still appreciate to see [this great explanation](https://github.com/eclipse-apoapsis/ort-server/pull/5125#discussion_r3271737485) as part of the commit message.

dao/src/main/kotlin/repositories/resolvedconfiguration/DaoResolvedConfigurationRepository.kt
oheger-bosch: Added the explanation to the commit message.

PR #5136, opened 2026-05-21
feat(ui): Add buttons to rerun and delete runs to the admin runs table

mnonnenmacher:
Add the same buttons to the admin runs table that already exist in the repository runs table.

image


sschuberth:
Nit: I believe we usually say "rerun" instead of "retry".

sschuberth (approved):

PR #5165, opened 2026-05-27
Handle provenance resolution issues

oheger-bosch:
This PR makes sure that issues created by the scanner when provenance resolution fails are returned by the REST API and shown by the UI. Previously, such issues were only shown by the WebApp report.

### DIFF
--- a/workers/scanner/src/main/kotlin/ScannerWorker.kt
+++ b/workers/scanner/src/main/kotlin/ScannerWorker.kt
@@ -38,7 +38,9 @@ import org.eclipse.apoapsis.ortserver.workers.common.validateForProcessing
 
 import org.jetbrains.exposed.v1.jdbc.Database
 
+import org.ossreviewtoolkit.model.Issue as OrtIssue
 import org.ossreviewtoolkit.model.Provenance
+import org.ossreviewtoolkit.model.ProvenanceResolutionResult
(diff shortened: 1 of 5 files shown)


workers/scanner/src/main/kotlin/ScannerWorker.kt
```diff
@@ -159,16 +161,38 @@ class ScannerWorker(
...
         results.forEach { result ->
             idsByProvenance[result.provenance] = id.mapToModel()
         }
     }
 
-    return issues.flatMapTo(mutableSetOf()) { (provenance, issues) ->
-        issues.map { issue ->
-            issue.mapToModel(identifier = idsByProvenance[provenance], worker = "scanner")
+    issues.forEach { (provenance, issues) ->
+        issues.forEach { issue ->
```
nnobelis: Maybe rename the second lambda parameter form `issues` to something else, to avoid subtil shadowing errors.

nnobelis (dismissed):
LGTM, but feel free to wait for @mnonnenmacher review.

workers/scanner/src/main/kotlin/ScannerWorker.kt
oheger-bosch: Renamed.

nnobelis (approved):

PR #5179, opened 2026-05-28
Render plugin descriptions in Markdown

### LINKED ISSUE #5170: Plugin descriptions should be rendered as Markdown
Some plugin descriptions contain Markdown, so they should be rendered as such. See e.g.

Image

where the links should be clickable.

Etsija:
This aligns the rendering with `admin/plugins` section in the UI.

Screenshot from 2026-05-28 12-41-49

### DIFF
--- a/ui/src/components/form/multi-select-field.tsx
+++ b/ui/src/components/form/multi-select-field.tsx
@@ -25,6 +25,7 @@ import {
   UseFormReturn,
 } from 'react-hook-form';
 
+import { MarkdownRenderer } from '@/components/markdown-renderer';
 import { Checkbox } from '@/components/ui/checkbox';
 import {
   FormControl,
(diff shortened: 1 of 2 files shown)


ui/src/components/form/multi-select-field.tsx
```diff
@@ -129,9 +130,10 @@ export const MultiSelectField = <
               <div className='flex flex-col'>
                 <FormLabel className='font-normal'>{option.label}</FormLabel>
                 {option.description != null && (
-                  <div className='text-sm text-gray-500'>
-                    {option.description}
-                  </div>
+                  <MarkdownRenderer
+                    markdown={option.description}
```
sschuberth, line 134: Is this also about _plugin_ option descriptions? Just wondering as I would have expected only one place to make the change (the below). Can you explain the difference between these hunks?

Etsija: For reporters, `MultiSelectField` renders `plugin.description` via `option.description`. If plugin descriptions can contain Markdown, reporter descriptions need the same rendering as advisor / scanner descriptions use with `PluginMultiSelectField`.

sschuberth (approved):

ui/src/components/form/multi-select-field.tsx
Etsija: One note: we should probably refactor `ReporterFields` component to using `PluginMultiSelectField` as well, the same way scanner and advisor now has - but that would be a bigger refactor, so I decided to do it this way (this time) instead, as there's no harm in using Markdown component for fields which do not use Markdown. And I am not sure whether the reporter section of the run form yet supports plugins (I believe not).

PR #5111, opened 2026-05-18
Filter vulnerabilities with advisors

Etsija:
#### More than one vulnerability advisor chosen:

Screenshot from 2026-05-18 14-14-24

#### When only one vulnerability advisor is chosen to the run, no sense in rendering the advisor filter:

Screenshot from 2026-05-18 14-14-44

Please see the commits for details.

PR #5188, opened 2026-05-29
build(gradle): Prefer `buildHealth` over `projectHealth`

sschuberth:
Do not stop at the first issue found, but report all of them for the whole build.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -132,7 +132,7 @@ jobs:
       working-directory: website
       run: pnpm format:check
 
-  project-health:
+  build-health:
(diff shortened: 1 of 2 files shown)


sschuberth:
@mnonnenmacher, due to the job renaming, `project-health` needs to be removed from required jobs, and `build-health` needs to be added.

mnonnenmacher:
> @mnonnenmacher, due to the job renaming, `project-health` needs to be removed from required jobs, and `build-health` needs to be added.

I can do that once #5187 is merged.
As #5111 is now also in the merge queue I will wait for that as well with updating the config.

mnonnenmacher (approved):

PR #5213, opened 2026-06-03
build: Upgrade to Java 25 to benefit from various optimizations

sschuberth:
This is a preparation for https://github.com/oss-review-toolkit/ort/pull/11586.


sschuberth:
@eclipse-apoapsis/bosch, I'd like to merge this ASAP. Any objections?

sschuberth:
> @eclipse-apoapsis/bosch, I'd like to merge this ASAP. Any objections?

Ping again, esp. @MarcelBochtler and @oheger-bosch. We plan to merge this after today's release unless there are objections.

gradle/gradle-daemon-jvm.properties
```diff
@@ -1,3 +1,12 @@
...
+toolchainUrl.FREE_BSD.AARCH64=https\://api.foojay.io/disco/v3.0/ids/cf726b4a1c84b50457225f9bba6d7650/redirect
+toolchainUrl.FREE_BSD.X86_64=https\://api.foojay.io/disco/v3.0/ids/fa1e318c287360478e3c83a9a3ef1007/redirect
+toolchainUrl.LINUX.AARCH64=https\://api.foojay.io/disco/v3.0/ids/cf726b4a1c84b50457225f9bba6d7650/redirect
+toolchainUrl.LINUX.X86_64=https\://api.foojay.io/disco/v3.0/ids/fa1e318c287360478e3c83a9a3ef1007/redirect
+toolchainUrl.MAC_OS.AARCH64=https\://api.foojay.io/disco/v3.0/ids/c2dd35c9d0aaf0ba6ad0791320f99dfc/redirect
+toolchainUrl.MAC_OS.X86_64=https\://api.foojay.io/disco/v3.0/ids/e5810bd7fd1f8a586644409d395a7e55/redirect
+toolchainUrl.UNIX.AARCH64=https\://api.foojay.io/disco/v3.0/ids/cf726b4a1c84b50457225f9bba6d7650/redirect
+toolchainUrl.UNIX.X86_64=https\://api.foojay.io/disco/v3.0/ids/fa1e318c287360478e3c83a9a3ef1007/redirect
+toolchainUrl.WINDOWS.AARCH64=https\://api.foojay.io/disco/v3.0/ids/7b3c4877c0749019e6805bb61e421497/redirect
+toolchainUrl.WINDOWS.X86_64=https\://api.foojay.io/disco/v3.0/ids/d76df094a9cbbabd3b08251f9e61444a/redirect
```
MarcelBochtler, line 11: What is the purpose of this change, and why is it required?

Edit: And why wasn't it required to set the architecture specific URLs before?

sschuberth: > What is the purpose of this change, and why is it required?

We previously maintained this file manually, which is a bad practice. Instead, we should leave its generation to the `updateDaemonJvm` task and configure the task.

> And why wasn't it required to set the architecture specific URLs before?

It's still not required, but not doing so results in Gradle making API calls to the foojay service to search for matching Java toolchains, which can be slow, or even break (having the various foojay outages in mind). So the `updateDaemonJvm` task hard-codes resolved URLs here as a performance optimization. Also see [Configuring provisioning URLs](https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:configure_provision).

MarcelBochtler (approved):

Issue #4547, opened 2026-02-23
Use up-to-date image for `test-ui-e2e`


mnonnenmacher:
The end-to-end UI test currently uses the latest image of `core`, this means that it will only be noticed after merge if a change to `core` breaks the test (see https://github.com/eclipse-apoapsis/ort-server/pull/4188#issuecomment-3681770129). To fix this, change the workflow to build a current image of `core` for the test.

PR #5247, opened 2026-06-08
Add support for generating an offline token in the UI and using it in the CLI

mnonnenmacher:
Add a feature to the UI to generate offline tokens which can then be used by the CLI to authenticate to the server. This solves the problem that users from external identity providers or with 2FA enabled could not use the CLI in automation (like CI), because username/password authentication requires additional manual steps in those cases.


lamppu:
There is something a bit weird that happens if the user isn't yet signed in and navigates straight to the cli-token page and then signs in, there's first a screen that says "Unable to log in", but after a few seconds the page appears, though there's still an error shown.

Screenshot from 2026-06-09 17-35-18

scripts/docker/keycloak/master-realm.json
```diff
@@ -724,12 +724,12 @@
     "enabled" : true,
     "alwaysDisplayInConsole" : false,
     "clientAuthenticatorType" : "client-secret",
-    "redirectUris" : [ "/*" ],
-    "webOrigins" : [ "/*" ],
+    "redirectUris" : [ "http://localhost:8082/profile/cli-token", "http://localhost:5173/profile/cli-token" ],
```
sschuberth: > add the new route as explicit redirect URI.

Is there an "an" or "the" missing?

cli/src/commonMain/kotlin/model/AuthenticationStorage.kt
```diff
@@ -99,7 +99,7 @@ internal data class HostAuthenticationDetails(
     val clientId: String,
 
     /** The username used for the authentication. */
-    val username: String,
+    val username: String?,
```
sschuberth, line 102: > as alternative

as _an_ alternative

mnonnenmacher:
> There is something a bit weird that happens if the user isn't yet signed in and navigates straight to the cli-token page and then signs in, there's first a screen that says "Unable to log in", but after a few seconds the page appears, though there's still an error shown.

Good catch, I have reworked the flow to use a dedicated callback route and verified that both routes behave correctly when opened by an unauthenticated user.

ui/src/routes/profile/cli-token/callback/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+      canceled = true;
+    };
+  }, [cliClientId, createOidcClient, navigate, offlineToken]);
+
+  return (
+    <div className='mx-auto flex w-full max-w-4xl flex-col gap-6'>
+      <Card>
+        <CardHeader>
+          <CardTitle className='text-3xl font-bold tracking-tight'>
+            CLI Authentication Token
```
MarcelBochtler: Can this token also be used for other "automations" aside from the CLI?
I.e. a custom script that calls the API?

Maybe we should go with a more generic Token name, like API Authentication Token?

mnonnenmacher: What about "offline token"? Because I'm already using that term in a few places.
Technically this is a refresh token so I would like to avoid a name that resembles "API token" because then people might think they can use the token directly for authentication.

mnonnenmacher: On the other hand, "offline" could also be confusing for users. So I think I would like to keep the terminology for now and merge as is, we can improve the wording later based on user feedback.

MarcelBochtler: > What about "offline token"?

This is IMO to technical (same goes for the CLI option `--offline-token`). I think I like "CLI Authentication Token" better than that.
I have the feeling that you're trying to avoid confusion for users that know about authentication workflows (i.e. the token can't be used as `Bearer <token>`.
But with the disadvantage of making it less self-explanatory for users that don't know anything about authentication workflow (i.e. what is offline about the `--offline-token`?).

Maybe just token and `--token` respectively?

Other CLIs use something similar (For these tools I just use the CLI, I don't care if the token is a long-lived refresh token or an actual authentication token):

*Bitwarden CLI*
https://bitwarden.com/help/personal-api-key/#authenticate-using-your-api-key
`bw login --apikey`

*OpenShift CLI*
`oc login --token=`

sschuberth: > This is IMO to technical (same goes for the CLI option `--offline-token`).

I agree. As a user of other tools, I've never heard the term "offline token" before. It sounds as if it couldn't be used with "online" deployments of ORT Server in K8s, but only locally when running in Docker Compose.

cli/src/commonMain/kotlin/LoginCommand.kt
```diff
@@ -57,21 +89,20 @@ class LoginCommand : SuspendingCliktCommand(name = "login") {
...
-        help = "The password to authenticate with the ORT Server instance."
-    ).required()
+    private val userOptions by UserOptions().cooccurring()
+    private val tokenOptions by TokenOptions().cooccurring()
 
     override fun help(context: Context) = "Login to an ORT Server instance."
 
     override suspend fun run() {
+        if (userOptions != null && tokenOptions != null) {
+            throw MutuallyExclusiveGroupException(listOf("--offline-token", "--username and --password"))
```
sschuberth: Nit: In the condition and also in the declaration of options, `userOptions` comes first. So should also in the exception argument "--username and --password" come first?

cli/src/commonMain/kotlin/LoginCommand.kt
```diff
@@ -57,21 +89,20 @@ class LoginCommand : SuspendingCliktCommand(name = "login") {
...
+        }
+
+        if (userOptions == null && tokenOptions == null) {
+            throw UsageError("Either --offline-token or --username and --password must be provided.")
```
sschuberth: See above.

cli/src/commonMain/kotlin/LoginCommand.kt
```diff
@@ -91,7 +122,12 @@ class LoginCommand : SuspendingCliktCommand(name = "login") {
             clientId = oidcConfig.clientId
         )
 
-        val tokenInfo = authService.generateToken(username, password, setOf("offline_access"))
+        var username: String? = null
+
+        val tokenInfo = userOptions?.let {
+            username = it.username
+            authService.generateToken(it.username, it.password, setOf("offline_access"))
+        } ?: authService.refreshToken(checkNotNull(tokenOptions).offlineToken, setOf("offline_access"))
```
sschuberth: How about instead:

```kotlin
val (username, tokenInfo) = userOptions?.let {
    it.username to authService.generateToken(it.username, it.password, setOf("offline_access"))
} ?: (null to authService.refreshToken(checkNotNull(tokenOptions).offlineToken, setOf("offline_access")))
```

ui/src/routes/profile/cli-token/callback/index.tsx
mnonnenmacher: > the token can't be used as Bearer <token>

Yes, I would like to avoid that people think the token can be used like that. I think calling it "Authentication token" in the UI plus a short explanation on the page should be sufficient for that. But I would still scratch the "CLI" because it can be used in scripts as well, just not as an access token. So if you agree I would do these changes:

* Change the UI routes from `/profile/cli-token` and `/profile/cli-token/callback` to `/profile/token` and `/profile/token/callback`.
* Change the page title to just "Authentication token" and update the text to explain how it can be used with the CLI. And maybe also a short explanation how to use it without CLI (or just a link to some docs).
* Rename the CLI option to just `--token`.

cli/src/commonMain/kotlin/LoginCommand.kt
mnonnenmacher: I don't like the `null` in the else branch, it shouldn't have to care about that. But as this is now delayed anyway I will look into better ways of addressing the ugly null handling with delegate properties and maybe add a preliminary refactoring commit.

ui/src/routes/profile/cli-token/callback/index.tsx
MarcelBochtler: > * Change the UI routes from `/profile/cli-token` and `/profile/cli-token/callback` to `/profile/token` and `/profile/token/callback`.
> 
> * Change the page title to just "Authentication token" and update the text to explain how it can be used with the CLI. And maybe also a short explanation how to use it without CLI (or just a link to some docs).
> 
> * Rename the CLI option to just `--token`.

Sounds good.

cli/src/commonMain/kotlin/LoginCommand.kt
mnonnenmacher: I have added a few commits to the beginning of the PR but for the username I actually decided to just use `userOptions?.username` below which avoids the need for a `username` variable altogether.

ui/src/routes/profile/cli-token/callback/index.tsx
mnonnenmacher: All changes are applied.

mnonnenmacher:
The `test-ui-e2e` failure is not related to this change but to the Java 25 upgrade from #5213 because of #4547.

scripts/docker/keycloak/master-realm.json
```diff
@@ -724,12 +724,12 @@
     "enabled" : true,
     "alwaysDisplayInConsole" : false,
     "clientAuthenticatorType" : "client-secret",
-    "redirectUris" : [ "/*" ],
-    "webOrigins" : [ "/*" ],
+    "redirectUris" : [ "http://localhost:8082/profile/token/callback", "http://localhost:5173/profile/token/callback" ],
```
sschuberth, line 727: The commit message needs adjustments, it still refers to "offline token" and the "cli-token" route.

cli/src/commonMain/kotlin/model/AuthenticationStorage.kt
sschuberth, line 102: Also here, the commit message needs adjustments for the mention of "offline".

sschuberth (changes requested):

sschuberth (approved):

PR #5262, opened 2026-06-11
fix(docker): Upgrade the base image to Java 25

mnonnenmacher:
Also pin the Temurin base image to a specific digest as Renovate does not do this automatically.

This is a fixup for 2a75662.

### DIFF
--- a/docker/Base.Dockerfile
+++ b/docker/Base.Dockerfile
@@ -22,7 +22,7 @@
 # setting file permissions for the user, setting up certificates, etc. And, it defines the temurin base image.
 
 ARG CREDENTIAL_HELPER_VERSION=0.2.0
-ARG TEMURIN_VERSION=21.0.11_10-jdk-noble
+ARG TEMURIN_VERSION=25.0.3_9-jdk-noble@sha256:02aba7518e48cfed96403ac9634e357a40329d6ec9418feb0b32636e43b245a1
 
 FROM alpine:3.24 AS credential-helper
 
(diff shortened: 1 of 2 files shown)


sschuberth (dismissed):

MarcelBochtler (commented):
Can you also update the renovate configuration?
https://github.com/eclipse-apoapsis/ort-server/blob/d64787d3e7dc73a9362eade3da48ba146e59b5d4/renovate.json#L28-L29

mnonnenmacher:
It seems the PyPI issue is resolved, I hope the scanner image build now works again.

sschuberth:
> Can you also update the renovate configuration?
> 
> https://github.com/eclipse-apoapsis/ort-server/blob/d64787d3e7dc73a9362eade3da48ba146e59b5d4/renovate.json#L28-L29

BTW, we do we seem to allow < 25 (in the new code) as well? Or is the syntax just misleading?

sschuberth (approved):

mnonnenmacher:
> > Can you also update the renovate configuration?
> > https://github.com/eclipse-apoapsis/ort-server/blob/d64787d3e7dc73a9362eade3da48ba146e59b5d4/renovate.json#L28-L29
> 
> BTW, we do we seem to allow < 25 (in the new code) as well? Or is the syntax just misleading?

I'm not sure I understand the question correctly, but yes, this only prevents upgrades to >= 26, so in theory upgrading to 24.x would also be allowed by this rule. However, as we are at 25 already that effectively allows upgrading to any 25.x version. There are more advanced ways to configure this more precisely, but it's not worth it IMO because the syntax gets more complicated:
https://docs.renovatebot.com/configuration-options/#packagerulesallowedversions

PR #5268, opened 2026-06-12
Improve declared licenses filtering

### LINKED ISSUE #5161: Declared license filter in package view does not work for non-SPDX licenses
In contrast to _detected_ licenses, _declared_ licenses can be arbitrary strings (they're just "unmapped" then). In fact, for NuGet packages these are often URLs (to license texts). For such packages the license filter is not populated correctly, so that the user cannot filter for such "licenses", see

Image

However, "https://www.nuget.org/packages/CommandLineParser/2.9.1/license" _should_ be shown as a filter option, which helps to quickly identify such packages.

### LINKED ISSUE #5263: Package license filter options ignore package curations
`GET /runs/{runId}/packages/licenses` returns `processedDeclaredLicenses` from raw analyzer data, without applying package
curations. The package list endpoint does apply curations, so users see declared licenses in the UI (e.g. ANTLR-PD) that are absent from the filter dropdown — making the filter inconsistent with the list.

Fix: the license options endpoint should derive its values from the same post-curation data used by the package list
(`PackageService.listForOrtRunId()`), rather than reading directly from `ProcessedDeclaredLicensesTable`.

Etsija:
This PR aligns the detected license filter options in the UI with the license data shown in the package cards, by adding the unmapped licenses to the filter options, and fixing a bug, where package curations for licenses were not taken into account when constructing the filter options.

### Also unmapped licenses can be included in the detected license filter

Screenshot from 2026-06-12 07-32-47

### After the bug fix, curated licenses are properly shown in the filter

Screenshot from 2026-06-12 07-31-09

Screenshot from 2026-06-12 07-32-06

### DIFF
--- a/api/v1/model/src/commonMain/kotlin/Package.kt
+++ b/api/v1/model/src/commonMain/kotlin/Package.kt
@@ -54,8 +54,8 @@ data class PackageFilters(
     /** Substring filter for purl. Null if not set. */
     val purl: FilterOperatorAndValue<String>? = null,
 
-    /** Set of SPDX license expressions to filter with. Null if not set. */
-    val processedDeclaredLicense: FilterOperatorAndValue<Set<String>>? = null,
+    /** Set of declared licenses to filter with. Null if not set. */
+    val declaredLicense: FilterOperatorAndValue<Set<String>>? = null,
 
     /** Filter by dependency type. True = direct, false = transitive, null = all. */
     val isDirectDependency: Boolean? = null
--- a/core/src/main/kotlin/api/RunsRoute.kt
+++ b/core/src/main/kotlin/api/RunsRoute.kt
@@ -292,7 +292,8 @@ fun Route.runs() = route("runs") {
             route("licenses") {
                 get(getRunPackageLicenses, requireRunPermission()) {
                     val licenses = Licenses(
-                        packageService.getProcessedDeclaredLicenses(call.ortRun.id)
+                        processedDeclaredLicenses = packageService.getProcessedDeclaredLicenses(call.ortRun.id),
+                        unmappedDeclaredLicenses = packageService.getUnmappedDeclaredLicenses(call.ortRun.id)
                     )
 
                     call.respond(HttpStatusCode.OK, licenses)
--- a/core/src/main/kotlin/api/RunsRoute.kt
+++ b/core/src/main/kotlin/api/RunsRoute.kt
@@ -517,15 +517,17 @@ private fun ApplicationCall.filters(): OrtRunFilters =
     )
 
 /**
- * Extract the filter for the processed declared license from this [ApplicationCall]. If this filter is missing or
- * empty, return null. Otherwise, it is interpreted as a comma-delimited list of license expression strings to filter
- * the result by. If the first item on the list is a minus, the provided licenses will be excluded from the result.
+ * Extract the filter for declared licenses from this [ApplicationCall]. If this filter is missing or empty, return
+ * null. Otherwise, it is interpreted as a comma-delimited list of license strings to filter the result by. If the
+ * first item on the list is a minus, the provided licenses will be excluded from the result.
  */
-private fun ApplicationCall.processedDeclaredLicense(): FilterOperatorAndValue<Set<String>>? {
-    val parts = parameters["processedDeclaredLicense"]?.split(',')?.toMutableSet() ?: return null
+private fun ApplicationCall.declaredLicense(): FilterOperatorAndValue<Set<String>>? = licenseFilter("declaredLicense")
 
-    return if (parts.removeIf { it == "-" }) {
-        FilterOperatorAndValue(ComparisonOperator.NOT_IN, parts)
+private fun ApplicationCall.licenseFilter(parameterName: String): FilterOperatorAndValue<Set<String>>? {
+    val parts = parameters[parameterName]?.split(',')?.toSet() ?: return null
+
+    return if ("-" in parts) {
+        FilterOperatorAndValue(ComparisonOperator.NOT_IN, parts - "-")
     } else {
         FilterOperatorAndValue(ComparisonOperator.IN, parts)
     }
@@ -538,7 +540,7 @@ private fun ApplicationCall.packageFilters(): PackageFilters =
     PackageFilters(
         identifier = parameters["identifier"]?.let { FilterOperatorAndValue(ComparisonOperator.ILIKE, it) },
         purl = parameters["purl"]?.let { FilterOperatorAndValue(ComparisonOperator.ILIKE, it) },
-        processedDeclaredLicense = processedDeclaredLicense(),
+        declaredLicense = declaredLicense(),
         isDirectDependency = parameters["isDirectDependency"]?.lowercase()?.toBooleanStrictOrNull()
     )
 
--- a/core/src/main/kotlin/apiDocs/RunsDocs.kt
+++ b/core/src/main/kotlin/apiDocs/RunsDocs.kt
@@ -845,7 +845,9 @@ val getRunStatistics: RouteConfig.() -> Unit = {
 
 val getRunPackageLicenses: RouteConfig.() -> Unit = {
     operationId = "getRunPackageLicenses"
-    summary = "Get the licenses for packages found in an ORT run"
+    summary = "Get the declared licenses for packages found in an ORT run"
+    description = "The response contains independently deduplicated and alphabetically sorted processed and " +
+            "unmapped declared licenses."
     tags = listOf("Runs")
 
     request {
@@ -856,10 +858,14 @@ val getRunPackageLicenses: RouteConfig.() -> Unit = {
 
     response {
         HttpStatusCode.OK to {
+            description = "Success"
             jsonBody<Licenses> {
                 example("Get licenses for packages") {
                     value = Licenses(
-                        processedDeclaredLicenses = listOf("Apache-2.0", "MIT")
+                        processedDeclaredLicenses = listOf("Apache-2.0", "MIT"),
+                        unmappedDeclaredLicenses = listOf(
+                            "https://www.nuget.org/packages/CommandLineParser/2.9.1/license"
+                        )
                     )
                 }
             }
--- a/core/src/main/kotlin/apiDocs/RunsDocs.kt
+++ b/core/src/main/kotlin/apiDocs/RunsDocs.kt
@@ -538,10 +538,11 @@ val getRunPackages: RouteConfig.() -> Unit = {
                     "substring match."
         }
 
-        queryParameter<String>("processedDeclaredLicense") {
-            description = "Defines the processed declared licenses for which packages are to be retrieved. This is a " +
+        queryParameter<String>("declaredLicense") {
+            description = "Defines the displayed declared license values for which packages are to be retrieved, " +
+                    "including processed declared licenses and unmapped declared license strings. This is a " +
                     "comma-separated string. Add a minus as the first item to exclude packages with the specified " +
-                    "license expressions, e.g. '-,MIT'."
+                    "licenses, e.g. '-,MIT'."
         }
 
         queryParameter<Boolean>("isDirectDependency") {
@@ -629,7 +630,7 @@ val getRunPackages: RouteConfig.() -> Unit = {
                                 operator = ComparisonOperator.ILIKE,
                                 value = "pkg:maven/org.example/name@1.0"
                             ),
-                            processedDeclaredLicense = FilterOperatorAndValue(
+                            declaredLicense = FilterOperatorAndValue(
                                 operator = ComparisonOperator.IN,
                                 value = setOf("Apache-2.0")
                             ),
--- a/core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
+++ b/core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
@@ -1750,8 +1750,9 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
             }
         }
 
-        "allow filtering by processed declared license and identifier" {
+        "allow filtering by declared license" {
             integrationTestApplication {
+                val licenseUrl = "https://www.nuget.org/packages/CommandLineParser/2.9.1/license"
                 val ortRunId = dbExtension.fixtures.createOrtRun(
                     repositoryId = repositoryId,
                     revision = "revision",
@@ -1763,10 +1764,9 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
                     configuration = AnalyzerJobConfiguration()
                 ).id
 
-                val identifier1 = Identifier("Maven", "com.example", "example", "1.0")
-                val identifier2 = Identifier("Maven", "com.example", "example2", "1.0")
-                val identifier3 = Identifier("Maven", "com.example", "example3", "1.0")
-                val identifier4 = Identifier("NPM", "com.example", "example4", "1.0")
+                val identifier1 = Identifier("Maven", "com.example", "processed", "1.0")
+                val identifier2 = Identifier("Maven", "com.example", "unmapped", "1.0")
+                val identifier3 = Identifier("Maven", "com.example", "other", "1.0")
 
                 dbExtension.fixtures.createAnalyzerRun(
                     analyzerJobId,
@@ -1784,21 +1784,13 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
                             processedDeclaredLicense = ProcessedDeclaredLicense(
                                 "Apache-2.0",
                                 emptyMap(),
-                                emptySet()
+                                setOf(licenseUrl)
                             )
                         ),
                         dbExtension.fixtures.generatePackage(
                             identifier3,
                             processedDeclaredLicense = ProcessedDeclaredLicense(
-                                "Apache-2.0",
-                                emptyMap(),
-                                emptySet()
-                            )
-                        ),
-                        dbExtension.fixtures.generatePackage(
-                            identifier4,
-                            processedDeclaredLicense = ProcessedDeclaredLicense(
-                                "MIT",
+                                "BSD-2-Clause",
                                 emptyMap(),
                                 emptySet()
                             )
@@ -1808,24 +1800,22 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
 
                 val response = superuserClient.get("/api/v1/runs/$ortRunId/packages") {
                     url {
-                        parameters.append("processedDeclaredLicense", "-,Apache-2.0 OR LGPL-2.1-or-later,MIT")
-                        parameters.append("identifier", "example3")
+                        parameters.append("declaredLicense", "Apache-2.0 OR LGPL-2.1-or-later,$licenseUrl")
                     }
                 }
 
                 response shouldHaveStatus HttpStatusCode.OK
                 val packages = response.body<PagedSearchResponse<ApiPackage, PackageFilters>>()
 
-                packages.data shouldHaveSize 1
-                packages.data.first().processedDeclaredLicense.spdxExpression shouldBe "Apache-2.0"
-                packages.data.first().identifier shouldBe identifier3.mapToApi()
-
+                packages.data.map { it.identifier } should containExactly(
+                    identifier1.mapToApi(),
+                    identifier2.mapToApi()
+                )
                 packages.filters shouldBe PackageFilters(
-                    processedDeclaredLicense = FilterOperatorAndValue(
-                        ComparisonOperator.NOT_IN,
-                        setOf("Apache-2.0 OR LGPL-2.1-or-later", "MIT")
-                    ),
-                    identifier = FilterOperatorAndValue(ComparisonOperator.ILIKE, "example3")
+                    declaredLicense = FilterOperatorAndValue(
+                        ComparisonOperator.IN,
+                        setOf("Apache-2.0 OR LGPL-2.1-or-later", licenseUrl)
+                    )
                 )
             }
         }
--- a/core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
+++ b/core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
@@ -2925,8 +2925,9 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
     }
 
     "GET /runs/{runId}/packages/licenses" should {
-        "return the processed declared licenses found in packages" {
+        "return the processed and unmapped declared licenses found in packages" {
             integrationTestApplication {
+                val licenseUrl = "https://www.nuget.org/packages/CommandLineParser/2.9.1/license"
                 val ortRunId = dbExtension.fixtures.createOrtRun(
                     repositoryId = repositoryId,
                     revision = "revision",
@@ -2941,25 +2942,25 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
                         dbExtension.fixtures.generatePackage(
                             Identifier("Maven", "com.example", "example", "1.0"),
                             processedDeclaredLicense = ProcessedDeclaredLicense(
-                                "LGPL-2.1-or-later",
-                                emptyMap(),
-                                emptySet()
+                                spdxExpression = "LGPL-2.1-or-later",
+                                mappedLicenses = emptyMap(),
+                                unmappedLicenses = setOf("custom-license", licenseUrl)
                             )
                         ),
                         dbExtension.fixtures.generatePackage(
                             Identifier("Maven", "com.example", "example2", "1.0"),
                             processedDeclaredLicense = ProcessedDeclaredLicense(
-                                "Apache-2.0 OR LGPL-2.1-or-later",
-                                emptyMap(),
-                                emptySet()
+                                spdxExpression = "Apache-2.0 OR LGPL-2.1-or-later",
+                                mappedLicenses = emptyMap(),
+                                unmappedLicenses = setOf(licenseUrl)
                             )
                         ),
                         dbExtension.fixtures.generatePackage(
                             Identifier("NPM", "", "example", "1.0"),
                             processedDeclaredLicense = ProcessedDeclaredLicense(
-                                "MIT",
-                                emptyMap(),
-                                emptySet()
+                                spdxExpression = "MIT",
+                                mappedLicenses = emptyMap(),
+                                unmappedLicenses = setOf("unknown-license")
                             )
                         )
                     )
@@ -2976,6 +2977,7 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
                     "LGPL-2.1-or-later",
                     "MIT"
                 )
+                licenses.unmappedDeclaredLicenses shouldBe listOf("custom-license", licenseUrl, "unknown-license")
             }
         }
 
--- a/model/src/commonMain/kotlin/runs/Package.kt
+++ b/model/src/commonMain/kotlin/runs/Package.kt
@@ -51,8 +51,8 @@ data class PackageFilters(
     /** Substring filter for purl. Null if not set. */
     val purl: FilterOperatorAndValue<String>? = null,
 
-    /** Set of SPDX license expressions to filter with. Null if not set. */
-    val processedDeclaredLicense: FilterOperatorAndValue<Set<String>>? = null,
+    /** Set of declared license values to filter with. Null if not set. */
+    val declaredLicense: FilterOperatorAndValue<Set<String>>? = null,
 
     /** Filter by dependency type. True = direct, false = transitive, null = all. */
     val isDirectDependency: Boolean? = null
--- a/services/ort-run/src/main/kotlin/PackageService.kt
+++ b/services/ort-run/src/main/kotlin/PackageService.kt
@@ -53,7 +53,7 @@ import org.jetbrains.exposed.v1.core.eq
 import org.jetbrains.exposed.v1.core.inList
 import org.jetbrains.exposed.v1.core.inSubQuery
 import org.jetbrains.exposed.v1.core.not
-import org.jetbrains.exposed.v1.core.notInList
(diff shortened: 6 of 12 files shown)


services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -236,7 +238,7 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
                 }
         }
 
-    /** Get all distinct processed declared license expressions found in packages in an ORT run. */
+    /** Return distinct processed declared SPDX expressions for the ORT run. */
```
sschuberth: Nit: I believe we should keep the word "license" in here, i.e. "SPDX license expressions".

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -247,6 +249,20 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
...
                 .mapNotNull { it[ProcessedDeclaredLicensesTable.spdxExpression] }
         }
+
+    /** Return distinct unmapped declared license strings for the ORT run. */
+    suspend fun getUnmappedDeclaredLicenses(ortRunId: Long): List<String> =
+        db.dbQuery {
+            PackagesTable.joinAnalyzerTables()
+                .innerJoin(ProcessedDeclaredLicensesTable)
+                .innerJoin(ProcessedDeclaredLicensesUnmappedDeclaredLicensesTable)
+                .innerJoin(UnmappedDeclaredLicensesTable)
```
sschuberth, line 317: I cannot really comment on that as I know too little about the DB schema, but having three joins for getting just the unmapped licenses seems very complicated.

api/v1/model/src/commonMain/kotlin/Package.kt
```diff
@@ -57,6 +57,9 @@ data class PackageFilters(
     /** Set of SPDX license expressions to filter with. Null if not set. */
     val processedDeclaredLicense: FilterOperatorAndValue<Set<String>>? = null,
 
+    /** Set of declared license values to filter with. Null if not set. */
+    val declaredLicense: FilterOperatorAndValue<Set<String>>? = null,
```
sschuberth, line 58: When having ORT context in mind, the naming is a bit misleading, as it sounds like the input to the license processor which generates `processedDeclaredLicense`. I wonder whether we should explicitly call this "processedAndUnmappedDeclaredLicenses" (also: plural).

sschuberth: BTW, is after this commit `processedDeclaredLicense` even still used? Because if not, we could probably remove it from here, which would make it less confusing to keep the property's name at `declaredLicense`.

core/src/main/kotlin/api/RunsRoute.kt
```diff
@@ -521,8 +521,18 @@ private fun ApplicationCall.filters(): OrtRunFilters =
...
+
+/**
+ * Extract the filter for declared licenses from this [ApplicationCall]. If this filter is missing or empty, return
+ * null. Otherwise, it is interpreted as a comma-delimited list of license strings to filter the result by. If the
+ * first item on the list is a minus, the provided licenses will be excluded from the result.
+ */
+private fun ApplicationCall.declaredLicense(): FilterOperatorAndValue<Set<String>>? = licenseFilter("declaredLicense")
+
+private fun ApplicationCall.licenseFilter(parameterName: String): FilterOperatorAndValue<Set<String>>? {
+    val parts = parameters[parameterName]?.split(',')?.toMutableSet() ?: return null
```
sschuberth: Does this really need to be mutable? I believe `toSet()` should suffice.

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -35,11 +35,15 @@ import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.ProcessedDecl
 import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.ShortestDependencyPathsTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.UnmappedDeclaredLicensesTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.resolvedconfiguration.CuratedPackagesTable
+import org.eclipse.apoapsis.ortserver.dao.tables.shared.IdentifierDao
```
sschuberth, line 38: > The `/licenses` options endpoint and the `processedDeclaredLicense` and
`declaredLicense` package filters match the raw analyzer values

There's one "and" too much in here for my taste, probably replace the first with a comma.

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -80,6 +86,15 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
                 .select(PackagesTable.id)
                 .where { AnalyzerJobsTable.ortRunId eq ortRunId }
 
+            var curationsMap: Map<Identifier, List<Pair<String, PackageCuration>>>? = null
+            var overrides: List<DeclaredLicenseOverride>? = null
```
sschuberth: Using `var` instead of `val` might be a code smell. I wonder whether there's a different way.

core/src/main/kotlin/apiDocs/RunsDocs.kt
```diff
@@ -856,10 +856,15 @@ val getRunPackageLicenses: RouteConfig.() -> Unit = {
 
     response {
         HttpStatusCode.OK to {
+            description = "Success. The response contains independently deduplicated and alphabetically sorted " +
+                    "processed and unmapped declared licenses."
```
mnonnenmacher: This should probably go to the description of the endpoint instead of the example response.

services/ort-run/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -800,10 +800,47 @@ class PackageServiceTest : WordSpec() {
...
+                                emptyMap(),
+                                setOf(licenseUrl, "custom-license")
+                            )
+                        ),
+                        fixtures.generatePackage(
+                            Identifier("Maven", "com.example", "example2", "1.0"),
+                            processedDeclaredLicense = ProcessedDeclaredLicense(
+                                "MIT",
+                                emptyMap(),
+                                setOf(licenseUrl, "MIT")
```
mnonnenmacher: To keep the test data meaningful use something else than "MIT" which is a valid SPDX license and therefore would never appear in the list of unmapped licenses.

services/ort-run/src/main/kotlin/PackageService.kt
mnonnenmacher, line 38: > For code clarity and to keep "detekt" happy, also refactor part of the
service code into private data classes and functions.

Refactorings should happen in separate commits.

mnonnenmacher (changes requested):

services/ort-run/src/main/kotlin/PackageService.kt
Etsija: This just implicates the need of the read model: the joins are not arbitrary, but they follow the normalized db schema:
- packages belong to an ORT run via analyzer run tables
- a package has a `ProcessedDeclaredLicensesTable` row
- unmapped licenses are connected through the M:N relation table `ProcessedDeclaredLicensesUnmappedDeclaredLicensesTable` (yes, the name is a real monster!)
- the actual string then lives in `UnmappedDeclaredLicensesTable`

api/v1/model/src/commonMain/kotlin/Package.kt
Etsija: It was kept there for not making the commit a breaking one. But I guess we already decided that the filter parameter can be changed, because the UI is its only customer currently, so I did what you asked for and removed `processedDeclaredLicense` from package filters.

core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
```diff
@@ -2943,23 +2944,23 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
                             processedDeclaredLicense = ProcessedDeclaredLicense(
                                 "LGPL-2.1-or-later",
                                 emptyMap(),
-                                emptySet()
+                                setOf("custom-license", licenseUrl)
```
sschuberth: While at it, could you please make all arguments named in these three "processedDeclaredLicense = ProcessedDeclaredLicense()" calls, like in line 1627?

core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
```diff
@@ -2943,23 +2944,23 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
...
+                                setOf(licenseUrl)
                             )
                         ),
                         dbExtension.fixtures.generatePackage(
                             Identifier("NPM", "", "example", "1.0"),
                             processedDeclaredLicense = ProcessedDeclaredLicense(
                                 "MIT",
                                 emptyMap(),
-                                emptySet()
+                                setOf("MIT")
```
sschuberth: Using "MIT" as an unmapped license is unrealistic as it's an SPDX license. Please use something that really cannot be mapped to an SPDX license.

services/ort-run/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -784,7 +784,7 @@ class PackageServiceTest : WordSpec() {
                             processedDeclaredLicense = ProcessedDeclaredLicense(
                                 "MIT",
                                 emptyMap(),
-                                emptySet()
+                                setOf("MIT")
```
sschuberth: See above, "MIT" is a bad choice here.

services/ort-run/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -800,10 +800,47 @@ class PackageServiceTest : WordSpec() {
...
+
+        "getUnmappedDeclaredLicenses" should {
+            "return the distinct unmapped declared licenses found in packages in the ORT run" {
+                val licenseUrl = "https://www.nuget.org/packages/CommandLineParser/2.9.1/license"
+                val ortRunId = createAnalyzerRunWithPackages(
+                    setOf(
+                        fixtures.generatePackage(
+                            Identifier("Maven", "com.example", "example", "1.0"),
+                            processedDeclaredLicense = ProcessedDeclaredLicense(
+                                "Apache-2.0",
```
sschuberth: Please also used named arguments here and in the other two calls to `ProcessedDeclaredLicense()` below.

core/src/main/kotlin/apiDocs/RunsDocs.kt
```diff
@@ -538,10 +538,11 @@ val getRunPackages: RouteConfig.() -> Unit = {
...
         }
 
-        queryParameter<String>("processedDeclaredLicense") {
-            description = "Defines the processed declared licenses for which packages are to be retrieved. This is a " +
+        queryParameter<String>("declaredLicense") {
+            description = "Defines the displayed declared license values for which packages are to be retrieved, " +
+                    "including processed declared licenses and unmapped declared license strings. This is a " +
                     "comma-separated string. Add a minus as the first item to exclude packages with the specified " +
-                    "license expressions, e.g. '-,MIT'."
+                    "license values, e.g. '-,MIT'."
```
sschuberth: Using the term "values" here, I guess to encompass both processed license expressions and unmapped license strings, is a bit unusual IMO. Maybe just say "declared licenses", or even just "licenses"?

core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
```diff
@@ -1763,42 +1764,33 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
...
+                val identifier3 = Identifier("Maven", "com.example", "other", "1.0")
 
                 dbExtension.fixtures.createAnalyzerRun(
                     analyzerJobId,
                     packages = setOf(
                         dbExtension.fixtures.generatePackage(
                             identifier1,
                             processedDeclaredLicense = ProcessedDeclaredLicense(
-                                "Apache-2.0 OR LGPL-2.1-or-later",
+                                "Apache-2.0",
```
sschuberth: Why this change? Expressions with operators should still be supported.

core/src/test/kotlin/api/RunsRouteIntegrationTest.kt
```diff
@@ -1763,42 +1764,33 @@ class RunsRouteIntegrationTest : AbstractIntegrationTest({
...
                                 emptyMap(),
                                 emptySet()
                             )
                         ),
                         dbExtension.fixtures.generatePackage(
                             identifier2,
                             processedDeclaredLicense = ProcessedDeclaredLicense(
-                                "Apache-2.0",
+                                "MIT",
```
sschuberth: Why change this license here? For the sake of keeping the diff small, I'd prefer to not do any unnecessary changes.

model/src/commonMain/kotlin/runs/Package.kt
```diff
@@ -51,8 +51,8 @@ data class PackageFilters(
     /** Substring filter for purl. Null if not set. */
     val purl: FilterOperatorAndValue<String>? = null,
 
-    /** Set of SPDX license expressions to filter with. Null if not set. */
-    val processedDeclaredLicense: FilterOperatorAndValue<Set<String>>? = null,
+    /** Set of declared license values to filter with. Null if not set. */
```
sschuberth, line 54: Also here, "values" is a term that we usually do not use in the context of licenses.

services/ort-run/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -457,29 +457,30 @@ class PackageServiceTest : WordSpec() {
...
+            "allow filtering by declared license" {
+                val licenseUrl = "https://www.nuget.org/packages/CommandLineParser/2.9.1/license"
                 val ortRunId = createAnalyzerRunWithPackages(
                     setOf(
                         fixtures.generatePackage(
-                            Identifier("Maven", "com.example", "example", "1.0"),
+                            Identifier("Maven", "com.example", "processed", "1.0"),
                             processedDeclaredLicense = ProcessedDeclaredLicense(
-                                "Apache-2.0 OR LGPL-2.1-or-later",
+                                "Apache-2.0",
```
sschuberth: Also for this test, more than necessary changes in input data seem to be done.

services/ort-run/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -489,27 +490,65 @@ class PackageServiceTest : WordSpec() {
...
+                        declaredLicense = FilterOperatorAndValue(
                             ComparisonOperator.IN,
-                            setOf("MIT", "Apache-2.0 OR LGPL-2.1-or-later")
+                            setOf("Apache-2.0", licenseUrl)
                         )
                     )
                 )
 
-                results.data shouldHaveSize 2
+                results.data.map { it.identifier.name } shouldBe listOf("processed", "unmapped")
```
sschuberth: Prefer: `shouldContainExactly("processed", "unmapped")`

services/ort-run/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -489,27 +490,65 @@ class PackageServiceTest : WordSpec() {
...
-                    processedDeclaredLicense.spdxExpression shouldBe "Apache-2.0 OR LGPL-2.1-or-later"
-                }
+            "allow excluding by declared license" {
+                val licenseUrl = "https://www.nuget.org/packages/CommandLineParser/2.9.1/license"
+                val ortRunId = createAnalyzerRunWithPackages(
+                    setOf(
+                        fixtures.generatePackage(
+                            Identifier("Maven", "com.example", "processed", "1.0"),
+                            processedDeclaredLicense = ProcessedDeclaredLicense(
+                                "Apache-2.0",
```
sschuberth: Please use named arguments here an below.

services/ort-run/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -551,6 +551,65 @@ class PackageServiceTest : WordSpec() {
...
+                    ortRunId,
+                    filters = PackageFilters(
+                        declaredLicense = FilterOperatorAndValue(
+                            ComparisonOperator.IN,
+                            setOf("Apache-2.0")
+                        )
+                    )
+                )
+
+                results.data.map { it.identifier.name } shouldBe listOf("mapped")
```
sschuberth: Prefer `shouldContainExactly()`.

services/ort-run/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -551,6 +551,65 @@ class PackageServiceTest : WordSpec() {
...
+                    ListQueryParameters(listOf(OrderField("identifier", OrderDirection.ASCENDING))),
+                    PackageFilters(
+                        declaredLicense = FilterOperatorAndValue(
+                            ComparisonOperator.NOT_IN,
+                            setOf("custom-foo")
+                        )
+                    )
+                )
+
+                results.data.map { it.identifier.name } shouldBe listOf("mapped", "other")
```
sschuberth: Prefer `shouldContainExactly()`.

services/ort-run/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -841,6 +900,24 @@ class PackageServiceTest : WordSpec() {
...
+                    Identifier("Maven", "com.example", "mapped", "1.0"),
+                    declaredLicenses = setOf("custom-foo"),
+                    processedDeclaredLicense = ProcessedDeclaredLicense(null, emptyMap(), setOf("custom-foo"))
+                )
+
+                val ortRunId = createAnalyzerRunWithPackages(setOf(pkgMit, pkgMapped)).id
+
+                addLicenseMappingCuration(ortRunId, pkgMapped.identifier, mapOf("custom-foo" to "Apache-2.0"))
+
+                service.getProcessedDeclaredLicenses(ortRunId) shouldBe listOf("Apache-2.0", "MIT")
```
sschuberth: Prefer `shouldContainExactly()`.

services/ort-run/src/test/kotlin/PackageServiceTest.kt
```diff
@@ -879,9 +956,48 @@ class PackageServiceTest : WordSpec() {
...
+                    Identifier("Maven", "com.example", "other", "1.0"),
+                    declaredLicenses = setOf("other-unmapped"),
+                    processedDeclaredLicense = ProcessedDeclaredLicense(null, emptyMap(), setOf("other-unmapped"))
+                )
+
+                val ortRunId = createAnalyzerRunWithPackages(setOf(pkgMapped, pkgOther)).id
+
+                addLicenseMappingCuration(ortRunId, pkgMapped.identifier, mapOf("custom-foo" to "Apache-2.0"))
+
+                service.getUnmappedDeclaredLicenses(ortRunId) shouldBe listOf("other-unmapped")
```
sschuberth: Prefer `shouldContainExactly()`.

sschuberth (dismissed):

services/ort-run/src/test/kotlin/PackageServiceTest.kt
Etsija: Gone with fixing as per the previous comment, back to original test structure.

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -35,6 +35,7 @@ import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.ProcessedDecl
 import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.ShortestDependencyPathsTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.analyzerrun.UnmappedDeclaredLicensesTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.resolvedconfiguration.CuratedPackagesTable
+import org.eclipse.apoapsis.ortserver.dao.tables.shared.IdentifierDao
```
mnonnenmacher, line 38: Commit message:
* This commit changes only the `PackageService`, it's unclear what "`/licenses` option endpoint" the commit message refers to. 
* I don't like the "SQL fast path" wording, it sounds smart but doesn't clarify anything. What's the difference between packages with curations and packages on that "fast path"?

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -79,6 +81,8 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
                 .select(PackagesTable.id)
                 .where { AnalyzerJobsTable.ortRunId eq ortRunId }
 
+            val curationsMap by lazy { CuratedPackagesTable.getForOrtRunId(ortRunId) }
```
mnonnenmacher, line 90: Why is the `listForOrtRunId` function changed in this commit? According to the commit message this seems unrelated.

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -125,10 +129,66 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
...
+                    val unaffected = if (affectedIds.isEmpty()) {
+                        sqlFilter
+                    } else {
+                        (PackagesTable.id notInList affectedIds) and sqlFilter
+                    }
+                    if (affectedMatchingIds.isEmpty()) {
+                        unaffected
+                    } else {
+                        unaffected or (PackagesTable.id inList affectedMatchingIds)
+                    }
```
mnonnenmacher, line 168: There should be empty lines between the three code blocks in this `andWhere` for better overview.

Also, the new code is very hard to understand, there should be at least some explanatory code comments. Same in the functions below.

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -125,10 +129,66 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
...
+                            ) { acc, (_, curation) ->
+                                curation.mapToOrt().apply(acc)
+                            }
+
+                            dao.id to curatedOrtPackage.metadata.mapToModel().processedDeclaredLicense
+                        }
+                    }
+                }
+
+                val affectedIds = overrides.map { it.first }
```
mnonnenmacher: The variable needs a better name, "affected" by what? Same for `affectedMatchingIds` below. Same in the functions below.

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -125,10 +129,66 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
...
+                val affectedIdentifierIds = curationsMap
+                    .filterValues { list ->
+                        list.any { (_, curation) ->
+                            curation.data.declaredLicenseMapping.isNotEmpty()
+                        }
+                    }
+                    .keys
+                    .mapNotNull { IdentifierDao.findByIdentifier(it)?.id }
+
+                val overrides = if (affectedIdentifierIds.isEmpty()) {
```
mnonnenmacher: The variable needs a better name, what does `overrides` mean? Better name the variable after what data it actually contains than what it will be used for. Same in the functions below.

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -82,6 +87,7 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
                 .where { AnalyzerJobsTable.ortRunId eq ortRunId }
 
             val curationsMap by lazy { CuratedPackagesTable.getForOrtRunId(ortRunId) }
+            val overrides by lazy { computeDeclaredLicenseOverrides(ortRunId, curationsMap) }
```
mnonnenmacher: I don't get what's going on here, `listForOrtRunId` needs to apply all curations anyway, why is there special handling for curations with declared license mappings?

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -423,12 +282,120 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
...
+            overrides.flatMapTo(licenses) { it.curatedUnmappedLicenses }
 
             licenses.sortedWith(String.CASE_INSENSITIVE_ORDER)
         }
 }
 
+/**
+ * Layer the curation overlay on top of the SQL `declaredLicense` filter (which considers both the
+ * processed SPDX expression and unmapped declared license strings). Unaffected packages use the
+ * existing subquery; affected packages are matched against curated SPDX and curated unmapped values.
```
mnonnenmacher: These docs are weird:
* What's a "curation overlay"?
* What "existing subquery"?

They should be rewritten to explain in normal language what the function does.

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -423,12 +282,120 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
...
+        } else {
+            unaffected or (PackagesTable.id inList affectedMatchingIds)
+        }
+    }
+}
+
+/**
+ * Snapshot of the curated `processedDeclaredLicense` fields for a single package whose curations could
+ * change the values that SQL holds. Used to reconcile filters and license-option lists with what the
+ * package detail view actually shows, without materializing every package in the run.
```
mnonnenmacher: Same as with the docs above, the way this is written I don't find it helpful at all.

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -423,12 +282,120 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
...
+    val packageId: EntityID<Long>,
+    val curatedSpdxExpression: String?,
+    val curatedUnmappedLicenses: Set<String>
+)
+
+/**
+ * Compute curated `processedDeclaredLicense` overrides for the packages in [ortRunId] whose curations
+ * may change the declared-license SPDX expression or unmapped licenses. Only packages with at least one
+ * curation carrying a non-empty `declaredLicenseMapping` are materialized; everything else stays on the
+ * SQL fast path.
```
mnonnenmacher: Same issue again:
* What are "overrides"? What do they override and why?
* What does it mean that packages are "materialized"?
* What's the "SQL fast path"?

services/ort-run/src/main/kotlin/PackageService.kt
```diff
@@ -423,12 +282,120 @@ class PackageService(private val db: Database, private val ortRunService: OrtRun
...
+    val affectedPackageIds = PackagesTable.joinAnalyzerTables()
+        .select(PackagesTable.id)
+        .where {
+            (AnalyzerJobsTable.ortRunId eq ortRunId) and
+                (PackagesTable.identifierId inList affectedIdentifierIds)
+        }
+        .map { it[PackagesTable.id] }
+
+    return if (affectedPackageIds.isEmpty()) {
+        emptyList()
```
mnonnenmacher: Should use an early return as for `affectedIdentifierIds` above: `if (affectedPackageIds.isEmpty()) return emptyList()`

mnonnenmacher (changes requested):

services/ort-run/src/main/kotlin/PackageService.kt
Etsija: It was like that, but didn't pass `detekt`, which complained about too many returns in a function - and the original form of the function was very much more readable than this version. I am using `detekt` as my main guideline for back-end code style issues, so it needs some changes, as some of its settings seem unrealistically tight.

Etsija:
@mnonnenmacher the last refactoring commit was triggered by the LLM noticing that after the service fix commit, there were repeated code blocks, and `listForOrtRunId()` was too long (this was a `detekt` error). 

The AI refactored the code, removing repeated code structures and making the corresponding function shorter. I agree the refactoring was probably too ambitious and included a lot of AI slop, especially in wording and variable names - so I decided to drop the refactoring commit altogether, rename some variables, and introduce some clarifying code comments to the existing code.

There is now some repeated code in `PackageService`, so I could still do a minimal refactor to it if needed. As it is, the functionality and performance of the queries stays the same even without the refactor.

services/ort-run/src/main/kotlin/PackageService.kt
mnonnenmacher: It's ok to suppress such warnings sometimes, but in general they are a strong indication that the code needs refactoring. For this specific case I haven't checked what the configured limit is so I can't tell if it should be changed.

mnonnenmacher:
> @mnonnenmacher the last refactoring commit was triggered by the LLM noticing that after the service fix commit, there were repeated code blocks, and `listForOrtRunId()` was too long (this was a `detekt` error).
> 
> The AI refactored the code, removing repeated code structures and making the corresponding function shorter. I agree the refactoring was probably too ambitious and included a lot of AI slop, especially in wording and variable names - so I decided to drop the refactoring commit altogether, rename some variables, and introduce some clarifying code comments to the existing code.
> 
> There is now some repeated code in `PackageService`, so I could still do a minimal refactor to it if needed. As it is, the functionality and performance of the queries stays the same even without the refactor.

The functions are still very long now and contain some duplication so a refactoring would be beneficial. However, the code will stay complex until something like https://github.com/eclipse-apoapsis/ort-server/discussions/4937 is implemented and just moving code to functions does not necessarily improve readability. So I would suggest to make some small refactoring commits on top (I'm also fine do to them in a follow-up PR) that one-by-one address some unnecessary duplications in smaller easier to review commits. For example, this code that applies curations appears multiple times and should be moved to a function:

```kotlin
                        val modelPackage = dao.mapToModel()
                        val curations = curationsMap[modelPackage.identifier].orEmpty()
                        val curatedOrtPackage = curations.fold(
                            OrtCuratedPackage(modelPackage.mapToOrt())
                        ) { acc, (_, curation) ->
                            curation.mapToOrt().apply(acc)
                        }
```

Etsija:
Two small refactoring commits.

mnonnenmacher (approved):

Issue #501, opened 2024-06-25
Admin configuration for available third-party tools


mnonnenmacher:
Currently, the used third-party tools like package managers, advisors, scanners, or package curation providers can be freely configured via the API when creating an ORT run. This is problematic because:

* Some tool configurations support setting arbitrary URLs which can be a security issue.
* The server admins have no control over which user can use which tools.

The current solution to this issue is the parameter validation script executed by the config worker. This script allows to modify any incoming ORT run requests to set default values or replace or remove values. While this approach is very flexible it has also some downsides:

* Writing and maintaining the Kotlin script can be difficult for new adopters and requires implementing tests to ensure that it behaves correctly.
* For the user it is intransparent how the script operates: Without additional documentation the only way to figure out how requests are modified is try and error.
* It is not possible to implement a UI to configure the available tools.
* It is not possible to implement an API endpoint that provides information about how requests are modified.

To improve the situation the API should be changed to allow the server admin to configure the available tools, and to configure which organizations have access to which tools. This would allow to add API endpoints to requests the available tools and also to implement an admin UI for the configuration.

With the example of advisors, this is how it could work:

Currently, the advisors can freely be configured via the `advisors` and `config` properties of the `AdvisorJobConfiguration`. The `config` property exposes ORT's `PluginConfiguration` object which can for example allow to configure the URL of an advisor.

Instead, let the admin configure a list of available pre-configured advisors that the users can choose from:

* The available advisor configurations are stored in the database.
* The available advisor configurations can be requested via the API and the user can provide the identifier of the configuration in the `AdvisorJobConfiguration` (this endpoint could be used by the UI as well).
* The admins could configure which organizations have access to which advisor.

sschuberth:
Just sharing my thoughts how this should work from a UI perspective: An admin of an ORT server installation should be able to see all e.g. ORT advisor plugins that are available on the classpath in a list, and from this list the admin should be able to select which of these advisors are made available for each organization. The API should support this UI requirement.

mnonnenmacher:
> admin should be able to select which of these advisors are made available for each organization.

In addition to that it should also be possible to configure the plugin options and to have multiple different plugin configurations for the same plugin. This would be required, for example, to configure different FossId instances for different organizations.

An exception to that should be the package managers, here it should still be possible for users to configure them via the .ort.yml file (and probably this is good enough and there is no need to duplicate those configuration options in the API like it is currently the case).

[...]

PR #5270, opened 2026-06-12
fix(ui): Prefer fixed plugin option values over values from last run

mnonnenmacher:
If a plugin options is set to a fixed value in a plugin template, that value is enforced by the backend: If a different value for that option is sent to the API, the run request is rejected.

This could cause issues when starting a run with "rerun" and plugin templates were changed in the meantime, because a plugin option used for the last run may now not be allowed anymore.

To fix this, always prefer fixed plugin option values over values used by the last run.

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
@@ -255,7 +255,8 @@ export function defaultValues(
               packageCurationProviderFormValues.selectedPluginIds,
             packageCurationProviderConfig: mergePluginConfigs(
               packageCurationProviderFormValues.config,
-              packageCurationProviderPluginDefaultValues
+              packageCurationProviderPluginDefaultValues,
+              packageCurationProviderPlugins
             ),
             keepAliveWorker:
               (ortRun.jobConfigs.analyzer?.keepAliveWorker && isSuperuser) ||
@@ -273,7 +274,8 @@ export function defaultValues(
               baseDefaults.jobConfigs.advisor.advisors,
             config: mergePluginConfigs(
               ortRun?.jobConfigs?.advisor?.config,
-              advisorPluginDefaultValues
+              advisorPluginDefaultValues,
+              advisorPlugins
             ),
             keepAliveWorker:
               (ortRun.jobConfigs.advisor?.keepAliveWorker && isSuperuser) ||
@@ -299,7 +301,8 @@ export function defaultValues(
             ),
             config: mergePluginConfigs(
               ortRun.jobConfigs.scanner?.config,
-              scannerPluginDefaultValues
+              scannerPluginDefaultValues,
+              scannerPlugins
             ),
           },
           evaluator: {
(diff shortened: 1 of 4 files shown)


sschuberth:
> If a plugin options

"options" -> "option"

> To fix this, always prefer fixed plugin option values over values used by the last run.

Maybe explicitly add "... and do not reject the run".

Etsija:
I couldn't find anything suspicious from the UI code, but GPT-5.5 thinks it's found something:
                                                                                                                        
`ui/src/.../create-run/-components/plugin-utils.ts:175` — fixed required secrets are deleted and then fail client-side validation.                                                                                                          

The new fixed-option override deletes fixed SECRET values when `getPluginDefaultValues()` has no default for them. But `getPluginDefaultValues()` intentionally ignores all secret defaults, and `validateRequiredPluginOptions()` still requires selected required options even when `option.isFixed` is true. So a rerun with a selected plugin that has an admin-fixed required secret will now end up with `config.<plugin>.secrets.<name>` removed, the field is disabled in the UI, and form validation blocks submission. 

This can break reruns even when the template did not change.

Please either skip required validation for fixed options that are supplied by the server/template, or otherwise      
preserve/provide a valid representation for fixed required secrets, and add a regression test with `type: 'SECRET',   
isFixed: true, and isRequired: true`.

- - -
I did verify the finding by:

1. Specifying a secret for the SCANOSS API key _and_ setting it to final:

Screenshot from 2026-06-16 08-10-45

2. Now when I start a new run (using rerun, but I believe this would also happen for a new run from scratch), the API key secret is locked and empty, cannot be changed:

Screenshot from 2026-06-16 08-13-48

GPT 5.5 was wrong: the form doesn't fail validation. It allows the run to be created, but the result is, SCANOSS fails with this weird issue without an originator:

Screenshot from 2026-06-16 08-22-59

When the API key secret in the SCANOSS template is made non-final, I can choose the correct API key in the run form, and SCANOSS passes normally with results.

I believe we've already discussed that a real fix would probably be to prevent any secrets from being "final" in the plugin templates section?

mnonnenmacher:
> I couldn't find anything suspicious from the UI code, but GPT-5.5 thinks it's found something [...]

The handling of secret options was improved in #5502, now this is not a problem anymore. However, the plugin manager is still not finished, so expect inconsistent behavior or problems when using plugin templates before #501 is closed.

> I believe we've already discussed that a real fix would probably be to prevent any secrets from being "final" in the plugin templates section?

No, admins must be able to set secrets and disallow users to overwrite them. But with #5502 the UI now shows explicitly that an admin-provided secret is used and handles that correctly (in both cases, when the value is final or not final).

sschuberth:
> > If a plugin options
> 
> "options" -> "option"
> 
> > To fix this, always prefer fixed plugin option values over values used by the last run.
> 
> Maybe explicitly add "... and do not reject the run".

FYI @mnonnenmacher, I believe my comments are still valid.

mnonnenmacher:
> > > If a plugin options
> > 
> > 
> > "options" -> "option"
> > > To fix this, always prefer fixed plugin option values over values used by the last run.
> > 
> > 
> > Maybe explicitly add "... and do not reject the run".
> 
> FYI @mnonnenmacher, I believe my comments are still valid.

I forgot to address those.

sschuberth (approved):

PR #5295, opened 2026-06-16
ui: Add package curation providers section to analyzer form

### LINKED ISSUE #5237: Add support to configure package curation providers
The API allows to configure package curation providers at `jobConfigs.analyzer.packageCurationProviders`:
https://eclipse-apoapsis.github.io/ort-server/api/post-repository-run

Add support to configure them in the UI when starting a run. The list of providers and their options should be taken from the plugins endpoint, similar to how it is done for advisors and scanners.

Etsija:
### Select providers from the list - the options are carried out from plugin options

Form:

Screenshot from 2026-06-17 13-08-12

Run payload: 

Screenshot from 2026-06-17 13-08-34

### The order of applying providers can be changed by drag-and-drop

The order of curation providers matters in the run payload: the providers on top of the form override curations of earlier providers in the list. Here, Default File is applied with higher priority than ClearlyDefined, as it's higher on the list.

Form:

Screenshot from 2026-06-17 13-09-11

Run payload:

Screenshot from 2026-06-17 13-09-32

### DIFF
--- a/ui/src/components/form/plugin-multi-select-field.tsx
+++ b/ui/src/components/form/plugin-multi-select-field.tsx
@@ -295,9 +295,6 @@ export const PluginMultiSelectField = <
                               {option.type}
                             </Badge>
                           </FormLabel>
-                          <FormDescription>
-                            {option.description}
-                          </FormDescription>
                           <FormControl>
                             {option.type === 'BOOLEAN' ? (
                               <Checkbox
@@ -384,6 +381,9 @@ export const PluginMultiSelectField = <
                               />
                             )}
                           </FormControl>
+                          <FormDescription>
+                            {option.description}
+                          </FormDescription>
                           {option.isFixed && (
                             <FormDescription className='font-semibold text-yellow-700'>
                               This option is set by an administrator and cannot
--- a/ui/src/components/form/plugin-multi-select-field.tsx
+++ b/ui/src/components/form/plugin-multi-select-field.tsx
@@ -114,7 +114,7 @@ export const PluginMultiSelectField = <
       render={({ field }) => (
         <FormItem
           className={cn(
-            'mb-4 flex flex-col justify-between rounded-lg border p-4',
+            'flex flex-col justify-between rounded-lg border p-4',
             className
           )}
         >
@@ -223,14 +223,14 @@ export const PluginMultiSelectField = <
                   }}
                 />
               </FormControl>
-              <div className='flex flex-col space-y-2'>
+              <div className='flex flex-col'>
                 <FormLabel className='font-normal'>
                   {plugin.displayName}
                 </FormLabel>
                 {plugin.summary != null && (
                   <MarkdownRenderer
                     markdown={plugin.summary}
-                    className='text-muted-foreground max-w-none pb-4 [&_p]:my-0'
+                    className='text-muted-foreground max-w-none pb-1 [&_p]:my-0'
                   />
                 )}
                 {scannerScopeName && field.value?.includes(plugin.id) && (
@@ -285,7 +285,7 @@ export const PluginMultiSelectField = <
                         `${configName}.${plugin.id}.${option.type === 'SECRET' ? 'secrets' : 'options'}.${option.name}` as Path<TFieldValues>
                       }
                       render={({ field }) => (
-                        <FormItem className='ml-4 flex flex-col space-y-0.5 pb-4'>
+                        <FormItem className='ml-4 flex flex-col pb-4'>
                           <FormLabel>
                             {option.name}
                             <Badge
--- a/ui/src/components/form/plugin-multi-select-field.tsx
+++ b/ui/src/components/form/plugin-multi-select-field.tsx
@@ -17,7 +17,10 @@
  * License-Filename: LICENSE
  */
 
+import { DragDropProvider } from '@dnd-kit/react';
+import { isSortable, useSortable } from '@dnd-kit/react/sortable';
 import { CheckedState } from '@radix-ui/react-checkbox';
+import { GripVerticalIcon } from 'lucide-react';
 import React from 'react';
 import {
   FieldPathByValue,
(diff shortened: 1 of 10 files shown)


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
```diff
@@ -35,7 +36,8 @@ export function defaultValues(
   ortRun: OrtRun | null,
   advisorPlugins: PreconfiguredPluginDescriptor[],
   scannerPlugins: PreconfiguredPluginDescriptor[],
-  isSuperuser: boolean
+  isSuperuser: boolean,
+  packageCurationProviderPlugins: PreconfiguredPluginDescriptor[] = []
```
sschuberth: Why does this have a default argument, but e.g. lines 37 and 38 don't?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/run-schema.ts
```diff
@@ -35,7 +35,8 @@ import {
 
 export const createRunFormSchema = (
   advisorPlugins: PreconfiguredPluginDescriptor[],
-  scannerPlugins: PreconfiguredPluginDescriptor[]
+  scannerPlugins: PreconfiguredPluginDescriptor[],
+  packageCurationProviderPlugins: PreconfiguredPluginDescriptor[] = []
```
sschuberth: Also here, why a default value for this new parameter?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/run-schema.ts
```diff
@@ -48,56 +49,81 @@ export const createRunFormSchema = (
...
-            Yarn: packageManagerOptionsSchema,
-            Yarn2: packageManagerOptionsSchema,
-          })
-          .refine((schema) => {
-            // Ensure that not both Gradle and GradleInspector are enabled at the same time.
-            return !(schema.Gradle.enabled && schema.GradleInspector.enabled);
-          }, '"Gradle Legacy" and "Gradle" cannot be enabled at the same time.'),
-      }),
+      analyzer: z
+        .object({
```
sschuberth, line 63: I guess this reformatting is enforced by prettier? It's a bit unfortunate, as it creates a massive diff for no reason.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/analyzer-fields.tsx
```diff
@@ -95,7 +104,7 @@ export const AnalyzerFields = ({
     permissions,
   });
 
-  const secrets = useSecrets({
+  const environmentSecrets = useSecrets({
```
sschuberth: Instead of renaming this, would it make more sense to give the newly introduced parameter a distinguished name?

ui/src/components/form/plugin-multi-select-field.tsx
```diff
@@ -114,7 +114,7 @@ export const PluginMultiSelectField = <
       render={({ field }) => (
         <FormItem
           className={cn(
-            'mb-4 flex flex-col justify-between rounded-lg border p-4',
+            'flex flex-col justify-between rounded-lg border p-4',
```
sschuberth: Commit message nit: Remember that "feat" commits cause the minor version to increase. Is this change (viewed in isolation) really justifying this? I guess not... so I'd make it a "chore".

ui/src/components/form/plugin-multi-select-field.tsx
```diff
@@ -384,6 +381,9 @@ export const PluginMultiSelectField = <
                               />
                             )}
                           </FormControl>
+                          <FormDescription>
```
sschuberth, line 451: This commit should also just be a "chore".

ui/src/components/form/plugin-multi-select-field.tsx
```diff
@@ -17,7 +17,10 @@
  * License-Filename: LICENSE
  */
 
+import { DragDropProvider } from '@dnd-kit/react';
```
sschuberth, line 20: Commit message:

> In "ordering mode", it raises the selected items to top of list

Is that still correct? I though the outcome of our discussion was to not group enabled items.

sschuberth (commented):
Mostly questions and comments from my side.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/run-schema.ts
Etsija: It is, yes.

ui/src/components/form/plugin-multi-select-field.tsx
Etsija: To this and next: we have always been on opposite sides on this. I align with Mikko: if it causes a visible change to the app, it is a feat, not chore.

From https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13

Screenshot from 2026-06-17 16-15-48

ui/src/components/form/plugin-multi-select-field.tsx
Etsija: Same counter-comment as above, and even more for this commit than the one before, because this completely changes the order of quite many UI elements (for advisor, analyzer, and scanner).

ui/src/components/form/plugin-multi-select-field.tsx
sschuberth: > if it causes a visible change to the app

What about fixing typos in user-facing descriptions as part of the UI? "feat", "fix", or "docs"? Just saying that certainly not every visible change is a feature. Where to draw the line is up to @mnonnenmacher as a project lead.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
Etsija: The default initialization is intentional here: there are still callers of `defaultValues()` that do not pass `packageCurationProviderPlugins`, and without `= []` the function gets `undefined`, which then causes errors when building the plugin default values. This could be an idea for a future refactoring pass on the form functions, once again.

ui/src/components/form/plugin-multi-select-field.tsx
sschuberth: Maybe an AI answer convinces you more than some random unauthoritative source 😉 (Note that I deliberate was asking the question unbiased as to the result):

image

mnonnenmacher: Regarding "style", our [own docs](https://github.com/eclipse-apoapsis/ort-server?tab=contributing-ov-file#types) say:
> style: Changes that do not affect the meaning of the code, e.g., formatting changes.

Angular itself doesn't have a "style" type (the AI made that up, see https://github.com/angular/angular/blob/main/contributing-docs/commit-message-guidelines.md#type) and the conventional commit specs only define what "fix" and "feat" should be used for.

I would argue that changing the appearance of the UI is changing the meaning of the code and mixing it with code style changes doesn't make sense (because such changes should appear in a user facing change log while code style changes don't really have to). However, calling it a feature would also be an exaggeration.

Looking at our list of types none of them really matches this type of change, maybe we should add a type like "tweak" for such minor changes to existing features. In the meantime I would be fine with either of the suggested types.

Etsija: I like "tweak"! For now, I reverted back to `feat`.

sschuberth: > maybe we should add a type like "tweak" for such minor changes to existing features.

To me, that's pretty much exactly what a "chore" is.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
sschuberth: My point is: Why not pass `[]` in call callers (there seem to be only very few, most of them in tests) instead of setting a default argument here?

Default arguments should not primarily be used for convenience to avoid tying, but for cases where it semantically makes sense that a value is the default. And I do not see how it makes sense to mostly have no package curation providers.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/analyzer-fields.tsx
```diff
@@ -379,10 +381,17 @@ export const AnalyzerFields = ({
             configName='jobConfigs.analyzer.packageCurationProviderConfig'
             label='Package curation providers'
             description={
-              <>Select the package curation providers enabled for this run.</>
+              <>
+                Select the package curation providers enabled for this run. Drag
```
sschuberth: I don't think we need "select" _and_ "enable". Also "for this run" should be clear from the run-form context. I propose: "Configure the package curation providers to use. Providers higher in the list take precedence over lower providers. Change the order of providers via drag & drop."

ui/src/components/form/plugin-multi-select-field.tsx
```diff
@@ -90,6 +174,18 @@ type PluginMultiSelectFieldProps<
...
+   * displayed in a sortable list and selected plugins are stored in the same
+   * order in which they appear in the list.
+   */
+  enableReordering?: boolean;
+  /**
+   * Show selected plugins first in their form value order before the user has
+   * reordered the list. This is intended for reruns, where an existing payload
+   * order should be reflected in the UI.
+   */
+  showSelectedPluginsFirst?: boolean;
```
sschuberth, line 188: It's _still_ unclear to me why this property and the feature to show selected plugins first is implemented. IIUC what we agreed on, it should never be used, also not for reruns. Or do I remember wrongly, @Etsija and @mnonnenmacher?

Etsija: Well, the minute the user selects curation providers _and_ changes their order by dnd, what should the order then be when RERUN is used? It doesn't make sense to show them in the ORIGINAL order anymore, as the user decided to change the order - then payload and UI would clash with each other. The only thing that made sense to me was that on RERUN, the selected curation providers are presented in the top, in the exact order as they were on the base run.

Mind you, this only applies when using rerun. The items in the list are not moving at all when several are selected - if their order isn't changed.

Etsija: I can explain the technical reasons behind this choice for you in a huddle if you like - rather not explaining it here.

sschuberth: Why can't a rerun show all providers, no matter if enabled or disabled, in exactly the same order as in the "base" run? Is that because the payload only contains the enabled providers, so we would not know where to eventually "inject" disabled providers in that list of enabled providers? I.e. the original positions of disabled providers gets lost?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
Etsija: You're right, it was an easy fix - fixed for both places, `default-values` and `run-schema`.

sschuberth (approved):

PR #5306, opened 2026-06-18
Support multiple containers in the Kubernetes Transport implementation

oheger-bosch:
This PR extends the Kubernetes Transport to allow the declaration of multiple containers. This enables advanced use cases like defining init containers or sidecar containers. Refer to the single commits for further information.


oheger-bosch:
@mnonnenmacher, do you have an idea about the failing build health check? This may be related to the introduction of the  fluent API client dependency in c170f77b, but I have no idea why it now suddenly pops up.

Copilot says that `client-java-api-fluent` bundles model classes that are also contained in `client-java-api`, which is the cause of the failure, e.g.:

```
Source set: main
\--- compile classpath
     +--- io/kubernetes/client/openapi/models/V1EmptyDirVolumeSource is provided by multiple dependencies: [io.kubernetes:client-java-api-fluent:26.0.0, io.kubernetes:client-java-api:26.0.0]
```
According to the AI, there is no artifact available containing only the fluent builder classes; so no easy way to work around the duplicated classes issues except for avoiding the fluent API. It therefore suggests to rewrite the `KubernetesMessageSender` class to only use the basic API.

Do you agree to this approach or do you have a better proposal?

mnonnenmacher:
> @mnonnenmacher, do you have an idea about the failing build health check? This may be related to the introduction of the fluent API client dependency in [c170f77](https://github.com/eclipse-apoapsis/ort-server/commit/c170f77b0567c1995b02dc0fd31d37cd129f6a4a), but I have no idea why it now suddenly pops up.
> 
> Copilot says that `client-java-api-fluent` bundles model classes that are also contained in `client-java-api`, which is the cause of the failure, e.g.:
> 
> ```
> Source set: main
> \--- compile classpath
>      +--- io/kubernetes/client/openapi/models/V1EmptyDirVolumeSource is provided by multiple dependencies: [io.kubernetes:client-java-api-fluent:26.0.0, io.kubernetes:client-java-api:26.0.0]
> ```
> 
> According to the AI, there is no artifact available containing only the fluent builder classes; so no easy way to work around the duplicated classes issues except for avoiding the fluent API. It therefore suggests to rewrite the `KubernetesMessageSender` class to only use the basic API.
> 
> Do you agree to this approach or do you have a better proposal?

The warnings just are just noise that makes it harder to find the actual problem, I didn't find a way to suppress them:

>  Advice for :transport:kubernetes
>  Existing dependencies which should be modified to be as indicated:
>    api(libs.kubernetesClientApiFluent) (was implementation)

You can also auto-fix such errors using the `fixDependencies` task:
https://github.com/autonomousapps/dependency-analysis-gradle-plugin#fix-dependency-issues-automatically

The duplicate class issue was reported upstream but was auto-closed, so I don't expect that it will be fixed. But IIUC the classes are generated and identical so technically it should be fine.
https://github.com/kubernetes-client/java/issues/2874

oheger-bosch:
> > Advice for :transport:kubernetes
> > Existing dependencies which should be modified to be as indicated:
> > api(libs.kubernetesClientApiFluent) (was implementation)
> 
> You can also auto-fix such errors using the `fixDependencies` task: https://github.com/autonomousapps/dependency-analysis-gradle-plugin#fix-dependency-issues-automatically

Great, this did the trick! Many thanks.

transport/kubernetes/src/main/kotlin/KubernetesSenderConfig.kt
```diff
@@ -266,9 +301,32 @@ data class KubernetesSenderConfig(
...
+                name = "${name.lowercase()}-container",
+                imageName = System.getenv("${prefix}IMAGE_NAME") ?: mainContainer.imageName,
+                imagePullPolicy = System.getenv("${prefix}IMAGE_PULL_POLICY") ?: mainContainer.imagePullPolicy,
+                commands = System.getenv("${prefix}COMMANDS")?.splitAtWhitespace() ?: mainContainer.commands,
+                args = System.getenv("${prefix}ARGS")?.splitAtWhitespace() ?: mainContainer.args,
+                cpuLimit = System.getenv("${prefix}CPU_LIMIT"),
+                cpuRequest = System.getenv("${prefix}CPU_REQUEST"),
+                memoryLimit = System.getenv("${prefix}MEMORY_LIMIT"),
+                memoryRequest = System.getenv("${prefix}MEMORY_REQUEST"),
+                volumeMounts = System.getenv("${prefix}VOLUME_MOUNTS").splitAt(splitCommaListRegex)
```
MarcelBochtler: Shouldn't this also use the `Config` abstraction for consistency and more importantly it would allow to configure this in other means than env variables.

transport/kubernetes/src/main/kotlin/VolumeMounts.kt
```diff
@@ -0,0 +1,191 @@
...
+     * Populate the properties of the passed in [mount] according to the data stored in this object. Use the given
+     * [index] to generate properties that need to be unique, such as a volume name. The sender implementation calls
+     * this function when it constructs the Kubernetes manifest for the pod to create.
+     */
+    fun initializeVolumeMount(mount: V1VolumeMount, index: Int): V1VolumeMount
+
+    /**
+     * Populate the properties of the passed in [volume] according to the data stored in this object. Use the given
+     * [index] to generate properties that need to be unique, such as a volume name. The sender implementation calls
+     *      * this function when it constructs the Kubernetes manifest for the pod to create.
```
MarcelBochtler: Nit: Remove the additional `     * `

transport/kubernetes/src/main/kotlin/KubernetesSenderConfig.kt
oheger-bosch: With the `Config` interface, there is no mechanism I am aware of that could handle lists of objects with an arbitrary number of elements, which would be needed to define the additional containers and their properties. Therefore, I went for the approach to have one property with a list of container names, and the properties of these containers then have to be looked up via environment variables.

transport/kubernetes/src/test/kotlin/KubernetesMessageSenderFactoryTest.kt
```diff
@@ -98,17 +98,13 @@ class KubernetesMessageSenderFactoryTest : StringSpec({
             args should containsInOrder("run", "all tests", "fast")
             backoffLimit shouldBe BACKOFF_LIMIT
             restartPolicy shouldBe RESTART_POLICY
-            secretVolumes should containsInOrder(
+            volumeMounts shouldContainInOrder listOf(
```
mnonnenmacher: I would prefer to keep the shorter `should containsInOrder`.

transport/kubernetes/src/test/kotlin/KubernetesMessageSenderFactoryTest.kt
```diff
@@ -172,7 +168,7 @@ class KubernetesMessageSenderFactoryTest : StringSpec({
         val sender = MessageSenderFactory.createSender(AnalyzerEndpoint, configManager)
 
         sender.shouldBeTypeOf<KubernetesMessageSender<AnalyzerEndpoint>>()
-        sender.config.secretVolumes should containsInOrder(
+        sender.config.volumeMounts shouldContainInOrder listOf(
```
mnonnenmacher: I would prefer to keep the shorter `should containsInOrder`.

transport/kubernetes/src/test/kotlin/KubernetesMessageSenderFactoryTest.kt
```diff
@@ -193,7 +189,7 @@ class KubernetesMessageSenderFactoryTest : StringSpec({
 
         sender.shouldBeTypeOf<KubernetesMessageSender<AnalyzerEndpoint>>()
 
-        sender.config.pvcVolumes should containsInOrder(
+        sender.config.volumeMounts shouldContainInOrder listOf(
```
mnonnenmacher: I would prefer to keep the shorter `should containsInOrder`.

transport/kubernetes/src/test/kotlin/KubernetesMessageSenderFactoryTest.kt
```diff
@@ -214,7 +210,7 @@ class KubernetesMessageSenderFactoryTest : StringSpec({
 
         sender.shouldBeTypeOf<KubernetesMessageSender<AnalyzerEndpoint>>()
 
-        sender.config.emptyDirVolumes should containsInOrder(
+        sender.config.volumeMounts shouldContainInOrder listOf(
```
mnonnenmacher: I would prefer to keep the shorter `should containsInOrder`.

website/docs/admin-guide/infrastructure/transport/kubernetes.md
```diff
@@ -224,6 +227,31 @@ kubernetes.memory = 768M
...
+
+Declaring additional containers is done by providing a comma-separated list of container names in the `additionalContainers` property. While the properties of the main container are configured via the properties described above, additional containers take their configuration from environment variables: For each container, a set of environment variables starting with the name of the container in upper case followed by an underscore is expected. There are then corresponding variables for each property of the container, using snake-case for the property name instead of camelCase. For instance, for a container named _sidecar_, there can be variables like `SIDECAR_IMAGE_NAME` for the `imageName` property or `SIDECAR_CPU_LIMIT` for the `cpuLimit` property. These variables are optional; for missing variables, the value from the main container is used with the following exceptions:
```
mnonnenmacher: Please also describe where to find the list of supported config options.

mnonnenmacher (changes requested):

transport/kubernetes/src/test/kotlin/KubernetesMessageSenderFactoryTest.kt
oheger-bosch: Changed, although "should containsInOrder" reads strange to me.

website/docs/admin-guide/infrastructure/transport/kubernetes.md
oheger-bosch: I have restructured the documentation into sections for pod-related and container-related properties.

mnonnenmacher (approved):

PR #5327, opened 2026-06-22
Home screen

### LINKED ISSUE #1598: Rethink the home screen / allow to define custom favorites
The current home screen does not really "feel like home" as it's actually just the list of accessible organizations, but nothing customizable. To make users feel more accustomed, we should consider displaying "favorites", like "starred" runs / repositories / products / organizations on that page, and probably move the list of all accessible organizations elsewhere.

Background: I find myself often in the need to look at specific products in a given period of time, and while I could bookmark the deep link, having them accessible from the home screen would be more comfortable.

Etsija:
This PR implements a new user-tailored home page. It is now possible to mark organizations, products, repositories and runs as "favorite", and the favorite entities are shown in home page, along with recent runs created by the user.

Screenshot from 2026-06-23 10-58-15

Please see the commits for details. For reviewers, it is encouraged to take the PR branch to local for testing - the first implementation is front-end only, so that can easily be done without recompiling the back-end / building docker images.

### DIFF
--- a/ui/src/routes/-components/home-recent-runs-section.tsx
+++ b/ui/src/routes/-components/home-recent-runs-section.tsx
@@ -1,0 +1,209 @@
+/*
+ * Copyright (C) 2026 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import { useQuery } from '@tanstack/react-query';
+import { Link, type LinkProps } from '@tanstack/react-router';
+import { AxiosError } from 'axios';
+import { PlayCircle } from 'lucide-react';
+import { useEffect } from 'react';
+
+import { OrtRunStatus } from '@/api';
+import {
+  getRepositoryRunOptions,
+  getRunStatisticsOptions,
+} from '@/api/@tanstack/react-query.gen';
+import { ItemCounts } from '@/components/item-counts';
+import { OrtRunJobStatus } from '@/components/ort-run-job-status';
+import { RunDuration } from '@/components/run-duration';
+import { TimestampWithUTC } from '@/components/timestamp-with-utc';
+import { Badge } from '@/components/ui/badge';
+import {
+  Card,
+  CardContent,
+  CardDescription,
+  CardHeader,
+  CardTitle,
+} from '@/components/ui/card';
+import { config } from '@/config';
+import { getStatusBackgroundColor } from '@/helpers/get-status-class';
+import { isJobFinished } from '@/helpers/job-helpers';
+import { RecentRunItem, useHomeRecentRunActions } from '@/providers/home-data';
+import { HomeEmptyState } from './home-empty-state';
+
+const isNotFoundError = (error: unknown) =>
+  error instanceof AxiosError &&
+  (error.status ?? error.response?.status) === 404;
+
+// Stop polling once a run has reached a terminal state, matching the behavior
+// of the run details page.
+const isRunFinished = (status: OrtRunStatus | undefined) =>
+  status === 'FINISHED' ||
+  status === 'FINISHED_WITH_ISSUES' ||
+  status === 'FAILED';
+
+/** Render a recent run and remove it from the list if the run is deleted. */
+const RecentRunListItem = ({ recentRun }: { recentRun: RecentRunItem }) => {
+  const { removeRecentRun } = useHomeRecentRunActions();
+  const runQuery = useQuery({
+    ...getRepositoryRunOptions({
+      path: {
+        repositoryId: recentRun.repositoryId,
+        ortRunIndex: recentRun.runIndex,
+      },
+    }),
+    refetchInterval: (query) =>
+      isRunFinished(query.state.data?.status) ? false : config.pollInterval,
+    retry: (failureCount, error) => !isNotFoundError(error) && failureCount < 3,
+  });
+
+  useEffect(() => {
+    if (isNotFoundError(runQuery.error)) {
+      removeRecentRun(recentRun.id);
+    }
+  }, [recentRun.id, removeRecentRun, runQuery.error]);
+
+  const run = runQuery.data;
+  const statistics = useQuery({
+    ...getRunStatisticsOptions({
+      path: { runId: run?.id ?? recentRun.runId },
+    }),
+    enabled: Boolean(run?.jobs) && !isNotFoundError(runQuery.error),
+    refetchInterval: isRunFinished(run?.status) ? false : config.pollInterval,
+  });
+
+  const status = run?.status ?? recentRun.status;
+  const createdAt = run?.createdAt ?? recentRun.createdAt;
+  const params = {
+    orgId: recentRun.organizationId.toString(),
+    productId: recentRun.productId.toString(),
+    repoId: recentRun.repositoryId.toString(),
+    runIndex: recentRun.runIndex.toString(),
+  };
+
+  return (
+    <li className='space-y-1 rounded-lg border px-3 py-2 text-sm'>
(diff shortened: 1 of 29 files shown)


ui/src/routes/index.tsx
```diff
@@ -17,226 +17,49 @@
...
-            };
-          }}
-        />
-      </CardContent>
-    </Card>
+    <div className='mx-auto flex w-full max-w-7xl flex-col gap-4'>
+      <div>
+        <h1 className='text-3xl font-bold tracking-tight'>
+          Welcome to {productName}
+          {userDisplayName && `, ${userDisplayName}`}
```
sschuberth: Don't we always have a `userDisplayName`? If we don't, if there something we can and should fall back to here, like the email address?

sschuberth: BTW, I really like this friendly greeting, it's refreshing to have it in a "grave" compliance tool 😸 As such, I'd even consider adding an exclamation mark here!

Etsija: > Don't we always have a `userDisplayName`? If we don't, if there something we can and should fall back to here, like the email address?

Technically, the way back-end has been built, we don't: 
```typescript
  const userDisplayName =
    user.fullName || user.username || user.user?.profile.email;
```
is still `string || undefined`, even after this or operation.

Etsija: Moreover, each of `fullName`, `username`, and `email` can still be `undefined`, so there's nothing we can fall safely back to.

Etsija: > BTW, I really like this friendly greeting, it's refreshing to have it in a "grave" compliance tool 😸 As such, I'd even consider adding an exclamation mark here!

I had the same idea - done!

ui/src/routes/organizations/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+} from '@/components/ui/tooltip';
+import { useIsSuperuser } from '@/hooks/use-authorization';
+import { toastError } from '@/lib/toast';
+import {
+  filterByNameSearchParameterSchema,
+  paginationSearchParameterSchema,
+} from '@/schemas';
+import { useTablePrefsStore } from '@/store/table-prefs.store';
+
+// Fetch the default page size to loader from the store.
```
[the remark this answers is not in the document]
Etsija: Irrelevant for this PR.

ui/src/routes/organizations/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+  loader: async ({ context: { queryClient }, deps: { page, pageSize } }) => {
+    queryClient.prefetchQuery({
+      ...getOrganizationsOptions({
+        query: {
+          limit: pageSize || defaultPageSize,
+          offset: page ? (page - 1) * (pageSize || defaultPageSize) : 0,
+        },
+      }),
+    });
+  },
```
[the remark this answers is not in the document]
Etsija: Irrelevant for this PR.

mnonnenmacher:
There should be some hint to the user that the data is only stored in the browser to avoid wrong expectations and resulting frustration.

sschuberth:
> There should be some hint to the user that the data is only stored in the browser to avoid wrong expectations and resulting frustration.

I guess as only favorites are affected by this, that hint could go straight to the description in the favorites card, right?

ui/src/routes/organizations/$orgId/-components/organization-product-table.tsx
```diff
@@ -90,16 +98,28 @@ export const OrganizationProductTable = () => {
...
+                {row.original.name}
+              </Link>
+              {organization && (
+                <ProductFavoriteButton
+                  organization={organization}
+                  organizationId={row.original.organizationId}
+                  product={row.original}
+                  size='xs'
+                  variant='ghost'
+                  className='size-6 p-0'
```
mnonnenmacher, line 119: Why are the `size`, `variant`, and `className` parameters not defined within the buttons but repeated for each call? Not a change request, just a question out of curiosity because this is an omnipresent pattern in the UI code which contradicts DRY. For example, it's the same with link styles which are also not defined in a central place but repeated for every link.

mnonnenmacher:
> > There should be some hint to the user that the data is only stored in the browser to avoid wrong expectations and resulting frustration.
> 
> I guess as only favorites are affected by this, that hint could go straight to the description in the favorites card, right?

Maybe, but the recent runs list is also affected. I didn't make a concrete proposal because I had no good idea where this could go, I was hoping @Etsija has an idea.

sschuberth:
> Maybe, but the recent runs list is also affected.

Is it? I thought these are always determined dynamically, without storing anything.

Anyway, if more cards are affected, maybe putting the hint (which will be temporary only) to where @Etsija previously had the subtitle (i.e. directly below the title) could work.

mnonnenmacher:
> > Maybe, but the recent runs list is also affected.
> 
> Is it? I thought these are always determined dynamically, without storing anything.
> 
> Anyway, if more cards are affected, maybe putting the hint (which will be temporary only) to where @Etsija previously had the subtitle (i.e. directly below the title) could work.

I noticed for the runs the UI already contains such a hint:

> Runs started (max. 10) from this browser UI are stored locally and shown here.

Etsija:
Neither of the lists is persisted by back-end - there is no API support for it. The first implementation is local storage only.

ui/src/routes/organizations/$orgId/-components/organization-product-table.tsx
Etsija: shadcn-ui is a "headless" library, which means exactly what you're seeing: it wants to provide very minimal UI defaults, and lets the users specify their own very freely. So it is a deliberate design choice (and I like it - many UI libraries are too restricted and only provide some alternatives for the look and feel).

Usage of `className` for overriding many, many UI concepts is a very common, modern design pattern. `variant` meanwhile is used quite extensively on our UI.

By doing it like this, instead of hiding the look-and-feel behind the component API makes all code stay close to the UI element - I don't want to go and change the component from the library, it is much easier to read when these "modifiers" are close to where the component is used.

sschuberth:
> Neither of the lists is persisted by back-end - there is no API support for it. The first implementation is local storage only.

That's clear. The question (to me) is: What needs to be persisted for recent runs? I assumed nothing needs to be persisted on any storage as recent runs are just determined dynamically when the home page is loaded. Is that not the case?

Etsija:
> That's clear. The question (to me) is: What needs to be persisted for recent runs? I assumed nothing needs to be persisted on any storage as recent runs are just determined dynamically when the home page is loaded. Is that not the case?

I realized this wasn't made too clear on the commits, so I have reworded `feat(ui): Track recently started runs locally`. I also changed the wording on home page to make the "locally stored data only" -concept clearer to the user. Screenshot updated so we can discuss the exact wording.

ui/src/routes/-components/home-recent-runs-section.tsx
```diff
@@ -0,0 +1,209 @@
...
+  recentRuns: RecentRunItem[];
+}) => (
+  <Card>
+    <CardHeader>
+      <CardTitle className='flex items-center gap-2'>
+        <PlayCircle className='h-5 w-5 text-white' />
+        Recently started runs
+      </CardTitle>
+      <CardDescription>
+        Runs started recently (max. 10) from this browser UI are shown here.
```
sschuberth: With the new note at the top, the "from this browser UI" part can go away. I propose: "Runs started recently (max. 10) are shown here.".

ui/src/routes/-components/home-recent-runs-section.tsx
```diff
@@ -0,0 +1,209 @@
...
+              },
+            }}
+          />
+        )}
+      </div>
+    </li>
+  );
+};
+
+/** Render runs recently started from this browser UI. */
```
sschuberth, line 179: As a comment, it's probably ok to still refer to "this browser", but the "UI" part I still find awkward and would drop it.

ui/src/routes/index.tsx
```diff
@@ -17,226 +17,53 @@
...
-    </Card>
+    <div className='mx-auto flex w-full max-w-7xl flex-col gap-4'>
+      <div>
+        <h1 className='text-3xl font-bold tracking-tight'>
+          Welcome to {productName}
+          {userDisplayName && `, ${userDisplayName}`}!
+        </h1>
+        <p className='text-muted-foreground'>
+          Please note that the information displayed in this home page is only
+          shown and persisted in your current browser.
```
sschuberth: As discussed, this should be moved to a separate new card below the "Organizations" card, and be reworded like "Please note that the information displayed below is only persisted in your current browser.".

sschuberth:
> I assumed nothing needs to be persisted on any storage as recent runs are just determined dynamically when the home page is loaded. Is that not the case?

To make the answer more explicit here for myself: No, that's not the case. Due to a missing endpoint to query recent runs per user and / or performance reasons, local browser storage is used to "collect" runs at the point they are triggered to build a list of (max. 10) runs being shown in the UI.

sschuberth (approved):

PR #5333, opened 2026-06-22
ui: Add support to package configuration providers to ORT run create form

### LINKED ISSUE #5238: Add support to configure package configuration providers
The API allows to configure package configuration providers at `jobConfigs.evaluator.packageCurationProviders` and `jobConfigs.reporter.packageCurationProviders`:
https://eclipse-apoapsis.github.io/ort-server/api/post-repository-run

Add support to configure them in the UI when starting a run. The list of providers and their options should be taken from the plugins endpoint, similar to how it is done for advisors and scanners.

If both the evaluator and reporter jobs are enabled, only the package configuration providers from the evaluator job are taken into account by the backend. This should be reflected in the UI somehow, for example, by syncing both sections or by hiding the section in the reporter job if the evaluator job is enabled as well.

Etsija:
As requested, the reporter section is only rendered, when evaluator job is disabled. As with package curation providers, order of the package configuration providers matter, so drag-and-drop is supported and its behavior is similar to #5295.

Screenshot from 2026-06-23 05-56-14

Screenshot from 2026-06-23 05-56-37

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
@@ -37,7 +37,8 @@ export function defaultValues(
   advisorPlugins: PreconfiguredPluginDescriptor[],
   scannerPlugins: PreconfiguredPluginDescriptor[],
   isSuperuser: boolean,
-  packageCurationProviderPlugins: PreconfiguredPluginDescriptor[]
+  packageCurationProviderPlugins: PreconfiguredPluginDescriptor[],
+  packageConfigurationProviderPlugins: PreconfiguredPluginDescriptor[]
 ): CreateRunFormValues {
   /**
    * Constructs the default options for a package manager, either as a blank set of options
@@ -93,6 +94,16 @@ export function defaultValues(
   const packageCurationProviderFormValues = providerPluginConfigsToFormValues(
     ortRun?.jobConfigs.analyzer?.packageCurationProviders
   );
+  const packageConfigurationProviderPluginDefaultValues =
+    getPluginDefaultValues(packageConfigurationProviderPlugins);
+  const evaluatorPackageConfigurationProviderFormValues =
+    providerPluginConfigsToFormValues(
+      ortRun?.jobConfigs.evaluator?.packageConfigurationProviders
+    );
+  const reporterPackageConfigurationProviderFormValues =
+    providerPluginConfigsToFormValues(
+      ortRun?.jobConfigs.reporter?.packageConfigurationProviders
+    );
   const scannerDefaults = {
     scanners: ['ScanCode'],
     scannerScopes: {
@@ -170,12 +181,18 @@ export function defaultValues(
         copyrightGarbageFile: '',
         resolutionsFile: '',
         keepAliveWorker: false,
+        packageConfigurationProviders: [],
+        packageConfigurationProviderConfig:
+          packageConfigurationProviderPluginDefaultValues,
       },
       reporter: {
         enabled: true,
         formats: ['CycloneDX', 'SpdxDocument', 'WebApp'],
         deduplicateDependencyTree: false,
         keepAliveWorker: false,
+        packageConfigurationProviders: [],
+        packageConfigurationProviderConfig:
+          packageConfigurationProviderPluginDefaultValues,
       },
       notifier: {
         enabled: false,
@@ -289,6 +306,12 @@ export function defaultValues(
             enabled:
               ortRun.jobConfigs.evaluator !== undefined &&
               ortRun.jobConfigs.evaluator !== null,
+            packageConfigurationProviders:
+              evaluatorPackageConfigurationProviderFormValues.selectedPluginIds,
+            packageConfigurationProviderConfig: mergePluginConfigs(
+              evaluatorPackageConfigurationProviderFormValues.config,
+              packageConfigurationProviderPluginDefaultValues
+            ),
             keepAliveWorker:
               (ortRun.jobConfigs.evaluator?.keepAliveWorker && isSuperuser) ||
               baseDefaults.jobConfigs.evaluator.keepAliveWorker,
@@ -303,6 +326,12 @@ export function defaultValues(
               ) || baseDefaults.jobConfigs.reporter.formats,
             deduplicateDependencyTree:
               deduplicateDependencyTreeEnabled || undefined,
+            packageConfigurationProviders:
+              reporterPackageConfigurationProviderFormValues.selectedPluginIds,
+            packageConfigurationProviderConfig: mergePluginConfigs(
+              reporterPackageConfigurationProviderFormValues.config,
+              packageConfigurationProviderPluginDefaultValues
+            ),
             keepAliveWorker:
               (ortRun.jobConfigs.reporter?.keepAliveWorker && isSuperuser) ||
               baseDefaults.jobConfigs.reporter.keepAliveWorker,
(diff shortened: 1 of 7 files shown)


sschuberth:
For package configuration providers, like for [package curation providers](https://github.com/eclipse-apoapsis/ort-server/pull/5295), the order matters. Also see [this in ORT core](https://github.com/oss-review-toolkit/ort/blob/f517511332c853c80ab8206b49f42332aeefc35d/model/src/main/resources/reference.yml#L54-L59).

As such, I would have expected the drag-and-droppable UI in the screenshot also here. Or is the screenshot just outdated?

Etsija:
> For package configuration providers, like for [package curation providers](https://github.com/eclipse-apoapsis/ort-server/pull/5295), the order matters. Also see [this in ORT core](https://github.com/oss-review-toolkit/ort/blob/f517511332c853c80ab8206b49f42332aeefc35d/model/src/main/resources/reference.yml#L54-L59).
> 
> As such, I would have expected the drag-and-droppable UI in the screenshot also here. Or is the screenshot just outdated?

This is a misunderstanding, then, as I recall @mnonnenmacher saying the order didn't matter for these!

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
```diff
@@ -37,7 +37,8 @@ export function defaultValues(
   advisorPlugins: PreconfiguredPluginDescriptor[],
   scannerPlugins: PreconfiguredPluginDescriptor[],
   isSuperuser: boolean,
-  packageCurationProviderPlugins: PreconfiguredPluginDescriptor[]
+  packageCurationProviderPlugins: PreconfiguredPluginDescriptor[],
+  packageConfigurationProviderPlugins: PreconfiguredPluginDescriptor[] = []
```
sschuberth: I think we're going to have the same default argument discussion again here 😉

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
```diff
@@ -37,7 +37,8 @@ export function defaultValues(
   advisorPlugins: PreconfiguredPluginDescriptor[],
   scannerPlugins: PreconfiguredPluginDescriptor[],
   isSuperuser: boolean,
-  packageCurationProviderPlugins: PreconfiguredPluginDescriptor[]
+  packageCurationProviderPlugins: PreconfiguredPluginDescriptor[],
```
sschuberth, line 40: Commit message:

> Selected providers are submitted via
`jobConfigs.evaluator.packageConfigurationProviders`, or via
`jobConfigs.reporter.packageConfigurationProviders` when the evaluator job
is disabled.

So if both evaluator and reporter jobs are enabled, not both `jobConfigs.evaluator.packageConfigurationProviders` _and_ `jobConfigs.reporter.packageConfigurationProviders` are populated? Is that really how it's supposed to work, @mnonnenmacher?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/run-schema.ts
```diff
@@ -36,7 +36,8 @@ import {
 export const createRunFormSchema = (
   advisorPlugins: PreconfiguredPluginDescriptor[],
   scannerPlugins: PreconfiguredPluginDescriptor[],
-  packageCurationProviderPlugins: PreconfiguredPluginDescriptor[]
+  packageCurationProviderPlugins: PreconfiguredPluginDescriptor[],
+  packageConfigurationProviderPlugins: PreconfiguredPluginDescriptor[] = []
```
sschuberth: See above, no default argument should be passed here.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/run-schema.ts
```diff
@@ -244,3 +296,168 @@ export const flattenErrors = (
 
   return result;
 };
+
+if (import.meta.vitest) {
+  const { describe, expect, it } = import.meta.vitest;
+
+  const packageManagerIds = [
```
sschuberth: Do we really need to hard-code all of these just for testing?

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/run-schema.ts
```diff
@@ -244,3 +296,168 @@ export const flattenErrors = (
...
+          allowDynamicVersions: true,
+          skipExcluded: true,
+          keepAliveWorker: false,
+          packageCurationProviders: [],
+          packageCurationProviderConfig: {},
+          packageManagers: Object.fromEntries(
+            packageManagerIds.map((id) => [
+              id,
+              {
+                enabled: id !== 'Gradle',
```
sschuberth: Why this condition?

sschuberth (changes requested):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
Etsija: Oh s**t not again 🤣

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/run-schema.ts
Etsija: This was a nice catch, and the solution was quite elegant.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/run-schema.ts
Etsija: It's now removed once I fixed the test concept for your previous comment.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/create-run/-components/default-values.ts
Etsija: According to https://github.com/eclipse-apoapsis/ort-server/issues/5238 that's the case.

mnonnenmacher: Yes, to ensure that the evaluator and reporter output are consistent, they must use the same package configuration providers. Therefore, if the evaluator and reporter jobs are both enabled, only the package configuration providers configured for the evaluator job are taken into account and those configured in the reporter job are silently ignored. The logic is similar to the ORT CLI where the reporter only resolves package configurations if this hasn't happened before.

mnonnenmacher:
> As requested, the reporter section is only rendered, when evaluator job is disabled.

This could also be confusing to users in some situations. For example, if they configure the package configuration providers in the reporter job and then enable the evaluator job, will the configuration they made before just disappear? An alternative could be to keep the section in both jobs and automatically sync the content. However, we have to start somewhere, so to me it would be fine to keep the current approach and then iterate on that later if necessary.

Etsija:
> > As requested, the reporter section is only rendered, when evaluator job is disabled.
> 
> This could also be confusing to users in some situations. For example, if they configure the package configuration providers in the reporter job and then enable the evaluator job, will the configuration they made before just disappear? An alternative could be to keep the section in both jobs and automatically sync the content. However, we have to start somewhere, so to me it would be fine to keep the current approach and then iterate on that later if necessary.

I'm fine with keeping the current approach and collect user experience, then change later if it proves useful.

sschuberth (approved):

PR #5338, opened 2026-06-23
refactor: Move `Identifier` to shared API model

mnonnenmacher:
Components should not depend on the legacy API model but the shared API model instead.


sschuberth (approved):

oheger-bosch:
Just to make sure: Does this have any impact on the OpenAPI model and thus would be a breaking change?

Also, I was not aware that there is a "legacy API model". Are there plans for further restructurings?

mnonnenmacher:
> Just to make sure: Does this have any impact on the OpenAPI model and thus would be a breaking change?

No, that will stay the same.

> Also, I was not aware that there is a "legacy API model". Are there plans for further restructurings?

No, moving shared API classes is part of the component refactoring, this should have happened before the components started using the `Identifier`.

oheger-bosch (approved):

PR #5343, opened 2026-06-23
build(cli): Drop unused dynamic dependencies

mnonnenmacher:
Kotlin/Native inherently links platform-declared POSIX libraries like `libcrypt.so.1` regardless of whether the application actually needs them. This causes runtime failures on modern Linux distributions where legacy implementations like `libcrypt.so.1` have been deprecated and removed [1].

Instructing the linker to only include explicitly needed libraries safely strips out these dependencies, improving the portability of the Linux build of the CLI.

[1]: https://youtrack.jetbrains.com/projects/KT/issues/KT-55643

### DIFF
--- a/cli/build.gradle.kts
+++ b/cli/build.gradle.kts
@@ -40,7 +40,12 @@ kotlin {
         }
     }
 
-    linuxX64()
+    linuxX64 {
+        binaries.all {
+            linkerOpts("-Wl,--as-needed")
+        }
+    }
+
     macosArm64()
     @Suppress("deprecation") macosX64()
     mingwX64()


cli/build.gradle.kts
```diff
@@ -40,7 +40,12 @@ kotlin {
         }
     }
 
-    linuxX64()
+    linuxX64 {
+        binaries.all {
+            linkerOpts("-Wl,--as-needed")
```
sschuberth, line 45: Just as a side remark, very odd that this is not the default behavior... is something similar needed on other platforms, too?

sschuberth (approved):

cli/build.gradle.kts
mnonnenmacher: > Just as a side remark, very odd that this is not the default behavior... 

According to my research there is a risk that this could remove too much in certain scenarios.

> is something similar needed on other platforms, too?

Only if we see similar issues, like the one I found with `libcrypt.so.1`. Otherwise there is no reason to add this, the size of the binary barely goes down because these are dynamically linked libraries.

PR #5353, opened 2026-06-25
ui: Enable external unit tests while preserving inline tests

Etsija:
UI tests are inlined with the code and as such are scattered throughout the UI codebase in `helpers/`, `hooks/`, `lib/`, `providers/`, route local components, general components, ORT run creation helpers etc.

A simple way to divide the UI tests into two categories is by the file extension:
- tests in TypeScript (`.ts`) files are logical tests for pure TypeScript functions
- tests in JSX/React (`.tsx`) files are tests for rendering visual components

As these two categories use different Vitest testing principles and methods, the most natural and simple refactoring target directory structure for the current inlined tests is:
1. All TypeScript tests (tests for TypeScript function logic) are moved to `ui/tests/unit/logic`
2. All JSX tests (tests for component's rendering logic) are moved to `ui/tests/unit/components`

As the UI codebase already includes 126 directories, this is the simplest possible refactoring directory structure that doesn't add unnecessary complication - therefore, the author didn't see it worth while adding subdirectories under `unit/logic` and `unit/components` (currently, there are only 27 test files, so adding more directory structure for them would be an unnecessary overcomplication).

Whenever possible, the filename for the test refactored from the file it was inlined in has the same name, as in `calculate-duration.ts -> ui/tests/unit/logic/calculate-duration.test.ts` and `repository-runs-table.tsx -> ui/tests/unit/components/repository-runs-table.test.tsx`. This should make searching for a test easier.

This PR will start the series of PRs aimed for UI test refactoring by enabling the tests to be co-located in `ui/tests` and `ui/src`, and moving some tests to `ui/tests` as a proof of concept.


ui/src/helpers/calculate-duration.ts
```diff
@@ -19,7 +19,7 @@
 
 import { JobSummary, PagedResponseOrtRunSummary } from '@/api';
 
-function divmod(a: number, b: number): [number, number] {
+export function divmod(a: number, b: number): [number, number] {
```
sschuberth, line 22: > Prove the new test structure

Speaking of "proof": We should double-check that the _total_ number of tests before and after the full migration matches, to ensure that we did not accidentally exclude / not include any tests, e.g. due to wrong glob patterns.

Etsija: Yes, I do have the "before" count stored in my notes.

ui/tests/unit/logic/calculate-duration.test.ts
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (C) 2026 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
```
sschuberth: The copyright year in these "new" files should be the oldest copyright from any of the files where the code was moved from.

ui/tests/unit/logic/calculate-duration.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  expect(divmod(10, 3)).toStrictEqual([3, 1]);
+  expect(divmod(8, 2)).toStrictEqual([4, 0]);
+  expect(divmod(1, 1)).toStrictEqual([1, 0]);
+  expect(divmod(1, 0)).toStrictEqual([NaN, NaN]);
+  expect(divmod(-10, 3)).toStrictEqual([-3, -1]);
+  expect(divmod(10, -3)).toStrictEqual([-3, 1]);
+  expect(divmod(-10, -3)).toStrictEqual([3, -1]);
+});
+
+it('calculateDuration', () => {
```
sschuberth, line 24: I believe we should now start to maintain some ordering for tests. I propose alphabetical by function name under test / test title, somewhat mirroring the imports.

sschuberth: Same for other files as well, of course.

sschuberth (changes requested):

sschuberth (approved):

ui/tests/unit/logic/calculate-duration.test.ts
```diff
@@ -0,0 +1,51 @@
+/*
```
mnonnenmacher, line 1: Why not mirror the directory structure from `/src` and put these tests at `/tests/unit/helpers`? This should make it easier to find the matching tests for a file.

mnonnenmacher (dismissed):

ui/tests/unit/logic/calculate-duration.test.ts
Etsija: I prefer to keep this as simple as humanly possible, as our UI directory structure is already very (too) complicated, making it hard to find files (for example, in VS Code).

sschuberth: But how does using `ui/tests/unit/logic` make it any easier than using `/tests/unit/helpers`?

mnonnenmacher: I wanted to ask the same, isn't using `helpers` actually simpler in this case?

Etsija: I realized that as this PR starts a series of several, I should've provided more context, justification and information about the "end target", so I have now done it in the PR description. Hopefully that provides answers.

mnonnenmacher: > As the UI codebase already includes 126 directories, this is the simplest possible refactoring directory structure that doesn't add unnecessary complication - therefore, the author didn't see it worth while adding subdirectories under unit/logic and unit/components (currently, there are only 27 test files, so adding more directory structure for them would be an unnecessary overcomplication).

That sounds like just delaying the decision how to properly organize the tests until more tests are added. Anyway, I will remove my change request to not block the refactoring, I mainly added it to make sure my comment is seen before the PR is merged as it was already approved. I also think it would be good to get feedback from @lamppu so that the two main authors of the UI can make the decision together.

Etsija: I can wait for @lamppu to review the PR (I can explain the whole structure to her on Monday).

PR #5358, opened 2026-06-25
feat(compose): Migrate Keycloak setup to keycloak-config-cli

mnonnenmacher:
Replace the realm-import JSON file with a minimal, declarative realm configuration applied via keycloak-config-cli. The previous file mixed custom configuration with all of Keycloak's auto-generated defaults, making it hard to tell what was actually custom.

The setup now uses a dedicated `ort-server` realm (instead of the built-in `master` realm) defined in a concise YAML file containing only the custom clients, client scope, group, and users. A `keycloak-config` Compose service applies it to the running Keycloak, preserving Keycloak's own default entities.


website/docs/admin-guide/getting-started/docker-compose.md
```diff
@@ -125,17 +125,21 @@ Then execute the ORT Server in IntelliJ with the run configuration ["Run ORT Ser
...
 ```shell
-docker compose exec -it keycloak /bin/bash
-KC_HTTP_PORT=8081 /opt/keycloak/bin/kc.sh export --dir /opt/keycloak_init --users realm_file
+docker compose up keycloak-config
 ```
 
-Unfortunately, Keycloak keeps changing the order of entries in the exported JSON file, especially after version upgrades.
-To minimize the diff when making changes, it is recommended to export the configuration before making any changes and then again after applying the changes, and to put the changes in spearate commits.
-Otherwise, it can be hard to identify the actual changes.
+The `keycloak-config` service is idempotent: it runs once, reconciles the `ort-server` realm with the config file, and then exits.
```
sschuberth, line 138: Nit: Capitalized after ":"?

sschuberth (approved):

website/docs/admin-guide/getting-started/docker-compose.md
mnonnenmacher: I think lowercase is actually correct in this case.

PR #5382, opened 2026-06-30
Prepare Analyzer for execution in multiple phases

oheger-bosch:
This PR adds some functionality and refactorings that are needed to split the Analyzer execution into multiple phases with different configuration options. This split will be implemented in a follow-up PR.

### DIFF
--- a/workers/analyzer/src/main/kotlin/AnalyzerDownloader.kt
+++ b/workers/analyzer/src/main/kotlin/AnalyzerDownloader.kt
@@ -21,6 +21,8 @@ package org.eclipse.apoapsis.ortserver.workers.analyzer
 
 import java.io.File
 
+import kotlin.io.path.createTempDirectory
+
 import org.eclipse.apoapsis.ortserver.model.SubmoduleFetchStrategy
 
 import org.ossreviewtoolkit.downloader.VcsHost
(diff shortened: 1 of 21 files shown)


workers/common/src/main/kotlin/env/EnvironmentForkHelper.kt
```diff
@@ -109,12 +120,45 @@ object EnvironmentForkHelper {
         enableOrtStackTraces()
     }
 
+    /**
+     * Set up authentication information from the given [source] file and the provided [configManager]. Using this
+     * function, the data previously stored via [persistAuthenticationInfo] can be restored.
+     */
+    fun setupAuthentication(source: File, configManager: ConfigManager) {
+        val authInfo = source.inputStream().use { stream ->
```
sschuberth, line 128: Nit: Lambda parameter is a bit inconsistent compared to above, where "out" is used; so I would have expected "in" here.

workers/common/src/main/kotlin/env/ConfigFileBuilder.kt
```diff
@@ -190,7 +196,8 @@ class ConfigFileBuilder(
      * given [block].
      */
     suspend fun buildInUserHome(name: String, block: suspend PrintWriter.() -> Unit) {
-        build(File(System.getProperty("user.home"), name), block)
+        val target = userHomeDir?.let { File(it, name) } ?: File(System.getProperty("user.home"), name)
```
sschuberth: More compact: `val target = (userHomeDir ?: Os.userHomeDirectory) / name`

workers/common/src/main/kotlin/env/EnvironmentForkHelper.kt
oheger-bosch: `in` cannot be used, it is reserved in Kotlin.

workers/analyzer/src/main/kotlin/AnalyzerDownloader.kt
```diff
@@ -21,29 +21,62 @@ package org.eclipse.apoapsis.ortserver.workers.analyzer
...
+        /**
+         * Look up the directory of the download for the given [runId] under the given [root] directory based on the
+         * naming conventions applied by this class. Return *null* if the directory cannot be determined or an error
+         * occurs. This is useful if the download was done in a different phase of the analyzer run.
+         */
+        fun findDownloadDir(root: File, runId: Long): File? =
+            runCatching {
+                val dirName = downloadDirName(runId)
+
+                root.listFiles { file -> file.isDirectory && file.name.contains(dirName) }.single()
```
sschuberth: Prefer the `in` operator over `contains`.

workers/common/src/main/kotlin/env/EnvironmentForkHelper.kt
sschuberth: Well, `in` _could_ be used by putting it into backticks, not saying it's a good idea. My main point was about alignment / symmetry; then both lambda parameters should probably be called "stream", or just stick with the implicit "it" to begin with.

workers/common/src/main/kotlin/env/ConfigFileBuilder.kt
oheger-bosch: Changed. However, this required modifications on some tests that were expecting the user home directory to be obtained from system properties.

workers/common/src/main/kotlin/env/EnvironmentForkHelper.kt
oheger-bosch: TBH, my personal feeling for symmetry is not strongly hurt by this discrepancy. Do you insist?

workers/common/src/main/kotlin/env/ConfigFileBuilder.kt
sschuberth: I've seen the test changes, but these IMO look a bit simpler now as well 👍🏻

workers/common/src/main/kotlin/env/EnvironmentForkHelper.kt
sschuberth: I don't insist. I just don't understand the logic 😅

oheger-bosch: No, logic - just a combination of a small screen and a limited context.

workers/analyzer/src/main/kotlin/AnalyzerDownloader.kt
```diff
@@ -21,29 +21,62 @@ package org.eclipse.apoapsis.ortserver.workers.analyzer
...
         path: String = "",
-        submoduleFetchStrategy: SubmoduleFetchStrategy? = SubmoduleFetchStrategy.FULLY_RECURSIVE
+        submoduleFetchStrategy: SubmoduleFetchStrategy? = SubmoduleFetchStrategy.FULLY_RECURSIVE,
+        targetDir: File? = null,
+        runId: Long? = null
     ): DownloadResult {
         logger.info("Downloading repository '$repositoryUrl' revision '$revision'.")
 
-        val outputDir = createOrtTempDir("analyzer-worker")
+        val outputDir = createTempDirectory(targetDir?.toPath(), downloadDirName(runId)).toFile()
```
MarcelBochtler, line 79: What is the reason for changing the tmp dir path?
If a user mounts e.g. tmpfs to ORT's temp dir, this might impact them.
I don't think that it is problematic, I'm just curious why the change is required.

oheger-bosch: If the Analyzer runs in multiple phases, the clone of the repository has to be done on the shared volume. Cloning happens in the preparation phase, the analysis is then done in a later phase in a different container.

MarcelBochtler (approved):

PR #5399, opened 2026-07-01
ui: Fixes for dependency graph

### LINKED ISSUE #5392: The "Dependency Graph" renders vertical lines that lead to nowhere
In this view, there are several vertical lines / line segments being rendering even if no more siblings follow:

Image

Compare this to the WebApp rendering:

Image

### LINKED ISSUE #5393: The "Dependency Graph" does not render branch segments
See the marked "line segments":

Image

These should actually be rendered as "Branch segments" / "t-junctions" as in the WebApp report:

Image

### LINKED ISSUE #5394: The "Dependency Graph" search omits transitive dependencies of matches
It seems that only the path _to_ the matching dependency is rendered, but the transitive dependencies of the match are being omitted:

Image

Instead, it should look like in the WebApp report, where the graph is intact, and only the path to the finding is expanded, and all other branches (including transitive dependencies of matches) are collapsed:

Image

Etsija:
This PR fixes some bugs for the run dependency graph, while aligning it closer to the WebApp report.

### More tree-like view with branches and stateful +/- icons

Screenshot from 2026-07-01 10-03-31

### Search now opens all matches, while keeping the rest of the graph intact

Screenshot from 2026-07-01 10-04-33

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/dependencies/-components/dependency-tree-node.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/dependencies/-components/dependency-tree-node.tsx
@@ -27,7 +27,6 @@ import {
 import type { PackageIdType } from '@/schemas';
 import {
   formatDependencyGraphPackageLabel,
-  matchesSearch,
   type AdjacencyMap,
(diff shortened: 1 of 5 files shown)


sschuberth:
Screenshots LGTM, please @lamppu also look at the code.

lamppu (approved):

PR #5416, opened 2026-07-03
ui: Show API URL on auth token page

### LINKED ISSUE #5402: Show the API URL on the "Authentication Token" page
For user convenience, we should not only explain how to use OSC with `--token`, but also provide the correct `--url` on the "Authentication Token" page at the `/profile/token` route.

So instead of saying

>Use it to sign in with [ORT Server CLI](https://eclipse-apoapsis.github.io/ort-server/docs/user-guide/cli/getting-started) using the `--token` option.

we should say something like

> Use it to sign in with [ORT Server CLI](https://eclipse-apoapsis.github.io/ort-server/docs/user-guide/cli/getting-started) using the `--url https://api.ort-server.io` and `--token <token>` options.

Etsija:
Screenshot from 2026-07-03 12-30-12

### DIFF
--- a/ui/src/routes/profile/token/index.tsx
+++ b/ui/src/routes/profile/token/index.tsx
@@ -24,6 +24,7 @@ import { OidcClient } from 'oidc-client-ts';
 import { useCallback, useState } from 'react';
 
 import { getCliOidcConfigOptions } from '@/api/@tanstack/react-query.gen';
+import { CopyToClipboard } from '@/components/copy-to-clipboard';
 import { Button } from '@/components/ui/button';
 import {
   Card,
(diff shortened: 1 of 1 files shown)


ui/src/routes/profile/token/index.tsx
```diff
@@ -137,11 +139,16 @@ const TokenPage = () => {
...
-              option.
+              using{' '}
+              <span className='inline-flex items-center'>
+                <code className='bg-muted rounded px-1 py-0.5 font-mono text-xs'>
+                  {cliLoginCommand}
+                </code>
+                <CopyToClipboard
+                  copyText={cliLoginCommand}
+                  className='h-5 px-2 align-middle'
+                />
```
sschuberth, line 150: Nit: I'd prefer to keep the dot at the end of the sentence, probably after the copy-to-clipboard button.

Etsija: I left it out on purpose, because the `CopyToClipboard' has a padding around the icon, and putting a dot after that left it hanging weirdly on its own. Adjusted now.

sschuberth (approved):

PR #5427, opened 2026-07-06
fix(workers)!: Don't generate the NPM `always-auth` property anymore

nnobelis:
The .npmrc option always-auth is a configuration setting that has historically played an important role in how the npm CLI interacts with package registries. By default, npm only sends authentication details when performing write operations (like npm publish or npm deprecate) or when accessing explicit private endpoints. However, many private registry managers (such as JFrog Artifactory, Azure Artifacts, AWS CodeArtifact, or Sonatype Nexus) are configured behind strict security firewalls where even read-only requests (like looking up metadata during npm install) require authentication. Enabling always-auth=true ensured that npm would actively present its credentials for these read operations.
The always-auth option has a multi-stage history of deprecation and final removal [1]:
* It was functionally deprecated with npm v7
* Starting with npm v11, a CLI warning is now outputed: "npm warn Unknown user config "always-auth". This will stop working in the next major version of npm."

This commit removes the property from the environment generator.
This is a breaking change as the environment definition has changed.

[1]: https://github.com/actions/setup-node/issues/1305

### DIFF
--- a/workers/common/src/main/kotlin/env/NpmRcGenerator.kt
+++ b/workers/common/src/main/kotlin/env/NpmRcGenerator.kt
@@ -94,13 +94,6 @@ class NpmRcGenerator : EnvironmentConfigGenerator<NpmDefinition> {
                         GeneratorLogger.entryAdded(entry, NPMRC_FILE_NAME, definition.service)
                     }
                 }
-
-                if (definition.alwaysAuth) {
-                    "$uriFragment:always-auth=true".also { entry ->
-                        println(entry)
-                        GeneratorLogger.entryAdded(entry, NPMRC_FILE_NAME, definition.service)
-                    }
-                }
             }
 
             printProxySettings { proxyConfig ->
--- a/workers/common/src/main/kotlin/env/config/EnvironmentDefinitionFactory.kt
+++ b/workers/common/src/main/kotlin/env/config/EnvironmentDefinitionFactory.kt
@@ -163,8 +163,7 @@ class EnvironmentDefinitionFactory {
                 credentialsTypes = credentialsTypes(),
                 scope = getOptionalProperty("scope"),
                 email = getOptionalProperty("email"),
-                authMode = getEnumProperty("authMode", NpmAuthMode.PASSWORD),
-                alwaysAuth = getBooleanProperty("alwaysAuth", true)
+                authMode = getEnumProperty("authMode", NpmAuthMode.PASSWORD)
             )
         }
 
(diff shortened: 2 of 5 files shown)


nnobelis:
I would like a review from @eclipse-apoapsis/doubleopen please

mnonnenmacher:
> Enabling always-auth=true ensured that npm would actively present its credentials for these read operations.

@nnobelis Is this now always the case even if `always-auth` is not set?

nnobelis:
> > Enabling always-auth=true ensured that npm would actively present its credentials for these read operations.
> 
> @nnobelis Is this now always the case even if `always-auth` is not set?

I assume so. This is the answer AI gave me :

> With the overhaul of the npm registry fetch mechanics in npm v7, the underlying implementation made the global always-auth config flag obsolete and without effect.
Registries like Verdaccio explicitly updated their documentation to state: "Since npm 7.10, the config option always-auth has no effect. You can safely remove it from your configuration (https://www.verdaccio.org/docs/packages/)." 

I could not find extra information.

mnonnenmacher (changes requested):
Please also remove `alwaysAuth` from the UI, it appears in these two files:
* ui/src/lib/environment-definition-fields.tsx
* ui/src/lib/types.ts

Also, the `YarnRcGenerator` also supports an `alwaysAuth` flag, what about this?

nnobelis:
@mnonnenmacher it seems for Yarn it should be left:

> No, npmAlwaysAuth is not deprecated in Yarn (Berry / v2+). Unlike the npm CLI, Yarn explicitly supports and actively uses this configuration setting in modern versions.
Even though Yarn's modern architecture (v2, v3, and v4) is entirely separate from npm's CLI mechanics, it still interacts with the standard npm registry protocol.
Yarn requires this option for the exact same functional reason npm historically did: forcing credentials on read (GET) requests.
> - In npm: The global always-auth configuration was deprecated because the CLI shifted its focus toward scoped registry endpoints, relying on URL-specific configs 
> - In Yarn: The flag was kept and standardized as npmAlwaysAuth inside the .yarnrc.yml structure so you can easily enable or disable it on a per-scope basis (e.g., enabling it only for your private corporate scope while keeping it off for public npm packages) .

> If you are using modern Yarn (v2, v3, or v4) and are authenticating with a secure private registry, you should keep using npmAlwaysAuth: true. It is fully supported and is not slated for deprecation by the Yarn maintainers.

Also the UI usages you mentioned are both for the Yarn environment definition therefore they should not be changed.

mnonnenmacher (approved):

PR #5440, opened 2026-07-07
feat(Kubernetes): Improve detection of failed jobs

oheger-bosch:
The Kubernetes reaper task did not handle pods with multiple containers
correctly in which one container failed, but others were still running.
In this case, the status of the job is not yet updated. Fix this by
inspecting the pods of containers to make sure that all jobs in error
state are found.

### DIFF
--- a/tasks/src/main/kotlin/impl/kubernetes/JobHandler.kt
+++ b/tasks/src/main/kotlin/impl/kubernetes/JobHandler.kt
@@ -140,10 +140,11 @@ internal class JobHandler(
     private val recentJobsMutex = Mutex()
 
     /**
-     * Return a list with all currently existing jobs that have been completed before the given [time].
+     * Return a list with all currently existing jobs that have been completed before the given [time]. This includes
+     * failed jobs.
      */
-    fun findJobsCompletedBefore(time: OffsetDateTime): List<V1Job> =
-        listJobs(workerJobsLabelSelector).filter { it.completedBefore(time) }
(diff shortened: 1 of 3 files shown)


mnonnenmacher (approved):

tasks/src/main/kotlin/impl/kubernetes/JobHandler.kt
```diff
@@ -197,11 +198,11 @@ internal class JobHandler(
     /**
      * Find all pods that have been created for the job with the specified [jobName].
      */
-    private fun findPodsForJob(jobName: String): List<V1Pod> {
-        val selector = "job-name=$jobName"
+    private fun findPodsForJob(jobName: String?): List<V1Pod> = jobName?.let {
```
nnobelis, line 201: In which case is `jobName` null ? When a job has no metadata ? Can it happen ?

Wouldn't it be better to change the callsite to:
`findPodsForJob(job.metadata?.name?.orEmpty()` and make this parameter not nullable ?

sschuberth:
@nnobelis will put this in the queue.

tasks/src/main/kotlin/impl/kubernetes/JobHandler.kt
oheger-bosch: In practice, I have never seen a job without a name. However, in the data model of the Kubernetes client, this field is nullable. It was easier to handle this special case in this function instead in multiple places where it is called.

nnobelis (approved):

PR #5457, opened 2026-07-09
fix(workers): Resolved issues should contain an identifier

### LINKED ISSUE #5271: Regression for applying issue resolution
Although [applying resolutions to issues was implemented](https://github.com/eclipse-apoapsis/ort-server/pull/4716) and observed to work, it seem to have regressed now:

Image

Resolution are provided though [this `.ort.yml` file](https://github.com/nordic-institute/X-Road/blob/develop/.ort.yml#L196), and logs confirm that the resolutions have been applied. However, the issues are still visible in the UI / returned by the API.

@oheger-bosch, is it possible that https://github.com/eclipse-apoapsis/ort-server/pull/5125 introduced a regression in that regard?

nnobelis:
The commit b1fdb637d01bb62a4b4bef69dccfe075db812311 added an identifier check when matching issues while storing resolutions. This pull request makes sure that resolved issues contain an identifier, so that the aforementioned check returns the correct issues.

This fixes a bug that issues were appearing unresolved in the UI despite existing resolutions.

Fixes #5271.

### DIFF
--- a/workers/advisor/src/main/kotlin/AdvisorWorker.kt
+++ b/workers/advisor/src/main/kotlin/AdvisorWorker.kt
@@ -36,6 +36,7 @@ import org.eclipse.apoapsis.ortserver.workers.common.validateForProcessing
 
 import org.jetbrains.exposed.v1.jdbc.Database
 
+import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.Severity
 import org.ossreviewtoolkit.utils.ort.ORT_VERSION
 
@@ -84,9 +85,11 @@ internal class AdvisorWorker(
                 ).advisor
             ) { "ORT Adviser failed to create a result." }
 
-            val allIssues = buildList {
-                addAll(advisorRun.results.values.flatten().flatMap { it.summary.issues })
-                addAll(advisorRun.providerIssues)
+            val allIssues = buildMap {
+                advisorRun.results.forEach { (identifier, results) ->
+                    put(identifier, results.flatMap { it.summary.issues })
+                }
+                put(Identifier.EMPTY, advisorRun.providerIssues.toList())
             }
 
             val allVulnerabilities = advisorRun.results.values.flatten().flatMap { it.vulnerabilities }
@@ -102,7 +105,7 @@ internal class AdvisorWorker(
 
             // Apply resolutions using the common function for both issues AND vulnerabilities.
             val resolvedItems = resolutionProvider.matchResolutions(
-                issues = allIssues,
+                issuesByIdentifier = allIssues,
                 ruleViolations = emptyList(),
                 vulnerabilities = allVulnerabilities
             )
@@ -114,8 +117,9 @@ internal class AdvisorWorker(
             }
 
             // Calculate unresolved issues for logging.
-            val unresolvedIssues = allIssues.filter { issue ->
-                issue.mapToModel() !in resolvedItems.issues.keys
+            val unresolvedIssues = allIssues.flatMap { (ortIdentifier, issues) ->
+                val identifier = ortIdentifier.takeIf { it != Identifier.EMPTY }?.mapToModel()
+                issues.filter { it.mapToModel(identifier) !in resolvedItems.issues.keys }
             }
 
             // Calculate unresolved vulnerabilities for logging.
@@ -124,7 +128,7 @@ internal class AdvisorWorker(
             }
 
             logger.info(
-                "Advisor job ${job.id} finished with ${allIssues.size} total issues " +
+                "Advisor job ${job.id} finished with ${allIssues.values.flatten().size} total issues " +
                         "and ${unresolvedIssues.size} unresolved issues, " +
                         "${allVulnerabilities.size} total vulnerabilities " +
                         "and ${unresolvedVulnerabilities.size} unresolved vulnerabilities."
--- a/workers/analyzer/src/main/kotlin/AnalyzerWorker.kt
+++ b/workers/analyzer/src/main/kotlin/AnalyzerWorker.kt
@@ -41,6 +41,7 @@ import org.eclipse.apoapsis.ortserver.workers.common.validateForProcessing
 
 import org.jetbrains.exposed.v1.jdbc.Database
 
+import org.ossreviewtoolkit.model.Identifier as OrtIdentifier
 import org.ossreviewtoolkit.model.Severity
 import org.ossreviewtoolkit.utils.ort.ORT_VERSION
 
@@ -132,7 +133,7 @@ internal class AnalyzerWorker(
                 .mapTo(mutableSetOf()) { it.id.mapToModel() }
 
             // IMPORTANT: Use getAnalyzerIssues() to get ONLY analyzer issues, not all issues.
-            val allIssues = ortResult.getAnalyzerIssues().values.flatten()
+            val allIssues = ortResult.getAnalyzerIssues().mapValues { it.value.toList() }
 
             val resolutionProvider = OrtServerResolutionProvider.create(
                 context,
@@ -144,19 +145,20 @@ internal class AnalyzerWorker(
 
             // Apply resolutions using the common function.
             val resolvedItems = resolutionProvider.matchResolutions(
-                issues = allIssues,
+                issuesByIdentifier = allIssues,
                 ruleViolations = emptyList(),
                 vulnerabilities = emptyList()
             )
 
             // Calculate unresolved issues for logging.
-            val unresolvedIssues = allIssues.filter { issue ->
-                issue.mapToModel() !in resolvedItems.issues.keys
+            val unresolvedIssues = allIssues.flatMap { (ortIdentifier, issues) ->
+                val identifier = ortIdentifier.takeIf { it != OrtIdentifier.EMPTY }?.mapToModel()
+                issues.filter { it.mapToModel(identifier) !in resolvedItems.issues.keys }
             }
 
             logger.info(
                 "Analyzer job ${job.id} for repository ${repository.url} with revision ${ortRun.revision} " +
-                        "finished with ${allIssues.size} total issues " +
+                        "finished with ${allIssues.values.flatten().size} total issues " +
                         "and ${unresolvedIssues.size} unresolved issues."
             )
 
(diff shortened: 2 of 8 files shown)


workers/advisor/src/main/kotlin/AdvisorWorker.kt
```diff
@@ -84,9 +85,13 @@ internal class AdvisorWorker(
...
             ) { "ORT Adviser failed to create a result." }
 
-            val allIssues = buildList {
-                addAll(advisorRun.results.values.flatten().flatMap { it.summary.issues })
-                addAll(advisorRun.providerIssues)
+            val allIssues = buildMap {
+                putAll(
+                    advisorRun.results.mapValues { entry ->
+                    entry.value.flatMap { advisorResult -> advisorResult.summary.issues }
+                }
```
sschuberth: Wrong indentation.

workers/common/src/main/kotlin/resolutions/OrtServerResolutionProvider.kt
```diff
@@ -145,35 +147,42 @@ class OrtServerResolutionProvider(
...
+                            .map { it.mapToModel(ResolutionSource.REPOSITORY_FILE) }
+                    )
+                    addAll(
+                        managedIssueResolutions.filter { resolution ->
+                            checkNotNull(resolution.messageHash) == calculateResolutionMessageHash(issue.message)
+                        }
+                            .map { it.copy(source = ResolutionSource.SERVER) }
+                    )
+                }
+            }.filterValues { it.isNotEmpty() }.mapKeys { it.key.mapToModel(identifier.mapToModel()) }
```
sschuberth: Wrong indentation.

oheger-bosch:
I cannot claim to be an expert in the issues and resolutions model, but the approach used here seems plausible to me. I would, however, like to have some meaningful tests (not only using `Identifier.EMPTY`) that expose the bug and demonstrate that it is fixed.

workers/common/src/main/kotlin/resolutions/OrtServerResolutionProvider.kt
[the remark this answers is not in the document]
nnobelis: Yeah yeah this was on my radar :)

nnobelis:
> I cannot claim to be an expert in the issues and resolutions model, but the approach used here seems plausible to me. I would, however, like to have some meaningful tests (not only using `Identifier.EMPTY`) that expose the bug and demonstrate that it is fixed.

Yes this is the reason this PR is still a draft. I wanted to extend the current workers tests but this is not possible because everything is mocked, therefore I think I need a new test.

workers/advisor/src/main/kotlin/AdvisorWorker.kt
```diff
@@ -84,9 +85,13 @@ internal class AdvisorWorker(
...
-            val allIssues = buildList {
-                addAll(advisorRun.results.values.flatten().flatMap { it.summary.issues })
-                addAll(advisorRun.providerIssues)
+            val allIssues = buildMap {
+                putAll(
+                    advisorRun.results.mapValues { entry ->
+                        entry.value.flatMap { advisorResult -> advisorResult.summary.issues }
+                    }
+                )
+                put(Identifier.EMPTY, advisorRun.providerIssues.toList())
```
sschuberth, line 92: Maybe a bit simpler:

```kotlin
val allIssues = advisorRun.results.mapValues { entry ->
    entry.value.flatMap { advisorResult -> advisorResult.summary.issues }
} + (Identifier.EMPTY to advisorRun.providerIssues.toList())
```

Or:

```kotlin
val allIssues = mutableMapOf(Identifier.EMPTY to advisorRun.providerIssues.toList())

advisorRun.results.mapValuesTo(allIssues) { entry ->
    entry.value.flatMap { advisorResult -> advisorResult.summary.issues }
}
```

workers/advisor/src/main/kotlin/AdvisorWorker.kt
```diff
@@ -84,9 +85,10 @@ internal class AdvisorWorker(
                 ).advisor
             ) { "ORT Adviser failed to create a result." }
 
-            val allIssues = buildList {
-                addAll(advisorRun.results.values.flatten().flatMap { it.summary.issues })
-                addAll(advisorRun.providerIssues)
+            val allIssues = mutableMapOf(Identifier.EMPTY to advisorRun.providerIssues.toList())
+
+            advisorRun.results.mapValuesTo(allIssues) { entry ->
+                entry.value.flatMap { advisorResult -> advisorResult.summary.issues }
```
mnonnenmacher: You could use `buildMap`, I find this a bit cleaner and easier to read:

```kotlin
val allIssues = buildMap {
    advisorRun.results.forEach { (identifier, results) ->
        put(identifier, results.flatMap { it.summary.issues })
    }
    put(Identifier.EMPTY, advisorRun.providerIssues)
}
```

workers/advisor/src/main/kotlin/AdvisorWorker.kt
```diff
@@ -114,8 +116,11 @@ internal class AdvisorWorker(
             }
 
             // Calculate unresolved issues for logging.
-            val unresolvedIssues = allIssues.filter { issue ->
-                issue.mapToModel() !in resolvedItems.issues.keys
+            val unresolvedIssues = allIssues.flatMap { (identifier, issues) ->
+                val identifierServer = identifier.takeIf { it != Identifier.EMPTY }?.mapToModel()
```
mnonnenmacher: Nit: When mapping between models we usually use a prefix instead of a suffix and do that for the "external" model, so:
* `identifier` -> `ortIdentifier`
* `identifierServer` -> `identifier`

workers/advisor/src/main/kotlin/AdvisorWorker.kt
```diff
@@ -114,8 +116,11 @@ internal class AdvisorWorker(
...
+                issues.filter { issue ->
+                    issue.mapToModel(identifierServer) !in resolvedItems.issues.keys
+                }
```
mnonnenmacher: Nit: Could be a one-liner:

```kotlin
issues.filter { it.mapToModel(identifierServer) !in resolvedItems.issues.keys }
```

workers/analyzer/src/main/kotlin/AnalyzerWorker.kt
```diff
@@ -144,19 +145,22 @@ internal class AnalyzerWorker(
...
             )
 
             // Calculate unresolved issues for logging.
-            val unresolvedIssues = allIssues.filter { issue ->
-                issue.mapToModel() !in resolvedItems.issues.keys
+            val unresolvedIssues = allIssues.flatMap { (identifier, issues) ->
+                val identifierServer = identifier.takeIf { it != OrtIdentifier.EMPTY }?.mapToModel()
+                issues.filter { issue ->
+                    issue.mapToModel(identifierServer) !in resolvedItems.issues.keys
+                }
```
mnonnenmacher: Same nits as for `AdvisorWorker`.

workers/scanner/src/main/kotlin/ScannerWorker.kt
```diff
@@ -124,8 +130,11 @@ class ScannerWorker(
             }
 
             // Calculate unresolved issues for logging.
-            val unresolvedIssues = allOrtIssues.filter { issue ->
-                issue.mapToModel() !in resolvedItems.issues.keys
+            val unresolvedIssues = allOrtIssues.flatMap { (identifier, issues) ->
+                val identifierServer = identifier.takeIf { it != OrtIdentifier.EMPTY }?.mapToModel()
+                issues.filter { issue ->
+                    issue.mapToModel(identifierServer) !in resolvedItems.issues.keys
+                }
```
mnonnenmacher: Same nits as for `AdvisorWorker`.

mnonnenmacher (changes requested):

workers/advisor/src/main/kotlin/AdvisorWorker.kt
sschuberth: That's basically what @nnobelis had before, and [I proposed somewhat the other direction](https://github.com/eclipse-apoapsis/ort-server/pull/5457#discussion_r3557551370) 😬

nnobelis: @mnonnenmacher Should I change it back ?

sschuberth: I'm fine either way. Actually, what I disliked about the original code was the `putAll()` in combination with indentation. @mnonnenmacher's solution avoids that by using `put()` in the inner loop, which looks better than the original approach.

So yes, probably go for @mnonnenmacher's proposal.

mnonnenmacher: Well, I find it nicer because it's closer to the previous code and it doesn't make `allIssues` mutable which it shouldn't be. And I find it easier to understand. But I won't insist, you got two contradicting review comments, so I'd say it's up to you 😉

When keeping the current code, at least change this:

```kotlin
advisorRun.results.mapValuesTo(allIssues) { (_, results) ->
    results.flatMap { ... }
}
```

nnobelis: I went with an hybrid solution, also `buildMap` but with `putAll` that I find clearer.

sschuberth: Well, that's again the original code almost exactly now (modulo some lambda parameter renamings), with the [nasty indentation](https://github.com/eclipse-apoapsis/ort-server/pull/5457#discussion_r3558017783) 😄 

But I'll pull myself out of the discussion and leave the final decision to @mnonnenmacher.

mnonnenmacher: > I went with an hybrid solution, also `buildMap` but with `putAll` that I find clearer.

But that uses `mapValues` which creates an unnecessary intermediate map, using `forEach` with `put` inside avoids that.

nnobelis: TIL that `mapValues` constructs an intemediate map and is not a streaming operation. Fixed.

mnonnenmacher (approved):

PR #5459, opened 2026-07-09
Improve handling of `ENUM` options when creating runs

mnonnenmacher:
Show a dropdown for `ENUM` type options. Options of type `ENUM_LIST` will be handled in a separate PR.

image


ui/src/components/form/plugin-multi-select-field.tsx
```diff
@@ -371,6 +374,40 @@ export const PluginMultiSelectField = <
...
+                              >
+                                <SelectTrigger className='w-[280px]'>
+                                  <SelectValue placeholder='Select a value' />
+                                </SelectTrigger>
+                                <SelectContent>
+                                  {!option.isRequired && (
+                                    <SelectItem value={UNDEFINED_ENUM_VALUE}>
+                                      {option.defaultValue != null &&
+                                      option.defaultValue !== ''
+                                        ? 'Reset to default'
```
sschuberth, line 400: Nit: Why not just "Default" or "Default value"? What additional value does "Reset" bring?

sschuberth: Or actually, would it be possible to just mark the default value instead of creating a new entry? Like "V3 (default)" in this example?

mnonnenmacher: It's a bit tricky to explain, but if there is no default value, the option is called "Not defined" which is in line with the dropdown for secret options. However, if there is a default value, the UI automatically jumps back to the default value when unsetting the option which was weird. So as a quick fix I just renamed it to "Reset to default" in case there is a default value.
This definitely can be improved, and I think for all other option types there should also be a way to explicitly go back to the default value, but I'd rather first have feature completeness and then improve the UX.

sschuberth: I'm trusting your judgement here, I just wanted to get some explanation. So thanks for that.

sschuberth (approved):

PR #5460, opened 2026-07-09
Test chores

mnonnenmacher:


dao/src/test/kotlin/repositories/ortrun/DaoOrtRunRepositoryTest.kt
```diff
@@ -869,9 +868,8 @@ class DaoOrtRunRepositoryTest : WordSpec({
  * Check whether the given [timestamp] is rather close to the current system time.
  */
 private fun assertCurrentTime(timestamp: Instant?) {
-    val time = timestamp.shouldNotBeNull()
-
-    val delta = Clock.System.now() - time
-    delta.inWholeMilliseconds shouldBeGreaterThan 0
-    delta.inWholeMilliseconds shouldBeLessThan 3000
+    timestamp.shouldNotBeNull {
+        val delta = Clock.System.now() - this
```
sschuberth: Hmm. As `this` is only used here, why not just `val delta = Clock.System.now() - timestamp.shouldNotBeNull()`? Or actually:

```kotlin
with(Clock.System.now() - timestamp.shouldNotBeNull()) {
    inWholeMilliseconds shouldBeInRange 0L..3000L
}
```

dao/src/testFixtures/kotlin/DatabaseMigrationTestExtension.kt
```diff
@@ -1,61 +0,0 @@
-/*
```
sschuberth, line 1: Commit message nit: Maybe "usage" instead of "user" (which sounds like a person).

sschuberth (approved):

PR #5431, opened 2026-07-06
Rework handling of authentication for services declared via environment variables

oheger-bosch:
This is another preparation to allow the Analyzer to run in multiple phases. Refer to the single commits for further information.

PR #5463, opened 2026-07-10
Run Analyzer in multiple phases

oheger-bosch:
This PR implements the logic to optionally split the Analyzer into three phases: preparation, analysis, and result. The phase to execute is passed as command line argument to the endpoint. If no argument is provided, all phases are run - which is the same behavior as before. Refer to the single commits for further information.

This PR depends on https://github.com/eclipse-apoapsis/ort-server/pull/5431.

### DIFF
--- a/workers/analyzer/docker/scripts/await.sh
+++ b/workers/analyzer/docker/scripts/await.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+#
+# Copyright (C) 2026 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+# A script to support the synchronization between different containers if the Analyzer runs in multiple phases.
+# The script expects the path to a file as single argument. It waits until this file is created and then returns.
+
+if [ -z "$1" ]; then
+    echo "Error: No target file path provided." >&2
+    echo "Usage: $0 <target_file>" >&2
+    exit 1
+fi
+
+TARGET_FILE=$1
+echo "Waiting for file $TARGET_FILE."
+
+while [ ! -f "$TARGET_FILE" ]; do
+    sleep 5
+done
+
+echo "File $TARGET_FILE detected."
--- a/workers/analyzer/src/main/kotlin/AnalyzerComponent.kt
+++ b/workers/analyzer/src/main/kotlin/AnalyzerComponent.kt
@@ -45,23 +45,47 @@ import org.eclipse.apoapsis.ortserver.workers.common.ortRunServiceModule
 import org.koin.core.component.inject
 import org.koin.core.module.Module
 import org.koin.core.module.dsl.singleOf
+import org.koin.core.qualifier.named
 import org.koin.dsl.bind
 import org.koin.dsl.module
+
+/** The name of the preparation phase. */
+private const val PREPARATION_PHASE = "preparation"
+
+/** The name of the analysis phase. */
+private const val ANALYSIS_PHASE = "analysis"
+
+/** The name of the result phase. */
+private const val RESULT_PHASE = "result"
+
+/** The name of the full phase which includes all other phases. */
+private const val FULL_PHASE = "full"
+
+/** A set with the names of all supported phases. */
+private val VALID_PHASES = setOf(PREPARATION_PHASE, ANALYSIS_PHASE, RESULT_PHASE, FULL_PHASE)
 
 /**
  * The central entry point into the Analyzer service. The class processes messages to analyze repositories by
  * delegating to helper objects.
  */
-class AnalyzerComponent : EndpointComponent<AnalyzerRequest>(AnalyzerEndpoint) {
+class AnalyzerComponent(
+    /** The command line arguments passed to the Analyzer endpoint. */
+    private val args: Array<String>
+) : EndpointComponent<AnalyzerRequest>(AnalyzerEndpoint) {
     override val endpointHandler: EndpointHandler<AnalyzerRequest> = { message ->
         val analyzerWorker by inject<AnalyzerWorker>()
         val publisher by inject<MessagePublisher>()
-        // TODO: Obtain the current phase from the command line arguments.
-        val phase by inject<AnalyzerPhase>()
         val jobId = message.payload.analyzerJobId
 
         withMdcContext(AnalyzerEndpoint.jobMdcKey(jobId)) {
-            val response = when (val result = analyzerWorker.run(jobId, message.header.traceId, phase, emptyArray())) {
+            val result = analyzerWorker.run(
+                jobId,
+                message.header.traceId,
+                getPhase(),
+                args.drop(1).toTypedArray()
+            )
+
+            val response = when (result) {
                 is RunResult.Success -> {
                     logger.info("Analyzer job '$jobId' succeeded.")
                     Message(message.header, AnalyzerWorkerResult(jobId))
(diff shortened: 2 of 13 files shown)


oheger-bosch:
Note: The `keepAlive` flag is currently handled by the preparation phase. I plan to create a follow-up PR which allows configuring the phase which should stop.

workers/analyzer/src/main/kotlin/AnalyzerPhase.kt
```diff
@@ -95,3 +118,127 @@ internal class FullPhase(
...
+        val exchangeDir = exchangeDirectory(jobId, checkArgs(this, args, 1, 1))
+        val job = ortRunService.getValidAnalyzerJob(jobId)
+
+        val prepareExchange = contextFactory.withContext(job.ortRunId) { context ->
+            val configDir = exchangeDir.resolve(CONFIG_DIR).apply { mkdirs() }
+            val prepareResult = worker.prepare(context, job, ortRunService, environmentService, exchangeDir, configDir)
+
+            val authInfoFile = exchangeDir.resolve(AUTH_INFO_FILE)
+            logger.info("Writing auth info file '{}'.", authInfoFile)
+            EnvironmentForkHelper.persistAuthenticationInfo(authInfoFile)
```
MarcelBochtler, line 164: Not really relevant to this PR, but maybe as a follow up.
`EnvironmentForkHelper.persistAuthenticationInfo()` does not enforce any Unix permissions on the written file.

I think we should change that method to ensure `0600` permissions. Or would that propose an issue on another analyzer phase?

workers/analyzer/src/main/kotlin/AnalyzerPhase.kt
```diff
@@ -167,6 +175,96 @@ internal class PreparationPhase(
...
+    private fun copyConfigFiles(exchangeDir: File) {
+        val configDir = exchangeDir.resolve(CONFIG_DIR)
+        val userHomeDir = Os.userHomeDirectory
+        logger.info("Copying configuration files from '{}' to '{}'.", configDir, userHomeDir)
+
+        configDir.copyRecursively(target = userHomeDir)
+    }
+
+    /**
+     * Set up the current environment, so that the ORT Analyzer can run. Because this phase does not user a
```
MarcelBochtler: Nit: Typo: `Because this phase does not user a` -> `use`

workers/analyzer/src/main/kotlin/AnalyzerComponent.kt
```diff
@@ -102,10 +111,20 @@ class AnalyzerComponent : EndpointComponent<AnalyzerRequest>(AnalyzerEndpoint) {
...
         singleOf(::IssueResolutionEventStore)
         singleOf(::IssueResolutionService)
     }
+
+    /**
+     * Obtain the current phase for the Analyzer run from the dependency definitions based on the command line
+     * arguments. If no argument is specified, use the `FULL` phase.
+     */
+    private fun getPhase(): AnalyzerPhase =
+        getKoin().get(named(args.firstOrNull()?.lowercase() ?: "full"))
```
MarcelBochtler: A type in the phase name will lead to a confusing error message.
Can we add a check for all known phases?

workers/analyzer/docker/scripts/await.sh
```diff
@@ -0,0 +1,30 @@
...
+# The script expects the path to a file as single argument. It waits until this file is created and then returns.
+
+TARGET_FILE=$1
+echo "Waiting for file $TARGET_FILE."
+
+while [ ! -f "$TARGET_FILE" ]; do
+    sleep 5
+done
+
+echo "File $TARGET_FILE detected."
```
MarcelBochtler, line 36: What happens if the analyzer phase crashes or hangs without ever writing the `$TARGET_FILE`?

workers/analyzer/docker/scripts/await.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# A script to support the synchronization between different containers if the Analyzer runs in multiple phases.
+# The script expects the path to a file as single argument. It waits until this file is created and then returns.
+
+TARGET_FILE=$1
+echo "Waiting for file $TARGET_FILE."
+
+while [ ! -f "$TARGET_FILE" ]; do
+    sleep 5
+done
```
MarcelBochtler, line 34: If `$TARGET_FILE` is empty this will loop forever. Maybe add a check for this usage error.

MarcelBochtler (dismissed):
LGTM. Found some minor issues, but nothing that couldn't be addressed in a follow-up change.

workers/analyzer/src/main/kotlin/AnalyzerPhase.kt
oheger-bosch: Probably makes sense, although I am not fully sure whether this could cause problems with other phases. I am also not sure about the effect this would have: AFAIK, there is only the ORT user in the pod, and all code runs under this user - so would still have access to this file. And, the credentials contained in this file are also present in the package manager configuration files.

workers/analyzer/src/main/kotlin/AnalyzerComponent.kt
oheger-bosch: Added a check for valid phase names, and now a meaningful exception is generated for an unsupported name.

workers/analyzer/docker/scripts/await.sh
oheger-bosch: A crash of the preparation or analysis phase does not automatically cause this loop to end. Kubernetes keeps the pod alive while there is still a container running. For this reason, the Kubernetes job monitor was updated recently to detect this condition. It now checks whether the pod contains at least one failing container; then it kills the pod.

If a phase is hanging without crashing, eventually the long-running pod detection will kick in and kill the pod after the configured timeout. (This is the same behavior as without this change.)

workers/analyzer/docker/scripts/await.sh
oheger-bosch: Added a check.

MarcelBochtler (dismissed):

workers/analyzer/src/main/kotlin/AnalyzerPhase.kt
mnonnenmacher: In OpenShift random UIDs are used which are always member of GID 0 (root). So if permissions are changed, make sure that this group has the required permissions.

workers/analyzer/src/main/kotlin/Entrypoint.kt
```diff
@@ -32,13 +32,20 @@ private val logger = LoggerFactory.getLogger(AnalyzerComponent::class.java)
 /**
  * This is the entry point of the Analyzer worker. It calls the Analyzer from ORT programmatically by
  * interfacing on its APIs.
+ *
+ * Per default, a whole analysis is performed. This can be changed via command line arguments. The first argument is
+ * interpreted as the name of the analyzer phase to execute; further arguments are then passed to the phase. The
+ * following phases are supported (names are case-insensitive):
+ * - _preparation_ for the preparation phase
+ * - _analysis_ for the analysis phase
+ * - _result_ for the result phase
```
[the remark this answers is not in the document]
oheger-bosch: I originally deliberately left this out, since providing the _full_ argument is redundant. But it might be cleaner to document this technically valid parameter. Adapted the documentation.

workers/analyzer/src/main/kotlin/AnalyzerPhase.kt
```diff
@@ -0,0 +1,400 @@
...
+    /**
+     * Copy the package manager configuration files created during the preparation phase from the given [exchangeDir]
+     * to the user's home directory.
+     */
+    private fun copyConfigFiles(exchangeDir: File) {
+        val configDir = exchangeDir.resolve(CONFIG_DIR)
+        val userHomeDir = Os.userHomeDirectory
+        logger.info("Copying configuration files from '{}' to '{}'.", configDir, userHomeDir)
+
+        configDir.copyRecursively(target = userHomeDir)
```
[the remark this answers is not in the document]
oheger-bosch: Not strictly necessary, since we have full control over the image and know that it does not contain any configuration files in the home directory. But I have applied this suggestion to be on the safe side.

workers/analyzer/src/main/kotlin/AnalyzerPhase.kt
```diff
@@ -0,0 +1,400 @@
...
+    return File(exchangeRoot, "$JOB_DIR_PREFIX$jobId").also { it.mkdirs() }
+}
+
+/**
+ * Check whether the number of arguments for [phase] in the given [args] is in the valid range of [minCount], and
+ * [maxCount]. Throw an exception with a meaningful message if this is not the case.
+ */
+private fun checkArgs(phase: AnalyzerPhase, args: Array<String>, minCount: Int, maxCount: Int): Array<String> {
+    require(args.size in minCount..maxCount) {
+        "Unexpected arguments passed to phase '${phase.javaClass.simpleName}' phase: '${args.joinToString(" ")}'."
```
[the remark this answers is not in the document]
oheger-bosch: Indeed, that's a bummer.

MarcelBochtler (approved):

PR #5484, opened 2026-07-13
ci(renovate): Configure automatic tool upgrades for the scanner Dockerfile

mnonnenmacher:
Add a Renovate custom regex manager [1] to automate updating the tool versions in the `Scanner.Dockerfile`. To simplify matching the version args, use inline `# renovate:` annotations.

These annotations allow to specify the datasource and package name Renovate should use for the update, and optionally a custom regex to extract the version.

This is only implemented for the `Scanner.Dockerfile` to test how well the approach works before applying it to other Dockerfiles as well.

[1]: https://docs.renovatebot.com/modules/manager/regex/

### DIFF
--- a/workers/scanner/docker/Scanner.Dockerfile
+++ b/workers/scanner/docker/Scanner.Dockerfile
@@ -63,7 +63,7 @@ ARG RUBY_VERSION=3.4.4
 ARG SCANCODE_VERSION=32.5.0
 
 # Install Askalono.
-RUN curl -LOs https://github.com/amzn/askalono/releases/download/$ASKALONO_VERSION/askalono-Linux.zip && \
+RUN curl -LOs https://github.com/jpeddicord/askalono/releases/download/$ASKALONO_VERSION/askalono-Linux.zip && \
     mkdir /opt/askalono && \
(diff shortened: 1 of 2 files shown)


workers/scanner/docker/Scanner.Dockerfile
```diff
@@ -56,10 +56,15 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     zlib1g \
     zlib1g-dev
 
+# renovate: datasource=github-releases depName=amzn/askalono
```
sschuberth: This should actually say "jpeddicord/askalono".

sschuberth (changes requested):

workers/scanner/docker/Scanner.Dockerfile
sschuberth: And while at it, maybe also add a commit to fix it at https://github.com/eclipse-apoapsis/ort-server/blob/1fab9e2a20487f820dfa153f79145851ff0a19ce/workers/scanner/docker/Scanner.Dockerfile#L66

sschuberth (approved):

PR #5499, opened 2026-07-15
ci(github): Put all downloaded archives into the same directory

sschuberth:


.github/workflows/release.yml
```diff
@@ -161,7 +161,8 @@ jobs:
     - name: Create checksums
       shell: bash
       run: |
-        shasum --algorithm 256 ./artifacts/osc-cli-*/*.{tar.gz,zip} | sed 's|./artifacts/osc-cli-[^/]*/||' > checksums.txt
+        cd artifacts
+        sha256sum osc-cli-*.{tar.gz,zip} > checksums.txt
```
nnobelis: IIUC, the sed removed `./artifacts/osc-cli-*`  in the output of shasum.

Now you don't need the `./artifacts` because you are in the `artifacts` directory. But why `osc-cli-[^/]*` does not have to be sed anymore ?

sschuberth: Ah, you're right, I over-simplified things. I'll rework.

nnobelis (dismissed):

nnobelis (approved):

PR #5450, opened 2026-07-08
Improve handling of enum options for plugin templates

mnonnenmacher:
Use a dropdown for `ENUM` type options and a multi-select field for `ENUM_LIST` type options when creating or editing plugin templates. Also apply a bunch of smaller fixes and refactorings.

This PR only addresses the UI to manage plugin templates, similar changes for the run creation form will be done in a follow-up PR.

Dropdown for `ENUM` options:

image

Multi-select for `ENUM_LIST` options:

image

PR #5527, opened 2026-07-20
Extend the keep-alive mechanism to Analyzer phases

oheger-bosch:
Introduce a new property to the Analyzer job configuration that allows
specifying the phase(s) after which execution should wait. This enables
debugging of single phases.

### DIFF
--- a/api/v1/mapping/src/commonMain/kotlin/ApiMappings.kt
+++ b/api/v1/mapping/src/commonMain/kotlin/ApiMappings.kt
@@ -26,6 +26,7 @@ import org.eclipse.apoapsis.ortserver.api.v1.model.AdvisorJob as ApiAdvisorJob
 import org.eclipse.apoapsis.ortserver.api.v1.model.AdvisorJobConfiguration as ApiAdvisorJobConfiguration
 import org.eclipse.apoapsis.ortserver.api.v1.model.AnalyzerJob as ApiAnalyzerJob
 import org.eclipse.apoapsis.ortserver.api.v1.model.AnalyzerJobConfiguration as ApiAnalyzerJobConfiguration
+import org.eclipse.apoapsis.ortserver.api.v1.model.AnalyzerPhase as ApiAnalyzerPhase
 import org.eclipse.apoapsis.ortserver.api.v1.model.ComparisonOperator as ApiComparisonOperator
 import org.eclipse.apoapsis.ortserver.api.v1.model.EcosystemStats as ApiEcosystemStats
 import org.eclipse.apoapsis.ortserver.api.v1.model.EnvironmentConfig as ApiEnvironmentConfig
@@ -87,6 +88,7 @@ import org.eclipse.apoapsis.ortserver.model.AdvisorJob
 import org.eclipse.apoapsis.ortserver.model.AdvisorJobConfiguration
 import org.eclipse.apoapsis.ortserver.model.AnalyzerJob
 import org.eclipse.apoapsis.ortserver.model.AnalyzerJobConfiguration
+import org.eclipse.apoapsis.ortserver.model.AnalyzerPhase
 import org.eclipse.apoapsis.ortserver.model.EcosystemStats
 import org.eclipse.apoapsis.ortserver.model.EnvironmentConfig
 import org.eclipse.apoapsis.ortserver.model.EnvironmentVariableDeclaration
@@ -222,7 +224,8 @@ fun AnalyzerJobConfiguration.mapToApi() =
         packageManagerOptions = packageManagerOptions?.mapValues { it.value.mapToApi() },
         repositoryConfigPath = repositoryConfigPath,
         skipExcluded = skipExcluded,
-        keepAliveWorker = keepAliveWorker
+        keepAliveWorker = keepAliveWorker,
+        keepAlivePhases = keepAlivePhases.mapTo(mutableSetOf()) { it.mapToApi() }
     )
 
 fun ApiAnalyzerJobConfiguration.mapToModel() =
@@ -236,7 +239,8 @@ fun ApiAnalyzerJobConfiguration.mapToModel() =
         packageManagerOptions = packageManagerOptions?.mapValues { it.value.mapToModel() },
         repositoryConfigPath = repositoryConfigPath,
         skipExcluded = skipExcluded,
-        keepAliveWorker = keepAliveWorker
+        keepAliveWorker = keepAliveWorker,
+        keepAlivePhases = keepAlivePhases.mapTo(mutableSetOf()) { it.mapToModel() }
     )
 
 fun EvaluatorJob.mapToApi() =
@@ -776,6 +780,20 @@ fun ApiSubmoduleFetchStrategy.mapToModel() = when (this) {
     ApiSubmoduleFetchStrategy.DISABLED -> SubmoduleFetchStrategy.DISABLED
     ApiSubmoduleFetchStrategy.TOP_LEVEL_ONLY -> SubmoduleFetchStrategy.TOP_LEVEL_ONLY
     ApiSubmoduleFetchStrategy.FULLY_RECURSIVE -> SubmoduleFetchStrategy.FULLY_RECURSIVE
+}
+
+fun AnalyzerPhase.mapToApi() = when (this) {
+    AnalyzerPhase.PREPARATION -> ApiAnalyzerPhase.PREPARATION
+    AnalyzerPhase.ANALYSIS -> ApiAnalyzerPhase.ANALYSIS
+    AnalyzerPhase.RESULT -> ApiAnalyzerPhase.RESULT
+    AnalyzerPhase.FULL -> ApiAnalyzerPhase.FULL
+}
+
+fun ApiAnalyzerPhase.mapToModel() = when (this) {
+    ApiAnalyzerPhase.PREPARATION -> AnalyzerPhase.PREPARATION
+    ApiAnalyzerPhase.ANALYSIS -> AnalyzerPhase.ANALYSIS
+    ApiAnalyzerPhase.RESULT -> AnalyzerPhase.RESULT
+    ApiAnalyzerPhase.FULL -> AnalyzerPhase.FULL
 }
 
 fun ShortestDependencyPath.mapToApi() = ApiShortestDependencyPath(
--- a/api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
+++ b/api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
@@ -111,7 +111,13 @@ data class AnalyzerJobConfiguration(
      * Keep the worker alive after it has finished. This is useful for manual problem analysis directly
      * within the pod's execution environment.
      */
-    val keepAliveWorker: Boolean = false
+    val keepAliveWorker: Boolean = false,
+
+    /**
+     * If [keepAliveWorker] is *true*, this property lists the phases in which the worker should be kept alive. If no
+     * phase is provided, the worker will wait after all steps are finished.
+     */
+    val keepAlivePhases: Set<AnalyzerPhase> = emptySet()
 )
 
 /**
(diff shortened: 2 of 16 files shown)


oheger-bosch:
A new field for the keep-alive phases has been added to the UI:
Screenshot from 2026-07-21 07-35-52

It is only visible if the "Keep-alive worker" flag is set:
Screenshot from 2026-07-21 07-36-11

sschuberth:
Aren't the available phases known in advance? If so, I think this should be reflected in the UI by not allowing to input arbitrary strings, or using error-prone comma separation, but by using a similar UI like for [plugin option enum lists](https://github.com/eclipse-apoapsis/ort-server/pull/5450).

oheger-bosch:
How about this:
Screenshot from 2026-07-21 08-14-26

sschuberth:
> How about this:

Works as well for me!

Is there now a difference between disabling "Keep worker alive" and disabling all phases? I.e., do we still need the toggle button, or is it enough to only keep the checkboxes (and imply a disabled toggle button if all phases are disabled)?

oheger-bosch:
> Is there now a difference between disabling "Keep worker alive" and disabling all phases? I.e., do we still need the toggle button, or is it enough to only keep the checkboxes (and imply a disabled toggle button if all phases are disabled)?

If no phase is selected, the keep-alive flag is still effective, but the worker waits in the default phase (which is either _analysis_ or _full_, depending on the deployment). I think, this separation makes sense because in most cases, these default phases are suitable to debug a problem with a run. You typically only need to select a different phase to investigate special problems, for instance with cloning of the repository. So, in the typical scenario, the user can just enable the keep-alive checkbox and ignore the phases.

sschuberth:
@lamppu, can you maybe think of a more "cohesive" UI where the "Keep alive phases" card does not pop out of nowhere if "Keep worker alive" is enabled, but rather have the checkbox be shown inside the "Keep worker alive" card? Or is the current UI good enough in your opinion?

lamppu:
> @lamppu, can you maybe think of a more "cohesive" UI where the "Keep alive phases" card does not pop out of nowhere if "Keep worker alive" is enabled, but rather have the checkbox be shown inside the "Keep worker alive" card? Or is the current UI good enough in your opinion?

I maybe have a suggestion, but I'm not sure if I like it better or not so maybe we can take a look at it together and make a follow-up if needed :smile:

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
```diff
@@ -111,7 +111,13 @@ data class AnalyzerJobConfiguration(
...
      * within the pod's execution environment.
      */
-    val keepAliveWorker: Boolean = false
+    val keepAliveWorker: Boolean = false,
+
+    /**
+     * If [keepAliveWorker] is *true*, this property can contain a comma-separated list with the names of phases in
+     * which the worker should be kept alive. If unspecified, this defaults to the full phase and the result phase.
+     */
+    val keepAlivePhases: String? = null
```
sschuberth: Why isn't this an array of strings already at this level, making the API more safe to use by avoiding one pitfall of parsing errors?

workers/analyzer/src/main/kotlin/AnalyzerPhase.kt
```diff
@@ -86,6 +87,33 @@ internal sealed interface AnalyzerPhase {
...
+    /**
+     * Evaluate the keep-alive settings for this [AnalysisPhase] and [config]. If necessary, write a file that keeps
+     * the worker alive.
+     */
+    suspend fun handleKeepAlive(config: AnalyzerJobConfiguration) {
+        handleKeepAlive(config.keepAliveWorker, config.keepAlivePhases)
+    }
+
+    suspend fun handleKeepAlive(keepAliveWorker: Boolean, keepAlivePhases: String?) {
+        if (keepAliveWorker) {
```
sschuberth: If this does nothing for `keepAliveWorker == false`, why even bother calling the function in that case / passing `keepAliveWorker` as a parameter?

sschuberth (commented):
In general, I believe `keepAlivePhases` should be typed stronger. Ideally as a `enumSet`.

oheger-bosch:
> In general, I believe `keepAlivePhases` should be typed stronger. Ideally as a `enumSet`.

I was thinking about this, but am not sure whether it is worth the effort. This is not a feature for end users, but targets admins who should know what they are doing.

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
oheger-bosch: In typical scenarios, you provide only a single phase. Having multiple phases is mainly used to represent the defaults "analysis,full" - here both values need to be provided because the system does not know at this level whether it is configured to run the Analyzer in a single shot or in multiple phases.

Representing this property as an array would therefore only complicate its usage.

workers/analyzer/src/main/kotlin/AnalyzerPhase.kt
oheger-bosch: To handle the whole keep-alive logic in a central place.

sschuberth:
> > In general, I believe `keepAlivePhases` should be typed stronger. Ideally as a `enumSet`.
> 
> I was thinking about this, but am not sure whether it is worth the effort. This is not a feature for end users, but targets admins who should know what they are doing.

IMO it's still useful for API validation and UI usage, e.g. to automatically populate the UI with all phase names.

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
sschuberth: I don't see how adding type-safety would complicate the usage. Even for a single phase, adding semantics to what the string actually _means_ is a good practice, I believe.

oheger-bosch: I have reworked the model. There is now a new `enum` defining the supported phases. The `keepAlivePhases` property is a set of this type. So, everything is strongly typed now.

oheger-bosch:
@sschuberth, @lamppu, I would love to get this into the release we are going to create today. Is this UI-wise fine with you? I have also addressed all comments regarding the server-side model.

api/v1/mapping/src/commonMain/kotlin/ApiMappings.kt
```diff
@@ -222,7 +224,8 @@ fun AnalyzerJobConfiguration.mapToApi() =
         packageManagerOptions = packageManagerOptions?.mapValues { it.value.mapToApi() },
         repositoryConfigPath = repositoryConfigPath,
         skipExcluded = skipExcluded,
-        keepAliveWorker = keepAliveWorker
+        keepAliveWorker = keepAliveWorker,
+        keepAlivePhases = keepAlivePhases.map { it.mapToApi() }.toSet()
```
sschuberth: Please map straight to a set: `keepAlivePhases.mapTo(mutableSetOf()) { it.mapToApi() }`

api/v1/mapping/src/commonMain/kotlin/ApiMappings.kt
```diff
@@ -236,7 +239,8 @@ fun ApiAnalyzerJobConfiguration.mapToModel() =
         packageManagerOptions = packageManagerOptions?.mapValues { it.value.mapToModel() },
         repositoryConfigPath = repositoryConfigPath,
         skipExcluded = skipExcluded,
-        keepAliveWorker = keepAliveWorker
+        keepAliveWorker = keepAliveWorker,
+        keepAlivePhases = keepAlivePhases.map { it.mapToModel() }.toSet()
```
sschuberth: Please map straight to a set: `keepAlivePhases.mapTo(mutableSetOf()) { it.mapToApi() }`

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
```diff
@@ -111,7 +111,13 @@ data class AnalyzerJobConfiguration(
...
      * within the pod's execution environment.
      */
-    val keepAliveWorker: Boolean = false
+    val keepAliveWorker: Boolean = false,
+
+    /**
+     * If [keepAliveWorker] is *true*, this property lists the phases in which the worker should be kept alive. If no
+     * phase is provided, the worker will wait after all steps are finished.
+     */
+    val keepAlivePhases: Set<AnalyzerPhase> = emptySet()
```
sschuberth, line 120: Just checking: I guess you're not using an `EnumSet` here as it's JVM-only, but the project is KMP, correct?

workers/analyzer/src/main/kotlin/AnalyzerPhase.kt
sschuberth: How about then at least reducing indentation (and making more clear that nothing is done in this case) by returning early: `if (!keepAliveWorker) return`

workers/analyzer/src/main/kotlin/AnalyzerPhase.kt
```diff
@@ -195,6 +224,7 @@ internal class AnalysisPhase : AnalyzerPhase {
         val exchangeDir = exchangeDirectory(jobId, checkArgs(this, args, 1, 2))
 
         val prepareResult = createPrepareResult(exchangeDir, jobId)
+        handleKeepAlive(prepareResult.runnerConfig.keepAliveWorker, prepareResult.runnerConfig.keepAlivePhases)
```
sschuberth, line 231: Why not use the `handleKeepAlive(prepareResult.runnerConfig)` overload here?

sschuberth: Doing so makes the only caller of `handleKeepAlive(Boolean, Set<AnalyzerPhaseEnum>)` be the `AnalyzerWorkerTest`. So I really wonder whether that overload should be inlined in favor of only keeping `handleKeepAlive(AnalyzerJobConfiguration)`, and an `AnalyzerJobConfiguration` with only those two settings is easy enough to construct in the test:

```kotlin
AnalyzerJobConfiguration(keepAliveWorker = true, keepAlivePhases = EnumSet.of(AnalyzerPhaseEnum.RESULT))
```

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/analyzer-fields.tsx
```diff
@@ -393,28 +394,51 @@ export const AnalyzerFields = ({
...
+                      <InlineCode>full</InlineCode> and{' '}
+                      <InlineCode>analysis</InlineCode> phases.
+                    </>
+                  }
+                  options={[
+                    { id: 'PREPARATION', label: 'Preparation' },
+                    { id: 'ANALYSIS', label: 'Analysis' },
+                    { id: 'RESULT', label: 'Result' },
+                    { id: 'FULL', label: 'Full' },
+                  ]}
```
sschuberth: Can't we now get this somehow from the Kotlin enum instead of hard-coding? Also ping @lamppu.

sschuberth (changes requested):

api/v1/model/src/commonMain/kotlin/JobConfigurations.kt
oheger-bosch: Yes.

workers/analyzer/src/main/kotlin/AnalyzerPhase.kt
oheger-bosch: There is no overload for `AnalyzerRunnerConfig`, only for `AnalyzerJobConfiguration`. The reason for the additional `handleKeepAlive` version is that in `AnalysisPhase` no job configuration is available.

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/-components/analyzer-fields.tsx
oheger-bosch: Don't know. Are there other examples how enums are handled?

lamppu: You can add import 
`import { zAnalyzerPhase } from '@/api/zod.gen';` 

and use:
```
options={zAnalyzerPhase.options.map((phase) => ({
                    id: phase,
                    label: capitalize(phase),
                  }))}
 ```

oheger-bosch: Thanks! Changed accordingly.

workers/analyzer/src/main/kotlin/AnalyzerPhase.kt
sschuberth: > There is no overload for `AnalyzerRunnerConfig`, only for `AnalyzerJobConfiguration`.

Ah, I was indeed assuming `prepareResult.runnerConfig` to be a `AnalyzerJobConfiguration`.

BTW, really annoying that we use "Config" and "Configuration" as class name suffixes inconsistently.

sschuberth (approved):

sschuberth:
> > @lamppu, can you maybe think of a more "cohesive" UI where the "Keep alive phases" card does not pop out of nowhere if "Keep worker alive" is enabled, but rather have the checkbox be shown inside the "Keep worker alive" card? Or is the current UI good enough in your opinion?
> 
> I maybe have a suggestion, but I'm not sure if I like it better or not so maybe we can take a look at it together and make a follow-up if needed 😄

See https://github.com/eclipse-apoapsis/ort-server/pull/5555.

PR #5530, opened 2026-07-20
build(pnpm): Remove unused dependencies

sschuberth:
As fixed by `pnpm dlx knip --dependencies --fix --allow-remove-files`, modulo the false-positive for "@easyops-cn/docusaurus-search-local".

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -130,6 +130,10 @@ jobs:
       working-directory: ui
       run: pnpm build
 
+    - name: Check for unused dependencies
+      working-directory: ui
+      run: pnpm dlx knip --dependencies
+
     - name: Test UI
       working-directory: ui
       run: pnpm test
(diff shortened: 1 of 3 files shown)


lamppu:
Looks like the Docusaurus dependencies were explicitly added in #2904 because there was some Renovate problem

sschuberth:
> Looks like the Docusaurus dependencies were explicitly added in #2904 because there was some Renovate problem

Ah, thanks for digging that PR out. Then let me limit the scope of this to the `ui` project only.

sschuberth:
I've also added a CI check now, please have another look @lamppu.

.github/workflows/build-and-test.yml
```diff
@@ -126,6 +126,10 @@ jobs:
       working-directory: ui
       run: pnpm install
 
+    - name: Check for unused dependencies
+      working-directory: ui
+      run: pnpm dlx knip --dependencies
+
```
lamppu: This should go after the build step underneath so that the `@/api` imports are available.

lamppu (approved):

PR #5550, opened 2026-07-23
docs: Describe the secure Analyzer configuration

oheger-bosch:
Add a document that explains how the Analyzer can be configured to run in multiple phases exposing only a limited set of secrets to the phase that deals with potentially untrusted build scripts.

### DIFF
--- a/website/docs/admin-guide/guides/secure-analyzer-setup.md
+++ b/website/docs/admin-guide/guides/secure-analyzer-setup.md
@@ -1,0 +1,100 @@
+# Hardening the Analyzer against malicious Input
+
+This document describes a way how to configure the Analyzer worker to offer a reduced attack vector against malicious code in build scripts to be analyzed.
+
+## Problem
+
+When invoking package managers to analyze the dependencies of a project, it is not possible to prevent the execution of external code in all cases. For some package managers, build scripts are just code that gets executed; and even if this code comes from trusted sources (within the same company), there is always the danger that malicious code gets in that can do harm. Recent examples have shown that supply chain attacks become more and more sophisticated in their attempts to steal passwords and gather sensible information. This makes the environment in which the Analyzer operates especially insecure. Therefore, it is important to protected credentials as far as possible.
+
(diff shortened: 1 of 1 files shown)


website/docs/admin-guide/guides/secure-analyzer-setup.md
```diff
@@ -0,0 +1,100 @@
...
+
+The remaining part of this document describes how to achieve a secured Analyzer execution on Kubernetes based on this idea in detail.
+
+## Implementation
+
+_Note: ORT Server can be configured to use different secret storage implementations. Therefore, this section cannot provide detailed settings for this part of the configuration. It will focus on the concrete steps to secure the Analyzer worker and give general guidelines about which secrets need to be made available in which phase._
+
+The full configuration of the pod for the Analyzer worker is located in the deployment manifest for the _Orchestrator_. Here, a larger number of environment variables are defined that are evaluated by the Kubernetes Transport implementation and determine various aspects of the pod. If not noted otherwise, the code fragments shown below represent such variable definitions and therefore need to be placed into the `env:` section of the deployment manifest for Orchestrator. (To simplify the description, the variable values are provided eplicitly; it is typically good practice to define them in a `values.yaml` file.)
```
MarcelBochtler: `eplicitly` -> `explicitly`

website/docs/admin-guide/guides/secure-analyzer-setup.md
```diff
@@ -0,0 +1,100 @@
...
+- name: ANALYZERINIT_ARGS
+  value: "-c \"exec java $JAVA_OPTS -cp $( cat /app/jib-classpath-file ) @/app/jib-main-class-file\ preparation /mnt/exchange\""
+- name: ANALYZERRESULT_ARGS
+  value: "-c \"/etc/analyzer_scripts/await.sh /mnt/exchange/sync/ready && exec java $JAVA_OPTS -cp $( cat /app/jib-classpath-file ) @/app/jib-main-class-file\ result /mnt/exchange\""
+```
+
+This (admittedly verbose) configuration basically tells the three containers which phase to execute and where to find the shared volume for data exchange - by constructing the `java` command from the classpath and main class files and passing the corresponding command line arguments. It also addresses the issue that the result phase should not start before the analysis phase has finished. This is not directly supported by Kubernetes; therefore, the Analyzer container image contains a small shell script `await.sh` that waits for a specific file to be created in the shared volume. The analysis phase creates this file when it has finished, which then allows the result phase to start.
```
MarcelBochtler: I'd drop the `(admittedly verbose)` and the `basically` from the beginning of this sentence. IMO this technical documentation is not the right place for this "self-critical" language.

MarcelBochtler (approved):

PR #5589, opened 2026-07-29
ui: Show detected licenses and the corresponding findings for a package

### LINKED ISSUE #5466: Allow to show license findings by package
The current "Detected Licenses" view is license-centric, meaning that findings are grouped by license and then by package. This makes sense to generally identify problematic _licenses_, but in order to identify problematic _packages_ (due to licensing issues), we should have an additional view where the top-level entity is a package / project.

Instead of re-adding the complex grouping functionality we once had for some tables / columns, the idea is to have either radio buttons or tabs within the existing "Detected Licenses" view where a "by license" table shows the current table, and a "by package / project" table shows a new table that lists packages and projects.

This table should be similar to the existing "Projects" / "Packages" views in that a collapsed row should display the ID and (abbreviated) detected license below, and when expanding, license findings should be show like currently done in the license-centric view.

Etsija:
#### New licenses section in Packages table / package details

Screenshot from 2026-07-30 08-11-24

#### Clicking a detected license SPDX expression takes the user to the corresponding findings in the Detected Licenses table

Screenshot from 2026-07-29 13-07-32

### DIFF
--- a/components/license-findings/backend/src/routes/kotlin/routes/GetRunPackagesWithDetectedLicense.kt
+++ b/components/license-findings/backend/src/routes/kotlin/routes/GetRunPackagesWithDetectedLicense.kt
@@ -27,6 +27,8 @@ import org.eclipse.apoapsis.ortserver.components.authorization.routes.get
 import org.eclipse.apoapsis.ortserver.components.licensefindings.LicenseFindingService
 import org.eclipse.apoapsis.ortserver.components.licensefindings.PackageIdentifier
 import org.eclipse.apoapsis.ortserver.model.repositories.OrtRunRepository
+import org.eclipse.apoapsis.ortserver.model.util.ComparisonOperator
+import org.eclipse.apoapsis.ortserver.model.util.FilterOperatorAndValue
 import org.eclipse.apoapsis.ortserver.shared.apimappings.mapToApi
 import org.eclipse.apoapsis.ortserver.shared.apimappings.mapToModel
 import org.eclipse.apoapsis.ortserver.shared.apimodel.Identifier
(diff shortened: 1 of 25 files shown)


sschuberth:
Small UI comment based on the screenshot: It looks as if the "Licenses" section title is not in bold. I think it should be, to align with the other sections titles. For the nested "Declared License:" and "Detected Licenses:" sub-section titles / keys I believe we could omit "License(s)" for brevity and just say "Declared:" and "Detected:".

PS: It's nice that the normalized licenses are vertically aligned. This way one quickly sees the difference between declared and detected licenses of a package.

components/license-findings/backend/src/routes/kotlin/routes/GetRunPackagesWithDetectedLicense.kt
```diff
@@ -56,7 +56,11 @@ internal fun Route.getRunPackagesWithDetectedLicense(
                 description = "The URL-encoded SPDX license identifier (e.g. Apache-2.0, GPL-2.0-or-later)."
             }
             queryParameter<String>("identifier") {
-                description = "Filter by ORT identifier using a case-insensitive substring match."
+                description = "Filter by ORT identifier using a case-insensitive substring match. Mutually exclusive " +
+                    "with 'exactIdentifier'."
+            }
+            queryParameter<String>("exactIdentifier") {
+                description = "Filter by an exact ORT identifier. Mutually exclusive with 'identifier'."
             }
```
lamppu, line 66: Now that this PR introduces this kind of "exact vs. substring" match filtering, I wonder if we should think about the design a bit. I think I would prefer something where one field, like the `identifier` field here, would state the value to be filtered with, and another field, for example `identifierMatchType`, would describe what kind of matching should be used, like in this case `exact` or `substring` (this could also be an enum I guess).

```
queryParameter<String>("identifier") {
    description = "Filter by ORT identifier."
}
queryParameter<String>("identifierMatchType") {
    description = "Filter by identifier match type. Supported values are 'exact' and 'substring'. Defaults to 'substring'."
}
```

Although if this was used, then we probably should also discuss the default, I guess "exact" would be the most natural choice, but that would cause breaking changes.

As Martin and Sebastian are currently on vacation, we were wondering if maybe @oheger-bosch or @MarcelBochtler you might have a bit of time to chime in, if you have a strong preference, or a different idea.

Etsija: Refactoring done.

lamppu (approved):

PR #5631, opened 2026-08-04
feat(core): Improve validation of URLs

oheger-bosch:
So far, the validation of repository URLs was done using a utility function from Ktor. This function does a lenient parsing only and accepts some problematic URLs, for instance, URLs containing control characters. This may cause problems when processing such URLs and could also become a security risk. Therefore, replace the validation of URLs with a more strict implementation based on Java standard classes.

Since the validation is done in Kotlin multi-platform code, introduce a new `UrlValidator` interface. This allows injecting a Java-based implementation from the `core` module.


api/v1/model/src/commonMain/kotlin/validation/UrlValidator.kt
```diff
@@ -0,0 +1,66 @@
...
+         * Generate a validation error message for a URL with a user info component of the given optional [urlType].
+         */
+        fun userInfoMessage(urlType: String? = null): String = replaceUrlType(USER_INFO_MESSAGE, urlType)
+
+        /**
+         * Generate a validation error message based on the given [message] with the optional [urlType] replaced in the
+         * message.
+         */
+        private fun replaceUrlType(message: String, urlType: String?): String =
+            message.replace(URL_TYPE_PLACEHOLDER, urlType?.let { "$it " }.orEmpty())
```
mnonnenmacher, line 54: Nit (slightly shorter): `urlType?.plus(" ").orEmpty()`

mnonnenmacher (approved):

PR #5636, opened 2026-08-05
Fix config file path handling

mnonnenmacher:
Prevent path traversal outside of config directories. See the commit messages for details.

### DIFF
--- a/config/local/src/main/kotlin/LocalConfigFileProvider.kt
+++ b/config/local/src/main/kotlin/LocalConfigFileProvider.kt
@@ -29,6 +29,7 @@ import org.eclipse.apoapsis.ortserver.config.ConfigFileProvider
 import org.eclipse.apoapsis.ortserver.config.ConfigManager
 import org.eclipse.apoapsis.ortserver.config.Context
 import org.eclipse.apoapsis.ortserver.config.Path
+import org.eclipse.apoapsis.ortserver.config.resolveSecurely
 
 /**
  * An implementation of [ConfigFileProvider] that reads config files from a [local directory][configDir].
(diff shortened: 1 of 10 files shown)


MarcelBochtler (approved):
Found an correctness issue on macOS. But this can be solved in a follow-up PR.

Edit. This might also be an issue on non-macOS machines, if the user uses symlinks.

config/local/src/main/kotlin/LocalConfigFileProvider.kt
```diff
@@ -61,20 +62,20 @@ class LocalConfigFileProvider(
...
         val isDirectoryPath = path.path.endsWith("/")
-        val p = configDir.resolve(path.path)
+        val p = configDir.resolveSecurely(path)
 
         return (!isDirectoryPath && p.isFile) || (isDirectoryPath && p.isDirectory)
     }
 
     override fun listFiles(context: Context, path: Path): Set<Path> {
-        val dir = configDir.resolve(path.path)
+        val dir = configDir.resolveSecurely(path)
```
MarcelBochtler, line 78: Because `resolveSecurely()` canonicalizes the result, this breaks on MacOS, because e.g. `/tmp/...` will be changed to `/private/tmp/...`.

Due to this, tests fail on my machine.

mnonnenmacher:
> Found an correctness issue on macOS. But this can be solved in a follow-up PR.

Can you make the follow-up to make the test work on macOS?

> Edit. This might also be an issue on non-macOS machines, if the user uses symlinks.

IIUC this could be a problem if, for example, the `LocalConfigFileProvider` is used and the config directory is a symlink? I hope this is an edge case we can live with at least for now.

workers/reporter/src/main/kotlin/ReporterRunner.kt
```diff
@@ -416,9 +416,18 @@ private fun createAssetDirectory(asset: ReporterAsset, directory: File): File =
...
  * Transform this string to a [Path] for downloading a reporter template file. This requires removing the prefix
- * which marks this string as a template file.
+ * which marks this string as a template file. As the remaining value is user-controlled, reject absolute paths and
+ * paths containing ".." segments to prevent path traversal outside the config directory.
  */
-private fun String.toTemplatePath(): Path = Path(removePrefix(ReporterComponent.TEMPLATE_REFERENCE))
+internal fun String.toTemplatePath(): Path {
+    val raw = removePrefix(ReporterComponent.TEMPLATE_REFERENCE)
+
+    require(!File(raw).isAbsolute && raw.split('/', '\\').none { it == ".." }) {
```
MarcelBochtler, line 425: From an AI I got this feedback:

`toTemplatePath()` is also called during replacement for every comma-separated option segment (line 380), not only `reporter-template://` segments.

This newly rejects valid non-template absolute references in mixed values (e.g. `/opt/templates/custom.ftl,reporter-template://reporter/default.ftl`).

Validate only prefix-marked segments at the call site, or make this function require the prefix before applying the traversal checks.

MarcelBochtler:
> IIUC this could be a problem if, for example, the `LocalConfigFileProvider` is used and the config directory is a symlink? I hope this is an edge case we can live with at least for now.

Yes I think this would be the case.
And yes, I think that we can live with for now.

PR #5644, opened 2026-08-05
ui: Refactor remaining dropdowns to use infinite queries

### LINKED ISSUE #5383: Consider using infinite queries for UI dropdown content
Currently, these dropdowns which are populated from API endpoints use `ALL_ITEMS` for fetching:
- organization dropdowns in `admin/users/create-user`, `admin/plugins/...`, `move-repository`, `siblings`
- product/repository sibling dropdowns in `move-repository`, `siblings`

It should be investigated, whether fetching `ALL_ITEMS` in these dropdowns could be replaced with async searchable / infinite dropdowns, using this concept from TanStack Query: https://tanstack.com/query/v5/docs/framework/react/guides/infinite-queries

Etsija:
After this PR, all dropdowns that used to fetch large amount of data all at once, have been refactored to use infinite querying. Statistics:
- before this PR and #5637: 17 `ALL_ITEMS` usages in 10 files
- after: 10 `ALL_ITEMS` usages in 5 files left
- usages remaining: [use-secrets.ts](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/hooks/use-secrets.ts), [use-infrastructure-services.ts](https://github.com/eclipse-apoapsis/ort-server/blob/main/ui/src/hooks/use-infrastructure-services.ts), and the Issues, Projects, and Rule Violations tables of a run

All refactors have been verified by triggering the refactored dropdowns in the UI with a total of 104 organizations, and checking that when the dropdown list is scrolled from top to bottom, the paged queries are triggered exactly three times (50 + 50 + 4 = 104; page size for the queries is 50).

### DIFF
--- a/ui/src/components/siblings.tsx
+++ b/ui/src/components/siblings.tsx
@@ -17,86 +17,163 @@
  * License-Filename: LICENSE
  */
 
-import { useQuery } from '@tanstack/react-query';
+import {
+  useQuery,
+  type QueryFunction,
+  type QueryKey,
+  type SkipToken,
+} from '@tanstack/react-query';
 import { Link, useParams, useRouter } from '@tanstack/react-router';
 import { Check, ChevronDown, ExternalLink } from 'lucide-react';
+import { useState } from 'react';
 
 import {
+  getOrganizationProductsInfiniteOptions,
   getOrganizationProductsOptions,
+  getOrganizationsInfiniteOptions,
   getOrganizationsOptions,
+  getProductRepositoriesInfiniteOptions,
   getProductRepositoriesOptions,
   getRepositoryOptions,
   getRepositoryRunsOptions,
 } from '@/api/@tanstack/react-query.gen';
+import { PagingData } from '@/api/types.gen';
+import { SearchableInfiniteList } from '@/components/searchable-infinite-list';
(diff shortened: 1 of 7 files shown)


ui/src/components/siblings.tsx
```diff
@@ -55,48 +61,55 @@ export const Siblings = ({ entity, pathName }: SiblingsProps) => {
...
-  // Now, only the single query relevant to the clicked dropdown will be executed. Also ensure
-  // that queries that depend on a path parameter (like orgId) do not run until that parameter
-  // is actually present.
-  const {
-    data: organizations,
-    isPending: isOrgPending,
-    isError: isOrgError,
-    error: orgError,
-  } = useQuery({
-    ...getOrganizationsOptions({ query: { limit: ALL_ITEMS } }),
```
lamppu, line 67: > The drawback for this is that the chevron is now shown for every
organization, product and repository breadcrumb, where it used to appear
only when there was more than one entry to pick from. The reason is,
nothing is known about the number of entries before the dropdown is
opened.

If this becomes an issue though, this could just be found out by making a request with limit=1. But I don't think it's too bad the way it is now.

Etsija: Fixed it now, to avoid naggings from @sschuberth :smirk:

lamppu (approved):

Issue #5349, opened 2026-06-25
Issues, rule violations, projects endpoints need filters and additional sorting


Etsija:
In order to get rid of client-side data manipulation of the issues, rule violations, and projects tables in the UI - and, correspondingly, usage of ALL_ITEMS when fetching these items from the back-end, the following filters need to be added to the corresponding API endpoints:

1. Issues endpoint: package identifier, PURL, severity.
2. Rule violations endpoint: package identifier, PURL, severity, rule.
3. Projects endpoint: identifier, declared license, definition file path.

The tables need to be also sortable by the listed fields, so adding the filtering and sorting to the endpoints will enable the UI to switch to using server-side data manipulation in the corresponding tables.

Additionally needed: endpoints for all filters which are going to be used in the UI with select type of filtering. These endpoints will gather all elements found from the run for using in the select filter. A good example of such an existing endpoint is `/runs/:runId/packages/licenses`
1. Rule violations: rules.
2. Projects: declared licenses.

Refactoring each endpoint will need to be followed (as the last commit) by switching the corresponding UI table from client-side to server-side data manipulation.

The Issues table currently has a select filter for Category, but that's implemented purely in the UI with RegEx matching of the Issue messages. Implementing this properly on back-end would require changes in ORT Core, which are outside the scope of this issue. Therefore, Category filtering needs to be dropped completely from the UI, and there will be no filtering with Issue category when the Issues table is converted to server-side.

PR #5682, opened 2026-08-11
Add filtering and sorting to issues endpoint

Etsija:
This PR adds back-end filtering and sorting to issues endpoint and converts the issues table in the UI to using server-side data manipulation, as described in #5349. The two first commits are just refactors, which prepare for using helper functions for PURLs and severities for several endpoints.


mnonnenmacher (dismissed):

services/ort-run/src/main/kotlin/IssueService.kt
```diff
@@ -110,29 +131,98 @@ class IssueService(
...
+        issuesFilter.severity?.let { filter ->
+            require(filter.operator == ComparisonOperator.IN || filter.operator == ComparisonOperator.NOT_IN) {
+                "Unsupported operator for severity filter: ${filter.operator}"
+            }
+
+            query.andWhere {
+                when (filter.operator) {
+                    ComparisonOperator.IN -> IssuesTable.severity inList filter.value
+                    else -> IssuesTable.severity notInList filter.value
+                }
```
lamppu: There's an `applyFilter` overload in the Extensions for collections that could be used here

```
query.andWhere { IssuesTable.severity.applyFilter(filter.operator, filter.value) }
```

lamppu (approved):

PR #5685, opened 2026-08-11
Allow to set a root path for Scaleway secrets provider

### LINKED ISSUE #5626: Allow to set a base path for Scaleway Secrets Provider
An API key may only have permission to manage secrets in a specific path, so allow to configure a base path for the secrets.

lamppu:


secrets/scaleway/src/test/kotlin/ScalewaySecretsProviderTest.kt
```diff
@@ -72,10 +169,172 @@ class ScalewaySecretsProviderTest : WordSpec({
...
 private fun createProvider(config: ScalewayConfiguration = createConfig()) =
     ScalewaySecretsProvider(config)
+
+private const val PATH_PREFIX = "/secret-manager/v1beta1/regions/fr-par"
+private const val SECRETS_ENDPOINT = "$PATH_PREFIX/secrets"
+private const val SECRETS_ACCESS_ENDPOINT = "$PATH_PREFIX/secrets-by-path/versions/latest/access"
+private const val VERSIONS_ENDPOINT = "$SECRETS_ENDPOINT/secret-id/versions"
+private const val DELETE_ENDPOINT = "$SECRETS_ENDPOINT/secret-id"
+
+private fun WireMockServer.stubSecretAccess() {
```
mnonnenmacher, line 239: How did you get the actual response bodies, are they documented or did you do test requests? Could you add a comment with a short explanation how to get them in case they ever need to be updated?

secrets/scaleway/src/test/kotlin/ScalewaySecretsProviderTest.kt
```diff
@@ -72,10 +169,172 @@ class ScalewaySecretsProviderTest : WordSpec({
...
+}
+
+private fun WireMockServer.verifySecretDelete(count: Int = 1) {
+    verify(
+        count,
+        deleteRequestedFor(urlPathEqualTo(DELETE_ENDPOINT))
+    )
+}
+
+private fun secretResponse() =
```
mnonnenmacher: Could this be a `val` instead?

mnonnenmacher (changes requested):

secrets/scaleway/src/test/kotlin/ScalewaySecretsProviderTest.kt
lamppu: They're based on the model classes in `ScalewaySecretsModel`, those have comments with links to the  developer API docs.

lamppu: But of course if you'd prefer it, I can add links as comments separately here as well.

mnonnenmacher: No, then that's good enough for me.

mnonnenmacher (approved):

PR #5688, opened 2026-08-11
Refactor `admin-config`

mnonnenmacher:
Refactor the `admin-config` module with the goal to simplify the code as a preparation to add more config options. The main refactorings are:

* Replace the manual mapping of config file properties with Hoplite to reduce the amount of boilerplate code.
* Move default values to model classes to have a single source of truth. Previously, defaults were different when building the model in code vs. building the model from the config file.
* Move post-processing logic of config values to the domain model to improve separation of concerns and to ensure it is consistent also when creating the admin config in code.

To be able to implement the changes one by one without breaking anything, the PR creates some temporary code which is removed again by the end of the refactoring. To get an overall impression of the refactoring, best first look at the complete PR diff.


gradle/libs.versions.toml
```diff
@@ -39,6 +39,7 @@ commonsText = "1.15.0"
 exposed = "1.4.0"
 flyway = "13.2.0"
 hikari = "7.1.0"
+hoplite = "2.9.0"
```
Etsija, line 42: A reference to hoplite in the commit message which introduces it would be benefitial, and I believe has been a custom to our project.

oheger-bosch:
I have to admit that I am not that happy with introducing another configuration mechanism; specifically, as Hoplite does too much magic for my taste.

Does this change impact the way how environment variables are interpreted by the configuration?

mnonnenmacher:
> I have to admit that I am not that happy with introducing another configuration mechanism; specifically, as Hoplite does too much magic for my taste.
> 
> Does this change impact the way how environment variables are interpreted by the configuration?

By default the following [preprocessors](https://github.com/sksamuel/hoplite#built-in-preprocessors) are enabled (for the used version, the list on the website refers to the Hoplite 3 RC):
* EnvOrSystemPropertyPreprocessor
* RandomPreprocessor
* LookupPreprocessor

Also these [properties sources]() are enabled by default:
* EnvironmentVariableOverridePropertySource
* SystemPropertiesPropertySource
* UserSettingsPropertySource
* XdgConfigPropertySource

So any `${ENV_VAR_NAME}` values should be resolved as before.

sschuberth:
> introducing another configuration mechanism

I believe the goal is to fully migrate from Typesafe Config to Hoplite, and not to have a mixed state where parts of the code base use one configuration mechanism and others another mechanism.

So I regard this PR to only be the first step. But maybe we should track that whole effort in an issue, @mnonnenmacher?

In general, migrating to Hoplite makes sense to me as we're already familiar with it from ORT (where we did the same migration a long time ago) and Hoplite 3 will bring some nice new features and improvements.

Also, from the diff stats alone this seems to be a sensible change to reduce complexity:

image


> specifically, as Hoplite does too much magic for my taste.
> 
> Does this change impact the way how environment variables are interpreted by the configuration?

I believe @mnonnenmacher was able to allay those concerns with his answer.

components/admin-config/backend/src/main/kotlin/ReporterConfig.kt
```diff
@@ -284,3 +353,5 @@ data class ReporterConfig(
             }
     }
 }
+
+private fun String.ensureTrailingSlash() = takeIf { it.endsWith("/") } ?: plus("/")
```
sschuberth, line 357: ORT by now has

```kotlin
fun String.ensureSuffix(suffix: String) = takeIf { it.endsWith(suffix) } ?: "$this$suffix"
```

Maybe use that instead? Could be done as a follow-up.

components/admin-config/backend/src/main/kotlin/RuleSet.kt
```diff
@@ -19,24 +19,46 @@
...
 
     /** The path to the files with rules to use for the evaluation. */
-    val evaluatorRules: String
+    val evaluatorRules: String = DEFAULT_EVALUATOR_RULES_FILE
+) {
+    companion object {
+        internal const val DEFAULT_COPYRIGHT_GARBAGE_FILE = ORT_COPYRIGHT_GARBAGE_FILENAME
+        internal const val DEFAULT_LICENSE_CLASSIFICATIONS_FILE = ORT_LICENSE_CLASSIFICATIONS_FILENAME
+        internal const val DEFAULT_RESOLUTIONS_FILE = ORT_RESOLUTIONS_FILENAME
+        internal const val DEFAULT_EVALUATOR_RULES_FILE = ORT_EVALUATOR_RULES_FILENAME
```
sschuberth, line 54: Why have these mappings and not use straight ORT's constants?

workers/reporter/src/test/kotlin/ReporterRunnerTest.kt
```diff
@@ -785,9 +785,9 @@ class ReporterRunnerTest : WordSpec({
             runner.run(OrtResult.EMPTY, jobConfig, null, context)
 
             downloadedAssets.keys shouldHaveSize 2
-            downloadedAssets[Path("data")] shouldBe configDirectory
+            downloadedAssets[Path("data/")] shouldBe configDirectory
 
-            downloadedAssets[Path("images")] shouldNotBeNull {
+            downloadedAssets[Path("images/")] shouldNotBeNull {
```
sschuberth, line 790: It's a bit unfortunate that the trailing slashes are needed now. Can you think of a way to avoid it?

BTW, reading the `Path` docs it says

> A class representing a path of a property in the configuration of the application.

but that description does not seem to be precise to me, because if something like "data/" is allowed, that's not a property, but a group of properties, or?

oheger-bosch:
> I believe the goal is to fully migrate from Typesafe Config to Hoplite, and not to have a mixed state where parts of the code base use one configuration mechanism and others another mechanism.
> 
> So I regard this PR to only be the first step. But maybe we should track that whole effort in an issue, @mnonnenmacher?
> 

This does not seem to be a trivial task. Especially, since the wide-spread used `ConfigManager` class is closely related to Typesafe Config. How do you plan to deal with this?

components/admin-config/backend/src/main/kotlin/ReporterConfig.kt
mnonnenmacher: I did not want to introduce a dependency just for one small helper function, especially as ORT's common-utils depends on Spring core and commons compress.

workers/reporter/src/test/kotlin/ReporterRunnerTest.kt
mnonnenmacher: I guess `Path` was used here for convenience even though the usage does not match the class docs.

I actually liked that the slashes became visible here now, because that shows that the test data is aligned with the contract of the admin config. On the other hand I'm not sure if it is even required that directory paths end with a slash like it is currently enforced in `ReporterConfig`, but I wanted to minimize the functional changes in this refactoring.

components/admin-config/backend/src/main/kotlin/RuleSet.kt
mnonnenmacher: The `AdminConfigService` also uses the constants to determine if the values of a `RuleSet` are the default values, so I wanted to ensure that both classes use the same values.

mnonnenmacher:
> > I believe the goal is to fully migrate from Typesafe Config to Hoplite, and not to have a mixed state where parts of the code base use one configuration mechanism and others another mechanism.
> > So I regard this PR to only be the first step. But maybe we should track that whole effort in an issue, @mnonnenmacher?
> 
> This does not seem to be a trivial task. Especially, since the wide-spread used `ConfigManager` class is closely related to Typesafe Config. How do you plan to deal with this?

I would indeed really like to introduce Hoplite also for parsing the application.conf files, because it would allow to remove a lot of boilerplate code and it provides extension points to add custom behavior where required. But I have not made specific plans how to do it, because there are too many other more urgent tasks currently. I introduced it here because the admin config file is separate from the `application.conf` files and does not support features like different providers for secret properties.

sschuberth (approved):

oheger-bosch:
> > > I believe the goal is to fully migrate from Typesafe Config to Hoplite, and not to have a mixed state where parts of the code base use one configuration mechanism and others another mechanism.
> > > So I regard this PR to only be the first step. But maybe we should track that whole effort in an issue, @mnonnenmacher?
> > 
> > 
> > This does not seem to be a trivial task. Especially, since the wide-spread used `ConfigManager` class is closely related to Typesafe Config. How do you plan to deal with this?
> 
> I would indeed really like to introduce Hoplite also for parsing the application.conf files, because it would allow to remove a lot of boilerplate code and it provides extension points to add custom behavior where required. But I have not made specific plans how to do it, because there are too many other more urgent tasks currently. I introduced it here because the admin config file is separate from the `application.conf` files and does not support features like different providers for secret properties.

So, I guess we all know how this will work out: From now on up to the far future, there will be two different configuration libraries in use.

sschuberth:
> So, I guess we all know how this will work out: From now on up to the far future, there will be two different configuration libraries in use.

Personally, I'm not so pessimistic in the age of AI coding agents.

Issue #5004, opened 2026-05-06
Policy resolution in UI leads to technical issue


hanna-modica:
I created the following policy resolution in the UI:

<img width="1127" height="268" alt="Image" src="https://github.com/user-attachments/assets/d1a303ee-0b5e-4891-ac8b-e955e547f828" />

This lead to the following issue in the reporter job:

<img width="1169" height="832" alt="Image" src="https://github.com/user-attachments/assets/a0e41dc0-b5c5-477d-9dc1-b7f08f49a4b4" />

This is my reporter configuration:

```
"reporter": {
      "formats": [
        "CycloneDX",
        "SpdxDocument",
        "WebApp",
        "OrtResult"
      ],
      "config": {
        "SpdxDocument": {
          "options": {
            "outputFileFormats": "YAML,JSON"
          },
          "secrets": {}
        },
        "CycloneDX": {
          "options": {
            "outputFileFormats": "XML,JSON"
          },
          "secrets": {}
        }
      }
```

Maybe one of the report formats cannot handle the unescaped single quote.

sschuberth:
@hanna-modica, can you please narrow this down to the exact report format that failed? E.g. by looking at the logs, or enabling them one-by-one?

hanna-modica:
Hi @sschuberth,
I did a few reruns with different formats and it failed every time with the same error. I also tried new formats, like statichtml.

sschuberth:
Is anyone from @eclipse-apoapsis/bosch planning to work on this, as you seem to be able to reproduce it?

sschuberth:
For convenience, here's the text as text:

> Illegal character range near index 30
> Security vulnerabilities [CVE-2026-27903, CVE-2026-27904, CVE-2026-26996] of severities [HIGH, HIGH, HIGH], which are equal or greater than the configured threshold, found in
'NPM::minimatch:9.0.1'.

"index 30" is the "-" in "CVE-2026-27903". So apparently, the matcher string does not get Regex chars (like `[a-z]`) escaped when creating a resolution via the UI.

sschuberth:
I believe we should do several things to address this:
- The API / backend should continue to take a Regex to match ORT's "low-level" behavior.
- If an invalid Regex is sent to the API, the API should respond with an error (to fail early instead of only at reporter stage).
- The frontend should take care of escaping any reserved Regex chars in the message before sending it to the API.

MarcelBochtler:
[...]

PR #5706, opened 2026-08-13
fix(resolutions): Escape the message of managed resolutions

oheger-bosch:
The issue and rule violation resolutions managed by ORT Server do an exact match against the
message. The corresponding ORT resolutions in contrast interpret the message as
a regular expression. Therefore, when ORT Server resolutions are
converted to ORT resolutions, the message must be escaped. Otherwise,
special characters are interpreted by the Regex logic or could even cause
syntax errors.

### DIFF
--- a/services/ort-run/src/main/kotlin/OrtServerMappings.kt
+++ b/services/ort-run/src/main/kotlin/OrtServerMappings.kt
@@ -606,7 +606,9 @@ fun RuleViolation.mapToOrt() =
 
 fun RuleViolationResolution.mapToOrt() =
     OrtRuleViolationResolution(
-        message = message,
+        // ORT interprets the message as a regex, while Server-managed resolutions are exact string matches.
+        // Therefore, the message must be escaped.
+        message = message.takeUnless { source == ResolutionSource.SERVER } ?: Regex.escape(message),
         reason = reason.mapToOrt(),
         comment = comment
     )
--- a/services/ort-run/src/main/kotlin/OrtServerMappings.kt
+++ b/services/ort-run/src/main/kotlin/OrtServerMappings.kt
@@ -74,6 +74,7 @@ import org.eclipse.apoapsis.ortserver.model.runs.repository.PathInclude
 import org.eclipse.apoapsis.ortserver.model.runs.repository.ProvenanceSnippetChoices
 import org.eclipse.apoapsis.ortserver.model.runs.repository.RepositoryAnalyzerConfiguration
 import org.eclipse.apoapsis.ortserver.model.runs.repository.RepositoryConfiguration
+import org.eclipse.apoapsis.ortserver.model.runs.repository.ResolutionSource
 import org.eclipse.apoapsis.ortserver.model.runs.repository.Resolutions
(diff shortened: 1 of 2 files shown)


MarcelBochtler (approved):

sschuberth:
@oheger-bosch, @MarcelBochtler, should this have closed https://github.com/eclipse-apoapsis/ort-server/issues/5004?

oheger-bosch:
The fix was actually for another problem where resolutions were not effective. #5004 may be related, but this should be tested explicitly.

oheger-bosch:
BTW, I was quite surprised to find out that the server-managed resolutions are working differently with regard of matching than the other resolutions. Has this been a conscious decision? Couldn't this be a source of confusion for users?

sschuberth:
> BTW, I was quite surprised to find out that the server-managed resolutions are working differently with regard of matching than the other resolutions.

Can you please elaborate in more details what exactly you're referring to?

oheger-bosch:
The resolutions from ORT are using a regular expression to match against an issue message. The server-managed resolutions for issues and rule validations in contrast do an exact match against the message. So, here the message must be provided exactly. This happens behind the scenes when resolutions are created by the UI; but I am not sure whether this discrepancy is clear for users of the REST API.

sschuberth:
> The server-managed resolutions for issues and rule validations in contrast do an exact match against the message.

I do not recall whether this has been a conscious decision or not. I vaguely remember some discussion that for UI user's it's easier to copy & paste messages to create resolutions, but I feel that also from the UI it should be possible to create regex-based resolutions.

Do any of @mnonnenmacher, @Etsija, or @lamppu remember?

Anyway, in the UI we could have a radio button to toggle between "literal" and "regex" matching, and when the user chooses "literal", internally still regex-matching would be used, but with all regex-chars escaped, so that the matching in fact is literal.

oheger-bosch:
> Anyway, in the UI we could have a radio button to toggle between "literal" and "regex" matching, and when the user chooses "literal", internally still regex-matching would be used, but with all regex-chars escaped, so that the matching in fact is literal.

I fear it won't be that easy. In the current implementation of the server-managed resolutions, hashes are calculated on the messages to resolve. The matching between resolutions and issues/rule violations is done using those hashes. This does not support a pattern matching on messages.

mnonnenmacher:
> > The server-managed resolutions for issues and rule validations in contrast do an exact match against the message.
> 
> I do not recall whether this has been a conscious decision or not. I vaguely remember some discussion that for UI user's it's easier to copy & paste messages to create resolutions, but I feel that also from the UI it should be possible to create regex-based resolutions.
> 
> Do any of @mnonnenmacher, @Etsija, or @lamppu remember?

This was done because with the UI resolutions can be created for exactly one item to resolve, resolutions that match multiple items were out of scope when this was implemented. So users can only set the reason and comment, but not the matcher.

> Anyway, in the UI we could have a radio button to toggle between "literal" and "regex" matching, and when the user chooses "literal", internally still regex-matching would be used, but with all regex-chars escaped, so that the matching in fact is literal.

I would rather improve ORT itself to provide a better way for matching multiple items without using regular expressions.

PR #5712, opened 2026-08-14
fix(ui): Clarify that these are only the first fixed versions

sschuberth:
This also aligns with the internal name of the property.


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/runs/$runIndex/vulnerabilities/index.tsx
```diff
@@ -464,7 +464,7 @@ const VulnerabilitiesComponent = () => {
         >
           <AccordionItem value='fixed-versions'>
             <AccordionTrigger className='font-semibold'>
-              Fixed versions
+              First fixed versions
```
sschuberth, line 467: Or should we rather say "Versions introducing a fix", @willebra?

willebra: I like the word "first" being somehow involved. While "introducing" hints to further versions remaining fixed, it's not as express.

mnonnenmacher (approved):

PR #5717, opened 2026-08-14
build: Remove requirement for Docker when generating the OpenAPI spec

MarcelBochtler:
Generating the spec doesn't require a database connection, so avoid initializing the database extension and therefore also avoid the requirement for Docker.


sschuberth:
Just to double-check, did you read through the commit message of 2f1634e8611b0546c7801245161488e6dd7ceb0a? Is this part not relevant anymore:

> Upcoming changes will require that the database is already available when installing the Ktor modules.

mnonnenmacher:
> Just to double-check, did you read through the commit message of [2f1634e](https://github.com/eclipse-apoapsis/ort-server/commit/2f1634e8611b0546c7801245161488e6dd7ceb0a)? Is this part not relevant anymore:
> 
> > Upcoming changes will require that the database is already available when installing the Ktor modules.

If no database is provided for the test application, a mock will be used instead to fulfill that requirement. So as the tests pass, this shows that the database is indeed not required.

mnonnenmacher (approved):

PR #5735, opened 2026-08-17
refactor(workers): Remove unused `secretResolverService` argument

mnonnenmacher:
Remove the argument from the `workerContextModule` function because it was only used by a test that verified that the argument works.

### DIFF
--- a/workers/common/src/test/kotlin/context/WorkerContextModuleTest.kt
+++ b/workers/common/src/test/kotlin/context/WorkerContextModuleTest.kt
@@ -60,20 +60,6 @@ class WorkerContextModuleTest : KoinTest, StringSpec() {
                 secretResolverService.getSecretValue(secret) shouldBe secretValue
             }
         }
-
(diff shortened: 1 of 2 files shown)


workers/common/src/test/kotlin/context/WorkerContextModuleTest.kt
```diff
@@ -60,26 +60,15 @@ class WorkerContextModuleTest : KoinTest, StringSpec() {
...
     }
 
     /**
      * Prepare the test environment for a test of the given [workerModule] and then execute the given test [block].
      */
-    private suspend fun runModuleTest(workerModule: Module = workerContextModule(), block: suspend () -> Unit) {
+    private suspend fun runModuleTest(
+        workerModule: Module = workerContextModule(),
+        block: suspend () -> Unit
+    ) {
```
sschuberth: Unrelated formatting change?

sschuberth (approved):

PR #5743, opened 2026-08-18
Cap paged request limits at 100 (or a specified value)

### LINKED ISSUE #4073: Create an upper bound for the limit of paged requests
Currently, the API endpoints that support pagination do not define an upper bound for the limit parameter. This is problematic, because this means that clients can overload the backend by requesting a very large number of items. To fix this, introduce an upper bound for the limit parameter, and if a client requests a larger limit, coerce to that upper bound. The proposed upper bound is 100.

Before this upper bound can be introduced, there are several parts of the UI that need to be adapted that currently request a very large number of items. For each them, it needs to be decided if either the UX should be changed, or if a dedicated endpoint should be added, that provides the required information in an efficient way. The affected parts of the UI can be found by searching for usages of the `ALL_ITEMS` constant:
https://github.com/eclipse-apoapsis/ort-server/blob/bac0b7582c2773af8686a623f657b831f38e2bfe/ui/src/lib/constants.ts#L20-L26

Etsija:

### DIFF
--- a/shared/ktor-utils/src/main/kotlin/OpenApiHelpers.kt
+++ b/shared/ktor-utils/src/main/kotlin/OpenApiHelpers.kt
@@ -25,6 +25,8 @@ import io.github.smiley4.ktoropenapi.config.SimpleBodyConfig
 
 import io.ktor.http.ContentType
 
+import org.eclipse.apoapsis.ortserver.model.util.ListQueryParameters
+
 inline fun <reified T> RequestConfig.jsonBody(noinline block: SimpleBodyConfig.() -> Unit) =
     body<T> {
         mediaTypes = setOf(ContentType.Application.Json)
@@ -49,11 +51,15 @@ fun RequestConfig.standardSortQueryParameter() {
 }
 
 /**
- * Generate documentation for standard list query parameters.
+ * Generate documentation for standard list query parameters, using [maxLimit] as the maximum accepted limit.
  */
-fun RequestConfig.standardListQueryParameters() {
+fun RequestConfig.standardListQueryParameters(maxLimit: Int = ListQueryParameters.DEFAULT_MAX_LIMIT) {
+    require(maxLimit > 0) { "The maximum limit must be positive." }
+
     queryParameter<Int>("limit") {
-        description = "The maximum number of items to retrieve. If not specified at most 20 items are retrieved."
+        description = "The maximum number of items to retrieve. Values greater than $maxLimit are reduced to this " +
+                "maximum. If not specified, at most ${minOf(ListQueryParameters.DEFAULT_LIMIT, maxLimit)} items are " +
+                "retrieved."
     }
     queryParameter<Long>("offset") {
         description = "The offset of the first item in the result. Together with 'limit', this can be used to " +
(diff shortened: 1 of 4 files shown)


shared/ktor-utils/src/main/kotlin/PaginationHelpers.kt
```diff
@@ -31,15 +31,17 @@ import org.eclipse.apoapsis.ortserver.shared.apimodel.SortProperty
...
+ * maximum. If no offset is provided, 0 is used. If no sort order is provided, the [defaultSortProperty] is used.
  *
  * The default values ensure that reproducible results are returned and that large numbers of results are avoided.
  */
 fun ApplicationCall.pagingOptions(defaultSortProperty: SortProperty): PagingOptions {
     val sortProperties = parameters["sort"]?.let(::processSortParameter).orEmpty().takeIf { it.isNotEmpty() }
         ?: listOf(defaultSortProperty)
-    val limit = numberParameter("limit")?.toInt()?.takeIf { it > 0 } ?: ListQueryParameters.DEFAULT_LIMIT
+    val limit = numberParameter("limit")?.toLong()?.takeIf { it > 0 }
+        ?.coerceAtMost(ListQueryParameters.MAX_LIMIT.toLong())?.toInt()
```
oheger-bosch: I am fine with these changes in principle. Just thinking: Is this global limit appropriate in all cases? For large data structures like ORT runs, maybe a smaller limit would make sense. To leave this possibility open, the function could expect the limit as a parameter which defaults to `ListQueryParameters.MAX_LIMIT`.

Etsija: So adding a parameter for it to `pagingOptions()`, capping that also to `MAX_LIMIT`? I have nothing against that approach - it leaves a possibility to later on cap some paged requests to even lower limit. @mnonnenmacher what do you say?

mnonnenmacher: I agree, having the ability the adjust the limit for specific requests could be useful. I think capping it at `MAX_LIMIT` should not be necessary, maybe higher limits can also be useful in some places.

shared/ktor-utils/src/main/kotlin/OpenApiHelpers.kt
```diff
@@ -49,11 +51,15 @@ fun RequestConfig.standardSortQueryParameter() {
...
+ * Generate documentation for standard list query parameters, using [maxLimit] as the maximum accepted limit.
  */
-fun RequestConfig.standardListQueryParameters() {
+fun RequestConfig.standardListQueryParameters(maxLimit: Int = ListQueryParameters.DEFAULT_MAX_LIMIT) {
+    require(maxLimit > 0) { "The maximum limit must be positive." }
+
     queryParameter<Int>("limit") {
-        description = "The maximum number of items to retrieve. If not specified at most 20 items are retrieved."
+        description = "The maximum number of items to retrieve. Values greater than $maxLimit are reduced to this " +
+                "maximum. If not specified at most ${minOf(ListQueryParameters.DEFAULT_LIMIT, maxLimit)} items are " +
```
mnonnenmacher: I think there should be a comma after "specified".

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #5745, opened 2026-08-18
fix(evaluator): Apply vulnerability resolutions

oheger-bosch:
When setting up the resolution provider for the Evaluator run, also provide a `VulnerabilityResolutionService` instance. This is needed to make functions of the rule DSL like `hasVulnerability()` aware of server-managed vulnerability resolutions.

### DIFF
--- a/workers/evaluator/src/test/kotlin/EvaluatorRunnerTest.kt
+++ b/workers/evaluator/src/test/kotlin/EvaluatorRunnerTest.kt
@@ -21,6 +21,7 @@ package org.eclipse.apoapsis.ortserver.workers.evaluator
 
 import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.collections.beEmpty as collectionBeEmpty
 import io.kotest.matchers.collections.containExactly
 import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.maps.beEmpty
@@ -75,6 +76,7 @@ import org.ossreviewtoolkit.utils.spdxexpression.toSpdx
 
 const val SCRIPT_FILE = "/example.rules.kts"
 private const val PACKAGE_CONFIGURATION_RULES = "package-configurations.rules.kts"
+private const val VULNERABILITY_RULES = "vulnerability.rules.kts"
 private const val LICENSE_CLASSIFICATIONS_FILE = "/license-classifications.yml"
 private const val RESOLUTIONS_FILE = "/resolutions.yml"
 private const val UNKNOWN_RULES_KTS = "unknown.rules.kts"
@@ -300,6 +302,42 @@ class EvaluatorRunnerTest : WordSpec({
             }
         }
 
+        "not create a violation for a vulnerability that is resolved in the repository configuration" {
+            val ruleSet = testRuleSet.copy(evaluatorRules = VULNERABILITY_RULES)
+            adminConfigService.initRuleSet(ruleSet)
+
+            val result = runner.run(
+                OrtTestData.result,
+                EvaluatorJobConfiguration(),
+                createWorkerContext()
+            )
+
+            result.evaluatorRun.violations should collectionBeEmpty()
+        }
+
+        "not create a violation for a vulnerability that is only resolved in the resolved configuration" {
+            val ruleSet = testRuleSet.copy(evaluatorRules = VULNERABILITY_RULES)
+            adminConfigService.initRuleSet(ruleSet)
+
+            // Remove the resolution from the repository configuration so that the only remaining matching
+            // resolution is the one already present in OrtTestData.resolvedConfiguration.resolutions.vulnerabilities.
+            val ortResultWithResolutionInResolvedConfiguration = OrtTestData.result.copy(
+                repository = OrtTestData.repository.copy(
+                    config = OrtTestData.repository.config.copy(
+                        resolutions = OrtTestData.repository.config.resolutions.copy(vulnerabilities = emptyList())
+                    )
+                )
+            )
+
+            val result = runner.run(
+                ortResultWithResolutionInResolvedConfiguration,
+                EvaluatorJobConfiguration(),
+                createWorkerContext()
+            )
+
+            result.evaluatorRun.violations should collectionBeEmpty()
+        }
+
         "load managed rule violation resolutions from the server" {
             adminConfigService.initRuleSet(testRuleSet)
 
@@ -344,6 +382,9 @@ private fun createConfigManager(): ConfigManager {
 
         every { getFileAsString(resolvedConfigContext, Path(PACKAGE_CONFIGURATION_RULES)) } returns
                 File("src/test/resources/$PACKAGE_CONFIGURATION_RULES").readText()
+
+        every { getFileAsString(resolvedConfigContext, Path(VULNERABILITY_RULES)) } returns
+                File("src/test/resources/$VULNERABILITY_RULES").readText()
 
         every { getFile(resolvedConfigContext, Path(LICENSE_CLASSIFICATIONS_FILE)) } answers
                 { File("src/test/resources/license-classifications.yml").inputStream() }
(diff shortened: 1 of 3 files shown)


oheger-bosch:
To give some background: We have a rule in place that generates a violation if it finds a vulnerability with high severity (based on some threshold). This rule was firing, even if there was a server resolution for the vulnerability.

@Etsija, I would be interested in your opinion here. What was the reason why the `VulnerabilityResolutionService` was not used by the Evaluator runner when setting up the resolution provider? And may the `IssueResolutionService` need to be used as well?

workers/evaluator/src/test/kotlin/EvaluatorRunnerTest.kt
```diff
@@ -285,7 +295,7 @@ class EvaluatorRunnerTest : WordSpec({
             }
         }
 
-        "only resolve rule violations, not issues or vulnerabilities" {
+        "only resolve rule violations, not issues or vulnerabilities; but apply vulnerability resolutions" {
```
mnonnenmacher: The name is now contradictory regarding vulnerabilities, how about "resolve rule violations and vulnerabilities, but not issues"?

mnonnenmacher (changes requested):

Etsija:
> @Etsija, I would be interested in your opinion here. What was the reason why the `VulnerabilityResolutionService` was not used by the Evaluator runner when setting up the resolution provider? And may the `IssueResolutionService` need to be used as well?

That reasoning predates my change - it comes from c4d00d25, which set the rule "only the worker that adds items of a type needs that type's resolutions". I believe the rule breaks for the evaluator, though, as the rule is about what the worker writes, but the evaluator also reads resolutions. It probably went unnoticed because resolutions from the config files aways reached the evaluator - only the server-managed ones were missing.

I don't think `IssueResolutionService` is needed here - nothing in ORT's evaluator checks issue resolutions, and we pass an empy issue map to `matchResolutions()`.

mnonnenmacher:
> > @Etsija, I would be interested in your opinion here. What was the reason why the `VulnerabilityResolutionService` was not used by the Evaluator runner when setting up the resolution provider? And may the `IssueResolutionService` need to be used as well?
> 
> That reasoning predates my change - it comes from [c4d00d2](https://github.com/eclipse-apoapsis/ort-server/commit/c4d00d2513fcceb2bea62fecbded60479ac8b64e), which set the rule "only the worker that adds items of a type needs that type's resolutions". I believe the rule breaks for the evaluator, though, as the rule is about what the worker writes, but the evaluator also reads resolutions. It probably went unnoticed because resolutions from the config files aways reached the evaluator - only the server-managed ones were missing.

If the advisor already resolves vulnerabilities than those resolutions should already be stored in the `resolvedConfiguration` of the `OrtResult`, so it should not be necessary to resolve them again in later steps. Is there maybe some issue with storing those resolutions in the advisor worker or does the evaluator worker not load them correctly?

Or maybe it's again the issue from #5706 and the resolutions messages need to be escaped before adding them to the `resolvedConfiguration`. @oheger-bosch you could check if the resolutions are contained in the `resolvedConfiguration` of the `OrtResult` report.

oheger-bosch:
> If the advisor already resolves vulnerabilities than those resolutions should already be stored in the `resolvedConfiguration` of the `OrtResult`, so it should not be necessary to resolve them again in later steps. Is there maybe some issue with storing those resolutions in the advisor worker or does the evaluator worker not load them correctly?

I think the underlying problem is that there are now multiple sources for resolutions which are not always consistent. The vulnerability resolutions are correctly resolved and stored by the advisor worker. They are also loaded correctly and work as expected if resolution-related methods from `OrtResult` are called.

The `hasVulnerability()` method in `PackageRule` however uses the passed in resolution provider to apply resolutions to vulnerabilities. Here, the existing code passed an `OrtServerResolutionProvider` for which no vulnerability resolution service was set. Therefore, the server-managed vulnerability resolutions were not loaded and applied.

workers/evaluator/src/test/kotlin/EvaluatorRunnerTest.kt
oheger-bosch: Well, this name does not really fit. Vulnerabilities are not actually resolved - this has already been done by the advisor. The resolutions nevertheless need to be taken into account when processing rules. Therefore, I think the chosen test name expresses this complex situation quite accurate.

mnonnenmacher:
> I think the underlying problem is that there are now multiple sources for resolutions which are not always consistent. The vulnerability resolutions are correctly resolved and stored by the advisor worker. They are also loaded correctly and work as expected if resolution-related methods from `OrtResult` are called.
> 
> The `hasVulnerability()` method in `PackageRule` however uses the passed in resolution provider to apply resolutions to vulnerabilities. Here, the existing code passed an `OrtServerResolutionProvider` for which no vulnerability resolution service was set. Therefore, the server-managed vulnerability resolutions were not loaded and applied.

So the idea was that all workers apply resolutions for items they created, so that they resolutions are immediately visible in the UI. If this is implemented correctly, they should also be stored in the `resolvedConfiguration`. Then it should be sufficient to pass the `OrtResult` itself (which implements `ResolutionProvider` and provides resolutions from the `resolvedConfiguration`) as `ResolutionProvider` when calling ORT's `Evaluator` and `Reporter`. Except if server managed resolutions are not stored in the `resolvedConfiguration` for some reason, I don't remember the details right now.

oheger-bosch:
> > I think the underlying problem is that there are now multiple sources for resolutions which are not always consistent. The vulnerability resolutions are correctly resolved and stored by the advisor worker. They are also loaded correctly and work as expected if resolution-related methods from `OrtResult` are called.
> > The `hasVulnerability()` method in `PackageRule` however uses the passed in resolution provider to apply resolutions to vulnerabilities. Here, the existing code passed an `OrtServerResolutionProvider` for which no vulnerability resolution service was set. Therefore, the server-managed vulnerability resolutions were not loaded and applied.
> 
> So the idea was that all workers apply resolutions for items they created, so that they resolutions are immediately visible in the UI. If this is implemented correctly, they should also be stored in the `resolvedConfiguration`. Then it should be sufficient to pass the `OrtResult` itself (which implements `ResolutionProvider` and provides resolutions from the `resolvedConfiguration`) as `ResolutionProvider` when calling ORT's `Evaluator` and `Reporter`. Except if server managed resolutions are not stored in the `resolvedConfiguration` for some reason, I don't remember the details right now.

Then I find the API of `Evaluator` quite confusing. If `OrtResult` already is a resolution provider (which from an OO design perspective might be debatable), why does the `Evaluator` then support an additional resolution provider argument? Are there valid use cases where here a different provider is needed?

mnonnenmacher:
> > > I think the underlying problem is that there are now multiple sources for resolutions which are not always consistent. The vulnerability resolutions are correctly resolved and stored by the advisor worker. They are also loaded correctly and work as expected if resolution-related methods from `OrtResult` are called.
> > > The `hasVulnerability()` method in `PackageRule` however uses the passed in resolution provider to apply resolutions to vulnerabilities. Here, the existing code passed an `OrtServerResolutionProvider` for which no vulnerability resolution service was set. Therefore, the server-managed vulnerability resolutions were not loaded and applied.
> > 
> > 
> > So the idea was that all workers apply resolutions for items they created, so that they resolutions are immediately visible in the UI. If this is implemented correctly, they should also be stored in the `resolvedConfiguration`. Then it should be sufficient to pass the `OrtResult` itself (which implements `ResolutionProvider` and provides resolutions from the `resolvedConfiguration`) as `ResolutionProvider` when calling ORT's `Evaluator` and `Reporter`. Except if server managed resolutions are not stored in the `resolvedConfiguration` for some reason, I don't remember the details right now.
> 
> Then I find the API of `Evaluator` quite confusing. If `OrtResult` already is a resolution provider (which from an OO design perspective might be debatable), why does the `Evaluator` then support an additional resolution provider argument? Are there valid use cases where here a different provider is needed?

Yes, the ORT CLI works differently. Here the resolutions are applied by the evaluator, or by the reporter if the evaluator was not executed. The other tools like advisor or scanner apply resolutions only to determine the exit code but do not add them to the `resolvedConfiguration`. IIRC the main reason for this were `orthw` related workflows where it is common that new configuration like resolutions is added and then only part of the tools are executed again, and having them in the `resolvedConfiguration` too early would somehow break those workflows.

However, for the ORT server it's different as items appear in the API responses (and UI) immediately once a tool finished, so resolutions should also be applied immediately. And at least currently, there is no way to re-run only part of the jobs.

oheger-bosch:
Thanks for explaining. I have have changed `EvaluatorRunner` to use the `OrtResult` as `ResolutionProvider` and added a test to verify that the problematic scenario is fixed.

mnonnenmacher (approved):

PR #5747, opened 2026-08-18
Update test images with Renovate

mnonnenmacher:
Automate updating the Docker images used in tests with Renovate. Also pin all image versions to prevent unintended upgrades that could potentially break tests. See the commit messages for details.

### DIFF
--- a/utils/test/src/commonMain/kotlin/Images.kt
+++ b/utils/test/src/commonMain/kotlin/Images.kt
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2026 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


sschuberth (approved):

utils/test/src/commonMain/kotlin/Images.kt
```diff
@@ -0,0 +1,34 @@
...
+ * single location makes it easier to automatically update them with Renovate.
+ */
+object Images {
+    const val ARTEMIS = "apache/artemis:2.55.0"
+    const val AZURITE = "mcr.microsoft.com/azure-storage/azurite:3.35.0"
+    const val KEYCLOAK = "quay.io/keycloak/keycloak:26.6.0"
+    const val LOCALSTACK = "localstack/localstack:3.4.0"
+    const val POSTGRES = "postgres:15.19"
+    const val RABBITMQ = "rabbitmq:4.3.4"
+    const val VAULT = "hashicorp/vault:1.13.13"
```
sschuberth, line 33: @mnonnenmacher, as an afterthought, can we group these updates to void running into line-conflicts for each of these when updating in separate PRs?

mnonnenmacher: I think this should be less of a problem after the inital upgrade PRs. But adding KDoc for each constant would also solve the line-conflict problem, right?

sschuberth: > I think this should be less of a problem after the inital upgrade PRs.

Agreed. So let's just leave it as-is.

> But adding KDoc for each constant would also solve the line-conflict problem, right?

I don't think so. GitHub at least used to track conflicts on the whole file level.

PR #5771, opened 2026-08-20
ui: Start migrating the UI tables to TanStack Table v9

### LINKED ISSUE #5753: Create minimal migration for TanStack Table v9
Change the UI codebase to do minimal migration (no real v9 features) by adopting it according to instructions in https://tanstack.com/table/latest/docs/framework/react/guide/use-legacy-table. This enables the dependency update and provides a basis for the real migration to start.

At least these tasks should be done (but check the provided link to documentation for more):
- update TanStack Table to suitable v9 version in `package.json` (9.1.2 or newer if exists)
- note that all legacy imports come from '@tanstack/react-table/legacy' instead of '@tanstack/react-table'
- change all occurrences of `useReactTable()` to `useLegacyTable()` 
- change all row model imports into their legacy versions
- change all `Column`, `ColumnDef`, `Row` types into their legacy versions
- change all column helpers into their legacy versions
- rename all sorting functions

### LINKED ISSUE #5770: Sorting packages table with dependency type leads to 400 error
This was found while doing the TanStack Table v9 legacy migration, but it is not related to it, as the bug is found also on a deployed server.

Etsija:
This PR starts migrating the UI tables by implementing a minimal "legacy migration", which unblocks updating the dependency from v8 to v9.

The compatibility layer changes rendering internals (methods now live on prototypes, and the `useLegacyTable()` hook subscribes to the whole table state). For that reason, in addition to automated testing, the PR has been verified manually by:
- accessing every UI table in the browser and verifying that the UI doesn't break
- checking server-side pagination and page-size changes on repository runs table and vulnerabilities table
- checking multi-column sorting, including the third click that removes a column from the sort
- checking sorting reset in card-table sort menu
- checking column filters and the filter variants (text, regex, select, infinite select, single select)
- checking row expansion and sub-component rendering on the packages, projects, issues, rule-violations, and license-findings tables
- verifying responsive column sizing (`useTableSizing`) after a window resize, including a table that mixes `size`, `widthPercentage`, and `isGrow` columns

NOTE: it was tested that legacy migration increases the UI bundle size by 2.4% due to TanStack Table v9 core and the legacy compatibility layer. This overhead is temporary, until the tables are migrated to the native v9 API in upcoming PRs.

Please see the commits for details. For reviewers:
- commit 1 starts migration by updating  the dependency and breaking the UI build temporarily
- commits 2-3 prepare for the migration
- commits 4-12 migrate the tables group by group (mostly mechanic replacement of methods and types). Each commit is verified by checking that the TypeScript errors vanish for touched files. Each commit also records the errors left
- commit 13 finishes migration
- commit 14 fixes an existing bug for packages table that was found while testing the migration

### DIFF
--- a/ui/eslint.config.js
+++ b/ui/eslint.config.js
@@ -65,7 +65,7 @@ export default defineConfig([
       // value that is never nullish, so that the checks that do matter stand
       // out.
       '@typescript-eslint/no-unnecessary-condition': 'error',
-      // Silence the warnings about useReactTable() hook, because React Compiler
+      // Silence the warnings about useLegacyTable() hook, because React Compiler
       // is not used in this project.
       'react-hooks/incompatible-library': 'off',
     },
--- a/ui/package.json
+++ b/ui/package.json
@@ -43,7 +43,7 @@
     "@tanstack/react-query-devtools": "5.101.4",
     "@tanstack/react-router": "1.170.25",
     "@tanstack/react-router-devtools": "1.167.1",
-    "@tanstack/react-table": "8.21.3",
+    "@tanstack/react-table": "9.1.2",
     "ae-cvss-calculator": "1.0.13",
     "axios": "1.19.0",
     "class-variance-authority": "0.7.1",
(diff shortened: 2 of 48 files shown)


lamppu (approved):

lamppu:
Just for future reference, as you said that you had split this to so many commits for easier reviewing, I think commits 4-12 could have very well have been just one commit, especially when reviewing with the "Hide whitespace" setting you can see that it's the same kind of refactoring done everywhere, so IMO it would have been just fine to review it all in one commit.

Etsija:
> Just for future reference, as you said that you had split this to so many commits for easier reviewing, I think commits 4-12 could have very well have been just one commit, especially when reviewing with the "Hide whitespace" setting you can see that it's the same kind of refactoring done everywhere, so IMO it would have been just fine to review it all in one commit.

You're right, and I did consider it, just thought it was too many files in one commit :smile:

PR #5801, opened 2026-08-24
test(transport): Stop SQS receivers after each test

Etsija:
The SQS receiver tests share one queue, and every test started a receiver that was never stopped again. A leftover receiver could pick up a message meant for a later test and put it into its own abandoned queue. The message was then gone from SQS, and the waiting test failed on a timeout. This made the `test` job in `build-and-test.yml` fail often, usually passing on a rerun, in essence making the test task very flaky.

`startReceiver` now returns a `ReceiverHandle` that still behaves like the `BlockingQueue` holding the received messages, but also offers a `stop` function. The SQS test stops the receivers it started in `afterTest`, before the queue is deleted.

Bug was reproduced with 4/10 runs failing before the fix, and 0/25 after the fix.

### DIFF
--- a/transport/spi/src/testFixtures/kotlin/Utils.kt
+++ b/transport/spi/src/testFixtures/kotlin/Utils.kt
@@ -28,9 +28,11 @@ import java.util.concurrent.BlockingQueue
 import java.util.concurrent.LinkedBlockingQueue
 import java.util.concurrent.TimeUnit
 
-import kotlin.concurrent.thread
-
-import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.cancelAndJoin
+import kotlinx.coroutines.launch
 
 import org.eclipse.apoapsis.ortserver.config.ConfigManager
 import org.eclipse.apoapsis.ortserver.model.orchestrator.OrchestratorMessage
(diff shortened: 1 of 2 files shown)


transport/sqs/src/test/kotlin/SqsMessageReceiverFactoryTest.kt
```diff
@@ -63,17 +64,28 @@ class SqsMessageReceiverFactoryTest : StringSpec({
...
     beforeTest {
         client.createQueue(CreateQueueRequest { queueName = TEST_QUEUE_NAME })
         queueResponse = client.getQueueUrl(GetQueueUrlRequest { queueName = TEST_QUEUE_NAME })
     }
 
     afterTest {
+        // Stop the receivers before deleting the queue. Otherwise, a receiver of an already finished test keeps
+        // polling the queue and steals the messages that are meant for the next test.
+        receivers.forEach { receiver -> receiver.stop() }
+        receivers.clear()
```
mnonnenmacher, line 84: The commit message says "The SQS receiver tests share one queue", but actually the queue is deleted and recreated for each test case which does not seem to have any effect. If this doesn't work, the queue should be created only once after starting the test container and the code to to create and delete the queue for each test case should be removed (and it should be explained in the commit message that the old approach did not work, otherwise the "share one queue" is confusing when looking at the code).

I was wondering why other tests like `ArtemisMessageReceiverFactoryTest` and `RabbitMqReceiverFactoryTest` do not have the same issue. The reason seems to be that for them a new test container is started for each test case (and all of them are only stopped when the spec is finished, so they actually run multiple containers in parallel which is not great).

transport/spi/src/testFixtures/kotlin/Utils.kt
```diff
@@ -64,28 +66,45 @@ fun createConfigManager(
...
+ */
+class ReceiverHandle internal constructor(
+    private val job: Job,
+    queue: BlockingQueue<Message<OrchestratorMessage>>
+) : BlockingQueue<Message<OrchestratorMessage>> by queue {
+    /**
+     * Stop this receiver and wait until it has terminated. Tests that share a queue between test cases must do this
+     * at the end of each test case; otherwise the still running receiver of an already finished test case competes
+     * for the messages that are meant for the next test case. This only works for receivers that suspend while
+     * waiting for messages, which is currently the case for the SQS transport only.
```
mnonnenmacher: The last sentence is unclear to me, why was the `startReceiver` function changed to start a coroutine instead of a thread if this does only work for the SQS receiver? Could this stay a thread so that the solution is compatible with the other transports as well? Because based on my other comment they might need this feature as well to avoid that they run multiple test containers in parallel.

mnonnenmacher (changes requested):

transport/sqs/src/test/kotlin/SqsMessageReceiverFactoryTest.kt
Etsija: Regarding Artemis and RabbitMQ, Claude Opus 5 _did_ flag them as well, with this explanation:

"RabbitMQ and Artemis receivers still leak (they never call `stop()`), but they're harmless since each test gets its own broker. Their blocking JMS/AMQP receives also wouldn't respond to cancellation - noted in the `stop()` KDoc."

And one other thing flagged:

"`SqsMessageReceiverFactory` never closes its `SqsClient`. Pre-existing, out of scope here."

Etsija: Reply from Claude / Opus 5:

"Correct, recreating the queue doesn't isolate anything: same name, same URL, so a running receiver just keeps polling it. But it isn't useless. I tried creating the queue once and dropping the per-test create/delete, and 2 of 4 runs failed again. Stopping a receiver can leave behind a message that SQS already handed out but that was never processed or deleted - it reappears after the 30 s visibility timeout (this was verified by running tests with timing information) and turns up in a later test.

So stopping the receivers keeps the tests apart, and deleting the queue clears leftover messages. Both are now explained in the code and the commit message."

transport/spi/src/testFixtures/kotlin/Utils.kt
Etsija: Reply from Claude / Opus 5:

"The KDoc sentence was wrong. Stopping works for RabbitMQ too - it collects a cancellable flow and waits in `awaitClose`. Reworded the comments.

A thread wouldn't work for any of the three, though: none of the receivers react to `Thread.interrupt()`. The only places they can be stopped are the coroutine suspension points, so a thread would just wrap the same mechanism.

Artemis is the real exception, because of its own loop: `while (true)` around a blocking `receive()` with no timeout. Making it stoppable could be a follow-up if we want Artemis and RabbitMQ tests to share one container."

mnonnenmacher (approved):

PR #5803, opened 2026-08-24
Admin config improvements

mnonnenmacher:
Apply a bunch of refactorings to prepare for integrating the admin config with the plugin manager. See the commit messages for details.

### DIFF
--- a/components/admin-config/backend/src/main/kotlin/ScannerConfig.kt
+++ b/components/admin-config/backend/src/main/kotlin/ScannerConfig.kt
@@ -52,4 +52,22 @@ data class ScannerConfig(
     /** The source code origins to use, ordered by priority. The list must not be empty or contain any duplicates. */
     val sourceCodeOrigins: List<SourceCodeOrigin> =
         ortDefaultDownloaderConfig.sourceCodeOrigins.map { SourceCodeOrigin.valueOf(it.name) }
-)
+) {
+    /**
+     * Validate this [ScannerConfig] and return a list of found issues. If the returned list is empty, this config is
+     * valid.
+     */
+    fun validate(): List<String> =
+        buildList {
+            if (sourceCodeOrigins.isEmpty()) {
+                add("'sourceCodeOrigins' from scanner configuration must not be empty.")
+            }
+
+            if (sourceCodeOrigins.toSet().size != sourceCodeOrigins.size) {
+                add(
+                    "'sourceCodeOrigins' from scanner configuration must not contain duplicates. " +
+                            "Current value is $sourceCodeOrigins."
+                )
+            }
+        }
+}
--- a/components/admin-config/backend/src/test/kotlin/AdminConfigServiceTest.kt
+++ b/components/admin-config/backend/src/test/kotlin/AdminConfigServiceTest.kt
@@ -77,7 +77,7 @@ class AdminConfigServiceTest : WordSpec({
             }
 
             val exception = shouldThrow<ConfigException> {
-                service.loadAdminConfig(context, validate = true)
+                service.loadAdminConfig(context, validateConfigFiles = true)
             }
 
             exception.message shouldContain "non-default-garbage.yml"
@@ -97,7 +97,7 @@ class AdminConfigServiceTest : WordSpec({
             val service = createServiceWithConfig(config)
 
             val exception = shouldThrow<ConfigException> {
-                service.loadAdminConfig(context, validate = true)
+                service.loadAdminConfig(context)
             }
 
             exception.message shouldContain "'invalidReporter'"
@@ -113,7 +113,7 @@ class AdminConfigServiceTest : WordSpec({
             val service = createServiceWithConfig(config)
 
             val exception = shouldThrow<ConfigException> {
-                service.loadAdminConfig(context, validate = true)
+                service.loadAdminConfig(context)
             }
 
             exception.message shouldContain "'sourceCodeOrigins'"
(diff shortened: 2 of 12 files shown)


components/admin-config/backend/src/test/kotlin/ScannerConfigTest.kt
```diff
@@ -0,0 +1,40 @@
...
+class ScannerConfigTest : WordSpec({
+    "validate" should {
+        "return an issue if the list of source code origins is empty" {
+            ScannerConfig(sourceCodeOrigins = emptyList()).validate() should haveSize(1)
+        }
+
+        "return an issue if the list of source code origins contains duplicates" {
+            ScannerConfig(
+                sourceCodeOrigins = listOf(SourceCodeOrigin.VCS, SourceCodeOrigin.ARTIFACT, SourceCodeOrigin.VCS)
+            ).validate() should haveSize(1)
```
sschuberth: Why was the test made less specific than the original `exception.message shouldContain "'sourceCodeOrigins'"`?

PS: Kotest by now has `shouldBeSingle()` which could be useful here.

components/admin-config/backend/src/main/kotlin/ScannerConfig.kt
```diff
@@ -52,4 +52,22 @@ data class ScannerConfig(
     /** The source code origins to use, ordered by priority. The list must not be empty or contain any duplicates. */
     val sourceCodeOrigins: List<SourceCodeOrigin> =
         ortDefaultDownloaderConfig.sourceCodeOrigins.map { SourceCodeOrigin.valueOf(it.name) }
-)
+) {
+    /**
+     * Validate this [ScannerConfig] and return a list of found issues. If the returned list is empty, this config is
+     * valid.
+     */
+    internal fun validate(): List<String> =
```
sschuberth: Why not make this `public` now? My thinking is: The class is public, and every user of the class should be able to validate it.

components/admin-config/backend/src/test/kotlin/ReporterConfigTest.kt
```diff
@@ -119,10 +120,42 @@ class ReporterConfigTest : WordSpec({
...
+            ReporterConfig(
+                reportDefinitionsMap = mapOf("definition" to ReportDefinitionTemplate(pluginId = "non-existing"))
+            ).validate() should haveSize(1)
+        }
+    }
 })
 
 private const val REPORT_DEFINITION_NAME = "disclosure-document"
-private const val PLUGIN_ID = "PdfTemplateReporter"
+private const val PLUGIN_ID = "PdfTemplate"
```
sschuberth, line 242: Strictly, this change is unrelated, or?

components/admin-config/backend/src/test/kotlin/ReporterConfigTest.kt
```diff
@@ -119,10 +120,42 @@ class ReporterConfigTest : WordSpec({
             }
         }
     }
+
+    "validate" should {
```
sschuberth: In following commit you adhere to the convention to use a "()" suffix when testing functions. Why not also here? Are you aligning with whatever the existing test code does?

components/admin-config/backend/src/test/kotlin/ScannerConfigTest.kt
mnonnenmacher: I have added the checks using `shouldBeSingleton {}`.

components/admin-config/backend/src/main/kotlin/ScannerConfig.kt
mnonnenmacher: I agree, I don't see a reason why this and the other new functions should not be public.

components/admin-config/backend/src/test/kotlin/ReporterConfigTest.kt
mnonnenmacher: I have added the reason to the commit message (the old value was not a valid plugin ID).

components/admin-config/backend/src/test/kotlin/AdminConfigServiceTest.kt
```diff
@@ -1004,6 +798,167 @@ class AdminConfigServiceTest : WordSpec({
             }
         }
     }
+
+    "validateConfigFiles" should {
```
sschuberth: Nit: Here you're missing the conventional "()" still.

components/admin-config/backend/src/test/kotlin/AdminConfigTest.kt
```diff
@@ -0,0 +1,70 @@
...
+package org.eclipse.apoapsis.ortserver.components.adminconfig
+
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.collections.beEmpty
+import io.kotest.matchers.collections.containExactlyInAnyOrder
+import io.kotest.matchers.collections.haveSize
+import io.kotest.matchers.should
+
+class AdminConfigTest : WordSpec({
+    "getNonDefaultConfigFiles" should {
```
sschuberth: Nit: Here you're missing the conventional "()" still.

components/admin-config/backend/src/test/kotlin/AdminConfigTest.kt
```diff
@@ -0,0 +1,70 @@
...
+                reporterConfig = ReporterConfig(
+                    howToFixTextProviderFile = "non-default-5"
+                )
+            ).getNonDefaultConfigFiles() should containExactlyInAnyOrder(
+                "non-default-1", "non-default-2", "non-default-3", "non-default-4", "non-default-5"
+            )
+        }
+    }
+
+    "validate" should {
```
sschuberth: Nit: Here you're missing the conventional "()" still.

sschuberth (approved):

PR #5807, opened 2026-08-25
feat(analyzer)!: Allow overriding of the repository configuration

oheger-bosch:
So far, analyzer-related options in the repository configuration took precedence over options in the `AnalyzerJobConfiguration`. This was contrary to the handling of the environment configuration where the options in the job configuration could override the options in `.env.ort.yml`. Revert the priority, so that it is possible to override settings from `.ort.yml` when triggering a run. This also simplifies testing of options: The `.ort.yml` file in the repository defines default values, but those can now be overridden on a per-run basis.

The existing merge logic in ORT gave priority to the repository configuration. Therefore, add a new merge function with corresponding tests.

### DIFF
--- a/workers/analyzer/src/main/kotlin/AnalyzerRunner.kt
+++ b/workers/analyzer/src/main/kotlin/AnalyzerRunner.kt
@@ -39,6 +39,7 @@ import org.ossreviewtoolkit.analyzer.determineEnabledPackageManagers
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.ResolvedPackageCurations
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+import org.ossreviewtoolkit.model.config.RepositoryAnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.model.readValue
 import org.ossreviewtoolkit.model.readValueOrNull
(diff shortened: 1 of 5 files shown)


workers/analyzer/src/main/kotlin/AnalyzerRunner.kt
```diff
@@ -321,3 +311,52 @@ class AnalyzerRunner(
...
+        else -> {
+            val keys = sortedSetOf(String.CASE_INSENSITIVE_ORDER).apply {
+                addAll(ortPackageManagerOptions.keys)
+                addAll(filePackageManagers.keys)
+            }
+
+            val fromJob = sortedMapOf<String, OrtPackageManagerConfiguration>(String.CASE_INSENSITIVE_ORDER)
+                .apply { putAll(ortPackageManagerOptions) }
+            val fromFile = sortedMapOf<String, OrtPackageManagerConfiguration>(String.CASE_INSENSITIVE_ORDER)
+                .apply { putAll(filePackageManagers) }
```
mnonnenmacher: Slightly simpler:
```kotlin
val fromJob = ortPackageManagerOptions.orEmpty().toSortedMap(String.CASE_INSENSITIVE_ORDER)
val fromFile = filePackageManagers.orEmpty().toSortedMap(String.CASE_INSENSITIVE_ORDER)
```

workers/analyzer/src/main/kotlin/AnalyzerRunner.kt
```diff
@@ -321,3 +311,52 @@ class AnalyzerRunner(
...
+                    configFromFile == null -> checkNotNull(configFromJob)
+                    configFromJob == null -> configFromFile
+                    else -> configFromFile.merge(configFromJob)
+                }
+            }
+        }
+    }
+
+    return AnalyzerConfiguration(
+        allowDynamicVersions = allowDynamicVersions,
```
mnonnenmacher, line 354: Could you add a comment that explains that the `AnalyzerJobConfiguration` always has a value for `allowDynamicVersions` so a fallback to the value from `.ort.yml` is not possible?
I will probably create a follow-up ticket to change this to work the same ways as `skipExcluded` for the API use case (because if you provide no value in the API request it should use the value from `.ort.yml` like it does for `skipExcluded`).

mnonnenmacher (changes requested):

workers/analyzer/src/main/kotlin/AnalyzerRunner.kt
oheger-bosch: The compiler is event smart enough to deduce that the maps cannot be *null*.

workers/analyzer/src/main/kotlin/AnalyzerRunner.kt
oheger-bosch: I have added a corresponding TODO comment.

mnonnenmacher (approved):

PR #5818, opened 2026-08-26
ui: Require a valid URL when creating/editing a repository

### LINKED ISSUE #5812: Enable creating a repository only with a valid URL
In the initial state of the current "Add repository" form, where the URL is empty, the "Create" button should be disabled. It should get enabled if the "URL" field contains a syntactically valid URL (no check for an actually reachable URL should be done at this stage). The URL schema in the form is currently `ZodString` so that needs to be changed.

Etsija:
Screenshot from 2026-08-26 13-48-55

Screenshot from 2026-08-26 13-49-14

The same functionality is implemented also for the "edit repository" form, where the Submit button is disabled, until a valid URL is entered.

### DIFF
--- a/ui/src/schemas/index.ts
+++ b/ui/src/schemas/index.ts
@@ -22,9 +22,20 @@ import z from 'zod';
 import {
   zJobs,
   zOrtRunStatus,
+  zRepository,
   zSeverity,
   zVulnerabilityRating,
 } from '@/api/zod.gen';
+
+export const repositoryFormSchema = zRepository
+  .pick({
+    description: true,
+    name: true,
+    type: true,
+  })
+  .extend({ url: z.url({ protocol: /^https?$/ }) });
+
+export type RepositoryFormValues = z.infer<typeof repositoryFormSchema>;
 
 // Schema to validate that a string is a valid regular expression.
 export const regexSchema = z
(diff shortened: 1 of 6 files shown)


ui/src/schemas/index.ts
```diff
@@ -22,10 +22,21 @@ import z from 'zod';
...
   zVulnerabilityRating,
 } from '@/api/zod.gen';
 
+export const repositoryFormSchema = zRepository
+  .pick({
+    description: true,
+    name: true,
+    type: true,
+  })
+  .extend({ url: z.url({ protocol: /^https?$/ }) });
```
lamppu, line 36: There seems to also be `z.httpUrl()` maybe that could be used as well https://zod.dev/api?id=urls#urls

Etsija: `z.httpUrl()` also restricts the hostname to Zod's domain pattern, which rejects URLs using localhost or an IP address. These can be valid for private repositories and are accepted by the backend. Using `z.url()` with an HTTP(S) protocol restriction rejects arbitrary schemes without adding a hostname restriction. For the sake of unifying URL validation between back-end and front-end, I'd rather keep `z.url()`.

lamppu (approved):

ui/src/schemas/index.ts
sschuberth: Also, in theory Git repos may be cloned over `ssh://` URLs.

PR #5829, opened 2026-08-27
ui: Add a private repository easily

### LINKED ISSUE #5813: Enable to specify user credentials directly when adding a private repository
In the "Add repository" form, there should be a way to quickly add user credentials (username and password/PAT) for a private repository so that the repository can actually be reached by ORT Server. Feeding in these credentials when adding a repository should in the background create secrets for the username and password, and a matching infrastructure service, using the API. Both items should be created only after the repository has been created, and scoped to that created repository, not product or organization.

For easy management of repository secrets and infrastructure services, they should be named in a recognizable way. When the user adds a new private repository, we can assume it has no existing secrets or infrastructure services yet, so we could use hardcoded names for these such as `REPOSITORY_USER`, `REPOSITORY_PASSWORD`, `REPOSITORY_ACCESS` for them. If the added repository is of type `git`, `credentialsTypes` should be set to `GIT_CREDENTIALS_FILE`.

Possible linking a pre-existing (on org or prod level) infrastructure service to the added repository is out of scope of this issue - it will come later.

Etsija:
#### The user can specify credentials directly in the "Add repository" form

Screenshot from 2026-08-27 13-39-20

#### Repository secrets for credentials and an infrastructure service for repository access are automatically added

They are listed on the respective pages for review, but once the private repository has been added with credentials, the user is taken straight away to the "Create run" page, just as usual. The user doesn't have to even look at the secrets and infrastructure service but they are there for review:

Screenshot from 2026-08-28 06-20-18
Screenshot from 2026-08-27 13-41-04

#### The user can go on analyzing the new private repository right away

Screenshot from 2026-08-27 16-01-19

Tested manually in the UI with the author's credentials and a private git repository - works as expected.

### DIFF
--- a/website/docs/user-guide/concepts/external-services.md
+++ b/website/docs/user-guide/concepts/external-services.md
@@ -31,6 +31,16 @@ If Git should not be able to obtain the credentials from the `Netrc` file, one c
 In some cases, there could be conflicting services; for instance, if multiple repositories with different credentials are defined on the same repository server.
 Therefore, it is also possible to choose not to include the credentials in any files.
 
+### Repository Credentials
+
+When adding a repository in the UI, you can optionally provide a username and a password or personal access token (PAT).
+ORT Server then creates the repository-scoped Secrets `REPOSITORY-USERNAME` and `REPOSITORY-PASSWORD` and the repository-scoped Infrastructure Service `REPOSITORY-ACCESS`.
+The Infrastructure Service points to the repository URL and references the two Secrets.
+For a Git repository, its credentials are added to the Git credentials file; no credential types is set for other repository types.
+
+You can view and edit these resources on the repository's _Secrets_ and _Infrastructure Services_ pages.
+If credentials are shared by multiple repositories, create an Infrastructure Service at the organization or product level instead.
+
 ### `.ort.env.yml` Example
 
 ```yaml
--- a/ui/src/components/form/optional-input.tsx
+++ b/ui/src/components/form/optional-input.tsx
@@ -17,10 +17,23 @@
  * License-Filename: LICENSE
  */
 
+import type { ComponentProps } from 'react';
+
 import { Input } from '@/components/ui/input.tsx';
 
-export const OptionalInput = ({ ...props }) => {
+type OptionalInputProps = Omit<
+  ComponentProps<typeof Input>,
+  'placeholder' | 'value'
+> & {
+  placeholder?: string;
+  value?: ComponentProps<typeof Input>['value'] | null;
+};
+
+export const OptionalInput = ({
+  placeholder = '(optional)',
+  ...props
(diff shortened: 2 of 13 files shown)


sschuberth:
Nit: I believe the auto-generated secret name for the username should end with "USERNAME" instead of "USER", just to fully match the field name.

Also, should we prefer "\_" instead of "-" as a separate for secret names? Different providers might have different constraints regarding the name; might it that providers do not accept "-" or "\_" as part of secret names?

mnonnenmacher:
> Also, should we prefer "\_" instead of "-" as a separate for secret names? Different providers might have different constraints regarding the name; might it that providers do not accept "-" or "\_" as part of secret names?

Currently secret names are allowed to contains dashes but not underscores. AFAIK all supported providers allow dashes, I don't know if there was a specific reason to not allow underscores.

Etsija:
> > Also, should we prefer "_" instead of "-" as a separate for secret names? Different providers might have different constraints regarding the name; might it that providers do not accept "-" or "_" as part of secret names?
> 
> Currently secret names are allowed to contains dashes but not underscores. AFAIK all supported providers allow dashes, I don't know if there was a specific reason to not allow underscores.

That's exactly the case, and the reason why it's `-`, not `_`.

sschuberth:
UI LGTM from the screenshots, leaving the TS code review to others.

website/docs/user-guide/concepts/external-services.md
```diff
@@ -31,6 +31,16 @@ If Git should not be able to obtain the credentials from the `Netrc` file, one c
 In some cases, there could be conflicting services; for instance, if multiple repositories with different credentials are defined on the same repository server.
 Therefore, it is also possible to choose not to include the credentials in any files.
 
+### Repository Credentials
+
+When adding a repository in the UI, you can optionally provide a username and a password or personal access token (PAT).
+ORT Server then creates the repository-scoped Secrets `REPOSITORY-USER` and `REPOSITORY-PASSWORD` and the repository-scoped Infrastructure Service `REPOSITORY-ACCESS`.
```
sschuberth: This now also needs to say "REPOSITORY-USER*NAME*".

lamppu:
I really don't understand the need for refreshing the token, as the roles are now in the database. Does refreshing the token maybe accidentally help as it creates enough delay for the roles to be added to the table or something? Or why was it needed?

Etsija:
> I really don't understand the need for refreshing the token, as the roles are now in the database. Does refreshing the token maybe accidentally help as it creates enough delay for the roles to be added to the table or something? Or why was it needed?

I have completely removed the token refreshing from the PR. Tested that creating and analyzing a private repository still works as before.

lamppu (approved):

PR #5833, opened 2026-08-28
Refactor config worker

mnonnenmacher:
Refactor the config worker for better separation of concerns and add an integration test. Also fix the issue that the admin config was only validated if a validation script was present. See the commit messages for details.


workers/config/src/main/kotlin/Utils.kt
```diff
@@ -21,14 +21,46 @@ package org.eclipse.apoapsis.ortserver.workers.config
 
 import com.github.michaelbull.result.getOrElse
 
+import kotlin.time.Clock
```
sschuberth, line 24: Commit message: "all created issue*s*"

workers/config/src/main/kotlin/AdminConfigValidator.kt
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
```
sschuberth, line 2: Do you keep 2025 here due to copied code?

workers/config/src/test/kotlin/AdminConfigValidatorTest.kt
```diff
@@ -0,0 +1,123 @@
...
+            val adminConfig = AdminConfig()
+
+            val jobConfigs = JobConfigurations(
+                reporter = ReporterJobConfiguration(formats = listOf("invalid-format-1", "invalid-format-2"))
+            )
+
+            val issues = AdminConfigValidator.validate(adminConfig, jobConfigs)
+
+            issues.size shouldBe 2
+            issues.filter { it.message.contains("invalid-format-1") }.shouldBeSingleton()
```
sschuberth: I believe the combination of `filter {}` and `shouldBeSingleton()` can be simplified to `shouldHaveSingleElement {}`.

sschuberth: Also, prefer the `in` operator over `contains()`.

workers/config/src/main/kotlin/ConfigValidationResult.kt
```diff
@@ -32,6 +34,9 @@ import org.eclipse.apoapsis.ortserver.model.runs.Issue
 sealed interface ConfigValidationResult {
     /** A (possibly empty) list with issues detected during validation. */
     val issues: List<Issue>
+
+    /** The [issues] as [OptionalValue] that is [absent][OptionalValue.Absent] if there are no issues. */
+    val optionalIssues get() = if (issues.isEmpty()) OptionalValue.Absent else issues.asPresent()
```
sschuberth: Style nit: I recently did https://github.com/oss-review-toolkit/ort/pull/12365 as I believe `get()` is more readable on its own line.

workers/config/src/main/kotlin/ConfigValidationResult.kt
```diff
@@ -51,7 +51,10 @@ data class ConfigValidationResultSuccess(
 
     /** A map with labels to be added to the ORT run. */
     val labels: Map<String, String> = emptyMap()
-) : ConfigValidationResult
+) : ConfigValidationResult {
+    /** The [labels] as [OptionalValue] that is [absent][OptionalValue.Absent] if there are no labels. */
+    val optionalLabels get() = if (labels.isEmpty()) OptionalValue.Absent else labels.asPresent()
```
sschuberth: Same style nit as before.

workers/config/src/main/kotlin/ConfigValidationResult.kt
```diff
@@ -51,7 +51,10 @@ data class ConfigValidationResultSuccess(
 
     /** A map with labels to be added to the ORT run. */
     val labels: Map<String, String> = emptyMap()
-) : ConfigValidationResult
+) : ConfigValidationResult {
```
sschuberth, line 55: Commit message: Please provide a rationale. Something like "does not need to be recomputed".

workers/config/src/main/kotlin/ConfigValidationResult.kt
sschuberth: Actually, why does this have `get()` at all instead of just being assigned once, as `labels` do not change after construction?

workers/config/src/main/kotlin/ConfigValidator.kt
```diff
@@ -42,26 +40,16 @@ import org.slf4j.LoggerFactory
  */
 class ConfigValidator private constructor(
     /** The current [WorkerContext]. */
-    private val context: WorkerContext,
-
-    /** The service to access the admin configuration. */
-    private val adminConfigService: AdminConfigService
+    private val context: WorkerContext
```
sschuberth: > Before, the admin config validation was only executed when a validation
script file was found.

What's bad about that? Could you please improve the commit message to explain why validation still makes sense without a script? I.e. what's being validated if there is no script to execute?

workers/config/src/main/kotlin/ConfigValidator.kt
```diff
@@ -94,7 +82,7 @@ class ConfigValidator private constructor(
 
         return when (val result = executedScript.validationResult) {
             is ConfigValidationResultFailure -> result
-            is ConfigValidationResultSuccess -> result.validateAdminConfig()
+            is ConfigValidationResultSuccess -> result
```
sschuberth: Uhm, it it really correct that in both cases the same is returned? Why even have the case distinction then?

workers/config/src/main/kotlin/ValidationScriptRunner.kt
```diff
@@ -77,23 +59,21 @@ class ConfigValidator private constructor(
...
-     * available about the failure.
+     * Run the provided [script] and return the result. If script execution fails, return a
+     * [ConfigValidationResultFailure].
      */
-    fun validate(script: String): ConfigValidationResult = runCatching {
-        runScript(script)
-    }.getOrElse { exception ->
-        createScriptErrorResult(script, exception)
+    override fun runScript(script: SourceCode): ConfigValidationResult = runCatching {
+        (run(script).scriptInstance as ValidationScriptTemplate).validationResult
```
sschuberth, line 61: IMO this is a bit unreadable. How about doing

```kotlin
val validationScript = run(script).scriptInstance as ValidationScriptTemplate
```

first?

config/spi/src/testFixtures/kotlin/ConfigFileProviderFactoryForTesting.kt
```diff
@@ -72,7 +72,7 @@ class ConfigFileProviderFactoryForTesting : ConfigFileProviderFactory {
             if (context == ConfigManager.EMPTY_CONTEXT) {
                 File("src/testFixtures/resources/config-files")
             } else {
-                File(context.name)
+                File(context.name.removePrefix(RESOLVED_PREFIX))
```
sschuberth, line 75: Why did the bug not have any negative impact on testing?

workers/config/src/test/kotlin/ConfigWorkerIntegrationTest.kt
```diff
@@ -0,0 +1,310 @@
...
+
+            val result = configWorker.run(ortRun.id)
+
+            result shouldBe RunResult.Success
+
+            fixtures.ortRunRepository.get(ortRun.id).shouldNotBeNull {
+                resolvedJobConfigs.shouldNotBeNull {
+                    analyzer.enabledPackageManagers.shouldNotBeNull {
+                        this shouldNot beEmpty()
+                    }
```
sschuberth: Instead of nesting and referring to `this`, you could also use a one-liner:

```kotlin
analyzer.enabledPackageManagers.shouldNotBeNull() shouldNot beEmpty()
```

Similar below in lines 303-305.

sschuberth (changes requested):

workers/config/src/main/kotlin/AdminConfigValidator.kt
mnonnenmacher: Yes

workers/config/src/main/kotlin/ConfigValidationResult.kt
mnonnenmacher: GitHub doesn't show me which commit this refers to but I guess it's the one that added the `optionalLabels` property. I have added a rationale.

workers/config/src/main/kotlin/ConfigValidator.kt
mnonnenmacher: I have added an explanation (admin config and validation script are two different things).

workers/config/src/main/kotlin/ConfigValidator.kt
mnonnenmacher: It's correct, I didn't notice in this commit that the code is a bit odd now, but it's cleaned up in a later commit already.

workers/config/src/main/kotlin/ValidationScriptRunner.kt
mnonnenmacher: Really? I liked it.

config/spi/src/testFixtures/kotlin/ConfigFileProviderFactoryForTesting.kt
mnonnenmacher: Because the next commit adds the first test that actually uses the provider to resolve files.

workers/config/src/main/kotlin/ValidationScriptRunner.kt
sschuberth: I'm ok to keep it, just wanted to share my opinion.

sschuberth (approved):

PR #5836, opened 2026-08-28
Miscellaneous UI (wording) improvements

sschuberth:


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/users/index.tsx
```diff
@@ -129,8 +129,9 @@ const ManageUsers = () => {
         <CardTitle>Manage Users</CardTitle>
         <CardDescription className='flex flex-col gap-2'>
           <span>
-            Assign or remove users to different groups in repository:{' '}
-            <span className='font-semibold'>{repository.url}</span>
+            Assign or remove users to different groups in the{' '}
+            <span className='font-semibold'>{repository.url}</span>{' '}
```
mnonnenmacher: This should use `repository.name` if it is not empty and `repository.url` only as fallback. But also, why not just say "in this repository" (same below)?

mnonnenmacher (changes requested):

ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/users/index.tsx
sschuberth: Here, I found it quite useful to leave it as-is and repeat the name as it's otherwise not nearby in the UI, just to ensure the user is in the place (s)he thinks.

mnonnenmacher (approved):

PR #5841, opened 2026-08-30
docs(website): Split releasing out of contributing

sschuberth:
As most contributors do not do releases, split out creating releases into a dedicated section, and generally reword docs a bit.

### DIFF
--- a/website/docs/developer-guide/contributing.md
+++ b/website/docs/developer-guide/contributing.md
@@ -6,18 +6,6 @@ position: 2
 
 See the [contributing guide](https://github.com/eclipse-apoapsis/ort-server/?tab=contributing-ov-file) for general information on how to contribute to the ORT Server project.
 
-## Semantic versioning
-
-The version of a new release is determined by the types of the commit messages since the previous release, following the [semantic versioning](https://semver.org/) scheme.
-For this, the [git-semver-plugin](https://github.com/jmongard/Git.SemVersioning.Gradle) for Gradle is used.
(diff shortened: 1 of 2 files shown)


website/docs/developer-guide/contributing.md
```diff
@@ -29,3 +17,9 @@ You can also force `detekt` to automatically fix some issues (for example, impor
 
 For TypeScript, you can run `pnpm format:check` within the `ui/` and `/website` directories to check the code style.
 You can also automatically fix issues by running `pnpm format` within the respective directories.
+
+## Commit messages
+
+Commit messages need to follow the [semantic versioning](https://semver.org/) scheme as the version of a new release is determined by the types of the commit messages since the previous release.
```
mnonnenmacher: Did you forget to remove this section after moving to the other file? If not, this should rather emphasize to read the detailed description for PRs and commit messages in the contributing guide linked at the top.

sschuberth: No, I did not forget to remove it, as I felt it's relevant for both cases. So I kept it in both places with just slightly different wording.

I'll check how to link to the contributing guide in a meaningful way.

mnonnenmacher: Ok, I wasn't sure if it was intentional.

I think saying "commit messages need to follow the semantic versioning scheme" does not make sense as the commit messages are not versioned, I'd rather say something like "Commit messages need to follow the conventional commit specification as ..."

sschuberth: Agreed, that's actually what I wanted to say...

mnonnenmacher (approved):

PR #5850, opened 2026-08-31
fix(ui): Refer to "roles" instead of "groups" for managing users

sschuberth:
Note that this only cares about user-facing UI for now. Code should be adjusted as well, i.e. regarding function names.

### DIFF
--- a/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/users/index.tsx
+++ b/ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/users/index.tsx
@@ -104,7 +104,7 @@ const ManageUsers = () => {
         }),
       });
       toast.info('Add User', {
-        description: `User "${form.getValues().username}" added successfully to group "${form.getValues().groupId.toUpperCase()}".`,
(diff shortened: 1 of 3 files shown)


ui/src/routes/organizations/$orgId/products/$productId/repositories/$repoId/_repo-layout/users/index.tsx
```diff
@@ -132,7 +132,7 @@ const ManageUsers = () => {
         <CardTitle>Manage Users</CardTitle>
         <CardDescription className='flex flex-col gap-2'>
           <span>
-            Assign or remove users to different groups in the{' '}
+            Assign or remove roles to users to in the{' '}
```
Etsija: One "to" too much.

Etsija (changes requested):

Etsija:
What function names are you referring to? Organization users for example already uses an auto-generated query function `putOrganizationRoleToUserMutation()` I believe the naming is similar on all entity levels for user handling.

Answering to myself: in the code, `group` and `getGroupIcon()` are used, so yeah, these will then look a bit odd, after this PR.

Etsija (approved):

PR #5859, opened 2026-08-31
Git config file provider improvements

mnonnenmacher:
Prepare the `GitConfigFileProvider` to be used in `:core`, see the commit messages for details.

### DIFF
--- a/config/git/src/main/kotlin/GitConfigFileProvider.kt
+++ b/config/git/src/main/kotlin/GitConfigFileProvider.kt
@@ -26,6 +26,10 @@ import java.io.FilterInputStream
 import java.io.InputStream
 import java.util.HashMap
 
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+import kotlin.time.TimeMark
+import kotlin.time.TimeSource
 import kotlin.time.measureTime
 
 import org.eclipse.apoapsis.ortserver.config.ConfigException
@@ -33,6 +37,7 @@ import org.eclipse.apoapsis.ortserver.config.ConfigFileProvider
 import org.eclipse.apoapsis.ortserver.config.Context
 import org.eclipse.apoapsis.ortserver.config.Path
 import org.eclipse.apoapsis.ortserver.config.resolveSecurely
+import org.eclipse.apoapsis.ortserver.utils.config.getLongOrDefault
 import org.eclipse.apoapsis.ortserver.utils.config.getServiceUrl
 
 import org.ossreviewtoolkit.model.VcsInfo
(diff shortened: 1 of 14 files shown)


config/git/src/main/kotlin/GitConfigFileProvider.kt
```diff
@@ -110,7 +144,7 @@ class GitConfigFileProvider internal constructor(
      * the working tree is initialized first. The resolved revision is returned.
      */
     private fun updateWorkingTree(requestedRevision: String): String {
-        synchronized(this) {
+        synchronized(lock) {
```
sschuberth: Nit: `updateWorkingTree()` is now only called from functions that themselves have `synchronized(lock)`, so I  guess it _could_ be omitted from here.

config/git/src/main/kotlin/GitConfigFileProvider.kt
```diff
@@ -71,28 +73,60 @@ class GitConfigFileProvider internal constructor(
...
 
-    override fun resolveContext(context: Context): Context {
+    /**
+     * A lock guarding all access to the working tree in [configDir]. A single provider instance can be accessed
+     * concurrently by multiple threads, and all of them share this instance's one working tree. Therefore, the sequence
+     * of updating the tree to the requested revision and reading files from it must happen atomically. Otherwise, a
+     * concurrent call requesting a different revision could re-checkout the tree between the update and the read,
+     * causing files from the wrong revision, an inconsistent tree, or read failures to be observed.
+     */
+    private val lock = Any()
```
sschuberth, line 106: For the record: I'm personally not a fan of introducing such separate "dummy" `lock` objects if using `synchronized(this)` would suffice.

config/git/src/main/kotlin/GitConfigFileProvider.kt
```diff
@@ -49,25 +54,43 @@ import org.slf4j.LoggerFactory
...
         const val GIT_URL = "gitUrl"
 
+        /**
+         * Configuration property for the time-to-live (in seconds) of the [revisionCache] that maps a requested context
+         * (like a branch name) to its resolved revision.
+         */
+        const val GIT_REVISION_CACHE_TTL_SECONDS = "gitRevisionCacheTtlSeconds"
+
+        /** The default time-to-live of the [revisionCache] if [GIT_REVISION_CACHE_TTL_SECONDS] is unset. */
+        val DEFAULT_REVISION_CACHE_TTL_SECONDS: Duration = 60.seconds
```
sschuberth: Nit: The explicit type could be omitted.

config/git/src/main/kotlin/GitConfigFileProvider.kt
mnonnenmacher: I wasn't sure about keeping it maybe as a safety net, but I have removed it now and added a note to the function docs.

config/git/src/main/kotlin/GitConfigFileProvider.kt
mnonnenmacher: The reason for the private lock is encapsulation, it prevents that any other code with access to the instance could hold the lock and cause deadlocks or block other threads from using the instance. It's unlikely to be a problem here as we use the provider only from controlled code, but I think it's a best practice to use private locks.

sschuberth (approved):

PR #5873, opened 2026-09-02
ci(github): Use `pnpm ci` to install dependencies

sschuberth:
This is the default in CI environments, see [1].

[1]: https://pnpm.io/cli/install#--frozen-lockfile

### DIFF
--- a/.github/workflows/publish-website.yml
+++ b/.github/workflows/publish-website.yml
@@ -34,7 +34,7 @@ jobs:
 
     - name: Install dependencies
       working-directory: website
-      run: pnpm install --frozen-lockfile
+      run: pnpm ci
 
     - name: Test build
       working-directory: website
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-website.yml
```diff
@@ -34,7 +34,7 @@ jobs:
 
     - name: Install dependencies
       working-directory: website
-      run: pnpm install --frozen-lockfile
+      run: pnpm install
```
mnonnenmacher: I like being explicit about that and would prefer to keep it, even if it is technically not required.

mnonnenmacher: Another option would be to switch to https://pnpm.io/cli/ci.

sschuberth: I've switched to `pnpm ci` and verified that it indeed supports `--dev`, although not explicitly documented.

mnonnenmacher (approved):

PR #5875, opened 2026-09-02
Explicit config context

mnonnenmacher:
Introduce separate value classes for the requested and resolved config context. See the commit messages for details.

### DIFF
--- a/config/spi/src/main/kotlin/Context.kt
+++ b/config/spi/src/main/kotlin/Context.kt
@@ -20,13 +20,37 @@
 package org.eclipse.apoapsis.ortserver.config
 
 /**
- * A class representing a context for querying the application configuration.
+ * A class representing the requested config context. This context still needs to be resolved to a
+ * [ResolvedConfigContext] before it can be used to access the application config.
  *
- * When querying the configuration of ORT Server, clients have to provide a context, which uniquely defines the
- * configuration to be used. The exact meaning is specific to a concrete implementation of the service provider
- * interface, but in general, via the context a set of configuration properties can be selected if there are multiple
- * such sets. For instance, if the configuration is stored in a version control system, the context could be the
- * revision of the configuration to be checked out.
+ * The config context is used to access different versions of the configuration files provided by the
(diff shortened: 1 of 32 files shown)


workers/common/src/main/kotlin/context/WorkerContextImpl.kt
```diff
@@ -127,7 +127,11 @@ internal class WorkerContextImpl(
         createOrtTempDir(ortRunId.toString()).also(tempDirectories::add)
 
     /** Stores the resolved configuration context. */
-    private val currentContext by lazy { ortRun.resolvedJobConfigContext?.let(::Context) }
+    private val currentContext by lazy {
+        requireNotNull(ortRun.resolvedJobConfigContext) {
```
sschuberth: Nit: `require...` usually is for (API) user input. This looks more to me like you'd want to assert state, so probably `check...` is the better choice.

config/spi/src/main/kotlin/Context.kt
```diff
@@ -20,13 +20,37 @@
...
+        val EMPTY = RequestedConfigContext("")
+    }
+}
+
+/**
+ * A class representing the resolved config context. A resolved config context is required to read files from a
+ * [ConfigFileProvider]. Also see [RequestedConfigContext].
+ */
+@JvmInline
+value class ResolvedConfigContext(val name: String) {
```
sschuberth, line 48: How about making these also part of a `ConfigContext` sealed interface that has `val name: String`?

sschuberth (commented):
Note that there is a test failure. I didn't check whether it's related to your changes or not.

config/spi/src/main/kotlin/Context.kt
mnonnenmacher: I did not want to have a shared interface because there is not a single place in code where it does not matter if the context was resolved or not.

sschuberth:
> Note that there is a test failure. I didn't check whether it's related to your changes or not.

`DaoRepositoryRepositoryTest` still seems to fail.

sschuberth (approved):

PR #5879, opened 2026-09-03
ui: Improve creating new users

### LINKED ISSUE #5835: Improvements to the "Create User" form
1. The password should not be optional. Even if the user is forced to change the password on first login, still an initial password should be set that's securely shared with the user to ensure that only that user can sign up (and no anyone who guesses the username right).
2. The "Create" button should only be enabled once all criteria are fulfilled, i.e. username and password were entered ~~, and at least one organization was chosen~~.

Etsija:
This PR improves the "Create User" form, by fixing some logical flaws in it, making it unit-testable (which improves its upkeeping, as it has moderate amount of logic in its functionality), and improving the layout of the form to be more condensed.

#### Create button is only enabled after both username and password are entered

Note that the user doesn't need to be put to any organization, when creating - this was legacy Keycloak requirement, which is now dropped.

Screenshot from 2026-09-03 07-48-21

#### After the creation, the new user is immediately seen in the users table (a bug fixed here)

Screenshot from 2026-09-03 07-49-02

#### Admin gave the new user _only_ WRITE rights to this repository

Verified in the UI that the user can navigate to the repository from the root, but cannot create any new organizations, products, or repositories. Runs to the repository can be created.

Screenshot from 2026-09-03 07-51-55

NOTE: A good follow-up to this is immediately obvious: improve and streamline user creation by extending the form so that when creating a new user, all organization, product, and repository rights could be immediately set up for the user in this form.

Please see the commits for details. Commits 1 and 2 prepare for addition of new features, commits 3-5 implement them. Commit 7 has the UI layout polish.

### DIFF
--- a/ui/src/routes/admin/users/create-user/index.tsx
+++ b/ui/src/routes/admin/users/create-user/index.tsx
@@ -97,7 +97,10 @@ export const CreateUserForm = ({
       temporary: true,
       organizations: [],
     },
+    mode: 'onChange',
   });
+
+  const isSubmitting = isPending || form.formState.isSubmitting;
 
   return (
     <Card className='col-span-2 w-full'>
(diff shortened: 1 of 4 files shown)


sschuberth:
LGTM based on the screenshots! 👍🏻

lamppu (approved):

PR #5892, opened 2026-09-04
Include report definitions from admin config in available reporter plugins

### LINKED ISSUE #4315: Integrate admin-config for reporters with plugin manager
~~Integrate the admin-config that defines available report types with the plugin manager. The available report types should be treated like individual reporter plugins as this is how the API to trigger runs treats them.~~

~~This will allow to configure these types via the plugin manager and also to show them in the UI to trigger runs.~~

I have discarded the original plan above because of the problem that the admin config is not stored in a database but in a Git repository, and for each run a different revision of that repository could be used. This means that there is no way to ensure that the config from the plugin manager matches the admin config used in a run.

Instead, the plan is now to treat the admin config as a completely separate way to configure reporter plugins: The plugin manager will not allow to create plugin templates for custom report types from the admin config, but it will integrate them in the API response that is used to populate the forms on the create run page.

mnonnenmacher:
Include the report definitions from the admin config in the list of available reporter plugins. Also, generate forms for reporter plugin options in the UI as it is already done for other plugin types. See the commit messages for details.

**This contains a breaking change:**
The same config file provider that is used for the workers must now be configured for `core` as well.

### DIFF
--- a/components/admin-config/backend/src/main/kotlin/ReporterConfig.kt
+++ b/components/admin-config/backend/src/main/kotlin/ReporterConfig.kt
@@ -237,7 +237,7 @@ data class ReporterConfig(
      * [directory references][ReportDefinitionTemplate.assetDirectoriesRefs] to concrete [ReporterAsset]s. If a
      * reference cannot be resolved, it is recorded in [unresolvedAssetFilesRefs] or [unresolvedAssetDirectoriesRefs].
      */
-    private val resolvedReportDefinitions = reportDefinitionsMap.mapValues { (name, template) ->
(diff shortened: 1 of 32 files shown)


components/admin-config/backend/src/main/kotlin/ReporterConfig.kt
```diff
@@ -207,27 +207,6 @@ data class ReporterConfig(
             .map { it.id.id }
             .associateBy { it.lowercase() }
 
-        /**
```
sschuberth, line 210: Commit message: "reuqire" -> "require"

sschuberth (approved):

PR #5922, opened 2026-09-08
Search, page and sort admin/users in the back-end

### LINKED ISSUE #4505: User Management: Admin view of users should be searchable
To ease maintenance of user for an admin, the list of users should be searchable (by name). A name search could match any of the "Username", "First name", or "Last name" columns for convenience.

The UI part of it would put a search input field between the "Create user" button and the table, not to any column as a filter, as this search will match anything from those three columns. The search would probably need to commence only after enter is pressed or focus lost from the search field, not debounced. Example of such do exist in the UI codebase.

The filter should probably not be a regEx filter, but rather a substring filter.

Image

Etsija:
Issue #4505 asks for a back-end filter for the admin/users endpoint. Combining server-side filtering with client-side pagination would paginate only the subset returned by the server, causing incomplete pages, page counts, and inaccessible matching users. Therefore this PR moves pagination, sorting and filtering completely to the API list endpoint. This also brings the benefit of a table now not only filterable, but also sortable by "Username", "First name", "Last name", and "Email address" columns.

Note that this is a breaking change.

#### No filter

Screenshot from 2026-09-08 08-13-06

#### Filter matching both "Username" and "Last name"

Screenshot from 2026-09-08 08-14-17

#### Filter matching "Last name"

Screenshot from 2026-09-08 08-14-36

### DIFF
--- a/components/authorization/backend/src/main/kotlin/service/KeycloakUserService.kt
+++ b/components/authorization/backend/src/main/kotlin/service/KeycloakUserService.kt
@@ -64,9 +64,6 @@ class KeycloakUserService(
         keycloakClient.deleteUser(userId)
     }
 
-    override suspend fun getUsers(): Set<User> =
-        keycloakClient.getUsers().mapTo(mutableSetOf()) { it.toOrtUser() }
-
     override suspend fun listUsers(parameters: ListQueryParameters, search: String?): ListQueryResult<User> {
         val users = keycloakClient.getUsers().mapTo(mutableSetOf()) { it.toOrtUser() }
             .filter { user -> search.isNullOrBlank() || user.contains(search) }
--- a/components/authorization/backend/src/main/kotlin/service/KeycloakUserService.kt
+++ b/components/authorization/backend/src/main/kotlin/service/KeycloakUserService.kt
@@ -26,7 +26,12 @@ import kotlinx.coroutines.withContext
 import org.eclipse.apoapsis.ortserver.clients.keycloak.KeycloakClient
 import org.eclipse.apoapsis.ortserver.clients.keycloak.User as KeycloakUser
 import org.eclipse.apoapsis.ortserver.clients.keycloak.UserName
+import org.eclipse.apoapsis.ortserver.dao.QueryParametersException
 import org.eclipse.apoapsis.ortserver.model.User
+import org.eclipse.apoapsis.ortserver.model.util.ListQueryParameters
+import org.eclipse.apoapsis.ortserver.model.util.ListQueryResult
+import org.eclipse.apoapsis.ortserver.model.util.OrderDirection
+import org.eclipse.apoapsis.ortserver.model.util.OrderField
 
 /**
  * An implementation of the [UserService] interface that uses Keycloak as the backend user management system. As unique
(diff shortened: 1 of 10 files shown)


sschuberth:
@lamppu, please also have a look at the implementation.

components/authorization/backend/src/main/kotlin/service/UserService.kt
```diff
@@ -48,6 +50,14 @@ interface UserService {
      */
     suspend fun getUsers(): Set<User>
 
+    /**
+     * List current users according to the given [parameters] and optional [search].
```
mnonnenmacher: This should explain how `search` is used (that it is case-insensitive and which properties are taken into account).

components/authorization/backend/src/test/kotlin/service/KeycloakUserServiceTest.kt
```diff
@@ -106,6 +112,169 @@ class KeycloakUserServiceTest : WordSpec({
...
+    "listUsers" should {
+        "return a deterministic default page ordered by username" {
+            val keycloakUsers = (25 downTo 1).mapTo(mutableSetOf()) { index ->
+                createKeycloakUser(index, username = "user${index.toString().padStart(2, '0')}")
+            }
+            val service = createService(keycloakUsers)
+
+            val result = service.listUsers()
+
+            result.data.map(User::username) shouldContainExactly (1..20).map { index ->
```
mnonnenmacher: This should use `ListQueryParameters.DEFAULT_LIMIT` instead of the magic number `20`.

components/authorization/backend/src/test/kotlin/service/KeycloakUserServiceTest.kt
```diff
@@ -106,6 +112,169 @@ class KeycloakUserServiceTest : WordSpec({
...
+            )
+            val service = createService(keycloakUsers)
+            val expectedAscending = mapOf(
+                "username" to listOf("alpha", "bravo", "charlie"),
+                "firstName" to listOf("alpha", "bravo", "charlie"),
+                "lastName" to listOf("charlie", "alpha", "bravo"),
+                "email" to listOf("bravo", "alpha", "charlie")
+            )
+
+            expectedAscending.forEach { (field, expected) ->
```
mnonnenmacher: Please call this `sortField` instead of `field` to make it easier to understand what the test does.

components/authorization/backend/src/test/kotlin/service/KeycloakUserServiceTest.kt
```diff
@@ -106,6 +112,169 @@ class KeycloakUserServiceTest : WordSpec({
...
+        "load users from Keycloak only once" {
+            val keycloakUsers = setOf(createKeycloakUser(1), createKeycloakUser(2))
+            val client = mockk<KeycloakClient> {
+                coEvery { getUsers() } returns keycloakUsers
+            }
+            val service = KeycloakUserService(client)
+
+            service.listUsers(search = "user")
+
+            coVerify(exactly = 1) { client.getUsers() }
```
mnonnenmacher: I would drop this test case, it just makes it harder to refactor the implementation and provides little value.

ui/tests/unit/components/admin-user-search.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+    expect(getLastUsersRequest().query?.sort).toBe('username');
+    expect(
+      screen
+        .getAllByRole('row')
+        .slice(1)
+        .map((row) => within(row).getAllByRole('cell')[0]?.textContent)
+    ).toEqual(['zulu', 'alpha', 'mike']);
+  });
+
+  it('does not allow sorting by superuser status', async () => {
```
mnonnenmacher: Do we really need a test to verify that a column is not sortable?

components/authorization/backend/src/main/kotlin/service/KeycloakUserService.kt
```diff
@@ -64,9 +64,6 @@ class KeycloakUserService(
         keycloakClient.deleteUser(userId)
     }
 
-    override suspend fun getUsers(): Set<User> =
```
mnonnenmacher, line 62: Commit message: The "BREAKING CHANGE" section is redundant, it just repeats the sentence before.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #5923, opened 2026-09-08
docs(workers): Improve log messages when running fails

sschuberth:
In particular, clarify that not processing of the message itself fails, but running the work as instructed by the message.

### DIFF
--- a/workers/advisor/src/main/kotlin/AdvisorWorker.kt
+++ b/workers/advisor/src/main/kotlin/AdvisorWorker.kt
@@ -143,12 +143,12 @@ internal class AdvisorWorker(
     }.getOrElse {
         when (it) {
             is JobIgnoredException -> {
-                logger.warn("Message with traceId '$traceId' ignored: ${it.message}")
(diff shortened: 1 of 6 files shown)


workers/advisor/src/main/kotlin/AdvisorWorker.kt
```diff
@@ -143,12 +143,12 @@ internal class AdvisorWorker(
     }.getOrElse {
         when (it) {
             is JobIgnoredException -> {
-                logger.warn("Message with traceId '$traceId' ignored: ${it.message}")
+                logger.warn("Not running any work as message '$traceId' got ignored: ${it.message}")
```
mnonnenmacher: How about being more specific in these messages:
"Not running the advisor as ..." / "Error while running the advisor ..."

By now the trace ID is part of the MDC so it could be removed, this message was probably written before we used the MDC.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

Issue #2323, opened 2025-03-21
Package configurations should get applied to license findings


sschuberth:
There were reports that package configurations (like e.g. [license findings curations](https://oss-review-toolkit.org/ort/docs/configuration/package-configurations#defining-path-excludes-and-license-finding-curations) for detected licenses) are not applied / being reflected correctly in the UI. This should first be verified to be the case, e.g. by

1. Creating a run with a project scanner enabled that creates detected license findings
2. Double-checking that these findings are correctly displayed in the UI
3. Creating a "dummy" [license findings curation](https://oss-review-toolkit.org/ort/docs/configuration/package-configurations#defining-path-excludes-and-license-finding-curations) that e.g. eliminates a license finding by setting `concluded_license` to `NONE`.
4. Verifying that a rerun does not show that detected license anymore.

mnonnenmacher:
There are currently no endpoints that return detected licenses, so I don't think this issue applies to any existing endpoint.

sschuberth:
> There are currently no endpoints that return detected licenses

Ah, thanks for pointing that out. I was indeed assuming such an endpoint to already be there, and just not being used yet by the "License Findings" dummy menu entry we already have in the UI.

sschuberth:
> > There are currently no endpoints that return detected licenses
> 
> Ah, thanks for pointing that out. I was indeed assuming such an endpoint to already be there, and just not being used yet by the "License Findings" dummy menu entry we already have in the UI.

See https://github.com/eclipse-apoapsis/ort-server/issues/2074.

mnonnenmacher:
@Etsija Could you please check if the endpoint you added in #2074 already applies package configurations (license finding curations and path excludes)? If yes, this issue could be closed.

Etsija:
There's clear evidence from data that the endpoint in question _does not_ apply package configurations yet: I followed the instructions from @sschuberth and added a `.ort.yml` with a dummy license finding curation to it, ran the full scan, and the database shows the curation in `license_finding_curations` table:

<img width="1252" height="42" alt="Image" src="https://github.com/user-attachments/assets/706e10af-3ccc-4e41-9d24-a52893a469dd" />

Yet the UI still shows the license finding as before:

<img width="1165" height="336" alt="Image" src="https://github.com/user-attachments/assets/672bef64-765e-4e68-8241-c2c464c8fb3d" />

Etsija:
[...]

PR #5924, opened 2026-09-08
ui: Add license finding curation snippet

Etsija:
When starting to prepare initial testing of #2323, a need came up for a small "helper component": the license findings sub-table actually contains a lot of useful information to produce a "license finding curation template" to the clipboard:

Screenshot from 2026-09-08 13-38-11

Pressing the `CopyToClipboard` component produces this snippet to the clipboard, to be easily edited by the user, in case they want to add a package configuration, or a curation to a repository's `.ort.yml`:

```yaml
id: "NPM::chokidar:3.6.0"
license_finding_curations:
- path: "package/README.md"
  start_lines: "306"
  line_count: 3
  detected_license: "MIT AND LicenseRef-scancode-unknown-license-reference"
  reason: "CODE|DATA_OF|DOCUMENTATION_OF|INCORRECT|NOT_DETECTED|REFERENCE"
  comment: ""
  concluded_license: "<SPDX expression>|NONE"
```

### DIFF
--- a/ui/src/components/copy-to-clipboard.tsx
+++ b/ui/src/components/copy-to-clipboard.tsx
@@ -41,22 +41,19 @@ export const CopyToClipboard = ({
   tooltipContentClassName,
   className,
 }: CopyToClipboardProps) => {
-  const [isCopied, setIsCopied] = useState(false);
+  const [copyStatus, setCopyStatus] = useState<'idle' | 'copied' | 'failed'>(
+    'idle'
+  );
 
-  async function copyTextToClipboard(text: string) {
-    return await navigator.clipboard.writeText(text);
-  }
+  const handleCopyClick = async () => {
+    try {
+      await navigator.clipboard.writeText(copyText);
+      setCopyStatus('copied');
+    } catch {
+      setCopyStatus('failed');
+    }
 
-  // onClick handler function for the copy button
-  const handleCopyClick = () => {
-    // Asynchronously call copyTextToClipboard
-    copyTextToClipboard(copyText).then(() => {
-      // If successful, update the isCopied state value
-      setIsCopied(true);
-      setTimeout(() => {
-        setIsCopied(false);
-      }, 1500);
-    });
+    setTimeout(() => setCopyStatus('idle'), 1500);
   };
   return (
     <TooltipProvider delayDuration={0}>
@@ -70,12 +67,20 @@ export const CopyToClipboard = ({
             onClick={handleCopyClick}
           >
             <span className='fg-slate-300'>
-              {isCopied ? <Check color='gray' /> : <Copy color='gray' />}
+              {copyStatus === 'copied' ? (
+                <Check color='gray' />
+              ) : (
+                <Copy color='gray' />
+              )}
             </span>
           </Button>
         </TooltipTrigger>
         <TooltipContent className={tooltipContentClassName}>
-          {isCopied ? 'Copied!' : tooltipText}
+          {copyStatus === 'copied'
+            ? 'Copied!'
+            : copyStatus === 'failed'
+              ? 'Copy failed'
+              : tooltipText}
         </TooltipContent>
       </Tooltip>
     </TooltipProvider>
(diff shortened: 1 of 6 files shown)


ui/src/lib/license-finding-curation.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+const LICENSE_FINDING_CURATION_REASONS = [
+  'CODE',
+  'DATA_OF',
+  'DOCUMENTATION_OF',
+  'INCORRECT',
+  'NOT_DETECTED',
+  'REFERENCE',
+] as const;
```
sschuberth, line 27: Don't we get auto-generated Zod types for these? If not, why, shouldn't we?

ui/src/lib/license-finding-curation.ts
```diff
@@ -0,0 +1,70 @@
...
+      `  start_lines: ${quoteYamlString(startLine.toString())}`,
+      `  line_count: ${endLine - startLine + 1}`
+    );
+  }
+
+  lines.push(
+    `  detected_license: ${quoteYamlString(detectedLicense)}`,
+    `  reason: ${quoteYamlString(REASON_PLACEHOLDER)}`,
+    '  comment: ""',
+    '  concluded_license: "A-valid-license|NONE"'
```
sschuberth: Instead of `A-valid-license` I propose `<SPDX expression>`. Note the `<>` which I added to signal that this, in contrast to `NONE`, is a template value.

ui/src/lib/license-finding-curation.ts
Etsija: Simple reason: the `reason` is not exposed in the OpenAPI, thus, we are not seeing it in the UI. `LicenseFindingCuration.reason` is a plain `string` on the back-end side. Just one of those not-so-good aspects of the OpenAPI spec.

sschuberth: Ouch, I guess the root cause is that

https://github.com/eclipse-apoapsis/ort-server/blob/cb4a779a6e596d0602afd104c4d32a49fbc945e1/dao/src/main/kotlin/repositories/repositoryconfiguration/LicenseFindingCurationsTable.kt#L43

is a `text()` instead of `enumerationByName<>()`. Any idea whether this is just an oversight, @mnonnenmacher?

mnonnenmacher: I don't remember if there was a reason to not make this an enum, but it looks like an oversight.

sschuberth: Would making this an enum (plus doing all implied Kotlin code changes) already solve the problem of not having a auto-generated Zod types for the license finding curation reasons?

Etsija: If the enum becomes exposed to the OpenAPI spec as a type, it will become visible to the UI as a Zod type.

ui/src/components/copy-to-clipboard.tsx
```diff
@@ -30,12 +30,14 @@ import {
...
   tooltipContentClassName?: string;
   className?: string;
 };
 
 export const CopyToClipboard = ({
   copyText,
+  tooltipText = 'Copy to clipboard',
   tooltipContentClassName,
   className,
 }: CopyToClipboardProps) => {
```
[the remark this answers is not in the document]
Etsija: Not really related to this PR, but a nice find anyway, so I added a fix commit.

ui/src/components/copy-to-clipboard.tsx
```diff
@@ -73,7 +75,7 @@ export const CopyToClipboard = ({
           </Button>
         </TooltipTrigger>
         <TooltipContent className={tooltipContentClassName}>
-          {isCopied ? 'Copied!' : 'Copy to clipboard'}
+          {isCopied ? 'Copied!' : tooltipText}
         </TooltipContent>
```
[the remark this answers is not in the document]
Etsija: Not really related to this PR, but a nice find anyway, so I added a fix commit.

sschuberth (approved):

PR #5947, opened 2026-09-10
Add `OrtServerAuthenticator` to `:core`

mnonnenmacher:
Install the `OrtServerAuthenticator` in `:core` to enable authentication with private repositories when using the `GitConfigFileProvider`. See the commit messages for details.


core/src/test/kotlin/plugins/AuthenticatorTest.kt
```diff
@@ -0,0 +1,111 @@
+/*
+ * Copyright (C) 2025 The ORT Server Authors (See <https://github.com/eclipse-apoapsis/ort-server/blob/main/NOTICE>)
```
sschuberth: Nit: Shouldn't the year here be 2026?

config/git/README.md
```diff
@@ -59,6 +59,9 @@ configManager {
 ```
 The concrete secret values are then queried from the _secrets provider_.
 
+For this to work, the `OrtServerAuthenticator` must be installed and the URL with the secret references must be configured via an environment variable.
+The authenticator will read the secret references from the environment variable and return the resolved secrets when JGit asks for credentials for the given URL.
```
sschuberth, line 63: Just checking, this is really only ever JGit, not the `git` CLI, correct?

mnonnenmacher: Yes, TBH I don't know how that works with the `git` CLI, probably the special credentials provider is needed in that case. But the CLI is mainly used for submodules and sparse checkouts which are not supported. And as a fallback in case fetching with JGit fails, but I tested with all kinds of revisions as input and the fallback was never triggered.

sschuberth (approved):

PR #5967, opened 2026-09-14
fix(storage/s3): fail fast on missing bucket name configuration

nnobelis:
Throw a descriptive IllegalArgumentException in S3StorageProviderFactory when the 's3BucketName' property is absent, instead of letting a null bucket name propagate into PutObjectRequest and surface deep inside the AWS SDK as a cryptic "Required value was null." error at write time.


mnonnenmacher:
@nnobelis Are you sure the "Required value was null" error is not coming from the issue fixed in #5968?

nnobelis:
> @nnobelis Are you sure the "Required value was null" error is not coming from the issue fixed in #5968?

Yes it did. Thanks for the fix !

PR #5969, opened 2026-09-14
fix(ui): Consistenty use the term "Add" for hierarchy elements

sschuberth:
Instead of "Create", use the less technical "Add". This fixes some inconsistencies where both terms were mixed in the UI.


ui/src/routes/create-organization/index.tsx
```diff
@@ -91,7 +91,7 @@ const CreateOrganizationPage = () => {
 
   return (
     <Card className='mx-auto w-full max-w-4xl'>
-      <CardHeader>Create Organization</CardHeader>
+      <CardHeader>Add Organization</CardHeader>
```
lamppu, line 94: Nit: Typo in commit message Consistenty -> Consistently

lamppu:
There's some test failures in the `ui/tests/unit/components/create-repository-form.test.tsx` test because it doesn't find the button anymore, these can be fixed by replacing

`screen.getByRole('button', { name: 'Create' })`
with
`screen.getByRole('button', { name: 'Add' })`

lamppu (approved):

PR #5986, opened 2026-09-16
Persist and show issue how-to-fix messages

### LINKED ISSUE #5907: Show "How-to-fix"-messages for issues in the UI
We noticed that our "How-to-fix"-messages are shown in the web-app-report, but not in the UI, where we need the information as well.

Report:
Image

UI:
Image

Etsija:
#### How to fix messages for an ORT run are persisted in the database
Screenshot from 2026-09-16 11-20-29

#### ...and shown in the UI
Screenshot from 2026-09-16 10-43-49

The PR was tested by providing a dummy `reporter.how-to-fix-text-provider.kts` locally (in `scripts/compose/config`), matching some issue messages, and rerunning.


sschuberth:
@oheger-bosch, @nnobelis, @MarcelBochtler please review, if possible.

dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
```diff
@@ -224,6 +245,38 @@ class DaoOrtRunRepository(private val db: Database) : OrtRunRepository {
...
+    override fun updateIssueHowToFixTexts(ortRunId: Long, issues: Collection<Issue>): Int {
+        if (issues.isEmpty()) return 0
+
+        val distinctIssues = issues.groupBy { it.matchKey() }.values.map { matchingIssues ->
+            val howToFixTexts = matchingIssues.mapTo(mutableSetOf()) { it.howToFix }
+            val issue = matchingIssues.first()
+
+            require(howToFixTexts.size == 1) {
+                "Conflicting how-to-fix texts for issue '${issue.source}: ${issue.message}' at ${issue.timestamp}."
+            }
```
mnonnenmacher, line 263: I think it's unlikely that this ever happens in reality, because the `IssueMatchKey` contains all properties of ORT's `Issue` class and the how to fix text provider should return the same text for the same input. Nevertheless, the error message should contain all properties that are relevant for the `groupBy` to fully explain the problem, otherwise it could be very hard to understand what the problem is in case this ever happens at runtime.

In other words, the error message should explain that there are two issues with the same `timestamp`, `source`, `message`, `severity`, and `affectedPath`, but different how to fix texts (including all values).

services/ort-run/src/main/kotlin/OrtRunService.kt
```diff
@@ -643,6 +643,16 @@ class OrtRunService(
         }
     }
 
+    /**
+     * Set the how-to-fix text of all issue occurrences in the ORT run with [ortRunId] that match the content and
+     * timestamp of the given [issues], regardless of their identifier. Return the number of updated occurrences.
+     */
+    fun updateIssueHowToFixTexts(ortRunId: Long, issues: Collection<Issue>): Int {
+        val updatedOccurrences = ortRunRepository.updateIssueHowToFixTexts(ortRunId, issues)
+        logger.debug("Updated how-to-fix text for {} issue occurrences in ORT run {}.", updatedOccurrences, ortRunId)
```
mnonnenmacher: Nit: "issue occurrences" -> "issues"

dao/src/main/kotlin/repositories/ortrun/DaoOrtRunRepository.kt
```diff
@@ -29,7 +29,10 @@ import org.eclipse.apoapsis.ortserver.dao.mapAndDeduplicate
 import org.eclipse.apoapsis.ortserver.dao.repositories.product.ProductsTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.repository.RepositoriesTable
 import org.eclipse.apoapsis.ortserver.dao.repositories.userDisplayName.UserDisplayNameDao
+import org.eclipse.apoapsis.ortserver.dao.tables.shared.IssuesTable
```
mnonnenmacher, line 32: Commit message should make clear that this is only preparing the feature but not yet activating it. Currently it sounds like with this commit how to fix texts would already be stored.

workers/reporter/src/main/kotlin/ReporterRunner.kt
```diff
@@ -166,26 +168,87 @@ class ReporterRunner(
...
+        ortResult: OrtResult,
+        provider: HowToFixTextProvider
+    ): HowToFixTexts {
+        val issues = buildSet {
+            ortResult.getIssues().values.forEach(::addAll)
+            addAll(ortResult.getAdvisorProviderIssues())
+        }
+
+        if (provider === HowToFixTextProvider.NONE) {
+            return HowToFixTexts(issues.associateWith { null }, null)
```
mnonnenmacher: How to fix texts are resolved only once for each run, but never again. This means that there will never be a situation where an already stored how to fix text for an issue will be reset to `null`. Therefore the result should only contain issues for which a how to fix text was found, so that only those are updated in the database.

Otherwise, the queries from the previous commits that search for matching issues will do a lot of unnecessary work.

For this code this means that the fallback for `HowToFixTextProvider.NONE` should happen even before collecting the issues, and that the response should be empty. And the code below should filter issues for which no how to fix text was found.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #5988, opened 2026-09-16
Remove role assignments when deleting a user

### LINKED ISSUE #5973: Deleting users should also remove their role assignments
In my local dev server, I have added test users and assigned them roles. Later on I have deleted some of the test users. Currently, I have only one non-admin user `john.doe` in the server:

Image

However, `role_assignments` table still includes roles for `testuser` and `testuser2`, although the users have been deleted from Keycloak:

Image

[I'm seeing from code](https://github.com/eclipse-apoapsis/ort-server/blob/6db3a723379d6559537d3206f1fb5432f822833d/core/src/main/kotlin/api/AdminRoute.kt#L94-L99) that deleting a user does not remove the user's role assignments.

I believe this can impose a security risk, as recreating an earlier deleted username restores the old permissions to the user (this has been verified from the server UI). In general, Keycloak users and role assignments table will go badly out of sync in time.

@oheger-bosch was this a deliberate design choice, when the database role system was implemented (and if so, what was the reason), or is it a bug?

Removing both a user from Keycloak and their role assignments from `role_assignments` table requires careful failure handling, because Keycloak and PostgreSQL cannot share a transaction.

Etsija:
The PR has been verified using pgAdmin: when a user is deleted, all their roles are removed from `role_assignments` table. Also verified from the UI that as a consequence, no "earlier existing roles" pop up, if a user is recreated with the same username.

A follow-up PR will implement a task to clear out orphaned role assignment records from this table.


sschuberth:
@oheger-bosch, @nnobelis, @MarcelBochtler please review, if possible.

components/authorization/backend/src/test/kotlin/service/KeycloakUserServiceTest.kt
```diff
@@ -80,20 +82,111 @@ class KeycloakUserServiceTest : WordSpec({
...
+            coVerifySequence {
+                client.getUser(keycloakUser.username)
+                authorizationService.removeUserAssignments(username)
+                client.deleteUser(keycloakUser.id)
+            }
+            coVerify(exactly = 0) { authorizationService.assignRole(any(), any(), any()) }
+            confirmVerified(client, authorizationService)
+        }
+
+        "retry deletion after role assignments have already been removed" {
```
mnonnenmacher: Can you explain the value of this test case?

components/authorization/backend/src/test/kotlin/service/KeycloakUserServiceTest.kt
```diff
@@ -80,20 +82,111 @@ class KeycloakUserServiceTest : WordSpec({
...
+
+            coVerifySequence {
+                client.getUser(keycloakUser.username)
+                authorizationService.removeUserAssignments(username)
+            }
+            coVerify(exactly = 0) { client.deleteUser(any()) }
+            confirmVerified(client, authorizationService)
+        }
+
+        "not restore assignments if Keycloak deletion fails" {
```
mnonnenmacher: I find this test a bit weird, why test that a feature does not exist?

core/src/test/kotlin/api/AdminRouteIntegrationTest.kt
```diff
@@ -220,7 +234,101 @@ class AdminRouteIntegrationTest : AbstractIntegrationTest({
...
+                val recreatedUsersResponse = client.get("/api/v1/admin/users")
+                recreatedUsersResponse shouldHaveStatus HttpStatusCode.OK
+                recreatedUsersResponse.body<PagedResponse<UserWithSuperuserStatus>>().data
+                    .filter { it.user.username == username }.shouldBeSingleton {
+                        it.isSuperuser shouldBe false
+                    }
+            }
+        }
+
+        "leave the user and assignments unchanged when deletion is denied" {
```
mnonnenmacher: For all other endpoints we rely on the fact that permission checks happen before any action is executed, so the "require superuser role" test below is already sufficient.

core/src/test/kotlin/api/AdminRouteIntegrationTest.kt
```diff
@@ -220,7 +234,101 @@ class AdminRouteIntegrationTest : AbstractIntegrationTest({
...
+                }
+
+                val response = client.delete("/api/v1/admin/users") {
+                    parameter("username", username)
+                }
+
+                response shouldHaveStatus HttpStatusCode.NoContent
+                dbExtension.db.dbQuery {
+                    RoleAssignmentsTable.selectAll().where { RoleAssignmentsTable.userId eq username }.count()
+                } shouldBe 0
```
mnonnenmacher: I would not go down to implementation details here and verify the content of the assignments table, testing via the `authorizationService` is sufficient. Actually, just verifying that the newly created user does not have any roles assigned is enough from the perspective of this test.

mnonnenmacher (changes requested):

components/authorization/backend/src/test/kotlin/service/KeycloakUserServiceTest.kt
Etsija: LLMs _love_ writing tests! The test tried to simulate retrying a user deletion after the roles were removed but deleting the Kaycloak user failed. However, it only made the mocks to behave as if the roles had been removed: it did not verify that this actually happened or that repeating the cleanup was safe. Since the remaining checks are already covered by other tests, I removed the whole test.

components/authorization/backend/src/test/kotlin/service/KeycloakUserServiceTest.kt
Etsija: This made me change a skill for the ORT Server back-end implementation: test required behavior, not absent hypothetical features, because it doesn't increase coverage of tests.

mnonnenmacher (approved):

PR #5998, opened 2026-09-17
Enable HTTP response compression

### LINKED ISSUE #4381: Verify that compression for HTTP responses is enabled

Etsija:
Response compression was manually verified using locally built images and the vulnerabilities page in the server UI (with cache disabled). 
- in the browser's Network panel, the vulnerabilities API request to `localhost:8080`advertised gzip support via `Accept-Encoding`, and the server returned `200 OK` with `Content-Encoding: gzip` and `Vary: Accept-Encoding`
- the response transferred 12.58 kB for a decoded size of 46.81 kB, reducing the transferred size by approximately 73%

The server UI was also manually exercised and behaved normally. Downloads were also verified to work.


core/src/main/kotlin/plugins/HTTP.kt
```diff
@@ -35,6 +37,10 @@ fun Application.configureHTTP() {
     val config: ApplicationConfig by inject()
     val allowedHosts = config.property("ktor.cors.allowedHosts").getString()
 
+    install(Compression) {
+        // Enable response compression without changing how request bodies are handled.
+        mode = CompressionConfig.Mode.CompressResponse
+    }
```
mnonnenmacher, line 43: Add en empty line below.

core/src/main/kotlin/plugins/HTTP.kt
```diff
@@ -35,6 +37,10 @@ fun Application.configureHTTP() {
     val config: ApplicationConfig by inject()
     val allowedHosts = config.property("ktor.cors.allowedHosts").getString()
 
+    install(Compression) {
+        // Enable response compression without changing how request bodies are handled.
+        mode = CompressionConfig.Mode.CompressResponse
```
mnonnenmacher, line 42: After reading the docs I think we should set the minimum size for compression to 1024 to not compress data that already fits in a single TCP package:

https://ktor.io/docs/server-compression.html#configure_response_size

Also, before merging this I would like to understand if in our case a mitigation for this issue is required, but I do not have the time to do the research now:

https://ktor.io/docs/server-compression.html#security

mnonnenmacher (changes requested):

core/src/main/kotlin/plugins/HTTP.kt
Etsija: AI-assisted answer to both points (OpenAI / gpt-6-astra):

#### Setting minimum size for compression

Setting `minimumSize(1024)` requires some other changes as well, as adding a custom condition replaces Ktor's default conditions, including content-type exclusions. Those exclusions should therefore explicitly be retained, something like
```kotlin
install(Compression) {
    mode = CompressionConfig.Mode.CompressResponse
    minimumSize(1024)
    excludeContentType(
        ContentType.Video.Any,
        ContentType.Image.JPEG,
        ContentType.Image.PNG,
        ContentType.Audio.Any,
        ContentType.MultiPart.Any,
        ContentType.Text.EventStream
    )
}
```

#### BREACH mitigation

This attack works when a compressed response contains both a secret and text that an attacker can influence. The attacker repeatedly triggers requests using the victim's access and compares response sizes to try to discover the secret.

Out API uses JWT bearer tokens, which the UI explicitly adds to the Auth header. Unlike cookies, the browser does not automatically include these tokens in requests from malicious website. The Auth header is also not included in the compressed response body. This makes the usual BREACH attack unlikely for out authenticated API.

Before saying no protection is needed, we should still check public endpoints (`liveness/` is the only one?), downloads that use tokens in their URLs (`runs/:runId/downloads/report/:token` seems to be the only one?), and setups where a proxy handles authentication. Setting the minimum size to 1024 bytes avoids unnecessary compression of small responses, but does not prevent BREACH.

mnonnenmacher: > adding a custom condition replaces Ktor's default conditions

The Ktor docs are just disappointing, how can they not mention this? I mean, it's in the function docs, but something like this should be on the website as well...

@sschuberth FYI, we discussed in the daily today that we should have clarity if any mitigation for BREACH is required before merging this but there is no time to do the research today.

PR #6010, opened 2026-09-18
Apply plugin templates in config worker

### LINKED ISSUE #4317: Apply plugin templates in config worker

mnonnenmacher:
Apply the plugin templates to all run requests in the config worker. Before, plugin templates were only considered by the UI and it was validated that run requests did not conflict with plugin templates. See the commit messages for details.


components/plugin-manager/backend/build.gradle.kts
```diff
@@ -57,6 +57,15 @@ dependencies {
     implementation(ortLibs.reporter)
     implementation(ortLibs.scanner)
 
+    // Add all ORT plugins as runtime dependencies so that the plugin manager can discover them.
```
lamppu: Nit: Typo in commit message "dependenices" -> "dependencies"

lamppu, line 60: Nit: Typo in commit message "dependenices" -> "dependencies"

workers/config/src/main/kotlin/PluginTemplateApplicator.kt
```diff
@@ -0,0 +1,306 @@
+/*
```
lamppu, line 1: Another nit: the signed-off-by footer in this commit's message has a `.org` email (other have `.io`)

workers/config/src/main/kotlin/PluginTemplateApplicator.kt
```diff
@@ -0,0 +1,306 @@
...
+                    packageConfigTemplates
+                )
+            )
+        }
+    )
+}
+
+/** Return a map containing only the entries of this map with a non-null value. */
+private fun <K, V : Any> Map<K, V?>.filterNotNullValues(): Map<K, V> =
+    mapNotNullTo(mutableListOf()) { (key, value) -> value?.let { key to it } }.toMap()
```
sschuberth: A bit shorter, but maybe not so nice due to the unchecked cast (which needs to be suppressed by an annotation):

```kotlin
filterValues { it != null } as Map<K, V>
```

workers/config/src/main/kotlin/PluginTemplateApplicator.kt
```diff
@@ -0,0 +1,306 @@
...
+    /**
+     * Resolve the default package managers and apply the plugin templates applicable to [organizationId] to
+     * [jobConfigs], and return the updated [JobConfigurations]. Only plugins that are configured to run are considered,
+     * plugins that are not installed are ignored. Throws an [IllegalStateException] if a template lookup fails.
+     */
+    fun applyTemplates(jobConfigs: JobConfigurations, organizationId: OrganizationId): JobConfigurations {
+        val configs = jobConfigs.withDefaultPackageManagers(organizationId)
+
+        return applyTemplatesToConfigs(configs, fetchTemplates(configs, organizationId))
+    }
```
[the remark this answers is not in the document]
mnonnenmacher: `AdviceProviderFactory.ALL[it]` is not case-sensitive, it's backed by a `SortedMap` using `String.CASE_INSENSITIVE_ORDER`.

workers/config/src/main/kotlin/PluginTemplateApplicator.kt
mnonnenmacher: Such an old commit 😀

workers/config/src/main/kotlin/PluginTemplateApplicator.kt
mnonnenmacher: I did this to avoid the unchecked cast, but I have removed the mutable list because it was not necessary.

sschuberth (approved):

